PR #4, opened 2020-05-04
Introduce TestStore.scope

stephencelis:
This modifies the `TestStore` interface to be a little more inline with `Store`. Rather than expose a 5-argument initializer to `TestStore`'s auto-complete, we can instead disclose this transformation as a `scope` method. Should hopefully simplify things for folks that don't need to `scope` by only exposing a single initializer!

### DIFF
--- a/Sources/ComposableArchitectureTestSupport/TestStore.swift
+++ b/Sources/ComposableArchitectureTestSupport/TestStore.swift
@@ -111,9 +111,7 @@ import XCTest
 ///         $0.results = ["Composable Architecture"]
 ///       }
 ///     )
-public final class TestStore<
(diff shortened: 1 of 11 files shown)


Sources/ComposableArchitectureTestSupport/TestStore.swift
```diff
@@ -111,29 +111,14 @@ import XCTest
 ///         $0.results = ["Composable Architecture"]
 ///       }
 ///     )
-public final class TestStore<
-  State, LocalState: Equatable, Action: Equatable, LocalAction, Environment
-> {
+public final class TestStore<State, LocalState, Action: Equatable, LocalAction, Environment> {
```
stephencelis, line 114: Moved `LocalState` constraint into an extension that defines `assert`.

PR #11, opened 2020-05-04
Documentation clean up

alohaglenn:
Fixes some minor documentation issues.

### DIFF
--- a/Sources/ComposableArchitecture/CasePaths/EnumReflection.swift
+++ b/Sources/ComposableArchitecture/CasePaths/EnumReflection.swift
@@ -43,7 +43,7 @@ extension CasePath where Value == Void {
 /// - Parameters:
 ///   - embed: An enum case initializer.
 ///   - root: A root enum value.
-/// - Returns: Values iff they can be extracted from the given enum case initializer and root enum, otherwise `nil`.
+/// - Returns: Values if they can be extracted from the given enum case initializer and root enum, otherwise `nil`.
 public func extract<Root, Value>(case embed: (Value) -> Root, from root: Root) -> Value? {
   func extractHelp(from root: Root) -> ([String?], Value)? {
     if let value = root as? Value {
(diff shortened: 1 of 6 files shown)


Sources/ComposableArchitecture/CasePaths/EnumReflection.swift
```diff
@@ -43,7 +43,7 @@ extension CasePath where Value == Void {
 /// - Parameters:
 ///   - embed: An enum case initializer.
 ///   - root: A root enum value.
-/// - Returns: Values iff they can be extracted from the given enum case initializer and root enum, otherwise `nil`.
```
stephencelis, line 46: This one is actually okay. We meant it as "if and only if," which even the standard library uses occasionally! 😄

https://github.com/apple/swift/blob/9cdfb0edfcae6fa0baadee5df3690597138b5d5e/stdlib/public/core/ContiguousArrayBuffer.swift#L504

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -79,7 +79,7 @@ public struct Reducer<State, Action, Environment> {
   /// global state, action and environment. It accomplishes this by providing 3 transformations to
   /// the method:
   ///
-  /// * A writeable key path that can get/set a piece of local state from the global state
+  /// * A writeable key path that can get/set a piece of local state from the global state.
```
stephencelis, line 82: Nice! Noticed another typo while reading 😅 
```suggestion
  /// * A writable key path that can get/set a piece of local state from the global state.
```

Sources/ComposableArchitecture/CasePaths/EnumReflection.swift
```diff
@@ -43,7 +43,7 @@ extension CasePath where Value == Void {
...
+/// - Returns: Values if they can be extracted from the given enum case initializer and root enum, otherwise `nil`.
```
stephencelis, line 46:
```suggestion
/// - Returns: Values iff they can be extracted from the given enum case initializer and root enum, otherwise `nil`.
```

Sources/ComposableArchitecture/CasePaths/EnumReflection.swift
alohaglenn: Oh, TIL! 😄

stephencelis:
Gonna merge, thanks!

PR #15, opened 2020-05-05
Add strict initialization to Reducer

oskarek:
Add a `Reducer.strict` static func for initializing a reducer by providing a function that does not take an environment as input: only state and action. Instead, it returns a function that is given the environment. The motivation is that this makes it impossible to access the environment and invoke effects directly within the reducer, which shouldn't be done.

Inspired by [exercise 2](https://www.pointfree.co/episodes/ep98-ergonomic-state-management-part-1#exercises) from your *Ergonomic State Management: Part 1* episode.

### DIFF
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -50,6 +50,42 @@ public struct Reducer<State, Action, Environment> {
     self.reducer = reducer
   }
 
+  /// Creates a reducer from a reducer function signature that does not take the environment as
+  /// input, but instead returns a function that gets access to the environment. This makes it
+  /// *impossible* to access the environment and invoke any effects directly in the reducer, which
+  /// should never be done.
+  ///
+  /// For example:
+  ///
+  ///     struct MyState { var count = 0, text = "" }
+  ///     enum MyAction { case buttonTapped, textChanged(String) }
+  ///     struct MyEnvironment { var analyticsClient: AnalyticsClient }
+  ///
+  ///     let myReducer = Reducer<MyState, MyAction, MyEnvironment>.strict { state, action in
+  ///       switch action {
+  ///       case .buttonTapped:
+  ///         state.count += 1
+  ///         return { environment in
+  ///           environment.analyticsClient.track("Button Tapped")
+  ///         }
+  ///
+  ///       case .textChanged(let text):
+  ///         state.text = text
+  ///         return .none
+  ///       }
+  ///     }
+  ///
+  /// - Parameter reducer: A function signature that takes state and action, and returns a
+  ///   function that takes an environment and returns an effect.
+  /// - Returns: A reducer.
+  public static func strict(
+    _ reducer: @escaping (inout State, Action) -> (Environment) -> Effect<Action, Never>
+  ) -> Reducer {
+    .init { state, action, environment in
+      reducer(&state, action)(environment)
+    }
+  }
+
   /// A reducer that performs no state mutations and returns no effects.
   public static var empty: Reducer {
     Self { _, _, _ in .none }
--- a/Tests/ComposableArchitectureTests/ReducerTests.swift
+++ b/Tests/ComposableArchitectureTests/ReducerTests.swift
@@ -27,6 +27,16 @@ final class ReducerTests: XCTestCase {
     XCTAssertEqual(state, 1)
   }
 
+  func testStrictReducer() {
+    let reducer = Reducer<Int, Void, Void>.strict { state, _ in state += 1
+      return { _ in .none }
+    }
+
+    var state = 0
+    _ = reducer.callAsFunction(&state, ())
+    XCTAssertEqual(state, 1)
+  }
+
   func testCombine_EffectsAreMerged() {
     typealias Scheduler = AnySchedulerOf<DispatchQueue>
     enum Action: Equatable {


stephencelis:
Hey @oskarek! Thanks for PRing 😄

We actually had `strict` defined in the library but removed it at the last minute after @pteasima pointed out that its ergonomics were a bit lacking (`{ _ in .none }` everywhere!). So instead we have redefined it in one of the case studies demos:

https://github.com/pointfreeco/swift-composable-architecture/blob/b40959cb993d7c77e5da6253ae553880073b2b27/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-StrictReducers.swift#L25-L31

Higher-order reducers like this are super useful to have around! For now we want to keep the _core_ library surface area small with what we consider "bare essentials," which may change in the coming weeks with more feedback. In the meantime, luckily these reducers are easy to define in your own projects.

Projects like RxSwift maintain an RxCommunity of projects for additions like this, so maybe we should talk about starting a ComposableArchitectureCommunity if folks are interested! 😄

oskarek:
Thanks for your response! And yeah I get what you're saying, it makes a lot of sense.

Just an idea, if the `{ _ in .none }` ugliness/verboseness was the main reason for not including the strict version in the library, how about making the returned function *optional*? Like:

```diff
   public static func strict(
-    _ reducer: @escaping (inout State, Action) -> (Environment) -> Effect<Action, Never>
+    _ reducer: @escaping (inout State, Action) -> ((Environment) -> Effect<Action, Never>)?
   ) -> Reducer {
     .init { state, action, environment in
-      reducer(&state, action)(environment)
+      reducer(&state, action)?(environment) ?? .none
     }
   }
```

Then you could, in your reducer, just do `return nil` (or even `return .none` just like for Effects 🤓) to signal that you don't want any effects to be run.

So, the example from the documentation for the method would look like:

```swift
struct MyState { var count = 0, text = "" }
enum MyAction { case buttonTapped, textChanged(String) }
struct MyEnvironment { var analyticsClient: AnalyticsClient }

let myReducer = Reducer<MyState, MyAction, MyEnvironment>.strict { state, action in
  switch action {
  case .buttonTapped:
    state.count += 1
    return { environment in
      environment.analyticsClient.track("Button Tapped")
    }

  case .textChanged(let text):
    state.text = text
    return .none // or `return nil`
  }
}
```

pteasima:
Hey @oskarek, sorry I wasnt able to reply earlier. This version returning an Optional is precisely what Im currently using in my project and it works well.

Dont expect this to get merged anytime soon tho (unless pointfreeco change their minds). They gave very good reasons to not have this in the core library. Might become part of "composable-architecture-extensions" if that ever becomes a thing. For now, just include it in your project.

oskarek-storytel:
@pteasima okay yeah, I’ll close this PR now then and just include this in my own project instead, as you said. Thanks for replying!

oskarek:
@pteasima okay yeah, I’ll close this PR now then and just include this in my own project instead, as you said. Thanks for replying!

stephencelis:
@oskarek Here's some context from a discussion in the beta repo:

> We're still determining the surface area of the library, but one thing you can also do is implement a > wrapper type that has as much ergonomics as you'd like:
> 
> ```swift
> struct EnvironmentalEffect<Environment, Effect> {
>   let run: (Environment) -> Effect
>   static var none: Self { .init { _ in .none } }
> }
> 
> extension Reducer {
>   static func strict(
>     _ reducer: @escaping (inout State, Action) -> EnvironmentalEffect<Environment, Action>
>   ) -> Reducer {
>     .init { state, action, environment in
>       reducer(&state, action).run(environment)
>     }
>   }
> }
> 
> let counterReducer = Reducer<CounterState, CounterAction, CounterEffect>.strict { state, action in
>   switch action {
>   case .incrementButtonTapped:
>     state.count += 1
>     return .none // no more `{ _ in .none }`!
>   case .decrementButtonTapped where state.count == 1:
>     return .init { environment in .fireAndForget { environment.beep() } }
>   case .decrementButtonTapped:
>     state.count -= 1
>     return .none
>   }
> }
> ```
> 
> But maintaining these two branches in the core repo comes at a cost. Do you introduce all the Effect helpers on EnvironmentalEffect so that for example .fireAndForget can get nested less above? I think our general recommendation is to not use the .strict reducer, and maybe we should even consider it on the chopping block by moving it over to be an example of another higher-order reducer that you can define yourself quite simply!

And so we ended up removing it to keep the core library focused. As more and more folks use the library we can refine the core surface area.

One thing that could fix the above issue is method key paths, which would unlock dynamic member lookup of methods on wrapper types, but till Swift supports this kind of thing we think we'll leave it to the community to come up with a nice design around strict reducers.

pteasima:
Hope its fine to keep the discussion going even though this is closed.

I just found out that Swift5.2 can take something that Swift5.1 couldn't:

```
@dynamicMemberLookup struct StrictEffect<Environment, Output, Error: Swift.Error> {
  let perform: (Environment) -> Effect<Output, Error>
  
  static var none: StrictEffect {
    Self { _ in .none }
  }
  static subscript(dynamicMember keyPath: KeyPath<Environment, () -> Effect<Output, Error>>) -> () -> StrictEffect {
    { Self { env in env[keyPath: keyPath]() } }
  }
  static subscript<Input>(dynamicMember keyPath: KeyPath<Environment, (Input) ->    Effect<Output, Error>>) -> (Input) -> StrictEffect {
    { a in Self { env in env[keyPath: keyPath](a) } }
  }
  
// TODO: needs more overrides for multiple params
// TODO: it still doesnt like keyPaths of lenght > 1, but afaik with even more overloads that lead to an intermediate type this could be doable
}
```

```
struct Environment {
  var getURL: (String) -> Effect<Data, URLError>
}

let x: StrictEffect<Data, URLError> = StrictEffect.getURL("www.pointfree.co") // now compiles
let x: StrictEffect<Data, URLError> = (StrictEffect.getURL)("www.pointfree.co") // this was needed before
```

All of a sudden, that `EffectWrapper`/`StrictEffect` type is much more compelling. Its usage can look like regular Swift function calls. 

We can still reach into the underlying effect using `map`. This is a little inconvenient, not sure if we should duplicate Combine's operator directly on StrictEffect instead.

```
extension StrictEffect {
  func map<NewOutput, NewError: Swift.Error>(transform: @escaping (Effect<Output, Error>) -> Effect<NewOutput, NewError>) -> StrictEffect<NewOutput, NewError> {
    .init { env in transform(self.perform(env)) }
  }
}
```

Pullback is imho also possible but cant find a usecase for it.

Heres the new `.strict`:

```
extension Reducer {
  static func strict(
    _ reducer: @escaping (inout State, Action) -> StrictEffect<Environment, Action, Never>
  ) -> Reducer {
    Self { state, action, environment in
      reducer(&state, action).perform(environment)
    }
  }
}
```
@oskarek I will be playing with this some more, can you try to use it and give feedback?

oskarek:
That's pretty cool! Yes, I will definitely play with it a bit, and I'll get back to you once I've done that!

pteasima:
Here's that intermediate type to allow longer KeyPaths btw:

```
@dynamicMemberLookup struct StrictEffectProxy<Environment, IntermediateEnvironment, Output, Error: Swift.Error> {
  let environmentTransform: KeyPath<Environment, IntermediateEnvironment>
  
  subscript(dynamicMember keyPath: KeyPath<IntermediateEnvironment, () -> Effect<Output, Error>>) -> () -> StrictEffect<Environment, Output, Error> {
  { StrictEffect { env in env[keyPath: self.environmentTransform][keyPath: keyPath]() } }
  }
  subscript<Input>(dynamicMember keyPath: KeyPath<IntermediateEnvironment, (Input) -> Effect<Output, Error>>) -> (Input) -> StrictEffect<Environment, Output, Error> {
  { a in StrictEffect { env in env[keyPath: self.environmentTransform][keyPath: keyPath](a) } }
  }
  
  subscript<NewEnvironment>(dynamicMember keyPath: KeyPath<IntermediateEnvironment, NewEnvironment>) -> StrictEffectProxy<Environment, NewEnvironment, Output, Error> {
    .init(environmentTransform: environmentTransform.appending(path: keyPath))
  }
}

extension StrictEffect {
  static subscript<NewEnvironment>(dynamicMember keyPath: KeyPath<Environment, NewEnvironment>) -> StrictEffectProxy<Environment, NewEnvironment, Output, Error> {
    StrictEffectProxy(environmentTransform: keyPath)
  }
}
```

Its a pain in the ass to implement but Im surprised by how well it works, no type inference issues.
So now not only can you do `StrictEffect.someEffect("someParam")` but also `StrictEffect.someService.someEffect("someParam")` and even `StrictEffect.someService.anotherNestedService.someEffect("someParam")`

oskarek:
@pteasima Okay, so I have played a bit with it now, and have some thoughts.

First of all, I think it's nice to realize that what you call `StrictEffect` is basically a specialized version of the `Reader` monad, so it might make sense to express it as a specialization of that. So first:

```swift
// The Reader monad
@dynamicMemberLookup struct Reader<Environment, Output> {
  let run: (Environment) -> Output
  func callAsFunction(_ env: Environment) -> Output { self.run(env) }
  
  static subscript(dynamicMember keyPath: KeyPath<Environment, () -> Output>) -> () -> Reader {
    { Self { env in env[keyPath: keyPath]() } }
  }
  static subscript<Input>(dynamicMember keyPath: KeyPath<Environment, (Input) -> Output>) -> (Input) -> Reader {
    { a in Self { env in env[keyPath: keyPath](a) } }
  }
  static subscript<I1, I2>(dynamicMember keyPath: KeyPath<Environment, (I1, I2) -> Output>) -> (I1, I2) -> Reader {
    { a, b in Self { env in env[keyPath: keyPath](a, b) } }
  }
}

// The proxy stuff
@dynamicMemberLookup struct ReaderProxy<Environment, IntermediateEnvironment, Output> {
  let environmentTransform: KeyPath<Environment, IntermediateEnvironment>
  
  subscript(dynamicMember keyPath: KeyPath<IntermediateEnvironment, () -> Output>) -> () -> Reader<Environment, Output> {
  { Reader { env in env[keyPath: self.environmentTransform][keyPath: keyPath]() } }
  }
  subscript<Input>(dynamicMember keyPath: KeyPath<IntermediateEnvironment, (Input) -> Output>) -> (Input) -> Reader<Environment, Output> {
  { a in Reader { env in env[keyPath: self.environmentTransform][keyPath: keyPath](a) } }
  }
  
  subscript<NewEnvironment>(dynamicMember keyPath: KeyPath<IntermediateEnvironment, NewEnvironment>) -> ReaderProxy<Environment, NewEnvironment, Output> {
    .init(environmentTransform: self.environmentTransform.appending(path: keyPath))
  }
}

extension Reader {
  static subscript<NewEnvironment>(dynamicMember keyPath: KeyPath<Environment, NewEnvironment>) -> ReaderProxy<Environment, NewEnvironment, Output> {
    ReaderProxy(environmentTransform: keyPath)
  }
}
```

Being a monad, it of course has the functions:

```swift
extension Reader {
  func map<T>(_ f: @escaping (Output) -> T) -> Reader<Environment, T> {
    .init { f(self($0)) }
  }
  static func zip<T1, T2>(_ lhs: Reader<Environment, T1>, _ rhs: Reader<Environment, T2>) -> Reader<Environment, (T1, T2)> {
    .init { (lhs($0), rhs($0)) }
  }
  func flatMap<T>(_ f: @escaping (Output) -> Reader<Environment, T>) -> Reader<Environment, T> {
    .init { f(self($0))($0) }
  }
}
```
(don't know if any other than `map` is actually useful for our purposes of using it for effects)

And you can introduce the `.none` convenience static property like:

```swift
// don't know about the name though haha
protocol Noneable {
  static var none: Self { get }
}
extension Effect: Noneable {}

extension Reader: Noneable where Output: Noneable {
  static var none: Reader { Self { _ in .none } }
}
```

And then just:

```swift
typealias StrictEffect<Environment, Output, Error: Swift.Error> =
  Reader<Environment, Effect<Output, Error>>
```
if you want to name it something different. But, speaking of the name: the _type itself_ is not _strict_ in some special way, we just happen to be using it in a way that makes the _reducer_ more strict. So I'm not sure about the name "StrictEffect" 🤔 Personally, I think just using `Reader<Environment, Effect<Output, Error>>` as it is, is pretty nice and clean. But maybe it's not a great API for people unfamiliar with Reader.

Anyway, it's really cool how you're using the `@dynamicMemberLookup` feature, feels like magic 😄
However, I do have some concerns/questions about the usefulness of this type as return type from reducers:

1. Autocompletion doesn't work (at least not for me on my computer?). So typing `StrictEffect` (or whatever name you're choosing) and then hitting `.` you don't see the members of the Environment type. I think that might make usage of the type a little confusing

2. You will still pretty much always have to use the `map` method, because you will want to do some transformation on the Effect (like map on it, to turn it into an action). And as soon as you have to reach for the  `map` function, I think the ergonomics are not any better than the much simpler solution that you and I are both currently using in our projects (with the return type of the reducer being an optional function). Compare:

```swift
return { environment in
  environment.loadSomethingFromAPI(state.key)
    .map { MyAction.someAction($0) }
}
```
with
```swift
return StrictEffect.loadSomethingFromAPI(state.key)
  .map { effect in
    effect.map { MyAction.someAction($0) }
  }
}
```

As you mentioned, this could maybe be solved by duplicating Combine's operators directly on StrictEffect or something like that. I don't know how well that could work, but it might be worth exploring

3. How do you do when you need some information from the environment in order to do a transformation on the effect? So for example:

```swift
environment.loadSomethingFromAPI(state.key)
  .map { MyAction.someAction($0) }
```

```swift
StrictEffect.loadSomethingFromAPI(state.key)
  .map { effect in
    effect
      .map { MyAction.someAction($0) }
      .receive(on: ???)
  }
}
```

In this case, you need access to the actual environment when doing the transformation.

Of course, in both the cases from concern 2 and 3, one could use the init directly, like:
```swift
StrictEffect { environment in
...
}
```
but that's not any better than the simpler solution with an optional function as return type from the reducer.


Wow, this became a very long message, sorry about that 😅

Anyway, it would be very cool if it was possible to make this work in a nice way!

pteasima:
Thanks @oskarek , all very good thoughts.

 - `StrictEffect` was a dumb working name, I should have done better. Stephen's `EnvironmentalEffect` is the only one making any sense, but lets also keep in mind that you have to type it out every-time (cant use dynamicMemberLookup with just a leading `.` atm). I wonder if something like `Reducer.Eff` is an option. Afaik client-side code could just say `Eff` since we're already in a `Reducer`.
 - You point about `Reader` is a good one, but Im all for getting newbies excited about FP, not scaring them away 😀 So lets make sure its at least properly hidden behind a `typealias`
 - using the raw `.init {...` in some special cases is to be expected, cant have convenience methods/subscripts for everything.
 - Yes, autocompletion and the constant need for a `.map` to reach in for the underlying `Effect` are the **biggest problems** of this approach and I **dont think they're solvable 😢**. I will keep using it for a bit but might go back to the Optional closure approach.

PR #25, opened 2020-05-06
Assorted fixes for collections of state

### LINKED ISSUE #21: Double Nested ForEachStore removal Bug
I have bug that I'm not sure if it's my implementation or CA. In the original app I found it I have a list of textfields that can have a carousel of images attached to them, as the user adds images they are added to the carousel. But they can also be removed, when removing the last item from the carousel it seems like there is an action that is still in the system looking for a now missing object in a `ForEachStore` Below is a minimum reproduction repo. 

https://github.com/ryantstone/LoopTest

stephencelis:
Fixes #21 and a few other things, which I will review.

### DIFF
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -194,6 +194,17 @@ public struct Reducer<State, Action, Environment> {
       guard let (index, localAction) = toLocalAction.extract(from: globalAction) else {
         return .none
       }
+      // NB: This does not need to be a fatal error because of the index subscript that follows it.
+      assert(
+        index < globalState[keyPath: toLocalState].endIndex,
+        """
+        Index out of range. This can happen when a reducer that can remove the last element from \
+        an array is then combined with a "forEach" from that array. To avoid this and other \
+        index-related gotchas, consider using an "IdentifiedArray" of state instead. Or, combine \
+        your reducers so that the "forEach" comes before any reducer that can remove elements from \
+        its array.
+        """
+      )
       return self.reducer(
         &globalState[keyPath: toLocalState][index],
         localAction,
@@ -238,7 +249,7 @@ public struct Reducer<State, Action, Environment> {
       guard let (id, localAction) = toLocalAction.extract(from: globalAction) else { return .none }
       return self.optional
         .reducer(
-          &globalState[keyPath: toLocalState][id],
+          &globalState[keyPath: toLocalState][id: id],
           localAction,
           toLocalEnvironment(globalEnvironment)
         )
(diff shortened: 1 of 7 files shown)


Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -194,6 +194,17 @@ public struct Reducer<State, Action, Environment> {
...
+      assert(
+        index < globalState[keyPath: toLocalState].endIndex,
+        """
+        Index out of range. This can happen when a reducer that can remove the last element from \
+        an array is then combined with a "forEach" from that array. To avoid this and other \
+        index-related gotchas, consider using an "IdentifiedArray" of state instead. Or, combine \
+        your reducers so that the "forEach" comes before any reducer that can remove elements from \
+        its array.
+        """
+      )
```
stephencelis, line 207: This error message should #21. We can finesse the wording if any of it could be improved.

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -238,7 +249,7 @@ public struct Reducer<State, Action, Environment> {
       guard let (id, localAction) = toLocalAction.extract(from: globalAction) else { return .none }
       return self.optional
         .reducer(
-          &globalState[keyPath: toLocalState][id],
+          &globalState[keyPath: toLocalState][id: id],
```
stephencelis, line 252: I've updated the interface to take an explicit `id` parameter to avoid any ambiguity for `Int`-based ids.

Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
```diff
@@ -92,12 +92,15 @@ where ID: Hashable {
...
+  public subscript(id id: ID) -> Element? {
     get {
       self.dictionary[id]
     }
     _modify {
       yield &self.dictionary[id]
+      if self.dictionary[id] == nil {
+        self.ids.removeAll(where: { $0 == id })
+      }
     }
```
stephencelis, line 104: We weren't properly cleaning up the `ids` array when `nil`-ing out elements.

Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
```diff
@@ -115,6 +118,7 @@ where ID: Hashable {
...
   }
 
+  @discardableResult
```
stephencelis, line 121: Stumbled upon this warning when writing tests.

Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
```diff
@@ -138,7 +142,7 @@ where ID: Hashable {
   }
 
   public mutating func remove(atOffsets offsets: IndexSet) {
-    for offset in offsets {
+    for offset in offsets.reversed() {
```
stephencelis, line 145: Tests caught a bug here, as well!

PR #26, opened 2020-05-06
Update CI to use Xcode 11.4.1

stephencelis:
Makes local development a bit friendlier, too.


stephencelis:
Gonna merge since we need to update the repo status checks alongside this name change.

PR #31, opened 2020-05-07
Fix Xcode previews crash

stephencelis:
#25 inadvertently broke Xcode previews, which would crash during compilation due to a `_modify` block on `IdentifiedArray` that yielded _and_ accessed the value that it just yielded.

We can explore different avenues for optimizing `IdentifiedArray` in the future.

### DIFF
--- a/Examples/TicTacToe/Sources/Core/GameCore.swift
+++ b/Examples/TicTacToe/Sources/Core/GameCore.swift
@@ -87,14 +87,15 @@ extension Array where Element == [Player?] {
 
   func hasWin(_ player: Player) -> Bool {
     let winConditions = [
-      [0,1,2], [3,4,5], [6,7,8],
-      [0,3,6], [1,4,7], [2,5,8],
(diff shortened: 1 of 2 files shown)


Examples/TicTacToe/Sources/Core/GameCore.swift
```diff
@@ -87,14 +87,15 @@ extension Array where Element == [Player?] {
 
   func hasWin(_ player: Player) -> Bool {
     let winConditions = [
-      [0,1,2], [3,4,5], [6,7,8],
-      [0,3,6], [1,4,7], [2,5,8],
-      [0,4,8], [6,4,2]
+      [0, 1, 2], [3, 4, 5], [6, 7, 8],
+      [0, 3, 6], [1, 4, 7], [2, 5, 8],
+      [0, 4, 8], [6, 4, 2],
```
stephencelis, line 92: Still don't have the format GitHub action quite working as I'd expect...

Issue #24, opened 2020-05-06
Binding textfield issue? 


donmcallister:
It seems binding doesn’t work with some of the other SwiftUI TextField overloads that use Value and formatting parameters.

stephencelis:
Hey @donmcallister! Thanks for the report.

Can you flesh this issue about a bit more? In particular, can you provide some code that shows the issue, and can you explain what you expect to happen vs. what actually happens?

It can also help to try the same thing in a vanilla SwiftUI app to see if this is an issue with the library or just how SwiftUI behaves.

donmcallister:
```
/*
 This doesn't work for me, no formatting, not binding and storing the value. I switch to another screen and when come back to this screen, the value entered doesn't store: 
                         
                         TextField(<#T##title: StringProtocol##StringProtocol#>, value: <#T##Binding<T>#>, formatter: <#T##Formatter#>) 
*/
                        
                        TextField(
                            "Enter Price",
                            value: viewStore.binding(get: \.purchasePrice, send: SharedStateAction.FundingAction.purchasePriceChange),
                            formatter: formatter
                        )
                            .keyboardType(.numberPad)
                        
// This does work, but don't have a way to use the formatter. And I'm probably overlooking something, but so far can't figure out how to get a currency format as the user enters in numbers into text field:
                        TextField(
                            "Enter Price",
                            text: viewStore.binding(get: \.purchasePrice, send: SharedStateAction.FundingAction.purchasePriceChange))
                            .keyboardType(.numberPad)`
```

my reducer:

```
let sharedStateFundingReducer = Reducer<SharedState.CalculatorState, SharedStateAction.FundingAction, Void> { state, action, _ in
    switch action {
        
    case let .purchasePriceChange(price):
        state.purchasePrice = price
        return .none
    case let .arvPriceChange(arvPrice):
        state.afterRepairValue = arvPrice
        return .none
    }
}
```

stephencelis:
@donmcallister Can you paste in a full example or link to a project we can clone? That way we'll have something to play with or PR back to you. Formatter-based bindings only call when editing finishes (unlike every key stroke, like the non-formatted-based text field), so it works a bit differently. We could maybe add a case study for this soon.

donmcallister:
[...]

PR #32, opened 2020-05-07
Formatted text field binding case study

stephencelis:
Adding a quick example of using the `TextField.init(_:value:formatter:)` API for reference. Inspired by #24.


stephencelis:
Chatted about this with @mbrandonw and we're going to close this PR unmerged in the interest of keeping the case study focused on bindings and not surface some of the quirks of `TextField`'s formatter-based API.

PR #33, opened 2020-05-07
Only run swift-format when PRs are merged

stephencelis:
From https://github.community/t5/GitHub-Actions/Run-action-only-when-a-PR-is-merged/m-p/56155/highlight/true#M9769

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,13 +1,16 @@
 name: Format
 
 on:
-  push:
-    branch: 'master'
+  pull_request:
+    types: [closed]
+  branches:
+    - master
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/format.yml
```diff
@@ -1,13 +1,16 @@
 name: Format
 
 on:
-  push:
-    branch: 'master'
```
stephencelis, line 5: Maybe this `branch` was the issue. Looks like it should be `branches`... 🤔

mbrandonw: lol, should we start using [dhall](https://github.com/vmchale/github-actions-dhall)? 😬

Issue #43, opened 2020-05-08
API for constructing stateless ViewStore


LK:
I find that I occasionally need to use `WithViewStore` only for the purpose of being able to send actions, without ever accessing any state on the view store directly (generally this happens when I introduce some more local store somewhere down the tree, with a `ForEachStore` or an `IfLetStore`). For example, if I'm rendering a table section with some list elements, I might do something like this:

```swift
WithViewStore(self.store) { viewStore in
    Section(header: Text("Header Text")) {
    ForEachStore(self.store.scope(state: \.books.elements, action: ReadingListAction.rowInteraction)) { bookStore in
            WithViewStore(bookStore) { bookViewStore in
                BookRow(title: bookViewStore.title)
                    .onAppear { viewStore.send(.rowAppeared(bookViewStore.id)) }
                    .onDisappear { viewStore.send(.rowDisappeared(bookViewStore.id)) }
            }
        }
    }
}
```

In this case, I want the `rowAppeared` and `rowDisappeared` actions to be sent to the top-level reducer, so I need the global `viewStore`, however I never need to touch any of its state directly.

This works fine, however I'm now re-rendering the entire section every time any of my rows are updated, which is unnecessary. What I actually want is to have the top-level `viewStore` ignore any state changes and just provide an outlet to send actions.

In these cases, it seems like the correct thing to do is to `scope` the store down to an empty state struct. Doing so is a bit cumbersome, so I've been using a custom helper:

```swift
public extension Store {
    struct Stateless: Equatable {}
    func stateless() -> Store<Stateless, Action> {
        return self.scope(state: { _ in Stateless() })
    }
}
```

There's a handful of different ways this could be expressed in the API (a `WithStatelessViewStore` helper might be better, because I see no other purpose for a stateless store other than to feed a `ViewStore`). Maybe the problem I'm trying to solve is a code smell and I should just refactor my actions or views to avoid the need for this altogether; either way, curious to hear thoughts.

Odrakir:
`WithViewStore` can take a closure to remove duplicates, you could maybe make it so it always return true? Even creating a `WithStatelessViewStore` wrapper that does this for you. 

This is a hacky solution, so don’t take it too seriously, I just wanted to share my two cents.

stephencelis:
@LK Thanks for starting this discussion! This is definitely a common enough problem that it would be nice to have a dedicated helper to solve it. We just haven't quite figured out the name that feels "right" yet.

[...]

PR #45, opened 2020-05-08
Helpers for making stores stateless/actionless

stephencelis:
To address #43.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -179,6 +179,15 @@ public final class Store<State, Action> {
     }
   }
 
+  public var stateless: Store<Void, Action> {
+    self.scope(state: { _ in () })
+  }
+
+  public var actionless: Store<State, Never> {
+    func absurd<A>(_ never: Never) -> A {}
+    return self.scope(state: { $0 }, action: absurd)
+  }
+
   private init(
     initialState: State,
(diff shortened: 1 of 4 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -179,6 +179,15 @@ public final class Store<State, Action> {
     }
   }
 
+  public var stateless: Store<Void, Action> {
```
stephencelis, line 182: Todo:

- [ ] Document

Sources/ComposableArchitecture/Store.swift
```diff
@@ -179,6 +179,15 @@ public final class Store<State, Action> {
...
+    self.scope(state: { _ in () })
+  }
+
+  public var actionless: Store<State, Never> {
```
stephencelis, line 186: Todo:

- [ ] Document

Sources/ComposableArchitecture/SwiftUI/WithViewStore.swift
```diff
@@ -76,6 +76,21 @@ extension WithViewStore where State: Equatable {
   }
 }
 
+extension WithViewStore where State == Void {
```
stephencelis, line 79: Helper for viewing `Void` stores.

stephencelis:
Gonna merge as is to get a feel for things. @LK mentioned naming still not feeling quite right but maybe it'll feel good with use, and if not we can keep workshopping!

PR #46, opened 2020-05-08
Allow motion manager to be demo'd in SwiftUI preview

mbrandonw:
One bummer about the MotionManager demo is that it can only be run on a device since the `Motion` APIs are not available to simulators or SwiftUI previews.

However, we can easily create a mock motion client that emits whatever motion data we want. So, I've updated the SwiftUI preview to output sinusoidal motion data on a timer so that you can demo the app as if it was actually running on a device.

Here's a demo of what it looks like in Xcode:

May-08-2020 09-23-32

Should we do the same for the app running in the simulator too? 🤔


Examples/MotionManager/MotionManager.xcodeproj/project.pbxproj
```diff
@@ -10,8 +10,6 @@
 		5A405A2E244F410F002D4C76 /* SceneDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5A405A2D244F410F002D4C76 /* SceneDelegate.swift */; };
 		5A405A30244F410F002D4C76 /* MotionManagerView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5A405A2F244F410F002D4C76 /* MotionManagerView.swift */; };
 		5A405A32244F4110002D4C76 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 5A405A31244F4110002D4C76 /* Assets.xcassets */; };
-		5A405A35244F4110002D4C76 /* Preview Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 5A405A34244F4110002D4C76 /* Preview Assets.xcassets */; };
-		5A405A38244F4110002D4C76 /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 5A405A36244F4110002D4C76 /* LaunchScreen.storyboard */; };
```
mbrandonw, line 14: just deleting things that don't need to be in the project

Examples/MotionManager/MotionManager/MotionManagerView.swift
```diff
@@ -2,6 +2,18 @@ import ComposableArchitecture
 import CoreMotion
 import SwiftUI
 
+private let readMe = """
+This demonstrates how to work with the MotionManager API from Apple's Motion framework.
```
mbrandonw, line 6: this project didn't have a read me, so adding one, and making it's focus on the mockability of dependencies in TCA

Examples/MotionManager/MotionManager/MotionManagerView.swift
```diff
@@ -63,11 +75,13 @@ struct AppView: View {
...
+        Text(readMe)
+          .multilineTextAlignment(.leading)
+          .layoutPriority(1)
+
         Spacer(minLength: 100)
 
-        ZStack {
-          plot(buffer: viewStore.z, scale: 40)
-        }
+        plot(buffer: viewStore.z, scale: 40)
```
mbrandonw, line 84: surrounding `ZStack` didn't seem needed

Examples/MotionManager/MotionManager/MotionManagerView.swift
```diff
@@ -117,15 +131,37 @@ func plot(buffer: [Double], scale: Double) -> Path {
 
 struct AppView_Previews: PreviewProvider {
   static var previews: some View {
-    AppView(
+    // Since MotionManager isn't usable in SwiftUI previews or simulators we create one that just
+    // sends a bunch of data on some sine curves.
+    var isStarted = false
+    let mockMotionClient = MotionClient(
```
mbrandonw, line 137: just a really simple mock motion client that sets up a timer to emit values, and does a simple `.filter` in order to control if it sends value out or not (so that we can start and stop it from the UI).

PR #52, opened 2020-05-08
Demos Xcode 11.3.1 support

stephencelis:
Makes sense to give as many folks an opportunity to run the examples as possible.


stephencelis:
11.3 test runs fails occasionally due to what appears to be a bug in SPM (package targets should not depend on each other internally, like ComposableArchitectureTestSupport).

We could merely build the projects on 11.3 (and not run tests) to avoid failing the build, but it looks like moving ComposableArchitectureTestSupport to its own package/repo might be the better workaround, especially if #57 is a related problem.

Having 2 repos isn't an ideal user experience, but it may be out of our hands till SPM can support this use case itself.

PR #71, opened 2020-05-11
Merge Test Support Module

### LINKED ISSUE #70: Merge the core library with the test support library
(NB: This issue subsumes https://github.com/pointfreeco/swift-composable-architecture/issues/57 and https://github.com/pointfreeco/swift-composable-architecture/issues/60)

Due to an SPM/Xcode bug we cannot have separate `ComposableArchitecture` and `ComposableArchitectureTestSupport` libraries. The bug is that SPM packages cannot vend two libraries where one depends on the other. The bug manifests itself by simply not compiling (if you split your app into multiple modules), or crashing the test suite (this happens most often on iOS 11.3, as seen in #52).

We've tried our hardest to work around this bug, but each thing we do seems to just cause another problem.

So, until the bug is fixed we have two options:

* Split the test support library into own repository 
* Combine the test support code into the core library

The first option is a non-starter for us because it would be quite an extra burden for us to maintain two libraries that need parallel releases, and it would complicate how we structure the example apps in this repo.

The second option seems possible, but we are not sure how to accomplish it. The problem is that we can't put any `XCTest` code inside `ComposableArchitecture` since the library is meant to be deployed to devices. Fortunately, we are only using one single symbol from `XCTest`, and that is `XCTFail`. We've spent a few hours trying to figure out if it's possible to dynamically invoke `XCTFail` without importing `XCTest` or mentioning the symbol, but everything we've tried has come up short (including using `dlopen` and `dlsym` 😬).

Does anyone have any ideas on a path forward?

stephencelis:
An attempt to fix #70.


ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,16 +0,0 @@
-{
```
stephencelis, line 1: Deleting an old file.

Examples/CaseStudies/CaseStudies.xcodeproj/project.pbxproj
```diff
@@ -24,6 +24,8 @@
 		CAF7FD82245C83B40081E4C3 /* IfLetSubscript.swift in Sources */ = {isa = PBXBuildFile; fileRef = CAF7FD81245C83B40081E4C3 /* IfLetSubscript.swift */; };
 		DC07231724465D1E003A8B65 /* 02-Effects-TimersTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DC07231624465D1E003A8B65 /* 02-Effects-TimersTests.swift */; };
 		DC072322244663B1003A8B65 /* 03-Navigation-Sheet-LoadThenPresent.swift in Sources */ = {isa = PBXBuildFile; fileRef = DC072321244663B1003A8B65 /* 03-Navigation-Sheet-LoadThenPresent.swift */; };
+		DC13940E2469E25C00EE1157 /* ComposableArchitecture in Frameworks */ = {isa = PBXBuildFile; productRef = DC13940D2469E25C00EE1157 /* ComposableArchitecture */; };
```
stephencelis, line 27: A bit of churn in every Xcode project since we have to remove the old, dynamic SPM library and use the new static one.

Makefile
```diff
@@ -59,6 +59,6 @@ test-workspace:
 		-quiet
 
 format:
-	swift format --in-place --recursive .
+	swift format --ignore-unparsable-files --in-place --recursive .
```
stephencelis, line 62: Swift-format trips up on `@convention(c)`.

Sources/ComposableArchitecture/TestSupport/TestStore.swift
```diff
@@ -414,3 +412,24 @@ extension TestStore {
...
+      .map({ unsafeBitCast($0, to: XCTFailureHandler.self) })
+    else { return }
+
+  guard
+    let _XCTCurrentTestCase = dlsym(_XCTest, "_XCTCurrentTestCase")
+      .map({ unsafeBitCast($0, to: XCTCurrentTestCase.self) })
+    else { return }
+
+  _XCTFailureHandler(_XCTCurrentTestCase(), true, "\(file)", line, message, nil)
+}
```
stephencelis, line 435: This is the dark magic that allows us to hook into XCTest dynamically.

Sources/ComposableArchitecture/TestSupport/XCTestShims.swift
```diff
@@ -0,0 +1,3 @@
+typealias XCTCurrentTestCase = @convention(c) () -> AnyObject
+typealias XCTFailureHandler
+  = @convention(c) (AnyObject, Bool, UnsafePointer<CChar>, UInt, String, String?) -> Void
```
stephencelis, line 3: Moved these into their own file since swift-format breaks on them and we still would like TestStore.swift to be formatted.

Sources/ComposableArchitectureTestSupport/Reducer.swift
```diff
@@ -1,24 +0,0 @@
...
-extension Reducer where Environment == Void {
-  public func callAsFunction(
-    _ state: State,
-    _ action: Action
-  ) -> State {
-    var state = state
-    _ = self.callAsFunction(&state, action)
-    return state
-  }
-}
```
stephencelis, line 24: Deleted these convenience functions...it seems rare to test a reducer by calling it directly so it seems like surface area we can trim.

Sources/ComposableArchitecture/TestSupport/XCTestShims.swift
stephencelis: Actually turns out that swift-format tripped on them only when they were referenced inline `(@convention(c) (...) -> ...).self`.

They format fine as type aliases!

Sources/ComposableArchitecture/TestSupport/TestStore.swift
```diff
@@ -414,3 +412,33 @@ extension TestStore {
...
+      .flatMap({ $0.executablePath })
+      .flatMap({ dlopen($0, RTLD_NOW) }),
+    let _XCTFailureHandler = dlsym(XCTest, "_XCTFailureHandler")
+      .map({ unsafeBitCast($0, to: XCTFailureHandler.self) }),
+    let _XCTCurrentTestCase = dlsym(XCTest, "_XCTCurrentTestCase")
+      .map({ unsafeBitCast($0, to: XCTCurrentTestCase.self) })
+  else {
+    assertionFailure(
+      """
+      Couldn't load XCTest. Are you driving a test store in application code?"
```
mbrandonw, line 435:
```suggestion
      Couldn't load XCTest. Are you using a test store in application code?"
```

Issue #69, opened 2020-05-11
Combine enum case reducers


slashmo:
In my project, I have a `Content` enum with the cases `.group` and `.request`. I also have both `GroupAction` and `RequestAction`. Here is the `contentReducer` I came up with:

``` swift
let contentReducer = Reducer<RequestCollection.Content, ContentAction, Void> { state, action, _ in
  switch state {
  case var .group(group):
    guard case let .group(groupAction) = action else { return .none }
    switch groupAction {
    case let .nameChanged(name):
      group.name = name
      state = .group(group)
    }
  case var .request(request):
    guard case let .request(requestAction) = action else { return .none }
    switch requestAction {
    case let .nameChanged(name):
      request.name = name
      state = .request(request)
    }
  }
  return .none
}
```

What I don't like about this is that I have to `guard case let` on the `ContentAction` where a `.group` case should only be possible to be sent to a `Group`. In an ideal scenario I'd like to compose this reducer from both `groupReducer` and `requestReducer`.
I'll try and prototype a higher order reducer for that purpose.

Did someone else already encounter something similar?

slashmo:
I could "reduce" the `contentReducer` by calling out to `groupReducer` and `requestReducer` manually:

``` swift
let contentReducer = Reducer<RequestCollection.Content, ContentAction, Void> { state, action, _ in
  switch state {
  case var .group(group):
    guard case let .group(groupAction) = action else { return .none }
    let effect = groupReducer(&group, groupAction).map(ContentAction.group)
    state = .group(group)
    return effect
  case var .request(request):
    guard case let .request(requestAction) = action else { return .none }
    let effect = requestReducer(&request, requestAction).map(ContentAction.request)
    state = .request(request)
    return effect
  }
}
```

boudavid:
If I understood correctly, your implementation of actions looks like this :

```swift
enum ContentAction {
  case group(GroupAction)
  case request(RequestAction)
}

enum GroupAction {
  case nameChanged(String)
}

enum RequestAction {
  case nameChanged(String)
}
```

Is it correct ? Can you also explain why the `switch` is on the `state` and not the `action` ? I'm just trying to understand what you want to achieve. 😃 

> In an ideal scenario I'd like to compose this reducer from both groupReducer and requestReducer.

Maybe the `Reducer.combine` method is what you're looking for ? Typically, it's used like this :

```swift
let someReducer = Reducer.combine(
[...]

PR #82, opened 2020-05-12
Better support enum state and struct actions with optional paths

stephencelis:
Pushing this forward from discussion in #69.

This PR introduces something we've been experimenting with in order to better support enum state (and struct actions). Alongside `CasePath` we need a new "path," which we call `OptionalPath`. It is a construct for optionally getting and always setting a value on some root. It is modeled after optional-chained property setters, where the getter is optional

```swift
user?.location.city // Optional("Brooklyn") // ✅
```

And the setter is non-optional:

```swift
user?.location.city = "Los Angeles" // ✅
user?.location.city = nil // 🛑 error: 'nil' cannot be assigned to type 'String'
```

We first explored this type in the exercises of our episode on case path properties here: https://www.pointfree.co/episodes/ep88-the-case-for-case-paths-properties#exercises

It is precisely the construct we need to compose reducers along enum state (which includes optional state) and struct actions.

All writable key paths can be promoted to optional paths:

```swift
OptionalPath(\GlobalState.localVar)
```

As can all case paths:

```swift
OptionalPath(/GlobalState.localCase)
```

And we've introduced a protocol to ensure that `pullback` can take a plain ole key path or case path, which means existing `pullback`s should still build, though please let us know if you encounter issues! If we encounter issues with this protocol we may want to try overloads instead.

We probably won't merge this branch for a bit to allow for any weird compiler issues, bugs, or ergonomic improvements to show themselves.


stephencelis:
Gonna make this a draft for now and will close in a bit since we don't intend to merge as-is. If you want to try this out in your own project, you should be able to copy and paste the following into a file:

<details>
<summary>Click to expand</summary>
```swift
public protocol WritablePath {
  associatedtype Root
  associatedtype Value
  func extract(from root: Root) -> Value?
  func set(into root: inout Root, _ value: Value)
}

extension WritableKeyPath: WritablePath {
  public func extract(from root: Root) -> Value? {
    root[keyPath: self]
  }

  public func set(into root: inout Root, _ value: Value) {
    root[keyPath: self] = value
  }
}

extension CasePath: WritablePath {
  public func set(into root: inout Root, _ value: Value) {
    root = self.embed(value)
  }
}

extension OptionalPath: WritablePath {}

public struct OptionalPath<Root, Value> {
  private let _extract: (Root) -> Value?
  private let _set: (inout Root, Value) -> Void

  public init(
    extract: @escaping (Root) -> Value?,
    set: @escaping (inout Root, Value) -> Void
  ) {
    self._extract = extract
    self._set = set
  }

  public func extract(from root: Root) -> Value? {
    self._extract(root)
  }

  public func set(into root: inout Root, _ value: Value) {
    self._set(&root, value)
  }

  public init(
    _ keyPath: WritableKeyPath<Root, Value?>
  ) {
    self.init(
      extract: { $0[keyPath: keyPath] },
      set: { $0[keyPath: keyPath] = $1 }
    )
  }

  public init(
    _ casePath: CasePath<Root, Value>
  ) {
    self.init(
      extract: casePath.extract(from:),
      set: { $0 = casePath.embed($1) }
    )
  }

  public func appending<AppendedValue>(
    path: OptionalPath<Value, AppendedValue>
  ) -> OptionalPath<Root, AppendedValue> {

    .init(
      extract: { self.extract(from: $0).flatMap(path.extract(from:)) },
      set: { root, appendedValue in
        guard var value = self.extract(from: root) else { return }
        path.set(into: &value, appendedValue)
        self.set(into: &root, value)
      }
    )
  }

  public func appending<AppendedValue>(
    path: CasePath<Value, AppendedValue>
  ) -> OptionalPath<Root, AppendedValue> {

    self.appending(path: .init(path))
  }

  public func appending<AppendedValue>(
    path: WritableKeyPath<Value, AppendedValue>
  ) -> OptionalPath<Root, AppendedValue> {

    .init(
      extract: { self.extract(from: $0).map { $0[keyPath: path] } },
      set: { root, appendedValue in
        guard var value = self.extract(from: root) else { return }
        value[keyPath: path] = appendedValue
        self.set(into: &root, value)
      }
    )
  }

  // TODO: Is it safe to keep this overload?
  public func appending<AppendedValue>(
    path: WritableKeyPath<Value, AppendedValue?>
  ) -> OptionalPath<Root, AppendedValue> {

    self.appending(path: .init(path))
  }
}

extension CasePath {
  public func appending<AppendedValue>(
    path: OptionalPath<Value, AppendedValue>
  ) -> OptionalPath<Root, AppendedValue> {

    OptionalPath(self).appending(path: path)
  }

  public func appending<AppendedValue>(
    path: WritableKeyPath<Value, AppendedValue>
  ) -> OptionalPath<Root, AppendedValue> {

    OptionalPath(self).appending(path: path)
  }

  // TODO: Is it safe to keep this overload?
  public func appending<AppendedValue>(
    path: WritableKeyPath<Value, AppendedValue?>
  ) -> OptionalPath<Root, AppendedValue> {

    OptionalPath(self).appending(path: path)
  }
}

extension WritableKeyPath {
  public func appending<AppendedValue>(
    path: OptionalPath<Value, AppendedValue>
  ) -> OptionalPath<Root, AppendedValue> {

    OptionalPath(
      extract: { path.extract(from: $0[keyPath: self]) },
      set: { root, appendedValue in path.set(into: &root[keyPath: self], appendedValue) }
    )
  }

  public func appending<AppendedValue>(
    path: CasePath<Value, AppendedValue>
  ) -> OptionalPath<Root, AppendedValue> {

    self.appending(path: .init(path))
  }
}

extension OptionalPath where Root == Value {
  static var `self`: OptionalPath {
    .init(.self)
  }
}

extension OptionalPath where Root == Value? {
  static var some: OptionalPath {
    .init(/Optional.some)
  }
}

extension Reducer {
  public func pullback<GlobalState, GlobalAction, GlobalEnvironment, StatePath, ActionPath>(
    state toLocalState: StatePath,
    action toLocalAction: ActionPath,
    environment toLocalEnvironment: @escaping (GlobalEnvironment) -> Environment
  ) -> Reducer<GlobalState, GlobalAction, GlobalEnvironment>
  where
    StatePath: WritablePath, StatePath.Root == GlobalState, StatePath.Value == State,
    ActionPath: WritablePath, ActionPath.Root == GlobalAction, ActionPath.Value == Action
  {

    return .init { globalState, globalAction, globalEnvironment in
      guard
        var localState = toLocalState.extract(from: globalState),
        let localAction = toLocalAction.extract(from: globalAction)
      else { return .none }
      let effect =
        self
        .reducer(&localState, localAction, toLocalEnvironment(globalEnvironment))
        .map { localAction -> GlobalAction in
          var globalAction = globalAction
          toLocalAction.set(into: &globalAction, localAction)
          return globalAction
        }
      toLocalState.set(into: &globalState, localState)
      return effect
    }
  }
}
```
</details>

slashmo:
Thanks for this @stephencelis 👍
I like the idea of abstracting the path component. In my current project I was already thinking about how one might pullback along nested nodes in a tree structure. Maybe this change could even allow me to cook up something like a `TreePath`.

stephencelis:
Gonna close this draft for now, but folks should feel free to discuss here (or on the forums) if they give things a spin!

AlexisQapa:
@stephencelis is there a reason this has been abandoned ?

While I can do:
`IfLetStore(store.scope(state: \.modalState).scope(state: /State.ModalState.someCase), then: SomeCaseView.init)`
without this PR, this isn't perfect and can't find a way to pull back reducers along a combination of keypath and casepath.

stephencelis:
@AlexisQapa We definitely haven't abandoned it, and have even pasted the code into some projects we're working with to get some more real world experience with it. Any chance you can do the same in your project? More usage would definitely help us figure out any issues before bringing the feature into the main library.

AlexisQapa:
That's my plan thanks ! 
On a side note, I've been using free functions like 
`func both<Whole, A, B>(_ a: @escaping (Whole) -> A?, _ b: @escaping (Whole) -> B?) -> (Whole) -> (A, B)?`
to merge states and things like that.
I wonder if those could be somehow added to this path handling

nmccann:
This is definitely something I'd like to see in the architecture. I'm currently struggling with a scenario that would be best represented with an enum-based state, essentially:

```
- Logged In
--- Initial
--- Province Selected
--- Province And City Selected
--- Province And City And Street Selected
----- Something that requires Province/City/Street
- Logged Out
```

Without the ability to represent these as mutually exclusive states, I'm having to have a bunch of optionals, but then have to deal with situations that shouldn't be possible (ex. if you clear the city, it should also clear the street and anything that relied upon it). These can be worked around, but increase the surface area for tests - if they could be represented as mutually exclusive then those situations wouldn't be possible, and would thus not need to be tested.

I'll consider pasting the relevant code into our project to try it out in a real world scenario, but I'm a bit hesitant to do so as it might put us in a situation where we're no longer able to easily update to future versions of the architecture.

stephencelis:
@nmccann Totally understandable. My hope is that you wouldn't get into this kind of situation, since the code is entirely additive. Once we come to a happy place where we can incorporate this functionality into TCA itself, you'll ideally be able to simply remove the code you paste in.

nmccann:
@stephencelis I've been experimenting with this in isolation [here](https://github.com/nmccann/ComposableArchitectureExperiments), and must say that I do like it. Some notes:

* I might be missing something as I wasn't able to implement it purely additively - the changes to `Reducer.swift` couldn't be implemented outside of the library (even when providing different names for otherwise conflicting properties/functions) as they make use of `reducer` which is private. I opted to change my SPM dependency to target this branch instead, but that also again means that we'd potentially be stuck on an out of date branch (unless it's periodically rebased on the latest).
* It's possible for parent reducers to get actions for states they are not currently in. This could be viewed as a feature or a bug I suppose. But if you're considering these to be truly mutually exclusive, it might be preferable to throw an error/warning if the reducer receives unexpected actions, particularly since the child reducer for the unexpected action doesn't receive it (from what I can tell)
** Can be worked around by having the parent reducer switch on a combination of the `state` and the `action`, and ignore actions that are not allowed for the current state.
* Would a `SwitchStore` be possible/desirable? In my example I was able to get the desired behaviour with a bunch of `IfLetStore`s, but that doesn't make it obvious to a reader that these are mutually exclusive conditions. I could also wrap the `IfLetStore`s in a normal switch statement, but that results in some redundancy since I am then unwrapping the state even though I know it is non-nil. There might be some obvious workaround for this that I am missing.

If this is something I can help with, either through working on an implementation, or expanding my experiment to cover more use cases, I'd be happy to do so. I think this would be a great feature to have in the architecture for a few reasons - my particular use case could be handled by a series of nested optional stores, but the nesting becomes cumbersome, as does the optionality. Being able to use an enum allows for more modularization - in my example, the `CountrySelector` is self-contained and doesn't know that it is pre-requisite for `Home`. Similarly, if I had something like deep-linking that could provide a country without having to manually select it, the app's state could be changed to `Home` without having to involve a `CountrySelector` at all.

AlexisQapa:
For you first point you can use Reducer.run. The switchStore is a nice parallel with IfLetStore but you can now have switch statements in you views so I'm not sure it's useful

stephencelis:
> * I might be missing something as I wasn't able to implement it purely additively - the changes to `Reducer.swift` couldn't be implemented outside of the library (even when providing different names for otherwise conflicting properties/functions) as they make use of `reducer` which is private. I opted to change my SPM dependency to target this branch instead, but that also again means that we'd potentially be stuck on an out of date branch (unless it's periodically rebased on the latest).

Ah yeah, as @AlexisQapa notes, you can update `reducer` to `run` and you should be good to go.

> * It's possible for parent reducers to get actions for states they are not currently in. This could be viewed as a feature or a bug I suppose. But if you're considering these to be truly mutually exclusive, it might be preferable to throw an error/warning if the reducer receives unexpected actions, particularly since the child reducer for the unexpected action doesn't receive it (from what I can tell)
>   ** Can be worked around by having the parent reducer switch on a combination of the `state` and the `action`, and ignore actions that are not allowed for the current state.

This should probably follow the pattern found in `optional` higher-order reducer, which currently assert in DEBUG mode:

https://github.com/pointfreeco/swift-composable-architecture/blob/fe06d95cff0dd7e68aa29a46e325340788f668c9/Sources/ComposableArchitecture/Reducer.swift#L305-L325

We have a branch changing this behavior with the introduction of an explicit `breakpointOnNil` parameter in #296, but we want to take a fresh pass at documentation to make things clearer before merging.

> * Would a `SwitchStore` be possible/desirable? In my example I was able to get the desired behaviour with a bunch of `IfLetStore`s, but that doesn't make it obvious to a reader that these are mutually exclusive conditions. I could also wrap the `IfLetStore`s in a normal switch statement, but that results in some redundancy since I am then unwrapping the state even though I know it is non-nil. There might be some obvious workaround for this that I am missing.

The nice thing about an explicit switch is that it lets you exhaustively switch over an enum, though I guess it could be problematic to introduce the whole enum of state to be observable just for that ability. That's why I think your solution is probably the best for now.

You might be able to do a `SwitchStore` that looks something like this, which just uses `IfLetStore` under the hood:

```swift
SwitchStore(
  self.store,
  .case(/AppState.loggedIn, then: LoggedInView.init(store:)),
  .case(/AppState.loggedOut, then: LoggedOutView.init(store:))
)

// vs.
IfLetStore(self.store.scope(state: /AppState.loggedIn), then: LoggedInView.init(store:)),
IfLetStore(self.store.scope(state: /AppState.loggedOut), then: LoggedOutView.init(store:))
```

But we still can't get compile-time exhaustivity (we could maybe do a runtime assertion when hitting an unhandled case), so I'm wondering whether it's worth it or not. Thoughts?

AlexisQapa:
The main issue I've hit with the changes on this branch is that it's clunky to mix keypaths and casepaths. While there is api support for it you can't really use the nice notation with a combination of both like /.keypath.casepath or \.keypath.casepath.

I wonder how what it should look like thought adding more symbols than "/" "\" or "." make things too far from the well know keypaths. 

This becomes quickly useful in more complex app. I use an enum for my modal states (like 15 different modals) and I need to scope / pullback thrice. Keypath from appstate to modal enum, then case path to the modal then keypath to the content.

nmccann:
Thank you for the feedback - I can confirm that changing the implementation to use `run` allowed me to implement it as an extension on `Reducer`, making it a purely additive change.

That said, perhaps it could be merged in as an experimental feature? It could be psuedo-namespaced under an `experimental` extension on `Reducer`.

To me being able to represent mutually exclusive state seems like a natural requirement for most projects. Admittedly the current syntax makes it a little clunky, but that could be refined with usage. But I understand that increasing the surface area of an open source project should not be taken lightly.

spekke:
Really like this optional paths concept. 

Currently we are using enums to drive navigation in our UIKit application. We experimented with this OptionalPath concept to remove a lot of boilerplate code, but ran into a minor problem.

Current solution without OptionalPath:
```
enum Navigation : Equatable {
    case subPage1(SubPage1State)
    case subPage2(SubPage2State)
}


struct MainState : Equatable {
    var navigation: Navigation?

    var subPage1State: SubPage1State? {
        set {
            navigation = newValue.map(Navigation.subPage1)
        }
        get {
            guard case .subPage1(let subPage1State) = navigation else {
                return nil
            }
            return subPage1State
        }
    }

    var subPage2State: SubPage2State? {
        set {
            navigation = newValue.map(Navigation.subPage2)
        }
        get {
            guard case .subPage2(let subPage2State) = navigation else {
                return nil
            }
            return subPage2State
        }
    }
}

let subPage1Reducer = Reducer<SubPage1State, Void, Void>.empty
let subPage2Reducer = Reducer<SubPage2State, Void, Void>.empty

let mainReducer: Reducer<MainState, MainAction, Void> = Reducer.combine(
    subPage1Reducer.optional().pullback(state: \.subPage1State, action: /MainAction.subPage1, environment: { $0 }),
    subPage2Reducer.optional().pullback(state: \.subPage2State, action: /MainAction.subPage2, environment: { $0 })
)
```

In UIViewController with Store<MainState, MainAction>:
```
store.scope(state: \.subPage1State, action: MainAction.subPage1)
    .ifLet(
        then: { store in
            let viewController = Page1ViewController(store: store)
            self.navigationController?.pushViewController(viewController, animated: true)
        },
        else {
            self.navigationController?.popToViewController(self, animated: true)
        }
    )
    .store(in: &disposeBag)
```

So the boilerplate we would like to get rid of is the computed properties `subPage1State` and `subPage2State` in `MainState`.

Our problem is that we would like to find a nice and ergonomic way of expressing an OptionalPath to e.g `MainState.Navigation.subPage1`.
The only way we accomplished this was with the following:

```
let mainReducer: Reducer<MainState, MainAction, Void> = Reducer.combine(
    subPage1Reducer.pullback(state: (\MainState.navigation).appending(path: /Navigation.subPage1State), action: /MainAction.subPage1, environment: { $0 }),
    subPage2Reducer.pullback(state: (\MainState.navigation).appending(path: /Navigation.subPage2State), action: /MainAction.subPage2, environment: { $0 })
)
```

And for our scope in UIViewController:
```
store.scope(state: (\MainState.navigation).appending(path: /Navigation.subPage1State).extract, action: MainAction.subPage1)
```

Is there a better way of expressing our OptionalPaths?

mbrandonw:
Hey @spekke, what you've got right now is pretty much the way to do it. I can only think of a few small things to clean up some noise.

First, for the `pullback` you could use the `..` operator to compose paths instead of `.appending(path:)`:

```swift
let mainReducer: Reducer<MainState, MainAction, Void> = Reducer.combine(
    subPage1Reducer.pullback(state: \MainState.navigation .. /Navigation.subPage1State, action: /MainAction.subPage1, environment: { $0 }),
    subPage2Reducer.pullback(state: \MainState.navigation .. /Navigation.subPage2State, action: /MainAction.subPage2, environment: { $0 })
)
```

And then for the `scope` we could introduce an operator that "lifts" any path to a function. This is what we used to have to do with key paths before key path literals could be automatically promoted to functions by the compiler. We had a `^` prefix operator and a `get` free function so that you could do things like:

```swift
users.map(^\.id)
users.map(get(\.id))
```

So we could apply that same idea to optionals paths, with something like:

```swift
store.scope(state: ^(\MainState.navigation .. /Navigation.subPage1State), action: MainAction.subPage1)
store.scope(state: extract(\MainState.navigation .. /Navigation.subPage1State), action: MainAction.subPage1)
```

We also find that a healthy amount of newlines help make this code easier to grok:

```swift
let mainReducer: Reducer<MainState, MainAction, Void> = Reducer.combine(
  subPage1Reducer.pullback(
    state: \MainState.navigation .. /Navigation.subPage1State, 
    action: /MainAction.subPage1, 
    environment: { $0 }
  ),
  subPage2Reducer.pullback(
    state: \MainState.navigation .. /Navigation.subPage2State, 
    action: /MainAction.subPage2, 
    environment: { $0 }
  )
)

store.scope(
  state: extract(\MainState.navigation .. /Navigation.subPage1State), 
  action: MainAction.subPage1
)
```

Just some ideas! Really the best would be if Swift had first class support for these concepts so that the compiler could provide us some really nice, magical syntax.

We'll have more to say about optional paths soon.

spekke:
Thanks for the reply, @mbrandonw ! Looking forward to hear more about OptionalPaths! 

Regarding Store.scope and OptionalPaths - Is there a specific reason for scope not supporting optional paths directly like Reducer.pullback to avoid the extra extract part? 

To feedback on the `..` - operator.
```
\MainState.navigation .. /Navigation.subPage1State
```
It's quite hard to understand what it's doing - The scope, pullback and CasePaths are quite hard to grasp already. And the new syntax `..` and `\` are increasing complexity even more. Would it make sense to have an initializer that at least removes one operator?  We have been experimenting and we find following more readable:
`OptionalPath(\MainState.navigation, /Navigation.subPage1State)`
or
`OptionalPath(\MainState.navigation, case: Navigation.subPage1State)`

What do think?

lukeredpath:
Reading through this thread with interest, as we are looking to integrate this behaviour into our app as we're finding lots of use cases for mutually exclusive state modelled as enums.

Has anyone been running this in production for a while yet and run into any issues?

Just to address a few points - @nmccann interesting to see somebody else had the idea of a `SwitchStore` - I started a discussion about this recently. I think you might have seen this already as I think you reacted to one of my posts but for those who haven't seen it, I have a working implementation of such a view here:
https://github.com/pointfreeco/swift-composable-architecture/discussions/388#discussioncomment-387573

It's a fair point to be made about losing the ability to exhaustively switch over each case. If that's important to you then a normal switch/case statement might be better but for me I think the automatic scoping over each case makes it worth it.

Regarding path composition - I definitely find the `..` operator to be helpful here even though it takes some getting used to. That said, I couldn't find an implementation of this operator that works with the new `Path` protocol so in case anyone needs it:

```swift
extension WritableKeyPath {
    public static func .. <AppendedValue>(
        lhs: WritableKeyPath,
        rhs: OptionalPath<Value, AppendedValue>
    ) -> OptionalPath<Root, AppendedValue> {
        return lhs.appending(path: rhs)
    }

    public static func .. <AppendedValue>(
        lhs: WritableKeyPath,
        rhs: CasePath<Value, AppendedValue>
    ) -> OptionalPath<Root, AppendedValue> {
        lhs.appending(path: .init(rhs))
    }
}

extension OptionalPath {
    public static func .. <AppendedValue>(
        lhs: OptionalPath,
        rhs: OptionalPath<Value, AppendedValue>
    ) -> OptionalPath<Root, AppendedValue> {
        return lhs.appending(path: rhs)
    }

    public static func .. <AppendedValue>(
        lhs: OptionalPath,
        rhs: CasePath<Value, AppendedValue>
    ) -> OptionalPath<Root, AppendedValue> {
        lhs.appending(path: .init(rhs))
    }
}

extension CasePath {
    public static func .. <AppendedValue>(
        lhs: CasePath,
        rhs: OptionalPath<Value, AppendedValue>
    ) -> OptionalPath<Root, AppendedValue> {
        return lhs.appending(path: rhs)
    }
}
```

This should allow composition of optional paths with key paths and case paths.

nmccann:
@lukeredpath yes I had seen that discussion - I appreciate your continued work on this! Aside from my initial investigations, I haven't had much time to further test the `optional-paths` branch - and certainly haven't used it in production.

I hope to find some time this week to try out your implementation of a `SwitchStore` - mutually exclusive state has popped up numerous times in my current project, and a better way to handle it would be greatly appreciated. 

I don't think it's a deal breaker that we can't exhaustively check that all cases are used - `IfLetStore` also lacks that ability, and at least `SwitchStore` properly conveys the intent that each case is mutually exclusive.

PR #85, opened 2020-05-12
Added Reducer.combine instance method

alexito4:
As requested here there is this little method. Is a Reducer.combine as an instance method so you can chain a reducer with another. Quite useful when you want to chain a "main" reducer with a "child" reducer. The static Reducer.combine feels more adequate when all reducers have the same "importance". 

I've added a method that accepts a single other reducer as that's what my [original tweet](https://twitter.com/alexito4/status/1260279123907416064) was about but feel free to ping me if you prefer another API. I wouldn't add an overload with varargs because at that point you may want to use the static one.
I've also toyed around with a label ->  `mainReducer.combine(with: childReducer)`, not sure what you prefer. 
I've tried to follow your wording on the docs, but feel free to suggest edits. 
I've also added a simple test kinda following the other combine one but simpler, I'm not sure if you want some other type of coverage. 

FYI, running `make format` errors in my machine
```
▶ make format
swift format --in-place --recursive .
error: unable to invoke subcommand: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift-format (No such file or directory)
make: *** [format] Error 2
```
so I formatted the files manually, not sure if they are totally correct 😅

### DIFF
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -75,6 +75,15 @@ public struct Reducer<State, Action, Environment> {
     }
   }
 
+  /// Combines a reducer with another given reducer by running each one on the state, and
+  /// concatenating both of their effects.
+  ///
+  /// - Parameter reducers: Another reducer.
+  /// - Returns: A single reducer.
+  public func combine(_ other: Reducer) -> Reducer {
+    Reducer.combine(self, other)
+  }
+
   /// Transforms a reducer that works on local state, action and environment into one that works on
   /// global state, action and environment. It accomplishes this by providing 3 transformations to
   /// the method:
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -78,9 +78,9 @@ public struct Reducer<State, Action, Environment> {
   /// Combines a reducer with another given reducer by running each one on the state, and
   /// concatenating both of their effects.
   ///
-  /// - Parameter reducers: Another reducer.
+  /// - Parameter other: Another reducer.
   /// - Returns: A single reducer.
-  public func combine(_ other: Reducer) -> Reducer {
+  public func combined(with other: Reducer) -> Reducer {
     Reducer.combine(self, other)
   }
 
--- a/Tests/ComposableArchitectureTests/ReducerTests.swift
+++ b/Tests/ComposableArchitectureTests/ReducerTests.swift
@@ -87,7 +87,7 @@ final class ReducerTests: XCTestCase {
       return Effect.fireAndForget { mainEffectExecuted = true }
         .eraseToEffect()
     }
-    .combine(childReducer)
+    .combined(with: childReducer)
 
     let store = TestStore(
       initialState: 0,
--- a/Tests/ComposableArchitectureTests/ReducerTests.swift
+++ b/Tests/ComposableArchitectureTests/ReducerTests.swift
@@ -69,6 +69,42 @@ final class ReducerTests: XCTestCase {
     )
   }
 
+  func testCombine() {
+    enum Action: Equatable {
+      case increment
+    }
+    
+    var childEffectExecuted = false
+    let childReducer = Reducer<Int, Action, Void> { state, _, _ in
+      state += 1
+      return Effect.fireAndForget { childEffectExecuted = true }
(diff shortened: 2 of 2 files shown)


stephencelis:
Thanks for PRing!

> FYI, running `make format` errors in my machine
> 
> ```
> ▶ make format
> swift format --in-place --recursive .
> error: unable to invoke subcommand: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift-format (No such file or directory)
> make: *** [format] Error 2
> ```
> 
> so I formatted the files manually, not sure if they are totally correct 😅

`swift-format` doesn't ship with Swift (yet :pray:), so you would need to install it for `make format` to work (it's available via Homebrew). But never fear! We actually set up a GitHub action that automatically re-formats `master` on new commits. _E.g._:

https://github.com/pointfreeco/swift-composable-architecture/commit/52baef0b2d93b1a5c3b68a06e962e35c5c64c31d

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -75,6 +75,15 @@ public struct Reducer<State, Action, Environment> {
     }
   }
 
+  /// Combines a reducer with another given reducer by running each one on the state, and
+  /// concatenating both of their effects.
+  ///
+  /// - Parameter reducers: Another reducer.
+  /// - Returns: A single reducer.
+  public func combine(_ other: Reducer) -> Reducer {
```
stephencelis, line 83: We can all workshop the name a bit, but what do you think of this?
```suggestion
  public func combine(with other: Reducer) -> Reducer {
```
I also don't know if we should be looking to the naming guidelines and call this `combining` (or `combined`?) 🤔

Tests/ComposableArchitectureTests/ReducerTests.swift
```diff
@@ -69,6 +69,42 @@ final class ReducerTests: XCTestCase {
     )
   }
 
+  func testCombine() {
```
stephencelis, line 72: Love to see tests 😍

alexito4:
Apparently adding this causes the compiler to get confused with overloads 😢 
Specifically https://github.com/pointfreeco/swift-composable-architecture/blob/61ce0bec60c5304d158848964e5333f31eeed452/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-ReusableFavoriting.swift#L170

I find this quite weird since is pretty clear that is a static vs. instance method. 

And adding a label doesn't seem to help, which seems even weirder 😞

stephencelis:
Interesting. That failure is from an instance of `combine` that is only taking a single reducer, which isn't a great use of `combine` 😅 https://github.com/pointfreeco/swift-composable-architecture/blob/6b27b00f29965d0ee3c922a8d95313aaef020d1a/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-ReusableFavoriting.swift#L170-L176

We could fix the static API to require at least 2 reducers, though it might muddy up the autocomplete. We should also fix the example above by getting rid of the `combine` 😆

alexito4:
> We could fix the static API to require at least 2 reducers

I've quickly tried

```
    public static func combine(_ head: Reducer, _ reducers: Reducer...) -> Reducer {
    .combine([head] + reducers)
  }
```

but is still complaining 😂  kinda feels like a bug to me tbh.

Seeing that the compiler is being so picky we probably should just use a different name like you suggest https://github.com/pointfreeco/swift-composable-architecture/pull/85#discussion_r424018271

> We should also fix the example above by getting rid of the combine 😆

I can clean that up once I make sure the PR doesn't introduce issues ^^

stephencelis:
I think it'd need to be:

```swift
public static func combine(_ first: Reducer, _ second: Reducer, _ rest: Reducer...) -> Reducer {
```

But that's not great! @mbrandonw and I will chat about this in a bit.

stephencelis:
Chatted with @mbrandonw! Let's call it `combined(with:)`. That'll help with the ambiguity.

Sources/ComposableArchitecture/Reducer.swift
alexito4: We went for `combined(with:)` 👍

alexito4:
Name changed and everything is compiling now 👍 
I've also cleaned up the example in another PR.

PR #87, opened 2020-05-13
Stop vendoring CasePaths

mbrandonw:
With our latest SPM-bug work around (#70) it seems we can now properly depend on the CasePaths library rather than venturing it directly.


Package.swift
```diff
@@ -16,10 +16,15 @@ let package = Package(
       targets: ["ComposableArchitecture"]
     ),
   ],
+  dependencies: [
+    .package(url: "https://www.github.com/pointfreeco/swift-case-paths", from: "0.1.0")
```
mbrandonw, line 20: this version good for pinning?

Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
```diff
@@ -55,8 +55,8 @@ where SchedulerTimeType: Strideable, SchedulerTimeType.Stride: SchedulerTimeInte
   public func schedule(
     after date: SchedulerTimeType,
     interval: SchedulerTimeType.Stride,
-    tolerance: SchedulerTimeType.Stride,
-    options: SchedulerOptions?,
+    tolerance _: SchedulerTimeType.Stride,
+    options _: SchedulerOptions?,
```
mbrandonw, line 59: unrelated but snuck this in. while working on some episode stuff I realized it's kinda nice to make it clear we aren't using this info

Package.swift
stephencelis, line 20: Let's bump to 0.1.1 before merging. Also let's drop the `www`:

```suggestion
    .package(url: "https://github.com/pointfreeco/swift-case-paths", from: "0.1.0")
```

ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,16 +0,0 @@
-{
```
stephencelis: Should this file have been deleted? Wonder if it's needed for CI, but guess we'll find out!

PR #90, opened 2020-05-13
Minor tweaks to Effect's API for consistency

stephencelis:
A few things I noticed. Up for workshopping further.

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -89,14 +89,14 @@ public struct Effect<Output, Failure: Error>: Publisher {
   ///  If you need to deliver more than one value to the effect, you should use the `Effect`
   ///  initializer that accepts a `Subscriber` value.
   ///
-  /// - Parameter work: A closure that takes a `callback` as an argument which can be used to feed
-  ///   it `Output` values.
+  /// - Parameter attemptToFulfill: A closure that takes a `callback` as an argument which can be
+  ///   used to feed it `Result<Output, Failure>` values.
   public static func future(
-    work: @escaping (@escaping (Result<Output, Failure>) -> Void) -> Void
+    _ attemptToFulfill: @escaping (@escaping (Result<Output, Failure>) -> Void) -> Void
   ) -> Effect {
     Deferred {
       Future { callback in
-        work { output in callback(output) }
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Effect.swift
```diff
@@ -89,14 +89,14 @@ public struct Effect<Output, Failure: Error>: Publisher {
   ///  If you need to deliver more than one value to the effect, you should use the `Effect`
   ///  initializer that accepts a `Subscriber` value.
   ///
-  /// - Parameter work: A closure that takes a `callback` as an argument which can be used to feed
-  ///   it `Output` values.
+  /// - Parameter attemptToFulfill: A closure that takes a `callback` as an argument which can be
```
stephencelis, line 92: This matches Combine's API.

Sources/ComposableArchitecture/Effect.swift
```diff
@@ -158,13 +158,13 @@ public struct Effect<Output, Failure: Error>: Publisher {
...
   ///
-  /// - Parameter run: A closure that accepts a `Subscriber` value and returns a cancellable. When
-  ///   the `Effect` is completed, the cancellable will be used to clean up any
-  ///   resources created when the effect was started.
+  /// - Parameter work: A closure that accepts a `Subscriber` value and returns a cancellable. When
+  ///   the `Effect` is completed, the cancellable will be used to clean up any resources created
+  ///   when the effect was started.
   public static func async(
-    _ run: @escaping (Effect.Subscriber<Output, Failure>) -> Cancellable
+    _ work: @escaping (Effect.Subscriber<Output, Failure>) -> Cancellable
```
stephencelis, line 165: We use `run` some places and `work` other places. In fact I wonder if this entire thing could be renamed from `async` to `run` 🤔

`Effect.run { subscriber in ... }` has a nice ring to it...

mbrandonw: Dang, yeah. That’s pretty nice

stephencelis: Should we make the change in this PR? Separate? Deprecate with `renamed:`?

stephencelis: I'll merge as is. We can think about this a bit more.

PR #91, opened 2020-05-13
Add "run" function to Reducer

stephencelis:
This makes it a lil more discoverable and friendlier for Xcode 11.3.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,4 +61,4 @@ test-workspace:
 format:
 	swift format --in-place --recursive .
 
-.PHONY: format
+.PHONY: format test-all test-swift test-workspace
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -61,4 +61,4 @@ test-workspace:
 format:
 	swift format --in-place --recursive .
 
-.PHONY: format
+.PHONY: format test-all test-swift test-workspace
```
stephencelis, line 64: Sneaking in an unrelated change.

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -293,20 +293,26 @@ public struct Reducer<State, Action, Environment> {
...
+  public func run(
     _ state: inout State,
     _ action: Action,
     _ environment: Environment
   ) -> Effect<Action, Never> {
     self.reducer(&state, action, environment)
   }
-}
 
-extension Reducer where Environment == Void {
```
stephencelis, line 305: Killing this `callAsFunction` overload.

PR #92, opened 2020-05-13
Rename certain Effect APIs

stephencelis:
The `async` and `sync` APIs don't quite feel right when we also have `future` and `result`.

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -167,13 +167,6 @@ public struct Effect<Output, Failure: Error>: Publisher {
     AnyPublisher.create(work).eraseToEffect()
   }
 
-  @available(*, deprecated, renamed: "run")
-  public static func async(
-    _ work: @escaping (Effect.Subscriber<Output, Failure>) -> Cancellable
-  ) -> Self {
-    self.run(work)
-  }
-
   /// Concatenates a variadic list of effects together into a single effect, which runs the effects
   /// one after the other.
   ///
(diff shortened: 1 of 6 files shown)


Sources/ComposableArchitecture/Effect.swift
```diff
@@ -161,12 +161,19 @@ public struct Effect<Output, Failure: Error>: Publisher {
...
     AnyPublisher.create(work).eraseToEffect()
   }
 
+  @available(*, deprecated, renamed: "run")
+  public static func async(
+    _ work: @escaping (Effect.Subscriber<Output, Failure>) -> Cancellable
+  ) -> Self {
+    self.run(work)
+  }
+
```
stephencelis, line 176: Not sure it's necessary to deprecate, but may as well for a version or two.

Sources/ComposableArchitecture/Effect.swift
```diff
@@ -257,9 +264,14 @@ extension Effect where Failure == Swift.Error {
   ///
   /// - Parameter work: A closure encapsulating some work to execute in the real world.
   /// - Returns: An effect.
-  public static func sync(_ work: @escaping () throws -> Output) -> Self {
+  public static func catching(_ work: @escaping () throws -> Output) -> Self {
```
stephencelis, line 267: To match the argument name of the `Result` initializer.

Sources/ComposableArchitecture/Effect.swift
mbrandonw: Yeah might as well moving forward. Could even create a `Deprecations.swift` in `Internal` to put all of these?

PR #94, opened 2020-05-14
Fix ordering of simultaneous work in test scheduler.

mbrandonw:
There was a small bug in the test scheduler where if two timers have emissions that later coalesce to happen simultaneously, then it is possible for them to emit in the wrong order. We need to track the order in which the timers were started in order to fix this, and we can do that with a simple auto-incrementing id instead of a UUID.

I haven't run the full demo test suite locally to see if there are any places in our demo tests where we may have been accidentally depending on this logic. Will let CI do that for me...

### DIFF
--- a/Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
+++ b/Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
@@ -21,9 +21,7 @@ where SchedulerTimeType: Strideable, SchedulerTimeType.Stride: SchedulerTimeInte
   ///
   /// - Parameter stride: A stride.
   public func advance(by stride: SchedulerTimeType.Stride = .zero) {
-    self.scheduled.sort {
-      $0.date < $1.date || $0.id < $1.id
-    }
+    self.scheduled.sort { $0.date < $1.date || $0.id < $1.id }
 
     guard
       let nextDate = self.scheduled.first?.date,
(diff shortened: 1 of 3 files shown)


Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
```diff
@@ -52,24 +49,29 @@ where SchedulerTimeType: Strideable, SchedulerTimeType.Stride: SchedulerTimeInte
...
       return { [weak self] in
         action()
-        self?.scheduled.append((id, date, scheduleAction(date.advanced(by: interval))))
+        let nextDate = date.advanced(by: interval)
+        self?.scheduled.append((id, nextDate, scheduleAction(for: nextDate)))
       }
     }
 
-    self.scheduled.append((id, date, scheduleAction(date.advanced(by: interval))))
+    self.scheduled.append((id, date, scheduleAction(for: date)))
```
mbrandonw, line 69: this is unrelated clean up, but helped make the logic a little clearer to me.

Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
```diff
@@ -82,11 +79,16 @@ where SchedulerTimeType: Strideable, SchedulerTimeType.Stride: SchedulerTimeInte
...
+    self.scheduled.append((self.nextId(), date, action))
   }
 
   public func schedule(options _: SchedulerOptions?, _ action: @escaping () -> Void) {
-    self.scheduled.append((UUID(), self.now, action))
+    self.scheduled.append((self.nextId(), self.now, action))
+  }
+
+  private func nextId() -> Int {
+    self.lastId += 1
```
stephencelis, line 90: Should we care about overflow ever? Probably unlikely. Could use `UInt` to double the number of work that can be scheduled?

mbrandonw: done, but yeah, I bet Swift couldn't even execute that much work in a reasonable amount of time 😂

PR #95, opened 2020-05-14
Fatal error for identified array subscript misuse

stephencelis:
Figured it'd be better to catch these in debug builds early.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
@@ -93,8 +93,8 @@ where ID: Hashable {
   ///
   /// - Parameter id: The identifier of element to access. Must be a valid identifier for an element
   ///   of the array and will _not_ insert elements that are not already in the array, or remove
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
```diff
@@ -84,29 +84,55 @@ where ID: Hashable {
...
-  ///   elements when passed `nil`. Use `insert(_:at:)` and `remove(id:)` to insert and remove
-  ///   elements.
+  ///   elements when passed `nil`. Use `append` or `insert(_:at:)` to insert elements. Use
+  ///   `remove(id:)` to remove an element by its identifier.
   /// - Returns: The element.
-  public subscript(id id: ID) -> Element? {
-    get {
-      self.dictionary[id]
+  #if DEBUG
+    public subscript(id id: ID, file: StaticString = #file, line: UInt = #line) -> Element? {
```
stephencelis, line 100: Looks like default subscript args came with 5.2...having the right file and line number are pretty handy in crashes, but since this logic is limited to debug builds already I guess it doesn't matter as much, so will remove.

PR #100, opened 2020-05-15
A  throwing Effect.future variant.

jasdev:
Ended up needing something like this today when wrapping AVFoundation-related work in an Effect that mixed both synchronous `throw`ing calls and completion handlers. Curious if y’all have any thoughts on this overload — in both the yay or nay case!


jasdev:
(er, hmm never mind haha — it makes ergonomics a bit clunky after seeing the CI failures. 😄)

PR #101, opened 2020-05-15
Move framework specific debug code into relevant case study

kaandedeoglu:
Speech & CoreLocation related `CustomDebugOutputConvertible` conformances shouldn't really live in the library. 

This PR moves the `SFSpeechRecognizerAuthorizationStatus` conformance into the `SpeechRecognition` case study, and removes the `CLAuthorizationStatus` one as it doesn't seem to be used in any of the examples.


stephencelis:
Hey @kaandedeoglu! Because we own the `CustomDebugOutputConvertible` protocol (which is used in the debug output in test store assertion failures and when you tack `.debug()` onto a reducer), we thought it would be safe to define these in the framework so that folks implementing their own speech and core location clients would be in a debuggable environment. I can see the argument for moving them to the clients, though. I'm gonna think about it a bit.

mbrandonw:
Hey @kaandedeoglu, we are going to leave these conformances in the library because it gives people the best output when debugging and testing. Without these conformances anyone who uses these types in their state/actions will get very bad output. And as @stephencelis pointed out, it can be [dangerous](https://wiki.haskell.org/Orphan_instance) for people to conform 3rd party types to 3rd party protocols, and so that doesn't leave a lot of options for users of the library.

However, in #104 we did decide to wrap the conformances in `if #DEBUG` so that it at least will not be shipped with production code.

I'm going to close this for now. Thanks!

kaandedeoglu:
Sounds good @mbrandonw, I went on with the PR following the inductive reasoning that this isn't really scalable. The list goes on with `UNAuthorizationStatus, ALAuthorizationStatus, MPMediaLibraryAuthorizationStatus ` and many more. Perhaps it's a good idea to vendor all these conformances in a file of its own rather than just these two.

stephencelis:
> The list goes on with `UNAuthorizationStatus, ALAuthorizationStatus, MPMediaLibraryAuthorizationStatus ` and many more.

Definitely a lot of work to be done to support them all, but we'll happily take PRs along the way 😄

PR #102, opened 2020-05-15
Make throttling public

kaandedeoglu:
Throttling can be as useful as debouncing, there's no reason to keep one public and the other internal.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Throttling.swift
+++ b/Sources/ComposableArchitecture/Effects/Throttling.swift
@@ -13,7 +13,7 @@ extension Effect {
   ///     `false`, the publisher emits the first element received during the interval.
   /// - Returns: An effect that emits either the most-recent or first element received during the
   ///   specified interval.
-  func throttle<S>(
+  public func throttle<S>(
     id: AnyHashable,
     for interval: S.SchedulerTimeType.Stride,
     scheduler: S,


stephencelis:
@kaandedeoglu Unfortunately, while throttling should work as expected in live applications, we haven't figured out how it can be tested yet. But maybe having this opened will help!

To see the problem, clone the repo and, in **EffectThrottleTests**, move the unique `CancelToken` per test, _e.g._: https://github.com/pointfreeco/swift-composable-architecture/blob/a90c48924fed1748c407c8032827572fe70a5b63/Tests/ComposableArchitectureTests/Internal/EffectThrottleTests.swift#L15

And define it once inside the file, instead, so that it's a shared identifier. Use of this identifier will begin to leak across tests, and tests will begin to fail.

Ideally, the shared storage for throttles would be associated with specific schedulers: https://github.com/pointfreeco/swift-composable-architecture/blob/a90c48924fed1748c407c8032827572fe70a5b63/Sources/ComposableArchitecture/Internal/Throttling.swift#L50-L51

And then each test would use a fresh scheduler and thus fresh storage for their throttle. We haven't spent a ton of time trying to tackle this, but the main blocker is that `Scheduler` is just a protocol, so I haven't figured out a way of "identifying" a single instance reliably. Got any ideas?

kaandedeoglu:
@stephencelis Thank you for reply and the pointer, I've indeed moved the `CancelToken` to be shared and understand the problem. I've got no leads to solve this at `Throttle` level atm. However as you mentioned the there's nothing wrong with how `throttle` is working, but rather how XCTest cases are run one after the other. 

A solution with two possible implementation possibilities come to mind:

- A well documented `Effect.setup()` convenience method that flushes the `throttleTimes` & `throttleValues` that users can call in `XCTestCase.setup()` (the instance method, not the class one)
- A `class ComposableArchitectureTestCase: XCTestCase` subclass that takes care of similar cleanups automatically.

stephencelis:
@kaandedeoglu I think we wanna wait till we have something that works without exposing these kinds of details to library users. Another "fix" is to make `AnyScheduler` a `class` instead of a `struct`, but we want to avoid that since it shouldn't matter.

I think we'll eventually land on a fix but it's not something we're actively looking into right now. If you decide to tackle and figure it out, though, please let us know!

In the meantime, is it cool to close this out till we've figured out how to best address testability?

kaandedeoglu:
Completely agreed :) I’ve actually looked into AnyScheduler - turns out CombineExt already has one (not surprised :)), although it’s not trivial at all.

PR #104, opened 2020-05-15
Remove some old tests and guard debug code in #if DEBUG.

mbrandonw:
Few things we talked about this morning...


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
```diff
@@ -64,7 +64,6 @@ struct AnimationsView: View {
               )
           }
           .frame(maxWidth: .infinity, maxHeight: .infinity)
-          .background(Color.white)
```
mbrandonw, line 67: also a fix for dark mode for this demo

PR #107, opened 2020-05-16
Readme fixes

OfftheCode:
Hey spotted and removed additional parenthesis in README code example.

Also AppStore now needs to be `Equatable` as per error:
`Referencing initializer 'init(_:)' on 'ViewStore' requires that 'AppState' conform to 'Equatable'`


OfftheCode:
Thanks for insane timing! Glad I could help!

PR #114, opened 2020-05-19
Add location manager examples to readmes.

mbrandonw:


Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-ResuableOfflineDownloads/DownloadComponent.swift
```diff
@@ -148,10 +148,6 @@ extension Reducer {
   }
 }
 
-private struct ThrottleId<ID>: Hashable where ID: Hashable {
```
mbrandonw, line 151: unrelated, but removing some old code.

PR #117, opened 2020-05-19
Gracefully handle recursive state debug output

stephencelis:
While state is generally a value type that cannot recursively reference itself, state still could contain the occasional reference type, which could contain a recursive reference to itself, so let's do what `dump` does and keep track of recursive items.

### DIFF
--- a/Tests/ComposableArchitectureTests/DebugTests.swift
+++ b/Tests/ComposableArchitectureTests/DebugTests.swift
@@ -373,6 +373,22 @@ final class DebugTests: XCTestCase {
     )
   }
 
+  func testRecursiveOutput() {
+    class Foo {
+      var foo: Foo?
+    }
+    let foo = Foo()
+    foo.foo = foo
+    XCTAssertEqual(
+      debugOutput(foo),
+      """
+      Foo(
+        foo: Foo()
+      )
+      """
+    )
+  }
+
   func testEffectOutput() {
     //    XCTAssertEqual(
     //      Effect<Int, Never>(value: 42)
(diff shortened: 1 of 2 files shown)


Tests/ComposableArchitectureTests/DebugTests.swift
```diff
@@ -373,6 +373,22 @@ final class DebugTests: XCTestCase {
...
+      var foo: Foo?
+    }
+    let foo = Foo()
+    foo.foo = foo
+    XCTAssertEqual(
+      debugOutput(foo),
+      """
+      Foo(
+        foo: Foo()
+      )
```
mbrandonw, line 387: awesome!

Tests/ComposableArchitectureTests/DebugTests.swift
```diff
@@ -373,6 +373,22 @@ final class DebugTests: XCTestCase {
...
+    class Foo {
+      var foo: Foo?
+    }
+    let foo = Foo()
+    foo.foo = foo
+    XCTAssertEqual(
+      debugOutput(foo),
+      """
+      Foo(
+        foo: Foo()
```
stephencelis, line 386: Not sure if we want to show somehow _which_ object this was...like:

```
Foo#1(
  foo: Foo#1()
)
```
Or if you have other ideas.

mbrandonw: huh, I'm down to leave as-is until we find a reason to change.

stephencelis: Some other thoughts (won't block merging, but could enhance later):

```
Foo(
  foo: Foo(...)
)
```

```
Foo(
  foo: Foo(↩︎)
)
```

PR #124, opened 2020-05-20
Add and remove map annotations only when necessary

kaandedeoglu:
Currently `MapView` removes and then re-adds all annotations regardless of if it's currently displayed or not. This causes an unpleasant flashing after each region change. 
This PR diffs the new point of interests with the existing ones to decide which annotations should be removed / added.

### DIFF
--- a/Examples/LocationManager/Common/AppCore.swift
+++ b/Examples/LocationManager/Common/AppCore.swift
@@ -2,7 +2,7 @@ import ComposableArchitecture
 import ComposableCoreLocation
 import MapKit
 
-public struct PointOfInterest: Equatable {
+public struct PointOfInterest: Equatable, Hashable {
   public let coordinate: CLLocationCoordinate2D
   public let subtitle: String?
   public let title: String?
@@ -224,5 +224,12 @@ extension PointOfInterest {
       && lhs.coordinate.longitude == rhs.coordinate.longitude
       && lhs.subtitle == rhs.subtitle
       && lhs.title == rhs.title
+  }
+
+  public func hash(into hasher: inout Hasher) {
+    hasher.combine(coordinate.latitude)
+    hasher.combine(coordinate.longitude)
+    hasher.combine(title)
+    hasher.combine(subtitle)
   }
 }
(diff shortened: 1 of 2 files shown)


Examples/LocationManager/Common/AppCore.swift
```diff
@@ -225,4 +225,11 @@ extension PointOfInterest {
       && lhs.subtitle == rhs.subtitle
       && lhs.title == rhs.title
   }
+
+  public func hash(into hasher: inout Hasher) {
```
mbrandonw, line 229: Could you move this down to the bottom of this file? That's where we also have the `==` implementation, and it just helps remove any unnecessary noise from the top of this file so people can focus on the important parts.

Examples/LocationManager/Common/MapView.swift
```diff
@@ -55,10 +55,16 @@ public struct MapView: ViewRepresentable {
       mapView.setRegion(region.asMKCoordinateRegion, animated: true)
     }
 
-    mapView.removeAnnotations(mapView.annotations)
-    mapView.addAnnotations(
-      self.pointsOfInterest.map(PointOfInterestAnnotation.init(pointOfInterest:))
-    )
+    let currentlyDisplayedPOIs = mapView.annotations.compactMap { $0 as? PointOfInterestAnnotation }.map(\.pointOfInterest)
```
mbrandonw, line 58: Can we make this change so that this example continues building for iOS 13.3, which sadly doesn't have key path functions.

```suggestion
    let currentlyDisplayedPOIs = mapView.annotations.compactMap { $0 as? PointOfInterestAnnotation }.map { $0.pointOfInterest }
```

Examples/LocationManager/Common/AppCore.swift
kaandedeoglu: This is already at the correct location I believe 👍

mbrandonw: Oh sorry! I misread the diff. Thanks!

Issue #125, opened 2020-05-22
IdentifiedArray.replaceSubrange causes a stack overflow


peterkovacs:
**Describe the bug**
Calling `replaceSubrange` on an identified array causes a stack overflow.

**To Reproduce**

```swift
  func testReplaceSubrange() {
    struct User: Equatable, Identifiable {
      let id: Int
      var name: String
    }

    var array: IdentifiedArray = [
      User(id: 3, name: "Blob Sr."),
      User(id: 2, name: "Blob Jr."),
      User(id: 1, name: "Blob"),
    ]

    array.replaceSubrange(0...1, with: [
      User(id: 4, name: "Flob IV"),
      User(id: 5, name: "Flob V")
    ])

    XCTAssertEqual(
      array,
      [User(id: 4, name: "Flob IV"), User(id: 5, name: "Flob V"), User(id: 1, name: "Blob")]
    )
  }
```

This results in repeated calls to 

```
#69778	0x0000000111c933ed in protocol witness for RangeReplaceableCollection.replaceSubrange<A>(_:with:) in conformance <> IdentifiedArray<A, B> ()
#69779	0x00007fff5116a0f7 in RangeReplaceableCollection.replaceSubrange<A, B>(_:with:) ()
```

**Environment**
 - Xcode 11.5
 - Swift 5.2.4

stephencelis:
Fixed by #126.

PR #126, opened 2020-05-22
Implement IdentifiedArray.replaceSubrange

peterkovacs:
Solves #125

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
@@ -240,7 +240,7 @@ where Element: Identifiable, ID == Element.ID {
 
   public mutating func replaceSubrange<C, R>(_ subrange: R, with newElements: C) where C : Collection, R : RangeExpression, Self.Element == C.Element, Self.Index == R.Bound {
     let replacingIds = self.ids[subrange]
-    let newIds = newElements.map(\.id)
+    let newIds = newElements.map { $0.id }
     ids.replaceSubrange(subrange, with: newIds)
 
     for element in newElements {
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
```diff
@@ -237,6 +237,20 @@ where Element: Identifiable, ID == Element.ID {
   public init() {
     self.init([], id: \.id)
   }
+
+  public mutating func replaceSubrange<C, R>(_ subrange: R, with newElements: C) where C : Collection, R : RangeExpression, Self.Element == C.Element, Self.Index == R.Bound {
+    let replacingIds = self.ids[subrange]
+    let newIds = newElements.map(\.id)
```
stephencelis, line 243: We'd like to support Xcode 11.3 for a bit, which contains Swift 5.1 and doesn't have this nice key path functionality. Let's use the old style instead:
```suggestion
    let newIds = newElements.map { $0.id }
```

Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
```diff
@@ -237,6 +237,20 @@ where Element: Identifiable, ID == Element.ID {
   public init() {
     self.init([], id: \.id)
   }
+
+  public mutating func replaceSubrange<C, R>(_ subrange: R, with newElements: C) where C : Collection, R : RangeExpression, Self.Element == C.Element, Self.Index == R.Bound {
```
stephencelis, line 241: Looks like we're also getting an 11.3 failure on this line for using `Self`. Can we try to omit it and see if that builds?
```suggestion
  public mutating func replaceSubrange<C, R>(_ subrange: R, with newElements: C) where C : Collection, R : RangeExpression, Element == C.Element, Index == R.Bound {
```

stephencelis:
Thanks!

PR #128, opened 2020-05-22
Implement sort() and sort(by:) for IdentifiedArray.

peterkovacs:
I'd love to figure out why `sort` didn't work without these methods, but it seemed like the elements were sorted correctly, but the ids were not.


stephencelis:
Thanks! Yeah it's puzzling why the methods we get "for free" don't appear to work. I'm sure there are more we've missed, so probably worth a full audit of the collection interface :/

PR #137, opened 2020-05-26
Avoid simultaneous access crash on double-cancel

### LINKED ISSUE #136: SwiftUICaseStudies app crashes
**Describe the bug**
The culprit is the `cleanup` function in Effects > Cancellation

**To Reproduce**
1. Navigate to Elm-like Subscriptions
2. Turn on the clock
3. Navigate back to the root view

**Screenshot**
Screenshot 2020-05-26 at 23 18 48

**Environment**
 - Xcode 11.5
 - Swift 5.1
 - iOS 13.5

stephencelis:
Fixes #136.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -45,9 +45,18 @@ extension Effect {
         .store(in: &cancellationCancellables[id, default: []])
       }
 
+      func cleanUp() {
+        cancellablesLock.sync {
+          guard !isCancelling.contains(id) else { return }
+          isCancelling.insert(id)
+          defer { isCancelling.remove(id) }
+          cancellationCancellables[id] = nil
+        }
+      }
+
       return subject.handleEvents(
-        receiveCompletion: { _ in cancellablesLock.sync { cancellationCancellables[id] = nil } },
-        receiveCancel: { cancellablesLock.sync { cancellationCancellables[id] = nil } }
+        receiveCompletion: { _ in cleanUp() },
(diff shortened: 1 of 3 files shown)


Tests/ComposableArchitectureTests/EffectCancellationTests.swift
```diff
@@ -223,17 +223,40 @@ final class EffectCancellationTests: XCTestCase {
 
     let expectation = self.expectation(description: "wait")
     effect
-      .sink(receiveCompletion: { _ in expectation.fulfill() }, receiveValue: { _ in })
+      .sink(receiveCompletion: { _ in
+        print("completed")
```
mbrandonw, line 227: some clean up to do here...

Sources/ComposableArchitecture/Effects/Cancellation.swift
```diff
@@ -76,12 +61,13 @@ extension Effect {
...
         cancellationCancellables[id] = nil
       }
     }
   }
 }
 
-var cancellationCancellables: [AnyHashable: [UUID: AnyCancellable]] = [:]
+var cancellationCancellables: [AnyHashable: Set<AnyCancellable>] = [:]
 let cancellablesLock = NSRecursiveLock()
+var isCancelling: Set<AnyHashable> = []
```
mbrandonw, line 73: I think this goes away now right?

Tests/ComposableArchitectureTests/EffectCancellationTests.swift
stephencelis: oh ha yeah

PR #140, opened 2020-05-27
refactor Step into enum instead of struct

ilyapuchka:
Swift now allows to use default values for enum associated values so we can get rid of nested step type enum and use cases instead of of static methods

### DIFF
--- a/Sources/ComposableArchitecture/TestSupport/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestSupport/TestStore.swift
@@ -196,8 +196,8 @@
       for step in steps {
         var expectedState = toLocalState(state)
 
-        switch step.type {
-        case let .send(action, update):
+        switch step {
+        case let .send(action, _, _, update):
           if !receivedActions.isEmpty {
             _XCTFail(
               """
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hey @ilyapuchka, thanks for taking the time to demonstrate this refactor!

Although it's nice that we get to leverage the default arguments for enum cases, it unfortunately ends up adding a bit of extra code in order to destructure the line and file from the enum. Because of this, we're not sure the change is worth it at this time.

So we think we'll leave it as-is for now until we have some public-facing benefits to the change.

Please let us know if you have other thoughts about this though!

ilyapuchka:
Technically we can replace these computed properties with variables that are assigned in the `switch step` (then they are being used in one of the follow up assert), but it's ok, not a necessary change of course.

stephencelis:
Yeah, it's a nice feature and it'd be cool to take advantage of it. Just a bummer that it introduces some ergonomic quirks in this case.

mbrandonw:
Sweet, will close for now, but good to know for future evolutions of this API. I could forsee some day having a fully extensible step system where people can make their own types of steps easily.

PR #143, opened 2020-05-27
Add debugging section to README

stephencelis:


README.md
```diff
@@ -93,15 +93,15 @@ struct AppState: Equatable {
 Next we have the actions in the feature. There are the obvious actions, such as tapping the decrement button, increment button, or fact button. But there are also some slightly non-obvious ones, such as the action of the user dismissing the alert, and the action that occurs when we receive a response from the fact API request:
 
 ```swift
-enum AppAction {
+enum AppAction: Equatable {
```
stephencelis, line 96: This is required for tests to work...could move the conformance to testing, though, if you think it's noisy here.

mbrandonw: Yeah, I had done that on purpose as a balance between ease of reading and explicitness, but down to add. maybe as a comment in the testing section though so to not do too much upfront?

PR #145, opened 2020-05-27
Small Optimization For Signposts

stephencelis:


Sources/ComposableArchitecture/Debugging/ReducerInstrumentation.swift
```diff
@@ -36,17 +36,20 @@ extension Reducer {
     let prefix = prefix.isEmpty ? zeroWidthSpace : "[\(prefix)] "
 
     return Self { state, action, environment in
+      var actionOutput: String!
       if log.signpostsEnabled {
-        os_signpost(.begin, log: log, name: "Action", "%s%s", prefix, debugCaseOutput(action))
+        actionOutput = debugCaseOutput(action)
```
stephencelis, line 41: May as well only compute this once...

Sources/ComposableArchitecture/Debugging/ReducerInstrumentation.swift
```diff
@@ -36,17 +36,20 @@ extension Reducer {
...
+        os_signpost(.begin, log: log, name: "Action", "%s%s", prefix, actionOutput)
       }
       let effects = self.run(&state, action, environment)
       if log.signpostsEnabled {
         os_signpost(.end, log: log, name: "Action")
+        return
```
stephencelis, line 47: We only wanna call `effectSignpost` in this `if`, right?

mbrandonw: yeah, that looks good

PR #147, opened 2020-05-28
Add toast on failures case study

sundeepgupta:
Demonstrates how a higher order reducer can be used to show a toast on all failures.

We're looking for this kind of functionality in our app and came up with this based on the other case studies. Looking for feedback and if suitable, hopefully helpful for others. 🙏 

Toasts

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-FailureToast.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-FailureToast.swift
@@ -45,11 +45,7 @@ enum ToastAction {
   case hide
 }
 
-struct ToastEnvironment {
-  var mainQueue: AnySchedulerOf<DispatchQueue>
-}
-
-let toastReducer = Reducer<ToastState, ToastAction, ToastEnvironment> { state, action, environment in
+let toastReducer = Reducer<ToastState, ToastAction, Void> { state, action, _ in
   switch action {
   case .show(let text):
     state.status = .showing(text)
@@ -112,7 +108,6 @@ enum AppAction {
 }
 
 struct AppEnvironment {
-  var mainQueue: AnySchedulerOf<DispatchQueue>
   var loadData: () -> Effect<[String], Error>
 }
 
@@ -159,7 +154,7 @@ let combinedFailureToastReducer = Reducer<AppState, AppAction, AppEnvironment>.c
   toastReducer.pullback(
     state: \AppState.toastState,
     action: /AppAction.toastAction,
-    environment: { _ in ToastEnvironment(mainQueue: DispatchQueue.main.eraseToAnyScheduler()) }
+    environment: { _ in () }
   )
 )
 
@@ -170,7 +165,6 @@ struct ToastView_Previews: PreviewProvider {
         initialState: AppState(),
         reducer: combinedFailureToastReducer,
         environment: AppEnvironment(
-          mainQueue: DispatchQueue.main.eraseToAnyScheduler(),
           loadData: {
             Fail(error: AppError.api)
               .delay(for: 1, scheduler: DispatchQueue.main)
--- a/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-FailureToast.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-FailureToast.swift
@@ -22,8 +22,49 @@ enum AppError: Error, LocalizedError {
 }
 
 struct AppState: Equatable {
-  var toastStatus: ToastStatus = .hiding
+  var toastState = ToastState()
   var data: [String] = []
+}
+
+enum AppAction {
+  case toastAction(ToastAction)
+  case loadData
+  case didLoadData([String])
+}
+
+struct AppEnvironment {
+  var mainQueue: AnySchedulerOf<DispatchQueue>
+  var loadData: () -> Effect<[String], Error>
+}
+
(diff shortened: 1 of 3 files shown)


Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-FailureToast.swift
```diff
@@ -0,0 +1,145 @@
...
+extension Reducer {
+  static func errorHandling(
+    _ reducer: @escaping (inout AppState, AppAction, AppEnvironment) -> Effect<AppAction, Error>
+  ) -> Reducer<AppState, AppAction, AppEnvironment> {
+    Reducer<AppState, AppAction, AppEnvironment> { state, action, environment in
+      reducer(&state, action, environment)
+        .catch { Just(.showToast($0.localizedDescription)) }
+        .eraseToEffect()
+    }
+  }
```
sundeepgupta, line 88: We're aware that this function isn't generic and tied to the `App` component, which is the only component in this example. We're interested in learning how this might work with sub-components as well in a nice way. We've looked at the `resending` technique which helps deal with duplicate actions needed in sub-components. But not sure if that's the right way. Ideally, this could be used on the root component only, and all sub-components would get it automatically also.

stephencelis:
This looks really cool! 😄 We're looking forward to dive in deeper when we have some time and lend any feedback we may have.

Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-FailureToast.swift
```diff
@@ -0,0 +1,145 @@
...
+      switch self {
+      case .showing(let text): return text
+      case .hiding: return ""
+      }
+  }
+}
+
+enum AppAction {
+  case showToast(String)
+  case hideToast
```
alexito4, line 50: I was expecting a `case toastAction(ToastAction)` so the feature action knows less about the reusable toast functionality (I think the favorites example does something like this)

Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-FailureToast.swift
alexito4: When I implemented reusable time travel functionality what I did was to have a generic state with what I needed, that nested the rest of the app state, same for the actions. Then the higher order function wraps the app state/action in the resusable functionality state/action. That makes your App feature completely unware of it. The small inconvenience is that you need a custom view that wraps your feature view too. 

If you wanted to show the toast on specific actions that would make it hard (then you would have to make your reusable feature aware of specific actions of your app) but since you're using the `catch` on effect (nice idea!) it should be fine.

grinder81: One of the cases we are hitting is that Reducer return type is `Effect<Action, Never>` and was wondering why we can't have `Reducer<State, Action, Error, Environment>`and return type of `Effect<Action, Error>`? Then we can bubble up the error using pullback.

Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-FailureToast.swift
sundeepgupta: Thank you - we've made the changes to make it more like the favourites example. 🙏

Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-FailureToast.swift
sundeepgupta: Also, we're looking at another approach that might work. Hoping to have something more concrete soon

sundeepgupta: To clear up the initial comment, we think that making the reducer generic is a separate concern from making this work with sub-components. 

Generics: 
We think we understand in general the technique you describe @alexito4. We looked around for the time travelling example and found a video on your site which we'll have to watch. 🙏 

Sub-components:
[We've experimented with a solution which seems to work](https://github.com/sundeepgupta/swift-composable-architecture/compare/error-toast-case-study...sundeepgupta:error-result-case-study). Essentially, we defined a `FailableReducer` which uses failable `Effect<Action, Error>. This way, the errors from sub-components can bubble up. We're looking to feedback on this approach and how much sense it might make for TCA's `Reducer` to work with failable effects. 🙏

sundeepgupta: Hi all, just wanted to bump this thread to get some thoughts on the failable reducer idea linked above... Any ideas/feedback is welcome. We need this kind of functionality in our app and wondering how good/bad of an idea this might be. Especially, considering potential future changes to TCA that might be on the roadmap

mbrandonw:
Hey @sundeepgupta, again thanks for these case studies! We're also going to close this one. We think that the thing that really needs to be demonstrated in this case study is how to generically introduce toasts at an app-wide level, so that you can show a toast from any layer inside your app. We may give some thought to this in the future, and if you have any ideas please feel free to let us know!

sundeepgupta:
Thanks @mbrandonw. Yes, I definitely agree with you and to this point, we're looking to your thoughts on the idea of global error handling and our current approach to it by using a failable reducer. We moved the thread discussion to [the forums](https://forums.swift.org/t/how-to-properly-bubble-up-state-or-global-error-handler/37335) if that's a better spot to discuss it.

PR #153, opened 2020-05-29
Allow references to missing items of IdentifiedArray.

peterkovacs:
I started some discussion of this on the [Swift Forums](https://forums.swift.org/t/allow-sending-actions-to-deleted-identifiedarray-items/36999/4).

After digging into the crash, and the implementation of `forEach`, I think it's just a bug in IdentifiedArray.

This pull request fixes the issue for DEBUG builds by allowing a newValue of `nil` if there is no key for id present in the dictionary.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
@@ -110,7 +110,7 @@ where ID: Hashable {
             """
           )
         }
-        if newValue == nil {
+        if newValue == nil, self.dictionary[id] != nil {
           fatalError(
             """
             Can't update element with identifier \(id) with nil.
--- a/Tests/ComposableArchitectureTests/IdentifiedArrayTests.swift
+++ b/Tests/ComposableArchitectureTests/IdentifiedArrayTests.swift
@@ -195,4 +195,10 @@ final class IdentifiedArrayTests: XCTestCase {
       ], array)
 
   }
+
+    func testReferenceToMissing() {
+        var array = IdentifiedArray<Int, Int>([1], id: \.self)
+        func takesReference(_ item: inout Int?) { }
+        takesReference(&array[id: 3])
+    }
 }


stephencelis:
@peterkovacs Thanks for taking the time to PR. Do you think you can provide us with a small demo showing the crash you were getting? There are some bugs in SwiftUI around how views access mutable state (like arrays as rows are removed and optionals that go `nil`) that TCA tries to work around, and is one of the reasons why `forEach` and `IdentifiedArray` currently work the way they do.

If the crash is due to SwiftUI behavior it'd definitely be nice to have a workaround, though we also wanna ensure that `IdentifiedArray` itself isn't interacted with in unexpected ways.

peterkovacs:
@stephencelis, I was getting the crash not in SwiftUI, but purely in my Reducers. In my app, I have a state reducer remove some substate based on the results of network request. Meanwhile, some sub-state reducers could still have `delay` or network Effects still active. The test in the PR shows the crash (if you revert the change I made to the subscript setter).

The non-DEBUG `_modify` accessor does not exhibit this behavior.  That is, an action mentioning a non-existent `id` does not cause a crash.

peterkovacs:
I'll work on generating a more realistic example of the crash.

stephencelis:
@peterkovacs The behavior of the subscript _is_ intentional to avoid accidental misuse, so I was wondering if seeing the full lifecycle outside the reducer (since actions get fed into then via view stores) would help me better understand the conditions.

In this case it would seem to me that a delayed effect should not try to directly access state that may not exist later on. I envision a few ways one could write code around this:

1. Make the effect cancellable and return `.cancel` with the corresponding ID for whatever action destroys that row.
2. Add `IdentifiedArray.contains(id:) -> Bool` for probing the array for this information in a deliberate, safe manner.

peterkovacs:
@stephencelis, I don't know if this counts as "realistic" but here is an example that crashes:

```swift
import ComposableArchitecture
import Foundation

struct State: Equatable {
    var data: IdentifiedArrayOf<Child>
}

struct Child: Equatable, Identifiable {
    let id: Int
    var value: Int
}

enum Action {
    case removeAll
    case child(id: Int, action: ChildAction)
}

enum ChildAction {
    case start
    case setValue(Int)
}

let childReducer = Reducer<Child, ChildAction, Void> { state, action, _ in
    switch action {
    case .start:
        return Effect(value: .setValue(10))
            .delay(for: 5, scheduler: DispatchQueue.main)
            .eraseToEffect()

    case .setValue(let value):
        state.value = value
        return .none
    }
}

let appReducer = Reducer<State, Action, Void> { state, action, _ in
    switch action {
    case .removeAll:
        state.data = []
        return .none
    case .child:
        return .none
    }
}
    .combined(with: childReducer.forEach(state: \.data, action: /Action.child, environment: { () }))


let store = Store(
    initialState: State(data: [Child(id: 1, value: 1)]),
    reducer: appReducer.debug(),
    environment: ()
)

let viewStore = ViewStore(store)

viewStore.send(.child(id: 1, action: .start))
viewStore.send(.removeAll)
// *crash* after 5 seconds
```

With the change, the action is received, but the store does nothing to it because of the `Reducer.optional` call in `Reducer.forEach`.

You are correct that I could `cancel` all of the outstanding Effects that refer to the id, but in my case, some of these fire off URLRequests, and the user expects them to be completed.

I thought about creating my own version of `forEach` that looked like this:

```swift
    func forEachIfPresent<GlobalState, GlobalAction, GlobalEnvironment, ID>(
        state toLocalState: WritableKeyPath<GlobalState, IdentifiedArray<ID, State>>,
        action toLocalAction: CasePath<GlobalAction, (ID, Action)>,
        environment toLocalEnvironment: @escaping (GlobalEnvironment) -> Environment
    ) -> Reducer<GlobalState, GlobalAction, GlobalEnvironment> {
        .init { globalState, globalAction, globalEnvironment in
            guard let (id, localAction) = toLocalAction.extract(from: globalAction) else { return .none }

            if globalState[keyPath: toLocalState][id: id] != nil {
                return
                    self.optional(&globalState[keyPath: toLocalState][id: id],
                                  localAction,
                                  toLocalEnvironment(globalEnvironment))
                        .map { toLocalAction.embed((id, $0)) }
            } else { return .none }
        }
    }
```

This works, but it's similarity to `forEach` made me think that its actually a bug in IdentifiedArray.

peterkovacs:
Based on your comments, here and [over on the Swift Forums](https://forums.swift.org/t/allow-sending-actions-to-deleted-identifiedarray-items/36999/7), I think I get what you're saying.

Perhaps, instead of this change, I could clean up the `forEachIfPresent` and make the `forEach` more explicitly check that `id` is present?

stephencelis:
@peterkovacs Thanks for taking the time to share this repro! Being able to play with the full picture has me considering things differently now. I'm inclined to think that this PR _should_ be merged, though I'd like to think through some of the repercussions, because silently ignoring actions can lead to difficult-to-debug behavior. For what it's worth, the `optional` higher-order reducer that comes with TCA has this exact same "problem" in that it fails to invoke its reducer when state is optional even when it receives a relevant action. Maybe these debugging issues can be unified and solved another way, though.

So let's think about this a bit more. I definitely think it should probably behave the same way `optional` does, so either:

- Both `optional` and `forEach` should ignore incoming actions (and ideally provide a debugging story for ignoring these), or
- both `optional` and `forEach` should assert when receiving actions with no associated state, but have clear, ergonomic workarounds.

pcolton:
FWIW, the fatal errors have helped me uncover more than one bug during my development [so far].

stephencelis:
Closing this in favor of #154 as well. Thanks!

PR #154, opened 2020-05-29
Make forEach explictly check that the ID is present in the IdentifiedArray

peterkovacs:
As we talked about in #153, This adds an explicit check to ensure that actions reference IDs present in IdentifiedArrays.

Currently there is a crash that seems unintentional. This enforces the semantics of `forEach` in a clear way.


stephencelis:
Thanks for this! Gonna merge and tweak the wording to match the others.

PR #155, opened 2020-05-29
Add `forEachIfPresent` which will ignore actions that reference Ids not present in IdentifiedArray

peterkovacs:
Adding as an optional followup to #154.

This provides some ergonomics to allow a `forEach`-type higher-order reducer that will ignore actions that reference ids not present in the underlying IdentifiedArray.


stephencelis:
@peterkovacs I think we're gonna close this, but it's nice to know that folks can define their own if they so wish to get around this functionality!

PR #156, opened 2020-05-29
Reset cancellation test environment in tearDown and fix testCancellationAfterDelay

klop:
`EffectCancellationTests` rely on `resetCancellables()` to reset global state before each test is run. Rather than reaching for the `cancellationCancellables` dictionary and clearing it, we can simply clear out the local `cancellables` set after each test and rely on clean up behaviour that's already present in `cancellable` -- the issue with tests failing without explicit clean up is a consequence of test cases existing in memory after they've finished running.

I've also updated some of the timings in `testCancellationAfterDelay()` since it cannot fail in its current form -- there's a `0.1 s` wait to assert that a value that's meant to arrive in `0.5 s` has not arrived, which would be true even without cancellation.


Tests/ComposableArchitectureTests/EffectCancellationTests.swift
```diff
@@ -5,10 +5,10 @@ import XCTest
...
 final class EffectCancellationTests: XCTestCase {
   var cancellables: Set<AnyCancellable> = []
-
-  override func setUp() {
-    super.setUp()
-    resetCancellables()
+  
+  override func tearDown() {
+    super.tearDown()
+    cancellables.removeAll()
```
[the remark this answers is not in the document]
stephencelis: Nah, `removeAll` is great!

stephencelis, line 11: We _do_ like explicit `self` though. 😆
```suggestion
    self.cancellables.removeAll()
```

Tests/ComposableArchitectureTests/EffectCancellationTests.swift
```diff
@@ -83,7 +83,7 @@ final class EffectCancellationTests: XCTestCase {
         .store(in: &self.cancellables)
     }
 
-    _ = XCTWaiter.wait(for: [self.expectation(description: "")], timeout: 0.1)
+    _ = XCTWaiter.wait(for: [self.expectation(description: "")], timeout: 0.3)
```
stephencelis, line 86: Was this bump necessary?

Tests/ComposableArchitectureTests/EffectCancellationTests.swift
```diff
@@ -4,15 +4,15 @@ import XCTest
 @testable import ComposableArchitecture
 
 final class EffectCancellationTests: XCTestCase {
+  struct CancelToken: Hashable {}
```
stephencelis, line 7: Ah good change. Local cancel tokens are nice for actual reducers, but not for tests!

Tests/ComposableArchitectureTests/EffectCancellationTests.swift
klop: I'm not sure what the tolerances are for writing timing tests that don't flake, but I do think that this timeout has to exceed the delay on `Just(1)`, otherwise it'd be impossible to write a test that fails even when cancellation works.

stephencelis: Oh I gotcha, good catch!

stephencelis:
Not sure why tests are failing. Re-running them now, but maybe GitHub is just having temporary problems.

klop:
Thanks!

Issue #152, opened 2020-05-29
Crash after remove


buscarini:
I'm not sure if this is a bug or if I'm doing something wrong. I'm trying to make a small macos app, which has a navigation view inside another navigation view, so it would be a 3 panes app, a groups list, an items list and a detail view. If I try to remove a group or an item the app crashes. I started with simple arrays and int indexes, but I switched to an IdentifiedArrayOf, but nothing works:

`
Fatal error: Can't update element with identifier B699DB06-EFA9-4957-B0E9-DAAE764EE938 with nil.

If you are trying to remove an element from the array, use the "remove(id:) method.":`

Rest assured, I am using the remove(id:) method.

I'm also removing the selection when I remove an item if it's the selected one, but that doesn't update the detail panel.

pcolton:
You may want to take a look at this solution, your's might be similar issue: https://forums.swift.org/t/complex-use-case-for-shared-state-and-null-exception/36942

stephencelis:
@buscarini Can you ensure that an element with that UUID still in the array? We could probably tweak the messaging of that error in `remove(id:)`, or make `remove(id:)` "safe" (we kept it unsafe to be inline with `remove(at:)` in the Standard Library).

Also, if an element with that UUID _is_ still in the array, can you please fill out our issue template? It's difficult to debug without a repro:

    **Describe the bug**
    Give a clear and concise description of what the bug is.

    **To Reproduce**
    Zip up a project that reproduces the behavior and attach it by dragging it here.

    ```swift
    // And/or enter code that reproduces the behavior here.

    ```

    **Expected behavior**
    Give a clear and concise description of what you expected to happen.

    **Screenshots**
    If applicable, add screenshots to help explain your problem.

    **Environment**
     - Xcode [e.g. 11.4.1]
     - Swift [e.g. 5.2.2]
     - OS (if applicable): [e.g. iOS 13]

    **Additional context**
    Add any more context about the problem here.

buscarini:
I'll try to extract a simple example to reproduce this, but in the meantime here's more info:

1. There isn't an element with that UUID in the array, the id mentioned is the item that was just removed, so it makes sense that it's not in the array.
2. It has something to do with the navigation links, if I directly embed the detail view in the list there's no crash.

buscarini:
I found it!
The problem is I was using 
`NavigationLink init(destination: Destination, isActive: Binding<Bool>, @ViewBuilder label: () -> Label)`

[...]

PR #157, opened 2020-05-29
Add assertion to Reducer.optional

stephencelis:
This was spurred after a bunch discussions today: #152, #153, #154, #155.

Right now, `optional` reducers will happily toss actions they like to process into the void. This can lead to very subtle bugs where we are sending actions to the reducer and expecting things to happen, but because some sub-state is `nil` nothing happens at all.

Adding this assertion even revealed a "bug" in our TicTacToe logic, where we handle `nil`ing out `LoginState` upon login _before_ letting the `loginReducer` handle its own action. If loginReducer were instrumented with some analytics, we'd be losing those. So I think this (or something like it) would be helpful in the future.

(cc @peterkovacs)

### DIFF
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -176,12 +176,12 @@ public struct Reducer<State, Action, Environment> {
           set their state to "nil". This ensures that optional reducers can handle their actions \
           while their state is still non-"nil".
 
+          * An active effect emitted this action while state was "nil". Make sure that effects for this \
+          optional reducer are canceled when optional state is set to "nil".
+
           * This action was sent to the store while state was "nil". Make sure that actions for \
           this reducer can only be sent to a view store when state is non-"nil". In SwiftUI \
           applications, use "IfLetStore".
-
-          * An effect returned this action while state was "nil". Make sure that effects for this \
-          optional reducer are canceled when optional state is set to "nil".
           """
         )
         return .none
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -171,12 +171,14 @@ public struct Reducer<State, Action, Environment> {
           "\(debugCaseOutput(action))" was received by an optional reducer when its state was \
           "nil". This can happen for a few reasons:
 
-          * The optional reducer was combined with another reducer that ran and set \
-          "\(State.self)" to "nil" when it handled the same action. Combine the optional reducer \
-          before this reducer to ensure the optional reducer can handle its case.
+          * The optional reducer was combined with another reducer that set "\(State.self)" to \
+          "nil" when it handled this action. Combine optional reducers before reducers that can \
+          set their state to "nil". This ensures that optional reducers can handle their actions \
+          while their state is still non-"nil".
 
           * This action was sent to the store while state was "nil". Make sure that actions for \
-          this reducer can only be sent to a view store when state is non-"nil".
+          this reducer can only be sent to a view store when state is non-"nil". In SwiftUI \
+          applications, use "IfLetStore".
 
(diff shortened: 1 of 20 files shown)


stephencelis:
Example error:

> Fatal error: ".loginResponse(.success)" was received by an optional reducer when its state was "nil". This can happen for a few reasons:
>
> * The optional reducer was combined with another reducer that ran and set "LoginState" to "nil" when it handled the same action. Combine the optional reducer before this reducer to ensure the optional reducer can handle its case.
>
> * This action was sent to the store while state was "nil". Make sure that actions for this reducer can only be sent to a view store when state is non-"nil".
>
> * An effect returned this action while state was "nil". Make sure that effects for this optional reducer are canceled when optional state is set to "nil".

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -165,7 +165,27 @@ public struct Reducer<State, Action, Environment> {
...
+          "nil" when it handled this action. Combine optional reducers before reducers that can \
+          set their state to "nil". This ensures that optional reducers can handle their actions \
+          while their state is still non-"nil".
+
+          * This action was sent to the store while state was "nil". Make sure that actions for \
+          this reducer can only be sent to a view store when state is non-"nil". In SwiftUI \
+          applications, use "IfLetStore".
+
+          * An effect returned this action while state was "nil". Make sure that effects for this \
+          optional reducer are canceled when optional state is set to "nil".
```
mbrandonw, line 184:
```suggestion
          "\(debugCaseOutput(action))" was received by an optional reducer when its state was \
          "nil". This can happen for a few reasons:

          * The optional reducer was combined with another reducer that set "\(State.self)" to \
          "nil" when it handled this action. Combine optional reducers before reducers that can \
          set their state to "nil". This ensures that optional reducers can handle their actions \
          while their state is still non-"nil".

          * An active effect emitted this action while state was "nil". Make sure that effects for this \
          optional reducer are canceled when optional state is set to "nil".

          * This action was sent to the store while state was "nil". Make sure that actions for \
          this reducer can only be sent to a view store when state is non-"nil". In SwiftUI \
          applications, use "IfLetStore".
```

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -208,11 +228,22 @@ public struct Reducer<State, Action, Environment> {
...
+        can handle their actions for their state at the intended index.
+
+        * This action was sent to the store while state contained no element at this index. Make \
+        sure that actions for this reducer can only be sent to a view store when state contains an \
+        element at this index. In SwiftUI applications, use `ForEachStore`.
+
+        * An effect returned this action while state contained no element at this index. Make sure \
+        that effects for this "forEach" reducer are canceled whenever elements are moved or \
+        removed from state. If your "forEach" reducer returns any long-living effects, you should \
+        use the identifier-based "forEach", instead.
```
mbrandonw, line 246:
```suggestion
        "\(debugCaseOutput(localAction))" was received by a "forEach" reducer at index \(index) \
        when state contained no element at this index. This can happen for a few reasons:

        * The "forEach" reducer was combined with another reducer that removed an element from \
        state when it handled this action. Combine index-based "forEach" reducers before reducers \
        that can move or remove elements from their state. This ensures that "forEach" reducers \
        can handle their actions for their state at the intended index.

        * An active effect emitted this action while state contained no element at this index. Make sure \
        that effects for this "forEach" reducer are canceled whenever elements are moved or \
        removed from state. If your "forEach" reducer returns any long-living effects, you should \
        use the identifier-based "forEach", instead.
        
        * This action was sent to the store while state contained no element at this index. Make \
        sure that actions for this reducer can only be sent to a view store when state contains an \
        element at this index. In SwiftUI applications, use `ForEachStore`.
```

PR #161, opened 2020-06-01
Fix scheduling for debounce

mbrandonw:
Fixes the bug we just uncovered. Turns out that `debounce` works by scheduling interval work that is cancelled, and so if you are not careful about when such work is inserted into the queue it won't get properly removed and will just become recurring work forever. Easiest way to reproduce this is to do `.debounce` followed by `.receive(on:)`, because after the first trigger of the debounce the `receive(on:)` work is queued, and then accidentally removed, which leaves the debounce work around to execute ad infinitum 😬

The fix is to keep around a unique identifier for removing work after it is executed (instead of assuming `scheduled.removeFirst()` will remove the right thing), and to insert the next interval work into the queue immediately before executing the work.

### DIFF
--- a/Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
+++ b/Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
@@ -5,10 +5,10 @@ import Foundation
 public final class TestScheduler<SchedulerTimeType, SchedulerOptions>: Scheduler
 where SchedulerTimeType: Strideable, SchedulerTimeType.Stride: SchedulerTimeIntervalConvertible {
 
-  private var lastId: UInt = 0
+  private var lastSequence: UInt = 0
   public let minimumTolerance: SchedulerTimeType.Stride = .zero
   public private(set) var now: SchedulerTimeType
-  private var scheduled: [(id: UInt, date: SchedulerTimeType, action: () -> Void)] = []
+  private var scheduled: [(id: UUID, sequence: UInt, date: SchedulerTimeType, action: () -> Void)] = []
 
   /// Creates a test scheduler with the given date.
   ///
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
```diff
@@ -34,9 +34,9 @@ where SchedulerTimeType: Strideable, SchedulerTimeType.Stride: SchedulerTimeInte
     let delta = self.now.distance(to: nextDate)
     self.now = nextDate
 
-    while let (_, date, action) = self.scheduled.first, date == nextDate {
+    while let (id, _, date, action) = self.scheduled.first, date == nextDate {
       action()
-      self.scheduled.removeFirst()
+      self.scheduled.removeAll(where: { $0.id == id })
```
mbrandonw, line 39: `removeFirst` was not correct here because in the process of executing `action()` we may trigger the work's cancellable (as is the case with `debounce`), which will remove the work from the `scheduled` queue, which will then cause us to accidentally remove the wrong work from the queue.

Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
```diff
@@ -56,20 +56,20 @@ where SchedulerTimeType: Strideable, SchedulerTimeType.Stride: SchedulerTimeInte
...
-    let id = self.nextId()
+    let sequence = self.nextSequence()
 
     func scheduleAction(for date: SchedulerTimeType) -> () -> Void {
       return { [weak self] in
-        action()
         let nextDate = date.advanced(by: interval)
-        self?.scheduled.append((id, nextDate, scheduleAction(for: nextDate)))
+        self?.scheduled.append((UUID(), sequence, nextDate, scheduleAction(for: nextDate)))
+        action()
```
mbrandonw, line 65: it also wasn't right to execute this interval work immediately and then append the next work onto the queue. We need to do the reverse so that if the work's cancellable is called while invoking `action` then we will clean up _all_ of the interval work and not accidentally leave some stuff behind.

PR #163, opened 2020-06-01
Replace recursive call in Store.send with a while loop to process synchronous effects

### LINKED ISSUE #160: Stack overflow when .concatenating many Effects
**Describe the bug**
When queueing up a bunch of Effects via .concatenate, it seems that the nesting of .send() calls inside of Store causes a stack overflow. Maybe the answer is just 'don't do that many', but I wanted to at least bring it up in case there's a non-recursive way in Store to achieve the queueing up of synchronous Effects.

**To Reproduce**
Zip up a project that reproduces the behavior: (this is a Mac console app):

[group-tasks.zip](https://github.com/pointfreeco/swift-composable-architecture/files/4708333/group-tasks.zip)

**Expected behavior**
non-crash / stack overflow

**Screenshots**
Example of the thread stack when it crashes....

Screen Shot 2020-05-31 at 12 53 36 PM

**Environment**
 - Xcode 11.5
 - OS: macOS 10.15.5

klop:
This fixes #160.


mbrandonw:
Awesome, thanks @klop! Could we add a quick test that would have failed with the old style? this one should do the job:

```swift
  func testLotsOfSynchronousActions() {
    enum Action { case incr, noop }
    let reducer = Reducer<Int, Action, ()> { state, action, _ in
      switch action {
      case .incr:
        state += 1
        return state >= 100_000 ? Effect(value: .noop) : Effect(value: .incr)
      case .noop:
        return .none
      }
    }

    let store = Store(initialState: 0, reducer: reducer, environment: ())
    store.send(.incr)
    XCTAssertEqual(ViewStore(store).state, 100_000)
  }
```

klop:
Thanks @mbrandonw, I've added the test.

klop:
@stephencelis @mbrandonw do you run `swift-format` locally when writing code, and if so, is there anywhere I could learn about how to do that with your configuration?

stephencelis:
We run it automatically when code is pushed to `master` with this GitHub action: https://github.com/pointfreeco/swift-composable-architecture/blob/c4d4201284dc3cd265107d0bb14da792230d5a3f/.github/workflows/format.yml#L14-L21

So it shouldn't be necessary for you to run :)

If you _do_ want to run it on your own code or are just curious, it's a `brew install swift-format` away!

PR #164, opened 2020-06-01
Speed up TestScheduler work removal

stephencelis:
This probably will have a negligible effect on most test code, but it could improve larger stress tests.

### DIFF
--- a/Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
+++ b/Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
@@ -37,7 +37,9 @@ where SchedulerTimeType: Strideable, SchedulerTimeType.Stride: SchedulerTimeInte
 
     while let (id, _, date, action) = self.scheduled.first, date == nextDate {
       action()
-      self.scheduled.removeAll(where: { $0.id == id })
+      if let idx = self.scheduled.firstIndex(where: { $0.id == id }) {
+        self.scheduled.remove(at: idx)
+      }
     }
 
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
```diff
@@ -8,7 +8,7 @@ where SchedulerTimeType: Strideable, SchedulerTimeType.Stride: SchedulerTimeInte
   private var lastSequence: UInt = 0
   public let minimumTolerance: SchedulerTimeType.Stride = .zero
   public private(set) var now: SchedulerTimeType
-  private var scheduled: [(id: UUID, sequence: UInt, date: SchedulerTimeType, action: () -> Void)] =
```
stephencelis, line 11: Guess we didn't really need `id` 😆

Sources/ComposableArchitecture/Scheduling/TestScheduler.swift
```diff
@@ -35,9 +35,9 @@ where SchedulerTimeType: Strideable, SchedulerTimeType.Stride: SchedulerTimeInte
     let delta = self.now.distance(to: nextDate)
     self.now = nextDate
 
-    while let (id, _, date, action) = self.scheduled.first, date == nextDate {
+    while let (_, date, action) = self.scheduled.first, date == nextDate {
+      self.scheduled.removeFirst()
```
stephencelis, line 39: Just needed to call this first.

PR #168, opened 2020-06-03
Add Loading UI case study

sundeepgupta:
A case study that shows how a loading UI could be handled centrally for long-running actions. Any feedback is appreciated. 🙏 

LoadingUI

### DIFF
--- a/Examples/CaseStudies/CaseStudies.xcodeproj/project.pbxproj
+++ b/Examples/CaseStudies/CaseStudies.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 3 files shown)


mbrandonw:
Hey @sundeepgupta, thanks again for this case study. We're trying to find the right balance of showing off core use cases versus showing everything that is possible in TCA. We think that maybe this one is a little too specific, and is mostly covered by some of the other case studies. So, we are going to close for now, but we really do appreciate the exploration! 

Also, if yer interested in creating your own repo with case studies then I think it'd be great to feature it on https://github.com/antranapp/awesome-tca

sundeepgupta:
Makes sense - and thanks for the link, I'll check it out

PR #175, opened 2020-06-08
Adds Codable support to Location wrapper in ComposableCoreLocation

alskipp:
I needed to persist Location data, so thought it'd be useful if the `Location` wrapper was `Codable`.

In this implementation the wrapper just proxies the properties of `CLLocation` using `@dynamicMemberLookup`. Let me know what you think.

(Note: I removed the `floor` init parameter as it seems to be a computed property with no useful way of creating one from scratch)

### DIFF
--- a/Sources/ComposableCoreLocation/Models/Location.swift
+++ b/Sources/ComposableCoreLocation/Models/Location.swift
@@ -4,9 +4,72 @@ import CoreLocation
 /// and write tests against its values.
 
 @dynamicMemberLookup
-public struct Location: Equatable {
+public struct Location {
   public let rawValue: CLLocation
 
+  public init(
+    altitude: CLLocationDistance = 0,
+    coordinate: CLLocationCoordinate2D = CLLocationCoordinate2D(latitude: 0, longitude: 0),
+    course: CLLocationDirection = 0,
+    horizontalAccuracy: CLLocationAccuracy = 0,
+    speed: CLLocationSpeed = 0,
+    timestamp: Date = Date(),
+    verticalAccuracy: CLLocationAccuracy = 0
+  ) {
+    self.rawValue = CLLocation(
+      coordinate: coordinate,
+      altitude: altitude,
+      horizontalAccuracy: horizontalAccuracy,
+      verticalAccuracy: verticalAccuracy,
+      course: course,
+      speed: speed,
+      timestamp: timestamp
+    )
+  }
+
+  public init(rawValue: CLLocation) {
+    self.rawValue = rawValue
+  }
+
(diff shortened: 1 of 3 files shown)


alskipp:
Looks like I’ve been bit with backwards incompatibility with Xcode 11.3 & CLLocation. 😭
I’m out and about currently so will take a look later. 

I guess the `#if compiler(>=5.2)` in the original was there for that purpose?

Sources/ComposableCoreLocation/Models/Location.swift
```diff
@@ -2,80 +2,131 @@ import CoreLocation
...
-
-    self.altitude = rawValue.altitude
-    self.coordinate = rawValue.coordinate
-    self.course = rawValue.course
-    self.floor = rawValue.floor
-    self.horizontalAccuracy = rawValue.horizontalAccuracy
-    self.speed = rawValue.speed
-    self.timestamp = rawValue.timestamp
-    self.verticalAccuracy = rawValue.verticalAccuracy
-    #if compiler(>=5.2)
```
mbrandonw, line 55: Yeah, we wanna maintain 11.3 compatibility so we need to bring back those compiler checks unfortunately.

Sources/ComposableCoreLocation/Models/Location.swift
```diff
@@ -2,80 +2,131 @@ import CoreLocation
...
-    self.course = course
-    self.courseAccuracy = courseAccuracy
-    self.floor = floor
-    self.horizontalAccuracy = horizontalAccuracy
-    self.speed = speed
-    self.speedAccuracy = speedAccuracy
-    self.timestamp = timestamp
-    self.verticalAccuracy = verticalAccuracy
+    if #available(iOS 13.4, macCatalyst 13.4, macOS 10.15.4, tvOS 13.4, watchOS 6.2, *) {
+      self.rawValue = CLLocation(
```
mbrandonw, line 22: I didn't realize that `CLLocation` was constructible. Most Core Location types are not so I must have assumed `CLLocation` wasn't either. Thanks for simplifying this!

We will still have to bring back the `#if compiler` logic, either within the initializer or perhaps better would be to just make multiple initializers.

alskipp:
Oh my! Im looking into this now. Looks pretty ugly ☹️.

Multiple initialisers is probably the correct way to go (though it makes things more verbose).

One initialiser with compiler switching inside is shorter, but then `speedAccuracy` & `courseAccuracy` parameters are ignored when using an older compiler, which could be pretty confusing.

I should have something shortly – and I wouldn't blame you if you decided to bail out on this idea when you see it! No offence would be taken!

alskipp:
Hmm, test failed, but I think it might have been simulator going haywire in CI?

PR #177, opened 2020-06-08
Update Naming of Navigation Case Studies

stephencelis:
The naming of the types in these case studies no longer reflect the language we use in describing them, so seems worth updating.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,6 @@ test-workspace:
 		-quiet
 
 format:
-	swift format --in-place --recursive .
+	swift format --in-place --recursive ./Package.swift ./Sources ./Tests
 
 .PHONY: format test-all test-swift test-workspace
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -79,6 +79,6 @@ test-workspace:
 		-quiet
 
 format:
-	swift format --in-place --recursive .
+	swift format --in-place --recursive ./Package.swift ./Sources ./Tests
```
stephencelis, line 82: I thought we made this change earlier. I had some locked artifacts in `.build` that it was tripping on again.

Sources/ComposableArchitecture/Effect.swift
```diff
@@ -188,7 +188,7 @@ public struct Effect<Output, Failure: Error>: Publisher {
 
     return
       effects
-      .suffix(effects.count - 1)
+      .dropFirst()
```
stephencelis, line 191: Small thing I noticed.

PR #178, opened 2020-06-09
Add a more ergonomic way to create effects from delegate protocol implementations

klundberg:
I was analyzing how you implemented the live `LocationManager` for ComposableCoreLocation as I was looking to integrate another delegate protocol for the app I'm working on. I thought that it might be nice to have a more ergonomic, less boilerplate way to integrate a delegate protocol into an effect, so this PR is what I came up with.

All protocols that need to take advantage of this will need to also conform to the new `EffectDelegate` protocol I created, which just stores an Effect.Subscriber. That delegate protocol implementation is then expected to have each of its delegate methods call `subscriber?.send()` directly, instead of having to create closure properties for each delegate method (which gets very unwieldy really fast in my opinion). After that, you can create your Effect using one of the `fromDelegate` static methods easily, without having to assign lots of closures to a delegate instance.

The benefit of this approach (aside from far less boilerplate) is that it's much harder to forget to implement an analogous action for a given delegate protocol method. I realized this as I was converting ComposableCoreLocation to demonstrate this new approach, as you can see that I've added a case to `LocationManger.Action` to cover the delegate method that was previously missing a corresponding action.

Let me know what you think of this! I do have some qualms with how I've written things so far, and I'd love to discuss what some changes might look like:

* I'm not the biggest fan of the `fromDelegate` method name (and the parameters might deserve better names too). I thought of using `Effect.for(delegate:)` but that doesn't read as nicely when you need to create a delegate in a closure, like ComposableCoreLocation does. Any suggestions here would be welcome.
* I've constrained `EffectDelegate`s to never be allowed to send failures, which makes intuitive sense to me but I might be missing a possible use case here. if so, we can definitely let the failure type on `EffectDelegate` be generic as well.
* I tried to document usage of the new protocol/methods well, but I'm not practiced at this so I'm sure improvements can be made here, and I'm happy to make any edits that anyone feels is appropriate.


mbrandonw:
Hey @klundberg, thanks for the PR! You've definitely stumbled upon a really great simplification of this delegate dance we do. However, I think there's a spot between what we have today and what you have which may be the sweet spot.

I believe the core of what you have discovered is that it makes far more sense to simply hand the delegate the `subscriber` rather than have the delegate expose a bunch of closures which then wrap the `subscriber` and call `.send` on the inside of the closure. And I think this insight vastly cleans up the style even without the protocols.

Would you be willing to update the PR to remove the protocols and just pass the `subscriber` directly to the delegate? If you don't have time for that update I'm more than happy to help out.

Thanks!

klundberg:
Sure, I can experiment with that! That'll obviate some of my more specific concerns with what I've done here so far, I'll see what I can do :)

klundberg:
This was pretty easy to do actually. I haven't tested this on device so I don't know if there's any other specific core location gotchas. Your suggestions make this much simpler and I think this'll be a great guide for people looking to do something similar for any protocols they need to wrap into their apps using the architecture :). Let me know if there's anything else you want me to change!

Issue #182, opened 2020-06-11
Voice Memos Demo crashes when deleting memo while it is playing


krawutzi:
**Describe the bug**
App crashes when deleting memo while playing it back.

**To Reproduce**
- Record a long memo. 
- Start playback of that memo. 
- Before it ends, try to remove the memo by swiping.

```VoiceMemos[22126:1295688] Assertion failed: ".timerUpdated" was received by a "forEach" reducer at index 0 when its state contained no element at this index. This is considered an application logic error, and can happen for a few reasons:

* This "forEach" reducer was combined with or run from another reducer that removed the element at this index when it handled this action. To fix this make sure that this "forEach" reducer is run before any other reducers that can move or remove elements from state. This ensures that "forEach" reducers can handle their actions for the element at the intended index.

* An in-flight effect emitted this action while state contained no element at this index. To fix this make sure that effects for this "forEach" reducer are canceled whenever elements are moved or removed from its state. If your "forEach" reducer returns any long-living effects, you should use the identifier-based "forEach", instead.

* This action was sent to the store while its state contained no element at this index. To fix this make sure that actions for this reducer can only be sent to a view store when its state contains an element at this index. In SwiftUI applications, use `ForEachStore`.: file /Users/___/github/swift-composable-architecture/Sources/ComposableArchitecture/Reducer.swift, line 228

```

**Expected behavior**
The app does not crash when deleting a voice memo, even when it is currently being played.
(There is no more action sent to the single memo reducer, when the bigger reducer has already removed it.)

**Screenshots**
Not applicable.

**Environment**
 - Xcode [11.5]
 - Swift [5.2.4]
 - OS (if applicable): [iOS 13]

**Additional context**
Add any more context about the problem here.

krawutzi:
I guess this is related to https://github.com/pointfreeco/swift-composable-architecture/pull/154

mbrandonw:
Good find!

This is indeed as a result of some changes we've made recently to `forEach`, but it's really just pointing out that there is an application bug. In particular, this bullet point of the assertion failure message explains what is happening:

> * An in-flight effect emitted this action while state contained no element at this index. To fix this make sure that effects for this "forEach" reducer are canceled whenever elements are moved or removed from its state. If your "forEach" reducer returns any long-living effects, you should use the identifier-based "forEach", instead.

[...]

PR #183, opened 2020-06-13
Fix voice memos crasher

mbrandonw:
As reported in #182, we have a crash in the voice memos demo when you delete a memo while it is playing. This is happening because there is an in-flight timer effect that keeps running even after the data is removed, and hence we try to run the reducer logic on a row that no longer exists.

The fix is easy enough, we just need to cancel some effects when removing the row, and in fact the assertion we have in `forEach` precisely describes what the problem is, so that's nice!

### DIFF
--- a/Examples/VoiceMemos/VoiceMemos/VoiceMemos.swift
+++ b/Examples/VoiceMemos/VoiceMemos/VoiceMemos.swift
@@ -28,6 +28,7 @@ struct VoiceMemo: Equatable {
 enum VoiceMemoAction: Equatable {
   case audioPlayerClient(Result<AudioPlayerClient.Action, AudioPlayerClient.Failure>)
   case playButtonTapped
+  case remove
   case timerUpdated(TimeInterval)
   case titleTextFieldChanged(String)
 }
@@ -56,7 +57,8 @@ let voiceMemoReducer = Reducer<VoiceMemo, VoiceMemoAction, VoiceMemoEnvironment>
         environment.audioPlayerClient
           .play(PlayerId(), memo.url)
           .catchToEffect()
-          .map(VoiceMemoAction.audioPlayerClient),
+          .map(VoiceMemoAction.audioPlayerClient)
+          .cancellable(id: PlayerId()),
 
         Effect.timer(id: TimerId(), every: 0.5, on: environment.mainQueue)
           .map {
@@ -76,6 +78,12 @@ let voiceMemoReducer = Reducer<VoiceMemo, VoiceMemoAction, VoiceMemoEnvironment>
           .fireAndForget()
       )
     }
+
+  case .remove:
+    return .merge(
+      .cancel(id: PlayerId()),
+      .cancel(id: TimerId())
+    )
 
   case let .timerUpdated(time):
     switch memo.mode {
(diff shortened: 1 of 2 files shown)


Examples/VoiceMemos/VoiceMemos/VoiceMemos.swift
```diff
@@ -28,6 +28,7 @@ struct VoiceMemo: Equatable {
 enum VoiceMemoAction: Equatable {
   case audioPlayerClient(Result<AudioPlayerClient.Action, AudioPlayerClient.Failure>)
   case playButtonTapped
+  case remove
```
mbrandonw, line 31: I decided to add a new action to the domain of the voice memo row so that it is responsible for cleaning up its effects when it is removed.

Examples/VoiceMemos/VoiceMemos/VoiceMemos.swift
```diff
@@ -121,7 +129,6 @@ enum VoiceMemosAction: Equatable {
   case alertDismissed
   case audioRecorderClient(Result<AudioRecorderClient.Action, AudioRecorderClient.Failure>)
   case currentRecordingTimerUpdated
-  case deleteVoiceMemo(IndexSet)
```
mbrandonw, line 124: can remove this because we have the action above

Examples/VoiceMemos/VoiceMemos/VoiceMemos.swift
```diff
@@ -140,6 +147,12 @@ struct VoiceMemosEnvironment {
 }
 
 let voiceMemosReducer = Reducer<VoiceMemosState, VoiceMemosAction, VoiceMemosEnvironment>.combine(
+  voiceMemoReducer.forEach(
```
mbrandonw, line 150: moved the `forEach` to be combined before the rest of the app logic.

Examples/VoiceMemos/VoiceMemos/VoiceMemos.swift
```diff
@@ -262,16 +271,14 @@ let voiceMemosReducer = Reducer<VoiceMemosState, VoiceMemosAction, VoiceMemosEnv
       }
       return .none
 
+    case let .voiceMemo(index: index, action: .remove):
```
mbrandonw, line 274: can now listen for the `.remove` action in the voice memos row to handle removing it from state.

Examples/VoiceMemos/VoiceMemosTests/VoiceMemosTests.swift
```diff
@@ -281,12 +281,44 @@ class VoiceMemosTests: XCTestCase {
...
     store.assert(
-      .send(.deleteVoiceMemo(IndexSet(integer: 1))),
-      .send(.deleteVoiceMemo(IndexSet(integer: 0))) {
+      .send(.voiceMemo(index: 0, action: .remove)) {
         $0.voiceMemos = []
       }
     )
   }
+
+  func testDeleteMemoWhilePlaying() {
```
mbrandonw, line 290: this was a failing test before the change.

Examples/VoiceMemos/VoiceMemosTests/VoiceMemosTests.swift
```diff
@@ -281,12 +281,44 @@ class VoiceMemosTests: XCTestCase {
     )
 
     store.assert(
-      .send(.deleteVoiceMemo(IndexSet(integer: 1))),
```
mbrandonw, line 284: I don't know why we were removing this index that doesn't even exist in the state. it didn't crash because apparently `remove(atOffsets:)` doesn't crash when you remove at indices that don't exist.

PR #185, opened 2020-06-15
Debug printing for action labels

ollitapa:
I made a small debug print helper to provide action label printing. Idea was to make actions one-liners so they are a bit easier to read than using the regular debug() reducer. This is specially helpful if one is interested in the action sequence order and not so much on the state changes.

### DIFF
--- a/Sources/ComposableArchitecture/Debugging/ActionDebugging.swift
+++ b/Sources/ComposableArchitecture/Debugging/ActionDebugging.swift
@@ -1,0 +1,49 @@
+import Foundation
+
+/// Prints out a short label of given action leaving out large structs and arrays:
+/// Like .action1(.action2(...))
+func shortActionLabel(_ a: Any) -> String {
+  let m = Mirror(reflecting: a)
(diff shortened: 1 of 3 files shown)


ollitapa:
Seems like some unrelated timer test failed. Can I rerun the checks somehow?

mbrandonw:
Hey @ollitapa, thanks for the PR! This is a really good suggestion, but we rather have this functionality packaged up int the existing `.debug` method rather than creating a new one.

So, we just created the PR https://github.com/pointfreeco/swift-composable-architecture/pull/187 to support the idea of printing only the labels of an action versus pretty-printing the action. What do you think?

mbrandonw:
We just merged that PR so I'm going to close this one. However, if you think there are still some things that PR doesn't address feel free to re-open this or open an issue to discuss further!

Thanks again for the suggestion!

ollitapa:
Looks good! :+1: I agree that it's better to keep the debug printing in one method.

PR #188, opened 2020-06-16
Print offending action on Store.send assertion

ollitapa:
Added a debug print of an action sent recursively to the store. Useful if the stacktrace is long or Xcode does not show the action in the debug-tooltip (nothing to do with this library, but a bug that happens to me quite frequently).

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -144,8 +144,10 @@ public final class Store<State, Action> {
       if self.isSending {
         assertionFailure(
           """
-          The store was sent an action while it was already processing another action. This can \
-          happen for a few reasons:
+          The store was sent an action while it was already processing another action.
+          Sent action: \(debugCaseOutput(action))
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -144,8 +144,10 @@ public final class Store<State, Action> {
       if self.isSending {
         assertionFailure(
           """
-          The store was sent an action while it was already processing another action. This can \
-          happen for a few reasons:
+          The store was sent an action while it was already processing another action.
+          Sent action: \(debugCaseOutput(action))
+
+          This can happen for a few reasons:
```
mbrandonw, line 150:
```suggestion
          The store was sent the action \(debugCaseOutput(action)) while it was already 
          processing another action.

          This can happen for a few reasons:
```

ollitapa: I applied your suggestion 👍

PR #193, opened 2020-06-21
Composable Core Motion

mbrandonw:
Adds a new library to the package that provides a TCA-friendly interface in front of `CMMotionManager`. There are some other API's in Core Motion that could be converted over too, but I think the motion manager is probably the most popular API in the framework.

/cc @joeblau


Examples/MotionManager/MotionManager/MotionManagerView.swift
```diff
@@ -1,34 +1,43 @@
...
+  the Composable Architecture models its dependencies and effects, it is trivial to substitute \
+  a mock MotionClient into the SwiftUI preview so that we can still play around with its basic \
+  functionality.
 
-  Here we are creating a mock MotionClient that simulates motion data by running a timer that emits \
-  sinusoidal values.
+  We also have the background of the screen change colors depending on if the screen is facing \
+  you or facing away. We do this by computing how much the device's attitude has changed from \
+  the moment you started recording, and then checking the device yaw to see which way it is \
+  facing.
```
mbrandonw, line 17: added a new feature to this demo where we detect whether or not the device is facing you or facing away from you in order to change the background color of the screen.

Sources/ComposableCoreMotion/Live.swift
```diff
@@ -0,0 +1,255 @@
...
+          assertionFailure(
+            """
+            You are attempting to create a motion manager with the id \(id), but there is already
+            a running manager with that id. This is considered a programmer error since you may
+            be accidentally overwriting an existing manager without knowning.
+
+            To fix you should either destroy the existing manager before creating a new one, or
+            you should not try creating a new one before this one is destroyed.
+            """)
+        }
```
mbrandonw, line 28: I've made the `MotionManager` a bit more strict than I did for `LocationManager`, and if we like this style then I may update the location library too.

Sources/ComposableCoreMotion/Live.swift
```diff
@@ -0,0 +1,255 @@
...
+
+private func couldNotFindMotionManager(id: Any) {
+  assertionFailure(
+    """
+    A motion manager could not be found with the id \(id). This is considered a programmer error.
+    You should not invoke methods on a motion manager before it has been created or after it
+    has been destroyed. Refactor your code to make sure there is a motion manager created by the
+    time you invoke this endpoint.
+    """)
+}
```
mbrandonw, line 255: here too...

Examples/MotionManager/MotionManager/MotionManagerView.swift
```diff
@@ -39,33 +48,64 @@ let appReducer = Reducer<AppState, AppAction, AppEnvironment> { state, action, e
...
+    return .none
 
   case .recordingButtonTapped:
     state.isRecording.toggle()
-    return state.isRecording
-      ? environment.motionClient.startDeviceMotionUpdates(id: MotionClientId())
-        .fireAndForget()
-      : environment.motionClient.stopDeviceMotionUpdates(id: MotionClientId())
-        .fireAndForget()
+    if state.isRecording {
```
joeblau, line 84: Would this be better format as?
```
switch state.isRecording {
case true:
  // do true stuff
case false:
  // do false stuff
}
```

mbrandonw: 👍 just updated!

Sources/ComposableCoreMotion/Models/AccelerometerData.swift
```diff
@@ -0,0 +1,22 @@
+import CoreMotion
+
+/// A data sample from the device's three accelerometers.
+///
+/// See the documentation for `CMAccelerometerData` for more info.
+public struct AccelerometerData: Equatable {
+  var acceleration: CMAcceleration
```
kaandedeoglu, line 7: This should probably be `public`?

mbrandonw: thanks, good catch! done!

Sources/ComposableCoreMotion/Models/Attitude.swift
```diff
@@ -0,0 +1,109 @@
...
+  @inlinable
+  public var rotationMatrix: CMRotationMatrix {
+    let q = self.quaternion
+
+    let s =
+      1
+      / (self.quaternion.w * self.quaternion.w
+        + self.quaternion.x * self.quaternion.x
+        + self.quaternion.y * self.quaternion.y
+        + self.quaternion.z * self.quaternion.z)
```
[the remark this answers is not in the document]
mbrandonw: ha yup

PR #197, opened 2020-06-25
Add support for annotating steps in .assert()

lukeredpath:
This PR adds support for passing in a custom step annotation that allows steps to be logged in the test console output and also as `XCTActivity` for a nice display in the Xcode test results window.

The PR includes the console and XCTActivity implementations as well as a null implementation (the default) and the ability to combine annotations.

In addition, I've added a new `.group` step which allows you to logically group some of your steps which I think can be useful in longer, integration-style asserts. Groups can be nested and the group level is available to annotations so they can format appropriately.

I'm reasonably happy with the underlying implementation but I think there's a few issues that need addressing:

* I'm not sure if the API of passing in the annotators to the `.assert()` method in the `annotateTo:` parameter is ideal ergonomically. Could it be passed into the initialiser of the store instead?

* The `Annotating.swift` file which defines the activity implementation imports `XCTest` which I've noticed you've gone to an effort to avoid doing in `TestStore.swift` - some advice on how to clean this up would be appreciated.

The proof of the pudding as they say...here's the test case showing it's use:

```swift
store.assert(
  .group("Incrementing steps",
    .send(.increment) {
      $0 = 1
    },
    .send(.increment) {
      $0 = 2
    }
  ),
  .group("Decrementing steps",
    .send(.decrement) {
      $0 = 1
    },
    .send(.decrement) {
      $0 = 0
    }
  ),
  .group("Send and receive with effects",
    .send(.produceEffect),
    .receive(.effectProduced)
  ),
  .send(.increment) {
    $0 = 1
  },
  annotateTo: .combine(.console, .activity)
)
```

The console output:
Screenshot 2020-06-25 at 13 22 21

The Xcode test report output using XCTActivity:
Screenshot 2020-06-25 at 13 28 25

### DIFF
--- a/Sources/ComposableArchitecture/TestSupport/Annotating.swift
+++ b/Sources/ComposableArchitecture/TestSupport/Annotating.swift
@@ -11,7 +11,7 @@ import XCTest
 
 extension TestStore.Annotating {
   public static var activity: Self {
-    Self { step, callback in
+    Self { step, _, callback in
       var activityName: String!
       
       switch step.type {
@@ -33,20 +33,25 @@ extension TestStore.Annotating {
   }
   
   public static var console: Self {
-    Self { step, callback in
+    Self { step, groupLevel, callback in
+      func console(_ string: String) {
+        let indent = String(repeating: "\t", count: groupLevel)
+        print("\(indent)\(string)")
+      }
+      
       switch step.type {
       case let .send(action, _):
-        print("\t send: \(action)")
+        console("send: \(action)")
       case let .receive(action, _):
-        print("\t receive: \(action)")
+        console("receive: \(action)")
       case let .group(name, _):
-        print("TestStore assert group: '\(name)' started at \(Date())")
+        console("TestStore assert group: '\(name)' started at \(Date())")
       default:
         return
       }
       
       callback() { stepPassed in
-        print("\t\t [\(stepPassed ? "PASS" : "FAIL")]")
+        console("\t [\(stepPassed ? "PASS" : "FAIL")]")
       }
     }
   }
--- a/Sources/ComposableArchitecture/TestSupport/Annotating.swift
+++ b/Sources/ComposableArchitecture/TestSupport/Annotating.swift
@@ -16,18 +16,18 @@ extension TestStore.Annotating {
       
       switch step.type {
       case let .send(action, _):
-        activityName = "send: \(action)"
+        activityName = "--> send: \(action)"
       case let .receive(action, _):
-        activityName = "receive: \(action)"
+        activityName = "<-- receive: \(action)"
       case let .group(name, _):
         activityName = name
       default:
-        callback()
+        callback() { _ in }
         return
       }
       
       XCTContext.runActivity(named: activityName) { _ in
-        callback()
+        callback() { _ in }
       }
     }
   }
@@ -45,7 +45,9 @@ extension TestStore.Annotating {
         return
       }
       
-      callback()
+      callback() { stepPassed in
+        print("\t\t [\(stepPassed ? "PASS" : "FAIL")]")
+      }
     }
   }
 }
--- a/Sources/ComposableArchitecture/TestSupport/Annotating.swift
+++ b/Sources/ComposableArchitecture/TestSupport/Annotating.swift
@@ -1,10 +1,3 @@
-//
-//  File.swift
-//  
-//
-//  Created by Luke Redpath on 25/06/2020.
-//
-
 #if DEBUG
 
 import XCTest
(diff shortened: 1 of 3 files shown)


lukeredpath:
Looks like there's an issue running tests that use `XCTActivity` on CI:

```
Fatal error: XCTContext.runActivity(named:block:) failed because activities are disallowed in the current configuration
```

mbrandonw:
Hey @lukeredpath, thanks for the PR, this is very cool!

Unfortunately we can't access `XCTest` directly in the `ComposableArchitecture` library because it gets built for applications, and that is the error you are seeing on CI. We previously had split out a `ComposableArchitectureTestSupport` library for holding these kinds of things, but due to a bug in SPM we had to merge the two (see #71).

So, the only way to get this merged is to figure out a way to not `import XCTest` and to dynamically invoke `XCTContext.runActivity`. Stephen and I just played around with doing it via `NSClassFromString` and `performSelector`, but we couldn't find the right magic to make it work.

Could you try that out to see if you can work around somehow, and if you succeed we can move forward?

Thanks again!

lukeredpath:
Thanks for the feedback - looks like this issue is a bit of a tricky one to resolve. I'm not sure I can really offer any solution to work around this other than extracting the test support into a completely separate package.

mbrandonw:
I think the only problem with that is providing the necessary hooks in `TestStore.assert` for an external package to be able to plug functionality into. Not sure we want to fully support that kind of escape hatch in the API right now.

Maybe we can get some help on this problem on twitter or the swift forums. Seems like it should be totally solvable.

lukeredpath:
@mbrandonw been thinking about this - for now, if you like the general idea, how do you feel about shipping the core functionality (group and the Annotating abstraction) and just the null and console implementations, and include the XCTest activity implementation as an example not included in the actual library? 

It's not ideal - it means people will need to drop the example into their project manually if they want to use it, but it can be added back in future if the underlying issue is solved.

With that in mind, do you think there's any improvements that can be made to the implementation or the API for using the annotations?

stephencelis:
@lukeredpath Chatted about this again today and we would love to figure out how to land it, but we do think it's important to land as a whole rather than just the console/null implementations, or we think that instead we should figure out how to make this API open to extension, instead, so that a feature like this can be implemented in another library.

I guess the sad reality is that SPM is in a very fragile state with regard to how dependencies are resolved statically/dynamically, and this creates some huge complications, as we're seeing here. If you're unable to tackle the dynamic linking issue, should we push pause on this PR for now?

lukeredpath:
I'd be happy to move implementations into a separate library if we can get the extensible part of the API into the core library - I'm already starting to think of other ways this could potentially be used (e.g. `XCTIssue`).

mbrandonw:
I just asked on twitter if anyone knows how to do this, maybe someone knows the right dose of magic! https://twitter.com/mbrandonw/status/1278331673785126914

mbrandonw:
I think I figured it out: https://github.com/pointfreeco/swift-composable-architecture/pull/197/commits/23c67809c3f7f610988272364b099746b1f3a4db

Will see what CI says 😄

mbrandonw:
Hm, nope 😞 

Works fine in Xcode, but crashes when running `swift test`.

mbrandonw:
So it works fine with `swift test` when using the Xcode 12 and Swift 5.3 command line, but not Xcode 11.4 and Swift 5.2.

We're thinking we could release this feature, we just wouldn't be able to have any tests in the library that demonstrate how to use it.

We'll think about it a bit more and come up with a plan forward.

lukeredpath:
@mbrandonw great work, could the tests be included in `#if compiled(>=5.3)` directives for now?

stephencelis:
Wanted to provide an update here. We chatted about this again this morning and were even able to push a small bit of logic to avoid the issue with `swift test` (though I still don't really understand why it's happening—we call to the same API in a non-dynamic way in our snapshot testing library and `swift test` behaves just fine).

We still wanna think through the API a bit before merging, though. We think maybe this feature should just always be on and not require the use to opt into it. Can you think of any reason why a user would want to opt out of or disable this functionality?

lukeredpath:
I think there's a good argument to be made for the XCTActivity annotation to be on by default, although I think it would be nice if it remained something you could hook into (the console logging seems like more of an opt-in thing to me).

If we moved this into the initialiser for the TestStore instead of an argument to the assert method, defaulted to to `.activity` then for most users that would be enough. If somebody wants to roll their own they can override (and still use `.combine` if they want to keep the XCTActivity feature). Alternatively, if you don't want the user to be able to opt-out of the XCTActivity functionality, you could just automatically combine whatever is passed in with `.activity`.

stephencelis:
@lukeredpath Given the recent changes to the TestStore DSL, what do you think of closing this till it can be revisited in a different form? We haven't deprecated `.assert` yet but likely will in the future.

lukeredpath:
@stephencelis makes sense to close this given recent changes. With the new API being normal method calls it seems like I could cook up my own abstraction around XCTActivity to group these calls together without needing to integrate with the TestStore now.

PR #198, opened 2020-06-25
Show how animation can be done from effect.

mbrandonw:


Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-AnimationsTests.swift
```diff
@@ -0,0 +1,63 @@
...
+      reducer: animationsReducer,
+      environment: AnimationsEnvironment(
+        mainQueue: self.scheduler.eraseToAnyScheduler()
+      )
+    )
+
+    store.assert(
+      .send(.rainbowButtonTapped),
+      .receive(.setColor(.red)) {
+        $0.circleColor = .red
```
stephencelis, line 22: Cool that colors are testable.

boudavid:
I'm facing the same problem with animations and I felt that I was cheating for using delays in order to have the animation played correctly before moving on. So it's nice to see that you're using the same technique 😂. I also didn't really like that I needed to match the delay duration to the animation duration.

mbrandonw:
> I'm facing the same problem with animations and I felt that I was cheating for using delays in order to have the animation played correctly before moving on. So it's nice to see that you're using the same technique 😂. I also didn't really like that I needed to match the delay duration to the animation duration.

haha, well we were just doing this for demonstration purposes. can you explain your situation a bit more? either here or on the swift forums?

boudavid:
Basically, I have a modal with two choices represented by two buttons. When you tap on a choice, it animates the text by scaling it down and scaling it up and then the modal is dismissed.

The problem was that without using a delay, the animation was never played because the modal was dismissed before you could see it. So I needed to not use the same state property for the animation and for the modal.

The way I managed to achieve this animation is a bit hacky :
1. First I needed to add a custom `ButtonStyle` for the "pressed effect" (scaling the text down) because you have access to an `isPressed` property in `ButtonStyleConfiguration`. So when I tap the button, it plays the first part of the animation. Not a big fan but it gets the job done while SwiftUI gets more mature and we have more Buttons APIs.
2. Second, in the reducer, for the `tappedChoice` action, I mutated a `tappedChoice` property to my state to the selected choice so the animation plays on the correct button.
3. The return Effect of the `tappedChoice` action is another action delayed by the animation duration that takes care of mutating the state property that is binded to the modal display/dismissal.

mbrandonw:
Ah yes I see, I think the delay is the best way to handle that unfortunately. I think you'd basically have to do the same thing even if you were using vanilla SwiftUI :/

PR #199, opened 2020-06-25
Don't use @Published for ViewStore state

stephencelis:
Combine's beta includes a new `assign(to:)` on publisher that can feed state updates to a `@Published` publisher. We don't want this, so let's ping `objectWillChange` directly.

https://developer.apple.com/documentation/combine/publisher/assign(to:)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   push:
     branches:
-      - master
+      - main
   pull_request:
     branches:
       - '*'
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -3,7 +3,7 @@ name: Format
 on:
   push:
     branches:
-      - master
+      - main
 
 jobs:
   swift_format:
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -66,7 +66,11 @@ public final class ViewStore<State, Action>: ObservableObject {
   }
 
   /// The current state.
-  @Published public internal(set) var state: State
+  public private(set) var state: State {
+    willSet {
+      self.objectWillChange.send()
+    }
+  }
 
   let _send: (Action) -> Void
 


DrewKiino:
I just updated my TCA to `0.6.0` and my SwiftUI `ViewStore` doesn't re-update it's state anymore. I did a small app with a `@ObservableObject` class and an inner variable with out `@Published` to mimic this new change and the view itself could not recognize the update until I added `@Published` back in. I think this breaks all iOS versions below `14.0`.

How is the View suppose to update if `@Published` is removed from the inner state for `ViewStore` now? Really need this because we're using it for our production apps.

mluisbrown:
> How is the View suppose to update if @Published is removed from the inner state for ViewStore now? Really need this because we're using it for our production apps.

All that @Published does is call `self.objectWillChange.send()` automatically before any change to the published variable belonging to a class conforming to `ObservableObject` so the change made in in this PR is just mimicking the functionality of `@Published` explicitly, therefore it should behave exactly as before.

I know this works since I had to do exactly the same thing in my ReactiveSwift version of the TCA.

mbrandonw:
@DrewKiino sorry to hear about that. All of the demo apps in the repo seem to work just as they did before.

And as @mluisbrown mentioned, we are just manually replicating what `@Published` gives to us for free, and we do that so that we can work around an escape each that Combine has introduced in iOS 14 allowing anyone from the outside to make changes to the state without sending an action.

Is it possible to share some code that would help us reproduce the problem you are seeing?

DrewKiino:
Thanks for the quick reply 👍🏼  I was able to figure it out, it turns out that `Xcode 11.4` doesn't support the new change for removing `@Published` and using `objectWillChange`, SwiftUI simply does not receive any updates. When I upgraded to `Xcode 11.5`, `0.6.0` started working for me so I think it was just a matter of Apple simply making a deprecated update that isn't as clear to most developers which was unfortunate.

https://www.hackingwithswift.com/quick-start/swiftui/how-to-send-state-updates-manually-using-objectwillchange
```
// quote from the above article
Note: SwiftUI actually provides us with a default objectWillChange property, but it doesn’t actually work in the current beta. So, it’s important to declare your own with the same name.
```
Which you can fix by adding this into your `ObservableObject` subclass 
```swift
let objectWillChange = ObservableObjectPublisher()
```

Another article with the changes -
https://sarunw.com/posts/swiftui-changes-in-xcode-11-beta-5/

So basically, `0.6.0` requires `Xcode 11.5` for the new changes to take effect due to Apple not making these changes backwards compatible.

mbrandonw:
Strange! I don't have Xcode 11.4 installed, but I do have Xcode 11.3 and all the case studies still seems to work, even without adding an `objectWillChange` to the `ViewStore` class. We want to support Xcode 11.3 and up, so if this is a problem I'd love to fix it. Can you describe how I might reproduce this locally?

mluisbrown:
@DrewKiino the posts you linked to are both referring to Xcode 11.0 *betas* from 2019, where this might have been a bug.

However in the ReactiveSwift version of ViewStore I [do explicitly create the `ObservableObjectPublisher`](https://github.com/trading-point/reactiveswift-composable-architecture/blob/master/Sources/ComposableArchitecture/ViewStore.swift#L79), and I don't remember why I did that 🤷 . If I remove it all the Case Studies still work fine, although I am using Xcode 11.5.

mluisbrown:
I also just found [this thread](https://forums.swift.org/t/question-about-valid-uses-of-observableobject-s-synthesized-objectwillchange/31141) on the Swift Forums which seem to indicate that at least in November 2019, if you didn't have any `@Published` members, you had to instantiate your own `ObservableObjectPublisher`.

This [comment](https://forums.swift.org/t/question-about-valid-uses-of-observableobject-s-synthesized-objectwillchange/31141/2) by @broadwaylamb really makes this clear:

> Since there's no `@Published` properties in your class, there's no storage `objectWillChange` could install the publisher into, so it just creates and returns a new instance every time you access `objectWillChange`.

I'm not sure when or if this behaviour changed, but IMO if you're not using any `@Published` properties, it would be safer to always instantiate your own `ObservableObjectPublisher`.

broadwaylamb:
@mluisbrown it has changed in the recent versions of Combine. It now falls back to a global lookup table if there are no @Published properties in the class.

mluisbrown:
Thanks for the really fast reply @broadwaylamb 🙏  Do you know in which version of Combine this changed, and when the change was made?

broadwaylamb:
Unfortunately, no.

PR #201, opened 2020-06-28
Generic alerts and action sheets

mbrandonw:
This creates some new generic state types that make it easy to work with alerts and action sheets in the TCA. The code is mostly straightforward, and has even cleaned up some code in our case studies, but I think a little more thought needs to be given to one part of the API.


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndActionSheets.swift
```diff
@@ -0,0 +1,148 @@
...
+      Form {
+        Section(header: Text(template: readMe, .caption)) {
+          Text("Count: \(viewStore.count)")
+
+          Button("Alert") { viewStore.send(.alertButtonTapped) }
+            .alert(
+              viewStore.alert,
+              send: viewStore.send,
+              dismissal: .alertCancelTapped
+            )
```
mbrandonw, line 121: this is the bit of API that I think can maybe be improved. Right now we need to pass along the state that represents the alert, the view store's `send` so that actions can be sent when interacting with buttons in the alert, and we further need a dismissal action to fire off when the alert is closed for non-button reasons (the system can close alerts without user interaction).

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndActionSheets.swift
```diff
@@ -0,0 +1,148 @@
...
+
+          Button("Action sheet") { viewStore.send(.actionSheetButtonTapped) }
+            .actionSheet(
+              viewStore.actionSheet,
+              send: viewStore.send,
+              dismissal: .actionSheetCancelTapped
+            )
```
mbrandonw, line 128: similarly this API could also be improved a bit

Makefile
```diff
@@ -67,6 +67,7 @@ test-workspace:
 		-destination platform="$(PLATFORM_IOS)"
 
 format:
-	swift format --in-place --recursive ./Package.swift ./Sources ./Tests
+	swift format --in-place --recursive \
+		./Examples ./Package.swift ./Sources ./Tests
```
mbrandonw, line 71: we weren't formatting `./Examples` so I added that and it added some noise to this pr

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-SharedState.swift
```diff
@@ -205,13 +206,9 @@ struct SharedStateCounterView: View {
...
       .navigationBarTitle("Shared State Demo")
       .alert(
-        item: viewStore.binding(
-          get: { $0.alert.map(PrimeAlert.init(title:)) },
-          send: .alertDismissed
-        )
-      ) { alert in
-        SwiftUI.Alert(title: Text(alert.title))
-      }
+        self.store.scope(state: \.alert),
```
mbrandonw, line 209: this may be fixed in one of your other branches, but to get things building on Xcode 11.3:

```suggestion
        self.store.scope(state: { $0.alert }),
```

PR #203, opened 2020-06-30
Add Headphone Motion Manager

stephencelis:
This adds iOS 14's new `CMHeadphoneMotionManager` to `ComposableCoreMotion`. An experiment in incrementally improving these APIs even with beta APIs.


Sources/ComposableCoreMotion/HeadphoneMotionManager/HeadphoneMotionManagerInterface.swift
```diff
@@ -0,0 +1,95 @@
+import ComposableArchitecture
+import CoreMotion
+
+/// A wrapper around Core Motion's `CMHeadphoneMotionManager` that exposes its functionality through
+/// effects and actions, making it easy to use with the Composable Architecture, and easy to test.
```
mbrandonw, line 5: should we add more docs here about usage like we've done with other support libraries?

stephencelis: Yeah...I tried making a demo but it looks like this API doesn't even work yet: https://developer.apple.com/forums/thread/650051

Converted to a draft that we can revisit in later betas.

andreyz:
Time to revisit and merge this piece of cool API?

stephencelis:
Got around to testing this on a device and it works! Closing this draft, though, in favor of a new PR.

PR #204, opened 2020-07-01
Pullback all case studies into mega reducer

mbrandonw:


Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-ResuableOfflineDownloads/DownloadComponent.swift
```diff
@@ -125,7 +125,7 @@ private let deleteAlert = AlertState(
 
 private let cancelAlert = AlertState(
   title: "Do you want to cancel downloading this map?",
-  primaryButton: .cancel(send: .cancelButtonTapped),
+  primaryButton: .destructive("Cancel", send: .cancelButtonTapped),
```
mbrandonw, line 128: this was a change from #201, but the cancel button should be a destructive button since it is cancelling the download, not just dismissing the alert.

PR #205, opened 2020-07-01
Make location getter work like Corelocation.location getter

mackoj:
In CoreLocation the location getter is optional and here it raise a fatal error.

https://developer.apple.com/documentation/corelocation/cllocationmanager/1423687-location

This PR make it optional.

### DIFF
--- a/Sources/ComposableCoreLocation/Interface.swift
+++ b/Sources/ComposableCoreLocation/Interface.swift
@@ -272,7 +272,7 @@ public struct LocationManager {
   @available(watchOS, unavailable)
   public var isRangingAvailable: () -> Bool = { _unimplemented("isRangingAvailable") }
 
-  var location: (AnyHashable) -> Location = { _ in _unimplemented("location") }
+  var location: (AnyHashable) -> Location? = { _ in nil }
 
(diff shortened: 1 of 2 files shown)


Sources/ComposableCoreLocation/Interface.swift
```diff
@@ -272,7 +272,7 @@ public struct LocationManager {
   @available(watchOS, unavailable)
   public var isRangingAvailable: () -> Bool = { _unimplemented("isRangingAvailable") }
 
-  var location: (AnyHashable) -> Location = { _ in _unimplemented("location") }
+  var location: (AnyHashable) -> Location? = { _ in nil }
```
mbrandonw, line 275: Oh, good catch!

It's actually ok to leave this as `_unimplemented` because we are supposed to override this endpoint in the live implementation:

```suggestion
  var location: (AnyHashable) -> Location? = { _ in _unimplemented("location") }
```

But, it seems like we forgot to set `manager.location` in `Live.swift`. Would you be willing to add this line to that file to finish this off?

```swift
manager.location = { id in dependencies[id]?.manager.location.map(Location.init(rawValue:)) }
```

PR #211, opened 2020-07-06
[Experiementation] Change the diff algo

mackoj:
##  Please do not merge this

Hi,

This is an experiment where I tried another [diff algo](https://github.com/krzysztofzablocki/Difference).

## Issue

When you have big and complex state, Xcode inlined red test error popup don't display the whole state. And that long time to 
open.

| description | image |
| --- | --- |
| the popup is not completely visible | Capture d’écran 2020-07-06 à 14 49 35 |
| it is impossible to go to the bottom of the popup | Capture d’écran 2020-07-06 à 14 49 51 |
| the only way is to go to the test navigator | Capture d’écran 2020-07-06 à 14 50 07 |

## Solution

Only display what changed inside the Xcode inlined red test error popup.

I tried using https://github.com/krzysztofzablocki/Difference for that.

Capture d’écran 2020-07-06 à 14 54 05

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -27,6 +27,7 @@ let package = Package(
   dependencies: [
     .package(url: "https://github.com/pointfreeco/combine-schedulers", from: "0.1.0"),
     .package(url: "https://github.com/pointfreeco/swift-case-paths", from: "0.1.1"),
+    .package(url: "https://github.com/krzysztofzablocki/Difference.git", .branch("master")),
   ],
   targets: [
     .target(
@@ -34,6 +35,7 @@ let package = Package(
       dependencies: [
         "CasePaths",
         "CombineSchedulers",
+        "Difference",
       ]
     ),
     .testTarget(
--- a/Sources/ComposableArchitecture/TestSupport/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestSupport/TestStore.swift
@@ -1,6 +1,7 @@
 #if DEBUG
   import Combine
   import Foundation
+  import Difference
 
   /// A testable runtime for a reducer.
   ///
@@ -225,13 +226,21 @@
           }
           let receivedAction = receivedActions.removeFirst()
           if expectedAction != receivedAction {
-            let diff =
-              debugDiff(expectedAction, receivedAction)
-              .map { ": …\n\n\($0.indent(by: 4))\n\n(Expected: −, Actual: +)" }
-              ?? ""
+//            let diff =
+//              debugDiff(expectedAction, receivedAction)
+//              .map { ": …\n\n\($0.indent(by: 4))\n\n(Expected: −, Actual: +)" }
+//              ?? ""
+            let nDiff : [String] = diff(
+              expectedAction,
+              receivedAction,
+              indentationType: Difference.IndentationType.tab,
+              skipPrintingOnDiffCount: false
+            )
+            let jDiff = nDiff.joined(separator: "\n")
+
             _XCTFail(
               """
-              Received unexpected action\(diff)
+              Received unexpected action\(jDiff)
               """,
               file: step.file,
               line: step.line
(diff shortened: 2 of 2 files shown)


stephencelis:
@mackoj Thanks for starting the conversation! There definitely seem to be times where Xcode's errors become a bit difficult to navigate, and you've nicely highlighted a few.

Difference is a nice library but we don't think it's the right solution here since a lot of the time you want the full context of your feature's state to troubleshoot test failures. We also think that the `+`/`-` diff format is a lil easier to read at-a-glance. We've worked with abbreviated/truncated diffs in the past in our SnapshotTesting library, but we aren't quite sure it's the right thing to do for the Composable Architecture assertions.

For "the popup is not completely visible" is there a chance you can share what's causing it to be hidden? We've experimented a bit with very long strings and as long as the window isn't narrow, the message wraps for us and is fully visible.

In general I think we consider these to be Xcode bugs. Xcode should allow us to scroll the contents of that popup around. We'll file feedback today and encourage you to do so, as well. In the meantime, the test navigator should always work as a workaround.

stephencelis:
(Submitted feedback FB7857972.)

mackoj:
Thanks for the response. 

**TLDR:** I understand your point of view but I still think that having a way to represent the error in a truncated way could be helpful.

> Difference is a nice library but we don't think it's the right solution here since a lot of the time you want the full context of your feature's state to troubleshoot test failures. We also think that the +/- diff format is a lil easier to read at-a-glance. We've worked with abbreviated/truncated diffs in the past in our SnapshotTesting library, but we aren't quite sure it's the right thing to do for the Composable Architecture assertions.

Having the full context can be really helpful when you run into an issue(when fixing an already existing test) but most of the time when I build tests having a truncated diffs help me to implement them with more ease. Maybe the best is to have an option to display full context or not ?

I use Difference daily for testing it and because it was already build and in order to quickly see if having less context was better for me or not. I do not think using a third party library is a good solution either.

> For "the popup is not completely visible" is there a chance you can share what's causing it to be hidden? We've experimented a bit with very long strings and as long as the window isn't narrow, the message wraps for us and is fully visible.

I think it's because the content is too tall and since it doesn't seems to support scrolling it is truncated by Xcode.

I will file a radar too.

stephencelis:
@mackoj We'll definitely take this under consideration and continue to think about it. We think there's a middle ground between the behavior today and Difference and hope to explore it in the future. If you have any ideas you want to flesh out, please continue the discussion!

mackoj:
We are still using this fork because it's not usefull for us to have everything displayed.
In some case we have to much data in our state. What can I do to help on this ?

Maybe we should have multiple output ?
- full diff aka the actual diff
- medium diff aka a few lines above and after the diff ?
- minimal diff with only what's differ and where ?

stephencelis:
@mackoj Almost a year later, but we believe a lot of your concerns are covered in our new [Custom Dump](https://github.com/pointfreeco/swift-custom-dump). Let us know what you think!

mackoj:
Thanks 🙏 for making it. It's great.
I can now use the `.debug()` on reducer and having a understanding of what is happening and I did make my state debug property wrapper use it.

```swift
import Foundation
import CustomDump

@propertyWrapper
public struct Diff<Value: Equatable>: Equatable {
  public static func == (lhs: Diff<Value>, rhs: Diff<Value>) -> Bool {
    lhs.value == rhs.value
  }

  private var value: Value
  private(set) var previous: Value? = nil

  public init(wrappedValue: Value) {
    self.value = wrappedValue
  }

  public var wrappedValue: Value {
    get { value }

    set {
      previous = value
      if let output = diff(value, newValue) { print(output) }
      value = newValue
    }
  }

  public var projectedValue: Self {
    return self
  }
}
```

stephencelis:
@mackoj Very cool! Looks like a useful property wrapper to have around 😄

PR #212, opened 2020-07-06
Added a read only binding

gkaimakas:
Not sure if this is exactly useful in many contexts but I found myself trying to access a read only binding today and though I would pitch it to the library


stephencelis:
Hey @gkaimakas! SwiftUI actually has support for this kind of thing out of the box, with `Binding.constant`. This means you can do something like:

```swift
TextField("Placeholder", text: .constant(viewStore.text))
  .disabled(true)
```

gkaimakas:
@stephencelis yes yes, you are right! 

I forgot that the body is called everytime state changes :)

PR #213, opened 2020-07-06
Add a missing optional related to #205

mackoj:
Add missing optional related to #205


stephencelis:
Ah good catch, thanks!

PR #215, opened 2020-07-07
More docs around Reducer.combine and ordering

stephencelis:


alexito4:
FYI, I wanted to keep the parent first (just because ¯\_(ツ)_/¯ ) so I started using 

```
public func combined(after other: Reducer) -> Reducer {
      .combine(other, self)
}
```

Probably not appropriate for the library itself but a nice little helper :)

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -56,17 +56,75 @@ public struct Reducer<State, Action, Environment> {
     Self { _, _, _ in .none }
   }
 
-  /// Combines many reducers into a single one by running each one on the state, and merging
+  /// Combines many reducers into a single one by running each one on state in order, and merging
   /// all of the effects.
   ///
+  /// It is important to consider the order of reducers passed to this function, especially when
+  /// combining a child domain with a parent domain that can modify the child domain by listening to
+  /// its actions.
```
mbrandonw, line 64: I'm thinking we need to explicitly say that `combine(a, b) != combine(b, a)` before diving into the `optional` example. 

```suggestion
  /// It is important to note that the order of combining reducers matter. Combining `reducerA` with 
  /// `reducerB` is not necessarily the same as combing `reducerB` with `reducerA`.
```

And should we also mention this also applies to `forEach`?

stephencelis: Added some more language around `forEach`.

PR #218, opened 2020-07-08
Added file and line to optional assert

alexito4:
As [suggested in the forums](https://forums.swift.org/t/reducer-optional-assert-with-more-info/38281?u=alejandro_martinez) this PR adds #file and #line default parameters to an `optional()` reducer method to add the information into the assertion message.

Questions I would like to ask:
- Are you happy if the function replaces the property (breaking change) or do you want to keep both?
- Feel free to suggest edits on the wording of the message.
- What other places would you like to see this change? `forEach` maybe?

Once this is clarified I will update the examples and tests 😄

### DIFF
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -137,59 +137,6 @@ public struct Reducer<State, Action, Environment> {
     }
   }
 
-  /// Transforms a reducer that works on non-optional state into one that works on optional state by
-  /// only running the non-optional reducer when state is non-nil.
-  ///
-  /// Often used in tandem with `pullback` to transform a reducer on a non-optional local domain
-  /// into a reducer on a global domain that contains an optional local domain:
-  ///
-  ///     // Global domain that holds an optional local domain:
-  ///     struct AppState { var modal: ModalState? }
-  ///     struct AppAction { case modal(ModalAction) }
-  ///     struct AppEnvironment { var mainQueue: AnySchedulerOf<DispatchQueue> }
-  ///
-  ///     // A reducer that works on the non-optional local domain:
-  ///     let modalReducer = Reducer<ModalState, ModalAction, ModalEnvironment { ... }
-  ///
-  ///     // Pullback the local modal reducer so that it works on all of the app domain:
-  ///     let appReducer: Reducer<AppState, AppAction, AppEnvironment> =
-  ///       modalReducer.optional.pullback(
-  ///         state: \.modal,
-  ///         action: /AppAction.modal,
-  ///         environment: { ModalEnvironment(mainQueue: $0.mainQueue) }
-  ///       )
-  ///
-  /// - See also: `IfLetStore`, a SwiftUI helper for transforming a store on optional state into a
-  ///   store on non-optional state.
-  /// - See also: `Store.ifLet`, a UIKit helper for doing imperative work with a store on optional
-  ///   state.
-  public var optional: Reducer<State?, Action, Environment> {
-    .init { state, action, environment in
-      guard state != nil else {
-        assertionFailure(
-          """
-          "\(debugCaseOutput(action))" was received by an optional reducer when its state was \
-          "nil". This can happen for a few reasons:
-
-          * The optional reducer was combined with or run from another reducer that set \
-          "\(State.self)" to "nil" before the optional reducer ran. Combine or run optional \
-          reducers before reducers that can set their state to "nil". This ensures that optional \
-          reducers can handle their actions while their state is still non-"nil".
-
-          * An active effect emitted this action while state was "nil". Make sure that effects for \
-          this optional reducer are canceled when optional state is set to "nil".
-
-          * This action was sent to the store while state was "nil". Make sure that actions for \
-          this reducer can only be sent to a view store when state is non-"nil". In SwiftUI \
-          applications, use "IfLetStore".
-          """
-        )
-        return .none
-      }
-      return self.reducer(&state!, action, environment)
-    }
-  }
-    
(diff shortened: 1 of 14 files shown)


stephencelis:
Thanks @alexito4!

> Questions I would like to ask:
> 
> * Are you happy if the function replaces the property (breaking change) or do you want to keep both?

Let's keep both but deprecate the property if possible. We keep all of our deprecations here for now:

https://github.com/pointfreeco/swift-composable-architecture/blob/a03611d45da994a706911ca024fa441b148e79cb/Sources/ComposableArchitecture/Internal/Deprecations.swift

> * What other places would you like to see this change? `forEach` maybe?

Definitely on `forEach`! Ideally any higher-order reducer that can assert.

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -189,6 +189,59 @@ public struct Reducer<State, Action, Environment> {
...
+            reducers before reducers that can set their state to "nil". This ensures that optional \
+            reducers can handle their actions while their state is still non-"nil".
+
+            * An active effect emitted this action while state was "nil". Make sure that effects for \
+            this optional reducer are canceled when optional state is set to "nil".
+
+            * This action was sent to the store while state was "nil". Make sure that actions for \
+            this reducer can only be sent to a view store when state is non-"nil". In SwiftUI \
+            applications, use "IfLetStore".
+            """
```
stephencelis, line 238: I think rather than embed the file/line in the message above, they'll be automatically added if you pass them along here:

```suggestion
            """,
            file: file,
            line: line
```

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -189,6 +189,59 @@ public struct Reducer<State, Action, Environment> {
...
+    ///         state: \.modal,
+    ///         action: /AppAction.modal,
+    ///         environment: { ModalEnvironment(mainQueue: $0.mainQueue) }
+    ///       )
+    ///
+    /// - See also: `IfLetStore`, a SwiftUI helper for transforming a store on optional state into a
+    ///   store on non-optional state.
+    /// - See also: `Store.ifLet`, a UIKit helper for doing imperative work with a store on optional
+    ///   state.
+    public func optional(_ file: String = #file, _ line: Int = #line) -> Reducer<State?, Action, Environment> {
```
stephencelis, line 219: I think `assertionFailure` parameters are `StaticString` and `Uint`...

```suggestion
    public func optional(_ file: StaticString = #file, _ line: UInt = #line) -> Reducer<State?, Action, Environment> {
```

alexito4: Yes you're totally right
Screenshot 2020-07-08 at 19 22 56

Sources/ComposableArchitecture/Reducer.swift
alexito4: Much better thanks.

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -189,6 +189,61 @@ public struct Reducer<State, Action, Environment> {
...
+    /// - See also: `IfLetStore`, a SwiftUI helper for transforming a store on optional state into a
+    ///   store on non-optional state.
+    /// - See also: `Store.ifLet`, a UIKit helper for doing imperative work with a store on optional
+    ///   state.
+    public func optional(_ file: StaticString = #file, _ line: UInt = #line) -> Reducer<State?, Action, Environment> {
+      .init { state, action, environment in
+        guard state != nil else {
+          assertionFailure(
+            """
+            "\(debugCaseOutput(action))" was received by an optional reducer at \(file):\(line) \
```
stephencelis, line 224: Assuming `file` and `line` make it into the console in a reasonable manner wanna change this messaging to be the same as the existing one?

alexito4: Yes that's what I was changing now but then I realised I don't see any difference passing file/line to the assert 🤔 I was expecting the error line to stop at the point of origin like what XCTest assertions do. 
Screenshot 2020-07-08 at 19 30 35

alexito4:
Thanks for the feedback.

- Made the optional property deprecated:
Screenshot 2020-07-08 at 19 21 18
- Updated examples to use the new optional function.
- Added file/line to all forEach variations. 
- I haven't found other high order reducers that can fail, not on the lib or in any example.
- Since Xcode doesn't seem to show up the file/line anywhere https://github.com/pointfreeco/swift-composable-architecture/pull/218#discussion_r451712447 should we add the info in the message itself?
- Should I mark the deprecation as unavailable as per #219 or is gonna be to early for that?
- swift-format run locally ^^

Sources/ComposableArchitecture/Reducer.swift
stephencelis: Isn't it showing up at the very end here?

image

stephencelis: And yeah it'd be nice for Xcode to trap on those lines instead, too :( Maybe feedback should be filed.

stephencelis:
> * Should I mark the deprecation as unavailable as per #219 or is gonna be to early for that?

We can wait a few versions to change this "deprecated" to "unavailable" so that folks have a chance to more slowly migrate to the new APIs rather than force them to all at once.

Sources/ComposableArchitecture/Reducer.swift
alexito4: You're right! I thought that was part of the message somehow. All good then 👍

alexito4:
All good then. Ping me if you need any further changes 👍

PR #220, opened 2020-07-09
Update docs for Effect.timer and TestStore.

mbrandonw:
I started by wanting to clarify the docs on `Effect.timer` that it's only appropriate for TCA features, but ended up updating the docs for the test store too.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Timer.swift
+++ b/Sources/ComposableArchitecture/Effects/Timer.swift
@@ -81,10 +81,10 @@ extension Effect where Failure == Never {
   ///   }
   ///
   /// - Note: This effect is only meant to be used with features built in the Composable
-  ///   Architecture, and returned from a reducer. If you want a testable alternative to Foundation's
-  ///   `Timer.publish` you can use the publisher `Publishers.Timer` that is included in this
-  ///   library via the [`CombineSchedulers`](https://github.com/pointfreeco/combine-schedulers)
-  ///   module.
+  ///   Architecture, and returned from a reducer. If you want a testable alternative to
+  ///   Foundation's `Timer.publish` you can use the publisher `Publishers.Timer` that is included
+  ///   in this library via the
+  ///   [`CombineSchedulers`](https://github.com/pointfreeco/combine-schedulers) module.
   ///
   /// - Parameters:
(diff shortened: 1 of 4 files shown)


Sources/ComposableArchitecture/Effects/Timer.swift
```diff
@@ -2,23 +2,89 @@ import Combine
...
+  ///
+  ///       .send(.stopButtonTapped)
+  ///     )
+  ///   }
+  ///
+  /// - Note: This effect is only meant to be used with features built in the Composable
+  /// Architecture, and returned from a reducer. If you want a testable alternative to Foundation's
+  /// `Timer.publish` you can use the publisher `Publishers.Timer` that is included in this
+  /// library via the [`CombineSchedulers`](https://github.com/pointfreeco/combine-schedulers)
+  /// module.
```
stephencelis, line 87:
```suggestion
  /// - Note: This effect is only meant to be used with features built in the Composable
  ///   Architecture, and returned from a reducer. If you want a testable alternative to
  ///   Foundation's `Timer.publish` you can use the publisher `Publishers.Timer` that is included
  ///   in this library via the
  ///   [`CombineSchedulers`](https://github.com/pointfreeco/combine-schedulers) module.
```

stephencelis: (This is the alignment that Apple uses in its docs.)

Sources/ComposableArchitecture/TestSupport/TestStore.swift
```diff
@@ -4,20 +4,55 @@
...
+  /// This object aids in writing expressive and exhaustive tests on your features built in the
+  /// Composable Architecture. It allows you to send a sequence of actions to the store, and each
+  /// step of the way you must assert exactly how state changed, and how effect emissions were fed
+  /// back into the system.
+  ///
+  /// There are multiple ways the test store forces you to exhaustively assert on how your feature
+  /// behaved:
+  ///
+  /// * After each action is sent you must describe precisely how the state changed from before the
+  /// action was sent to after it was sent.
```
stephencelis, line 16:
```suggestion
  ///   action was sent to after it was sent.
```

PR #221, opened 2020-07-09
Add tests for "Effect cancellation retains publisher" issue

darrarski:
I've created this PR as advised by @stephencelis. The issue is also described in a [thread on Swift Forums](https://forums.swift.org/t/ifletstore-effect-cancellation-retains-publisher/38306).

## Issue description

When creating an `Effect` from a custom publisher and later canceling it, the subscription is correctly canceled, but the publisher instance is retained in the memory. Even after releasing a `Store`, the publisher still seems to be held in the memory for some reason.

## Reproducing the issue

I've added a very basic `CustomPublisher` class that is used in three new tests:

- `testCustomPublisherRelease` - checks if `CustomPublisher` is correctly released from the memory after canceling the subscription
- `testEffectCancellationPublisherRelease` - checks if `CustomPublisher` is correctly released from the memory after canceling an `Effect` originating from it
- `testCombinedReducerEffectCancellationPublisherRelease` - checks if `CustomPublisher` is correctly released from the memory after canceling an `Effect` originating from it, when the `Effect` is returned from combined reducer

The first two tests pass without an issue, so I assume the problem is not connected to `CustomPublisher` implementation.

The second and the third test have exactly the same assertions and differs in how reducer is created (directly, or with `.combine` function).

The third test fails, because after `Effect` is canceled, the `CustomPublisher` instance still sits in the memory:

> ❌ should not retain publisher when canceled

...and even after releasing the store, the publisher is still retained:

> ❌ should not retain publisher when store is no longer referenced

## Discussion

I'm not 100% sure where exactly the source of the problem is located. It could be an issue not related to `TCA` at all. I am open to all ideas and hints that could help resolve the issue, either here or on [Swift Forums](https://forums.swift.org/t/ifletstore-effect-cancellation-retains-publisher/38306). There is also a demo project showcasing this issue [open-sourced on GitHub](https://github.com/darrarski/tca-ifletstore-effect-cancellation-retains-publisher-demo).

### DIFF
--- a/Tests/ComposableArchitectureTests/EffectCancellationTests.swift
+++ b/Tests/ComposableArchitectureTests/EffectCancellationTests.swift
@@ -312,13 +312,13 @@ final class EffectCancellationTests: XCTestCase {
       .store(in: &cancellables)
     publisher = nil
 
-    XCTAssertNotNil(weakPublisher, "should not release publisher after subscribing")
+    XCTAssertNotNil(weakPublisher, "should retain publisher after subscribing")
     XCTAssertEqual(receivedRequests, 1, "should receive request after subscribing")
     XCTAssertEqual(receivedCancels, 0, "should not receive cancel after subscribing")
 
     cancellables.removeAll()
 
-    XCTAssertNil(weakPublisher, "should release publisher after cancellation")
+    XCTAssertNil(weakPublisher, "should not retain publisher after cancellation")
     XCTAssertEqual(receivedRequests, 1, "should not receive request after cancellation")
     XCTAssertEqual(receivedCancels, 1, "should receive cancel after cancellation")
   }
--- a/Tests/ComposableArchitectureTests/EffectCancellationTests.swift
+++ b/Tests/ComposableArchitectureTests/EffectCancellationTests.swift
@@ -333,8 +333,8 @@ final class EffectCancellationTests: XCTestCase {
       case action
     }
 
-    var store: TestStore<State, State, Action, Action, Void>!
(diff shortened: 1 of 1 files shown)


darrarski:
When trying to investigate the problem using Xcode's memory manager, I can see that `CustomPublisher` instance is being held in the memory by `PublisherBox` when it should be already deinitialized:

Screenshot 2020-07-09 at 17 51 23

I've also noticed there is another leak detected by Xcode, connected to closures in `AnyCancellable`:

Screenshot 2020-07-09 at 17 52 16

Not sure if it's related to the problem though.

darrarski:
I just noticed the issue does not occur when using actual `Store`, not the `TestStore` 🤦‍♂️ I still see publishers being retained when using `Reducer.combine` function. I need to update this PR.

stephencelis:
@darrarski Thanks for sharing! Please keep us updated if you're able to track down a fix or if you're able to confirm it's a TCA bug vs. a Combine bug. We'll try to look at this soon but it may be a little while.

stephencelis:
@darrarski Did you ever manage to track this issue down? Could it also be that it was a bug in Combine that's been fixed?

darrarski:
@stephencelis I just checked it and it looks like the issue no longer occurs when using Xcode 12.1 (12A7403) with Apple Swift version 5.3 (swiftlang-1200.0.29.2 clang-1200.0.30.1) 🎉

stephencelis:
@darrarski Great! Gonna close this out :) Thanks again for reporting!

PR #222, opened 2020-07-09
Lifecycle case study

mbrandonw:
I moved our lifecycle experimentation stuff to a case study and fleshed it out a bit. I think we can let it marinate a bit more before we commit to it as a TCA feature.


Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-Lifecycle.swift
```diff
@@ -0,0 +1,167 @@
+import ComposableArchitecture
+import SwiftUI
+
+private let readMe = """
+  This demonstrates how to trigger effects when a view appears, and cancel effects when a view
+  disappears. This can be helpful for starting up a feature's long living effects, such as timers,
+  location managers, etc. when that feature is first presented.
+
+  To accomplish this we define a higher-order reducer that enhances any reducer with two additional
+  actions, `.onAppear` and `.onDisappear`, and a way to automate running effects when those actions
```
stephencelis, line 10:
```suggestion
  This demonstrates how to trigger effects when a view appears, and cancel effects when a view \
  disappears. This can be helpful for starting up a feature's long living effects, such as timers, \
  location managers, etc. when that feature is first presented.

  To accomplish this we define a higher-order reducer that enhances any reducer with two additional \
  actions, `.onAppear` and `.onDisappear`, and a way to automate running effects when those actions \
```

mbrandonw: ah yes, good catch. thanks

PR #223, opened 2020-07-10
Auto-generate and publish docs

lukeredpath:
I find it really useful if project code docs are easily accessible. To that end, this PR:

* Adds a make task for generating docs locally
* Adds a workflow for generating docs using Github Actions and automatically publishing them to Github Pages

I did look into doing this using the swift-doc tool (which is written in Swift and has it's own Github Action) but it didn't seem to like the source, so I'm using `jazzy` and the same workflow I already use on my own Swift library project. It's a Ruby gem which does mean adding a `Gemfile` to make it easier to install the gems in a local directory when the workflow runs.

The workflow uses OSX which already has a Ruby environment, installs and caches the gems (just `jazzy`), runs the `make docs` task and uses the publish to Github Pages action.

I've pushed an example to my own fork's gh-pages branch:

https://lukeredpath.github.io/swift-composable-architecture/


stephencelis:
Hey @lukeredpath! Thanks for looking into this. We'd love to have docs available and have spent a lil time on this in the past but never shipped anything. A few things to respond to:

> I did look into doing this using the swift-doc tool (which is written in Swift and has it's own Github Action) but it didn't seem to like the source

Can you provide more context here? We have a branch here that uses swift-doc and it seemed to generate things just fine: https://github.com/pointfreeco/swift-composable-architecture/tree/swift-docs

The branch is definitely just exploratory and not in a mergeable state, though. In particular, it's not as nice as your solution because it doesn't integrate with a GitHub Action nor does it commit to a `gh-pages` branch.

I think if possible, though, we'd like to use swift-doc and not jazzy, since it removes the dependency on Ruby/Bundler/etc., and because by default it renders things a bit more nicely, even though it's still in "beta." If possible would you like to explore swapping out jazzy for swift-doc? Maybe we can help troubleshoot any error you were seeing.

lukeredpath:
This is what I was seeing:

```
$ swift doc generate Sources --module-name ComposableArchitecture --output docs
Error: SwiftSyntax parser library isn't compatible
```

mattt:
> This is what I was seeing:
> 
> ```
> $ swift doc generate Sources --module-name ComposableArchitecture --output docs
> Error: SwiftSyntax parser library isn't compatible
> ```

@lukeredpath Could you please share the output of `swift --version`? Also, to clarify: are you able to generate documentation using the `swift-doc` GitHub action?

lukeredpath:
@mattt thanks for jumping in - this is my Swift version:

```
Apple Swift version 5.2.4 (swiftlang-1103.0.32.9 clang-1103.0.32.53)
Target: x86_64-apple-darwin19.5.0
```

I get the same error using Swift 5.3 that ships with the latest Xcode 12 beta too.

I haven't been able to try the Github action - I'm working on a branch and unless I'm missing something there doesn't seem to be a way to run a workflow that only exists on a branch?

mattt:
@lukeredpath Thanks for sharing that. Unfortunately, I wasn't able to reproduce this locally...

```terminal
$ brew install swift-doc
# 🕐🕑🕒😴
$ swift doc --version
1.0.0-beta.3
$ which swift-doc
/usr/local/bin/swift-doc
$ swift --version
Apple Swift version 5.2.4 (swiftlang-1103.0.32.9 clang-1103.0.32.53)
Target: x86_64-apple-darwin19.5.0
$ xcode-select -p
/Applications/Xcode.app/Contents/Developer
$ xed -v
xed version 11.5
$ swift doc generate --module-name SwiftDoc --format html Sources/SwiftDoc
# works on my machine 😭
```

Were you installing using Homebrew or `make install`?

> I haven't been able to try the Github action - I'm working on a branch and unless I'm missing something there doesn't seem to be a way to run a workflow that only exists on a branch?

You _should_ be able to trigger new workflows for a branch by adding a file `.github/workflows/`, but I'm not 100% sure about that. The one gotcha is that it needs to specify a trigger (top-level `on`) that matches that branch.

stephencelis:
Hey @lukeredpath! It took a bit to figure things out, but I was able to get swift-doc working with GitHub Actions, so I've opened another PR here: #241. Thanks again for the time you put into this. We definitely wanna make docs more readily available online and this was the push we needed.

lukeredpath:
Awesome. 👍🏻

PR #225, opened 2020-07-10
Fix issue with `ifLet` where `else` was being called too often

### LINKED ISSUE #216: Unexpected updates in ifLet for UIKit
I've slightly modified the TicTacToe UIKit example with the following code in `AppViewController`

```swift
self.store
  .scope(state: { $0.login }, action: AppAction.login)
  .ifLet(then: { [weak self] loginStore in
    self?.setViewControllers([LoginViewController(store: loginStore)], animated: false)
  },
  else: {
    print("else")
  })
```

After logging in you will notice that the else branch is triggered twice.

I've looked into this for a bit and the underlying problem seems to be  the combination of the two scope functions that are involved. Both of them trigger `handleEvents:receiveOutput` in the `ifLet` UIKit function. 

One other thing I've tried was verifying this behaviour in a unit test. But strangely I was not able to.

mluisbrown:
Pretty sure this fixes #216. 

I took the liberty of adding in the tests that @mbrandonw and @stephencelis added in [this branch](https://github.com/pointfreeco/swift-composable-architecture/compare/iflet) and they are now all passing 🎉 

FYI I got this working first in my ReactiveSwift port (not yet pushed) and then back ported it to here 😄 

cc: @fonkadelic

### DIFF
--- a/Sources/ComposableArchitecture/UIKit/IfLetUIKit.swift
+++ b/Sources/ComposableArchitecture/UIKit/IfLetUIKit.swift
@@ -43,16 +43,31 @@ extension Store {
     then unwrap: @escaping (Store<Wrapped, Action>) -> Void,
     else: @escaping () -> Void
   ) -> Cancellable where State == Wrapped? {
-    self
+    let elseCancellable = self
       .scope(
         state: { state in
           state
             .removeDuplicates(by: { ($0 != nil) == ($1 != nil) })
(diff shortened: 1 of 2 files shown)


mbrandonw:
Awesome! At first glance this looks like a really good solution! I want to spend a little more time poking around it to see if there are any other things to think about it, and then we'll merge.

Thanks for picking this up!

PR #226, opened 2020-07-10
Case study for tvOS focus.

mbrandonw:
Adding a new case study app for tvOS-specific features, and the first case study I included was how to control focus programmatically on tvOS.


mbrandonw:
I just pushed some hacky stuff to get iOS 13 building: https://github.com/pointfreeco/swift-composable-architecture/pull/226/commits/edabc760e00c633c05884f79d8445caab28dad6e

not sure if it's the right way to do this

Sources/ComposableCoreMotion/Internal/Exports.swift
```diff
@@ -1,2 +1,4 @@
+#if canImport(CoreMotion)
```
mbrandonw, line 1: these are necessary to get tvOS swiftui previews building. they for some reason try building ComposableCoreMotion, even though it isn't linked, and that fails since tvOS doesnt have core motion.

PR #243, opened 2020-07-29
Update documentation.yml

stephencelis:
The upstream PR was merged so we can drop this step.


mluisbrown:
@stephencelis just an FYI that when adding the documentation workflow to the ReactiveSwift TCA fork I found I still needed the `chown` step, otherwise I would get permission errors. See [here](https://github.com/JamesIves/github-pages-deploy-action/issues/388)

I noticed that the documentation workflow here hasn't been run since this step was removed. You will probably have to add it back.

stephencelis:
@mluisbrown Seeing this now, thanks! Will revert.

PR #245, opened 2020-08-05
Restore original cancellation logic

stephencelis:
Xcode 12 beta 4 seems to have fixed all the problems we were seeing in beta 3, so we should be able to safely restore the old logic and we could maybe migrate to the new logic whenever we drop iOS 13 and macOS 10.15 support.

Also looks like GitHub already has beta 4 on its runners, so adding CI.


stephencelis:
Ah never mind on CI! Was looking at the wrong build number. Hopefully updated soon tho.

PR #249, opened 2020-08-10
Alert bug exploration

stephencelis:
A user encountered a bug here: https://forums.swift.org/t/how-to-show-alerts-from-two-actions-one-after-the-other/39214

This branch should hopefully fix!

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
@@ -123,6 +123,11 @@ public struct ActionSheetState<Action> {
   public typealias Button = AlertState<Action>.Button
 }
 
+@available(iOS 13, *)
+@available(macCatalyst 13, *)
+@available(macOS, unavailable)
+@available(tvOS 13, *)
+@available(watchOS 6, *)
 extension ActionSheetState: CustomDebugOutputConvertible {
   public var debugOutput: String {
     let fields = (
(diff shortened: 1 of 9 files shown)


Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
```diff
@@ -105,6 +105,7 @@ import SwiftUI
 @available(tvOS 13, *)
 @available(watchOS 6, *)
 public struct ActionSheetState<Action> {
+  public let id = UUID()
```
stephencelis, line 108: I think it's safe to identify in this fashion.

Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
```diff
@@ -163,16 +163,11 @@ extension View {
     dismiss: Action
   ) -> some View {
 
-    let viewStore = ViewStore(store, removeDuplicates: { ($0 == nil) != ($1 == nil) })
```
stephencelis, line 166: This was the primary issue. Replacing alert state outright would consider the change a duplicate.

Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
```diff
@@ -127,23 +128,44 @@ public struct ActionSheetState<Action> {
 @available(macOS, unavailable)
 @available(tvOS 13, *)
 @available(watchOS 6, *)
-extension ActionSheetState: Equatable where Action: Equatable {}
+extension ActionSheetState: Equatable where Action: Equatable {
+  public static func == (lhs: Self, rhs: Self) -> Bool {
```
stephencelis, line 132: Have to do custom implementations of `Equatable`, `Hashable`, and `CustomDebugOutputConvertible` now to avoid the `UUID` being different in tests...maybe you have a better idea than hiding a UUID inside?

PR #251, opened 2020-08-10
Typo fixes

alskipp:
Just a couple of typos in the CaseStudies project.


alskipp:
Incidentally while looking through the CaseStudies project I made a few tweaks to replace closures with key paths, i.e.:

```
-              text: viewStore.binding(get: { $0.text }, send: BindingBasicsAction.textChange)
+              text: viewStore.binding(get: \.text, send: BindingBasicsAction.textChange)
```

Not sure how fussed you are about these cosmetic differences? I can make another PR with the amends if this would be useful.

Issue #253, opened 2020-08-13
IdentifiedArray Shuffle does not shuffle the ids


jeffersonsetiawan:
**Describe the bug**
IdentifiedArray.shuffled() doesn't shuffle the ids.

**To Reproduce**
Let say we have this struct and shuffle on the `items`, the ids order will not change.

```swift
struct Foo: Equatable, Identifiable {
    var id: Int
    var name: String
}

var items: IdentifiedArrayOf<Foo> = [
    Foo(id: 1, name: "A"),
    Foo(id: 2, name: "B"),
    Foo(id: 3, name: "C"),
    Foo(id: 4, name: "D"),
    Foo(id: 5, name: "E"),
]

items.shuffle()
print(items.ids) /// [1, 2, 3, 4, 5]
print(items.elements) /// Shuffled e.g. [id: 5, id: 3, id: 1, id: 4, id: 2]
```
The problem is when we want to remove by id, for example
```swift
items.remove(id: 2)
print(items.elements) // [id: 5, id: 1, id: 4, id: 2]
```
It will remove the id in the position before it got shuffled.

**Expected behavior**
the ids order change

**Screenshots**
![Screen Shot 2020-08-13 at 21 03 04](https://user-images.githubusercontent.com/14871122/90143936-62cbf780-dda8-11ea-9a72-8c84e3fc9f38.jpg)
As you can see in the screenshot, the removed item is with id 3, not 2

**Environment**
 - Xcode: 11.6
 - Swift:  5.2
 - OS: (13.6)

**Additional Info**
Current workaround: I uses this
```
items = IdentifiedArray(items.elements.shuffled())
```
Idea:
Implement shuffle method on the IdentifiedArray
```
public mutating func shuffle() {
        let idsAndElements = elements.shuffled().map { (id: $0[keyPath: id], element: $0) }
        ids = idsAndElements.map { $0.id }
        dictionary = Dictionary(idsAndElements, uniquingKeysWith: { $1 })
    }
```

stephencelis:
Hey @jeffersonsetiawan! Good find! I think your solution looks good and falls in line with other overloads we've had to define. Would you be able to submit a PR with a few tests?

jeffersonsetiawan:
Sure, I just tried that one too in my project, glad you suggest that as well, will try to do the unit test and make a PR, thanks for the suggestion @stephencelis !

PR #254, opened 2020-08-13
Add implementation of shuffle on IdentifiedArray

jeffersonsetiawan:
This PR fixed https://github.com/pointfreeco/swift-composable-architecture/issues/253

### DIFF
--- a/Tests/ComposableArchitectureTests/IdentifiedArrayTests.swift
+++ b/Tests/ComposableArchitectureTests/IdentifiedArrayTests.swift
@@ -195,4 +195,32 @@ final class IdentifiedArrayTests: XCTestCase {
       ], array)
 
   }
+  
+  func testShuffle() {
+    struct User: Equatable, Identifiable {
+      let id: Int
+      var name: String
+    }
+
+    var array: IdentifiedArray = [
+      User(id: 1, name: "Blob"),
+      User(id: 2, name: "Blob Jr."),
+      User(id: 3, name: "Blob Sr."),
+      User(id: 4, name: "Foo Jr."),
+      User(id: 5, name: "Bar Jr."),
+    ]
+    var lcrng = LCRNG(seed: 0)
+    array.shuffle(using: &lcrng)
+    XCTAssertEqual(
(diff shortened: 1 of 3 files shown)


Tests/ComposableArchitectureTests/IdentifiedArrayTests.swift
```diff
@@ -195,4 +195,32 @@ final class IdentifiedArrayTests: XCTestCase {
...
+        User(id: 1, name: "Blob"),
+        User(id: 3, name: "Blob Sr."),
+        User(id: 5, name: "Bar Jr."),
+        User(id: 4, name: "Foo Jr."),
+        User(id: 2, name: "Blob Jr.")
+      ],
+      array.elements
+    )
+    XCTAssertEqual([1, 3, 5, 4, 2], array.ids)
+  }
```
stephencelis, line 225: I think this will get things into building order:
```suggestion
  #if swift(>=5.2)
  func testShuffle() {
    struct User: Equatable, Identifiable {
      let id: Int
      var name: String
    }

    var array: IdentifiedArray = [
      User(id: 1, name: "Blob"),
      User(id: 2, name: "Blob Jr."),
      User(id: 3, name: "Blob Sr."),
      User(id: 4, name: "Foo Jr."),
      User(id: 5, name: "Bar Jr."),
    ]
    var lcrng = LCRNG(seed: 0)
    array.shuffle(using: &lcrng)
    XCTAssertEqual(
      [
        User(id: 1, name: "Blob"),
        User(id: 3, name: "Blob Sr."),
        User(id: 5, name: "Bar Jr."),
        User(id: 4, name: "Foo Jr."),
        User(id: 2, name: "Blob Jr.")
      ],
      array.elements
    )
    XCTAssertEqual([1, 3, 5, 4, 2], array.ids)
  }
  #endif
```

stephencelis:
Thanks again!

jeffersonsetiawan:
Thanks for the great reference and your help as well @stephencelis ! awesome library! 👍🏻

PR #256, opened 2020-08-17
Minor improvements of the Effect API 

maximkrouk:
I added some syntactic sugar to the Effect API.
Now reducers can use `.value(<some_action>)` instead of `.init(value: <some_action>)`

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -56,6 +56,11 @@ public struct Effect<Output, Failure: Error>: Publisher {
   public init(error: Failure) {
     self.init(Fail(error: error))
   }
+  
+  /// Initializes an effect that immediately emits the value passed in.
+  ///
+  /// - Parameter output: The value that is immediately emitted by the effect.
+  public static func value(_ output: Output) -> Effect { .init(value: output) }
 
   /// An effect that does nothing and completes immediately. Useful for situations where you must
   /// return an effect, but you don't need to do anything.


maximkrouk:
The only change I made is

[_Sources/ComposableArchitecture/Effect.swift_]
```swift
+  /// Initializes an effect that immediately emits the value passed in.
+  ///
+  /// - Parameter output: The value that is immediately emitted by the effect.
+  public static func value(_ output: Output) -> Effect { .init(value: output) }
```

So the PR shouldn't cause build failures 😣

AlexisQapa:
Nice, I always been annoyed by the previous syntax. 
Bike-shedding incoming, what do you think of a bit more discoverable name like .send(Value) (to use the store convention) or .dispatch(Value)

maximkrouk:
I think `send` may look even nicer 🙂

mbrandonw:
Thanks for the PR. Yeah I think `.send` could be pretty nice. @stephencelis and I will discuss this today and get back to you.

> So the PR shouldn't cause build failures 😣

Yeah ignore any failures for now, Xcode is being very flakey with running tests for some reason. It's something we need to look into. Scroll to the bottom of these [logs](https://github.com/pointfreeco/swift-composable-architecture/commit/36caf89de4b11db52f3ce5a4fc044bf87017824d/checks/995460402/logs) to see it complaining of not being able to build an Objective-C module even though there is no objc code in this repo 😕

mbrandonw:
Hey @maximkrouk, we chatted about this today and we remembered why the API is the way it is. What we like about how it is right now is that there is only a single way to construct an effect of the following flavors:

* Synchronous value: `.init(value:)`
* Synchronous failure: `.init(failure:)`
* Synchronous result: `.result { }`
* Asynchronous single value: `.future { }`
* Asynchronous multiple values: `.run { }`

We purposely tried to stay away from coming up with lots of ways to do the same thing with these constructors. 

That's not to say they are the correct choices. They could probably be improved. I personally liked `.send`, however such a constructor is most appropriate on `Effect` when `Failure` is constrained to `Never`, and so it doesn't replace `.init(value:)`, and that means we are back to having multiple ways to construct the same thing.

So right now we are leaning towards not bringing this in.

maximkrouk:
> That's not to say they are the correct choices. They could probably be improved. I personally liked .send, however such a constructor is most appropriate on Effect when Failure is constrained to Never

Yes, `.send` may be more appropriate for effects with `Never` failure type, so it's not the best syntax here.

>  it doesn't replace .init(value:), and that means we are back to having multiple ways to construct the same thing.

Any factory method here will lead to duplicate ways of instantiating objects. But I believe that they bring a lot to ergonomics at the call site. Such methods as `.value` and `.failure` would fit the API with `.result` `.future` and `.run` methods (not the trailing closure stuff, but general static factory pattern). There is also an option to remove public initializers, but I don't think it's necessary, this duplication is minor and there are only two methods to duplicate.

Anyway it's easy to make an internal extension to use such syntactic sugar, but I still think it worth being a first-class citizen of TCA 🙂

hanneskaeufler:
Just to weigh in with one more opinion: I find `.init(value:)` to be kind of un-intuitive and I had quite some difficulty in finding out how to dispatch an action from the reducer. Now that I know it's kind of fine, but discoverability was an issue for me. `.send` was what I had typed in hoping for autocompletion to tell me I guessed right as well ... :)

stephencelis:
Hey @maximkrouk! Thanks again for opening this. We do think there's room to improve these APIs, especially before a major release, but I'd like to widen the discussion outside of this PR. Going to close this for now, but if you'd like to restart the conversation over on the discussions tab, please do! You can link to your implementation here and we can start gathering feedback and more alternatives.

maximkrouk:
Hi, I'm not sure if this topic needs a separate discussion thread, so I'll just note here, that I do always make an extension I sent [here](https://github.com/pointfreeco/swift-composable-architecture/pull/256#issuecomment-675115152), I didn't like that sync effects API is different from all of the other ones with static methods, so I still think `.value(_:Action)` is much more ergonomic

Issue #47, opened 2020-05-08
API for EnvironmentLess Stores


natemann:
Following along with creating Stateless and Actionless viewStores in [#43 ](https://github.com/pointfreeco/swift-composable-architecture/issues/43), what are the thoughts on extending `Store` and `Reducer` with functions where `Environment` is Void? I am finding that for most stores, I don't need an `Environment` object.

I think it would be nice to have, but could bloat TCA quite a bit as you would have extensions for 3 possible scenarios:
- Environment is Void
- LocalEnvironment is void (for scopes)
-GlobalEnvironment is void (for pullbacks)

stephencelis:
This is worth having a discussion about. Thanks @natemann!

Early on we played with some helpers for `Void` environments, but in practice we found that these `Void` environments almost always became struct environments with dependencies, and that the work to make this happen was less cumbersome if we started with empty struct environments to begin with:

- Fewer places to rename `Void` to `XyEnvironment` (a type alias can help here, but that's even more work you have to undo)
- Fewer things to change in `pullback` transformations and similar

This is why all of the examples in the repo define struct environments, _even_ if they don't have any dependencies.

Of course, these are just the trade-offs we've currently landed at, and depending on how someone builds an application and how much they modularize, I can see some applications having a _lot_ of small reducers with no dependencies benefitting from using `Void` for their environments.

One more helper is a reducer initializer that doesn't need to return `Effect.none`:

- `Reducer.init(_: (inout State, Action) -> Void) where Environment == Void`

natemann:
@stephencelis. I see your point. 

I was not really thinking about a reducer that does not have to return `Effect.none`. I have a few reducers that don't have an environment, but still return an `Effect`. Mostly just the `fireAndForget` kind.

Secondly, I feel that we can still have an `Environment` , but have a reducer that does not return an `Effect`. I'm thinking of something like Environment.currentDate that just returns the current time. 

I know this is a little off topic, but your last comment intrigued me. My thoughts on `Environment == Void` was to not have to write `environment: { _ in () }` in pullbacks and scopes.

stephencelis:
> I have a few reducers that don't have an environment, but still return an `Effect`. Mostly just the `fireAndForget` kind.

Worth noting that you'll still want to introduce an environment to supply this work if you want to write tests for those actions. We have a few examples of these kinds of tests in the demo apps. Voice Memos, for one, has some fire-and-forget effects in its audio recorder and audio player dependencies.

[...]

PR #257, opened 2020-08-17
Custom pullbacks for Void-State && Void-Environment reducers

maximkrouk:
Related to [that disscusion](https://github.com/pointfreeco/swift-composable-architecture/issues/47). I think it can be handy, but whether to do it first-class stuff or not is up to you 🙂
(_Tests I wrote are more like the proof of concept, they check if code is compiling and not crushing at runtime, but they don't follow dependency injection principles_)

### DIFF
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -566,3 +566,104 @@ public struct Reducer<State, Action, Environment> {
     self.reducer(&state, action, environment)
   }
 }
+
+extension Reducer where State == Void {
+  /// Transforms a reducer that works on local state, action and environment into one that works on
+  /// global state, action and environment. It accomplishes this by providing 2 transformations to
+  /// the method:
+  ///
+  /// * A case path that can extract/embed a local action into a global action.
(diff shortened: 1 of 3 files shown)


maximkrouk:
For now, I use this workaround:
```swift
public protocol VoidProvider {
    var void: Void { get set }
}

extension VoidProvider {
    public var void: Void {
        get { () }
        set { }
    }
}

extension GlobalState: VoidProvider {}

...
localActionReducer.pullback(
    state: \.void,
    action: /GlobalAction.local
    environment: { $0 }
)

// Reducer<Void, LocalAction, GlobalEnvironment> -> Reducer<GlobalState, GlobalAction, GlobalEnvironment>
```

mbrandonw:
Hey @maximkrouk thanks for the PR! This would definitely improve the ergonomics of the library, unfortunately we are weary about bringing in more overloads. In fact, we already think we may need to soon cut some overloads for bindings and `WithViewStore`, `ForEach`, etc.

So I think we won't bring this in right now, although we will re-evaluate with each new Swift release to see if things have gotten better.

mbrandonw:
Will close this for now and we can reconsider in the future! Thanks again!

PR #258, opened 2020-08-17
Improve debugging info when some effects are still running issue happen

mackoj:
When you have a `Some effects are still running issue` error it can be really hard to debug.

In order to help understand what is happening I'm using the signpost data in order to have a better understanding of the Effect and Action that has been happening and that are not finished.

FWIW the implementation is pretty naive but working if you are interest in it I will take more time to improve it.

### What I have tried before making this

I have tried a various way to know witch was the effect that poses an issue for me but this way was the only really precise way to give me this info.

When I tried:
- to add `.debug()` on my reducer it was printing Effect that had the issue as if they where working normally
- or use the `.signpost()` I discover that I can't record a test when using the profiler and in normal mode it was working
- backtrace, breakpoint and the classic debug method didn't help either

### What it look like

As you can see on the image below `EngineContainerAction.searchEngineView(.launchLRSearch)` never finishes. 

| signpost(disabled) | signpost(enabled) | signpost(enabled) with a prefix(💥) | 
| --- | --- | --- |
| Capture d’écran 2020-08-18 à 01 04 20 | Capture d’écran 2020-08-18 à 01 18 06 | Capture d’écran 2020-08-18 à 01 19 03 | 

_I did remove the extra space above the new text on the first snapshot and the Optional()_

### How to use it

**signpost(disabled)**
```swift
    let localTestStore = TestStore(
      initialState: engineState,
      reducer: engineContainerReducer,
      environment: env
    )
``` 

**signpost(enabled)**
```swift
    let localTestStore = TestStore(
      initialState: engineState,
      reducer: engineContainerReducer.signpost(),
      environment: env
    )
```

Have a great day.

### DIFF
--- a/Sources/ComposableArchitecture/Debugging/ReducerInstrumentation.swift
+++ b/Sources/ComposableArchitecture/Debugging/ReducerInstrumentation.swift
@@ -13,6 +13,7 @@ struct SignpostData {
           \(log.joined(separator: "\n"))
       """
     }
+    return ""
   }
   
   var unfinishedContent : String {
@@ -23,6 +24,7 @@ struct SignpostData {
           \(unfinished)
       """
     }
+    return ""
   }
 }
 
--- a/Sources/ComposableArchitecture/Debugging/ReducerInstrumentation.swift
+++ b/Sources/ComposableArchitecture/Debugging/ReducerInstrumentation.swift
@@ -1,6 +1,33 @@
 import Combine
 import os.signpost
 
+struct SignpostData {
+  var log : [String] = []
+  var unfinished : [String:String] = [:]
+  
+  var logContent : String {
+    if !log.isEmpty {
+      return """
+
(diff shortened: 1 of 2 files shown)


mackoj:
My issue was `.receive(on: environment.mainQueue)` wasn't working has expected when called before `.catchToEffect()` if it help someone with the same issue.

stephencelis:
Hey @mackoj! We have plans on improving some of the experience here, and in the beta period would print out a description of the effects in flight, but weren't quite happy enough with the quality of the output, so we put it on the back burner. We definitely want to revisit this in the future, though!

mackoj:
Hi @stephencelis  I will re-open this I have upgraded the output and the usefulness of it.

mackoj:
@stephencelis I have re-open a new PR about this with a way more useful insight of how to fix it.

PR #260, opened 2020-08-18
Removed unused `currentTab` property

nmccann:
While reviewing how to handle shared state, I was confused as to why a nested state object was being passed the `currentTab`, and found that it was being unused. I could see some valid use cases in which you'd need to do this, but in all such cases I'd expect the nested state (or one of it's children) to use the property - so I'm guessing this was unintentional.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-SharedState.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-SharedState.swift
@@ -34,7 +34,6 @@ struct SharedState: Equatable {
   var profile: ProfileState {
     get {
       ProfileState(
-        currentTab: self.currentTab,
         count: self.counter.count,
         maxCount: self.counter.maxCount,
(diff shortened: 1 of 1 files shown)


stephencelis:
Hey @nmccann! Thanks for taking the time to PR. It's a lil subtle but we are in fact using `currentTab` in profile state. When you're in the profile screen and tap "reset", this not only resets the counter to zero, it will also flip you back to the counter tab:

https://github.com/pointfreeco/swift-composable-architecture/blob/main/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-SharedState.swift#L60-L61

nmccann:
Ah, you are totally right. That isn’t particularly obvious, but I should have realized it was intentional. I’ll close this PR and I’ll open another one that adds a test case at least for this particular situation (to avoid anyone else thinking something similar in the future), and I’ll try to add test cases for the other functionality if time allows.

Thank you again for your work on this architecture!

PR #261, opened 2020-08-18
Add a simple way to reduce diff content to the bare minimum

mackoj:
When you have a huge state diff between two state is not practical if you have only a few changes it could take you a long time to find them and sometime if the content is too big it will not even been shown.

This implementation is pretty naive but working if you are interest in it I will take more time to improve it.

It basically add a diffMode for state and actions in order to reduce them to only what is different.

| before | after |
| --- | --- |
| [incompleteDiff_480.mov.zip](https://github.com/pointfreeco/swift-composable-architecture/files/5093246/incompleteDiff_480.mov.zip) | Capture d’écran 2020-08-19 à 01 18 06 |

I still need to add a way to show the path of the property above each difference.

Hope it helps.

Cheers,

### DIFF
--- a/Sources/ComposableArchitecture/Internal/Debug.swift
+++ b/Sources/ComposableArchitecture/Internal/Debug.swift
@@ -144,9 +144,9 @@ func debugOutput(_ value: Any, indent: Int = 0) -> String {
   return debugOutputHelp(value, indent: indent)
 }
 
-func debugDiff<T>(_ before: T, _ after: T, printer: (T) -> String = { debugOutput($0) }) -> String?
+func debugDiff<T>(_ before: T, _ after: T, _ mode : DiffMode = .full, printer: (T) -> String = { debugOutput($0) }) -> String?
 {
-  diff(printer(before), printer(after))
+  diff(printer(before), printer(after), mode)
 }
 
 extension String {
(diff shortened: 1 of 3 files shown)


stephencelis:
We can definitely see customization here being useful, so thanks for sharing this draft. We'll discuss and think on this a bit.

wendyliga:
I can see this feature will be useful, not to mention, `xcode` XCTFail description is not scrollable, so you need to and manual new line on your test file. i feel that

wendyliga:
@mackoj should the `filtered` mode provide the parent scope of the context ?

example:
```swift
struct Verification {
   let id: Int
}
struct User {
   let id: Int
   let name: String
   let email: String
   let verification: Verification
}

let userBefore = User(id: 0, name: "Joe", email: "blob@email.com", verification: Verification(id: 100))
let userAfter = User(id: 0, name: "Joe", email: "blob@email.com", verification: Verification(id: 999))
let diff = debugDiff(userBefore, userAfter, mode: .filtered)
```

output will be
```
   - id: 100
   + id: 999
```

but this doesn't explain what id it is outside the indentation, maybe, for context inside indentation like verification id, we can provide the parent context ?

```
   Verfication:
     - id: 100
     - id: 999
```

mackoj:
Hi @wendyliga,

I totally agree that having the context of the properties in this case `User.Verification:`(represented above the diff) helps a lot. But I didn't take the time to implement it this PR, because the goal of this PR was to show how much this kind of feature could be helpful and doing minimal modification.

FWIW I did have a branch before that used to have it(https://github.com/pointfreeco/swift-composable-architecture/pull/211) I don't know if this branch still works 🤷‍♂️.

You are welcome to try to implement here if you want.

PR #262, opened 2020-08-19
Added some tests for Shared State case study.

nmccann:
This is a follow up to #260 - rather than removing something that I (erroneously) thought was unused, I've added tests to verify it's behaviour.

I attempted to write some other tests for the shared behaviour, for example by incrementing the counter, then asserting that the profile's `count` has the correct value, but was unable to test this because the setters for `ProfileState` are private. I had two ideas to address this:
1. Remove the private access modifier (but they may be there for some reason - ex. to further indicate that these should just "mirror" state from elsewhere rather than being changed directly)
2. Add a new type of Step that allows you to delegate to XCTAssert, similar to `do` but it would pass in the current state, allowing you to assert it's current values without having to construct a new instance of the state. Not sure how feasible/desirable this is however.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-SharedStateTests.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-SharedStateTests.swift
@@ -1,0 +1,77 @@
+import Combine
+import ComposableArchitecture
+import XCTest
+
+@testable import SwiftUICaseStudies
+
+class SharedStateTests: XCTestCase {
+  func testTabRestoredOnReset() {
+    let store = TestStore(
+      initialState: SharedState(),
+      reducer: sharedStateReducer,
+      environment: ()
+    )
+    
+    store.assert(
+      .send(.selectTab(.profile)) {
+        $0.currentTab = .profile
+      },
+      .send(.profile(.resetCounterButtonTapped)) {
+        $0.currentTab = .counter
+      })
+  }
+
+  func testTabSelection() {
+    let store = TestStore(
+      initialState: SharedState(),
+      reducer: sharedStateReducer,
+      environment: ()
+    )
+
+    store.assert(
+      .send(.selectTab(.profile)) {
+        $0.currentTab = .profile
+      },
+      .send(.selectTab(.counter)) {
+        $0.currentTab = .counter
+      })
+  }
+
+  func testIsPrimeWhenPrime() {
+    let store = TestStore(
+      initialState: SharedState.CounterState(alert: nil, count: 3, maxCount: 0, minCount: 0, numberOfCounts: 0),
+      reducer: sharedStateCounterReducer,
+      environment: ()
+    )
+
+    store.assert(
+      .send(.isPrimeButtonTapped) {
+        $0.alert = .init(
+          title: "👍 The number \($0.count) is prime!"
+        )
+      },
+      .send(.alertDismissed) {
+        $0.alert = nil
+      }
+    )
+  }
+
+  func testIsPrimeWhenNotPrime() {
+    let store = TestStore(
+      initialState: SharedState.CounterState(alert: nil, count: 6, maxCount: 0, minCount: 0, numberOfCounts: 0),
+      reducer: sharedStateCounterReducer,
+      environment: ()
+    )
+
+    store.assert(
+      .send(.isPrimeButtonTapped) {
+        $0.alert = .init(
+          title: "👎 The number \($0.count) is not prime :("
+        )
+      },
+      .send(.alertDismissed) {
+        $0.alert = nil
+      }
+    )
+  }
+}
--- a/Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-SharedStateTests.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-SharedStateTests.swift
@@ -15,9 +15,11 @@ class SharedStateTests: XCTestCase {
     store.assert(
       .send(.selectTab(.profile)) {
         $0.currentTab = .profile
+        $0.profile = .init(currentTab: .profile, count: 0, maxCount: 0, minCount: 0, numberOfCounts: 0)
       },
       .send(.profile(.resetCounterButtonTapped)) {
         $0.currentTab = .counter
+        $0.profile = .init(currentTab: .counter, count: 0, maxCount: 0, minCount: 0, numberOfCounts: 0)
       })
   }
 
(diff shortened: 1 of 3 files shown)


Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-SharedStateTests.swift
```diff
@@ -0,0 +1,77 @@
...
+  func testIsPrimeWhenNotPrime() {
+    let store = TestStore(
+      initialState: SharedState.CounterState(alert: nil, count: 6, maxCount: 0, minCount: 0, numberOfCounts: 0),
+      reducer: sharedStateCounterReducer,
+      environment: ()
+    )
+
+    store.assert(
+      .send(.isPrimeButtonTapped) {
+        $0.alert = .init(
```
nmccann, line 68: Another case where it might be nice to have a test Step that allows delegation to XCTAssert with access to the state (doesn't need to be XCTAssert, could just return true if step successful/false if not), is to allow fuzzy matching - ex. if you want to check that `alert` has some value, but you don't care exactly what value. This could allow for tests to be less brittle (exact wording of alert doesn't matter), or to delegate that precision to other tests (ex. one test already checks the exact wording, another test is checking something that requires the alert to be presented but doesn't care about the wording).

mbrandonw:
> * Remove the private access modifier (but they may be there for some reason - ex. to further indicate that these should just "mirror" state from elsewhere rather than being changed directly)

Yeah this is a good point. A way to accomplish this without `private` is to just make the properties `let` and have the mutating method just completely recreate the value:

```swift
  struct ProfileState: Equatable {
    let currentTab: Tab
    let count: Int
    let maxCount: Int
    let minCount: Int
    let numberOfCounts: Int

    fileprivate mutating func resetCount() {
      self = .init(currentTab: .counter, count: 0, maxCount: 0, minCount: 0, numberOfCounts: 0)
    }
  }
```

That should allow you to test and allow you to keep the invariant that this state is a mirror of some other state. If you want to try it out feel free!

> Add a new type of Step that allows you to delegate to XCTAssert, similar to do but it would pass in the current state, allowing you to assert it's current values without having to construct a new instance of the state. Not sure how feasible/desirable this is however.

Yeah this is intersting too. worth thinking about more!

mbrandonw:
Actually on second thought, even with the `private (set)` couldn't you write a test by reconstructing the value completely, i.e. `$0.profile = .init(...)`?

nmccann:
You're absolutely right - I'll push a commit with additional tests using that method.

Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-SharedStateTests.swift
```diff
@@ -0,0 +1,103 @@
...
+  func testSharedCounts() {
+    let store = TestStore(
+      initialState: SharedState(),
+      reducer: sharedStateReducer,
+      environment: ()
+    )
+
+    store.assert(
+      .send(.counter(.incrementButtonTapped)) {
+        $0.counter = .init(alert: nil, count: 1, maxCount: 1, minCount: 0, numberOfCounts: 1)
```
nmccann, line 53: Is there a way to send an action to the store without having to assert the changes it makes? I was originally sending a few actions, and only checking the state on the last action, but found that the tests failed because each `send` step said the state had unexpectedly changed. As a result I opted to check the state for each action, which is arguably more correct, but it would still be nice to do such things.

Oddly I was able to bypass the test failure by only checking the state of the counter or the profile, I didn't have to do both - I'm not quite sure why that is (though I ended up checking both anyways since that seems more correct, and makes it more obvious that the two are mirroring one another).

By this I mean:
```
.send(.counter(.incrementButtonTapped)) //This by itself caused a test failure due to unexpected state change

.send(.counter(.incrementButtonTapped)) {
$0.counter = .init(alert: nil, count: 1, maxCount: 1, minCount: 0, numberOfCounts: 1)
} //This didn't have a test failure, even though the profile also changed but wasn't explicitly checked

.send(.counter(.incrementButtonTapped)) {
$0.profile = .init(currentTab: .counter, count: 1, maxCount: 1, minCount: 0, numberOfCounts: 1)
} //This also didn't have a test failure, even though I didn't explicitly check the counter state, which also changed.
```

I'm confused why I can't just omit the closure altogether, yet I can just omit one of the state checks.

mbrandonw: The reason you don't have to do both is because `profile` isn't real state held in the struct, it's just a computed property. Any mutations to it are immediately reflected in the `counter` state, and any mutations to `counter` are immediately reflected in `profile`.

It does make the assertion a little confusing, but at least the assertion is still being exhaustive in what it is checking.

> Is there a way to send an action to the store without having to assert the changes it makes? I was originally sending a few actions, and only checking the state on the last action, but found that the tests failed because each send step said the state had unexpectedly changed. As a result I opted to check the state for each action, which is arguably more correct, but it would still be nice to do such things.

We don't allow this because we want the `assert` helper to be as exhaustive as possible, and for you to really prove that you know how state is mutated each step of the way. Further, if a step ever executed an effect that was fed back into the system you would need to assert against that with a `.receive` step, and we wouldn't want you to be able to skip that either.

Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-SharedStateTests.swift
```diff
@@ -0,0 +1,103 @@
...
+        $0.profile = .init(currentTab: .counter, count: 1, maxCount: 1, minCount: 0, numberOfCounts: 1)
+      },
+      .send(.counter(.decrementButtonTapped)) {
+        $0.counter = .init(alert: nil, count: 0, maxCount: 1, minCount: 0, numberOfCounts: 2)
+        $0.profile = .init(currentTab: .counter, count: 0, maxCount: 1, minCount: 0, numberOfCounts: 2)
+      },
+      .send(.counter(.decrementButtonTapped)) {
+        $0.counter = .init(alert: nil, count: -1, maxCount: 1, minCount: -1, numberOfCounts: 3)
+        $0.profile = .init(currentTab: .counter, count: -1, maxCount: 1, minCount: -1, numberOfCounts: 3)
+      })
```
mbrandonw, line 63: Here's another way to do this that might make it clearer how state is changing on a per-field basis:

```suggestion
    store.assert(
      .send(.counter(.incrementButtonTapped)) {
        $0.counter.count = 1
        $0.counter.maxCount = 1
        $0.counter.numberOfCounts = 1
      },
      .send(.counter(.decrementButtonTapped)) {
        $0.counter.count = 0
        $0.counter.numberOfCounts = 2
      },
      .send(.counter(.decrementButtonTapped)) {
        $0.counter.count = -1
        $0.counter.minCount = -1
        $0.counter.numberOfCounts = 3
      })
```

Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-SharedStateTests.swift
mbrandonw: Yeah, we would still like people to try to be as exhaustive as possible with their assertions, but I can definitely see a use case for something like that. It would be nice if the steps were more extensible than they are now. Something to think about for the future.

Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-SharedStateTests.swift
nmccann: Thank you for that clarification. I think this shows a bit of a flaw with doing shared state - it introduces a bit of "magic" that is hard to reason about. It's good that the architecture is able to support this, but I think I'd try hard to avoid using it in my own applications (or perhaps limit it to read only shared state, as opposed to bidirectional sharing, like the "reset profile" action does).

In regards to the assertions, I acknowledge that it is good to guide users towards more exhaustive tests, but that could also have the downside of users skipping some tests due to the hoops they would have to jump through to handle them.

Perhaps a future improvement could be an alternative less strict assert (maybe `softAssert`? `fuzzyAssert`? `unsafeAssert`?). That said, the testability of this architecture is a big reason why I became interested in it.

Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-SharedStateTests.swift
nmccann: That is more readable, though the downside is that it doesn't mention the `profile` at all, so it may not be clear to readers that the changes to counter are also expected to change the profile.

PR #265, opened 2020-08-20
Adds Hashable conformance to a bunch of our motion/location types

mbrandonw:


Sources/ComposableCoreLocation/Models/Beacon.swift
```diff
@@ -46,14 +46,4 @@ public struct Beacon: Equatable {
...
-
-  public static func == (lhs: Self, rhs: Self) -> Bool {
-    return lhs.accuracy == rhs.accuracy
-      && lhs.major == rhs.major
-      && lhs.minor == rhs.minor
-      && lhs.proximity == rhs.proximity
-      && lhs.rssi == rhs.rssi
-      && lhs.timestamp == rhs.timestamp
-      && lhs.uuid == rhs.uuid
-  }
```
mbrandonw, line 58: this wasn't even necessary, all these types are equatable just fine

Sources/ComposableCoreLocation/Models/Heading.swift
```diff
@@ -47,14 +47,4 @@ public struct Heading: Equatable {
...
-
-  public static func == (lhs: Self, rhs: Self) -> Bool {
-    lhs.headingAccuracy == rhs.headingAccuracy
-      && lhs.magneticHeading == rhs.magneticHeading
-      && lhs.timestamp == rhs.timestamp
-      && lhs.trueHeading == rhs.trueHeading
-      && lhs.x == rhs.x
-      && lhs.y == rhs.y
-      && lhs.z == rhs.z
-  }
```
mbrandonw, line 59: same here

Sources/ComposableCoreLocation/Models/Location.swift
```diff
@@ -36,20 +35,16 @@ public struct Location {
...
-    let speedAccuracyIsEqual: Bool
     let courseAccuracyIsEqual: Bool
-
     #if compiler(>=5.2)
       if #available(iOS 13.4, macCatalyst 13.4, macOS 10.15.4, tvOS 13.4, watchOS 6.2, *) {
         courseAccuracyIsEqual = lhs.courseAccuracy == rhs.courseAccuracy
       } else {
         courseAccuracyIsEqual = true
       }
-      speedAccuracyIsEqual = lhs.speedAccuracy == rhs.speedAccuracy
```
mbrandonw, line 50: I think this didn't need to be guarded in the `#if`, seems perfectly available on all the platforms/OS's we support

stephencelis: Hm tests are failing with

> Value of type 'CLLocation' has no member 'speedAccuracy'

mbrandonw: huh strange, docs say `speedAccuracy` was introduced in iOS 10.0, but maybe that is wrong. will change this bit.

PR #266, opened 2020-08-20
Update CI to include Xcode 12

stephencelis:
Looks like CI finally has beta 4!


stephencelis:
Hm maybe the docs are wrong and it's still beta 3 :/

PR #272, opened 2020-08-28
Swift Package Index fixes

stephencelis:
A few CI fixes to get SPI in better shape.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,12 @@
 PLATFORM_IOS = iOS Simulator,name=iPhone 11 Pro Max
 PLATFORM_MACOS = macOS
 PLATFORM_TVOS = tvOS Simulator,name=Apple TV 4K (at 1080p)
+PLATFORM_WATCHOS = watchOS Simulator,name=Apple Watch Series 4 - 44mm
 
 default: test
 
 test:
-	instruments -s devices
+	xcrun xctrace list devices
 	xcodebuild test \
 		-scheme ComposableArchitecture \
 		-destination platform="$(PLATFORM_IOS)"
@@ -15,6 +16,9 @@ test:
 	xcodebuild test \
 		-scheme ComposableArchitecture \
 		-destination platform="$(PLATFORM_TVOS)"
+	xcodebuild \
+		-scheme ComposableArchitecture_watchOS \
+		-destination platform="$(PLATFORM_WATCHOS)"
 	xcodebuild test \
 		-scheme ComposableCoreLocation \
 		-destination platform="$(PLATFORM_IOS)"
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -15,6 +15,9 @@ test:
 	xcodebuild test \
 		-scheme ComposableArchitecture \
 		-destination platform="$(PLATFORM_TVOS)"
+	xcodebuild \
+		-scheme ComposableArchitecture_watchOS \
+		-destination platform="$(PLATFORM_WATCHOS)"
```
stephencelis, line 20: While watchOS doesn't have XCTest we should still have CI build it...

Makefile
```diff
@@ -1,11 +1,11 @@
 PLATFORM_IOS = iOS Simulator,name=iPhone 11 Pro Max
 PLATFORM_MACOS = macOS
 PLATFORM_TVOS = tvOS Simulator,name=Apple TV 4K (at 1080p)
+PLATFORM_WATCHOS = watchOS Simulator,name=Apple Watch Series 4 - 44mm
 
 default: test
 
 test:
-	instruments -s devices
```
stephencelis, line 8: This produces a warning on Xcode 12 and we don't really need it so let's get rid of it for now.

Tests/ComposableCoreMotionTests/ComposableCoreMotionTests.swift
```diff
@@ -1,50 +1,52 @@
...
-    XCTAssertEqual(q2.yaw, Double.pi)
-
-    XCTAssertEqual(q3.roll, 0)
-    XCTAssertEqual(q3.pitch, 0)
-    XCTAssertEqual(q3.yaw, Double.pi)
-
-    XCTAssertEqual(q4.roll, 0)
-    XCTAssertEqual(q4.pitch, 0)
-    XCTAssertEqual(q4.yaw, 0)
+#if canImport(CoreMotion)
```
stephencelis, line 1: These tests need the same guard as the library.

Makefile
stephencelis: (This builds all 3 libraries for it, btw.)

Makefile
```diff
@@ -27,6 +30,12 @@ test:
 	xcodebuild test \
 		-scheme ComposableCoreMotion \
 		-destination platform="$(PLATFORM_IOS)"
+	xcodebuild test \
+		-scheme ComposableCoreMotion \
+		-destination platform="$(PLATFORM_MACOS)"
+	xcodebuild test \
+		-scheme ComposableCoreMotion \
+		-destination platform="$(PLATFORM_TVOS)"
```
stephencelis, line 38: As long as we support building core motion for any target we should probably also run the tests.

PR #275, opened 2020-08-31
Localizable Alerts and Action Sheets

### LINKED ISSUE #237: Localizable Alert and ActionSheet buttons
I am currently using `AlertState` and `ActionSheetState` TCA provided to display my Alert and ActionSheets. Unfortunately, both `AlertState<Action>.Button` and `ActionSheetState<Action>.Button` does not support localization since the function only accepts `String` as its argument and not `LocalizedStringKey`. Will there be any updates to support localization on both of these buttons?

stephencelis:
Fixes #237.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
@@ -4,7 +4,7 @@ import SwiftUI
 /// `Action` generic is the type of actions that can be sent from tapping on a button in the sheet.
 ///
 /// This type can be used in your application's state in order to control the presentation or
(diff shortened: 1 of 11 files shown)


Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
```diff
@@ -107,12 +107,12 @@ import SwiftUI
...
-  public var message: String?
-  public var title: String
+  public var message: LocalizedStringKey?
+  public var title: LocalizedStringKey
 
   public init(
-    title: String,
-    message: String? = nil,
+    title: LocalizedStringKey,
+    message: LocalizedStringKey? = nil,
```
stephencelis, line 115: Dunno if we want `S: StringProtocol` overloads for all these initializers, too...

PR #276, opened 2020-09-03
Add reverse implementation in IdentifiedArray

jeffersonsetiawan:
# Why
IdentifiedArray.reverse() just reverse the elements but doesn't reverse the `ids`.


stephencelis (approved):
Thanks! We'll have to look into avoiding these overloads, though, since they're definitely gotchas.

PR #278, opened 2020-09-03
Drive Test Store with a real Store

### LINKED ISSUE #277: Test stores can process actions from effects in the wrong order
**Describe the bug**
A [post on the forums](https://forums.swift.org/t/how-to-test-concatenated-effects-which-call-other-effects/39995) has alerted us to an issue with how a test store can process actions from effects in the wrong order.

**To Reproduce**
I've pushed a failing test here: https://github.com/pointfreeco/swift-composable-architecture/compare/test-store-concat-effect-bug

The expected order can be verified by running the test and checking the console, where an actual store prints debug info that verifies the order.

**Expected behavior**
The test store should process actions from effects in the same order as a live store.

**Environment**
TCA 0.7.0.

**Additional context**
We've tried to unify the concept of store and test store in the past and haven't been able to, and this is why the behaviors can stray from one another. Maybe we can come up with a better solution, though.

stephencelis:
Previously, test stores maintained similar logic to real stores, but we can take advantage of higher-order reducers and local mutable state to drive test stores with real stores under the hood.

Fixes #277.

### DIFF
--- a/Sources/ComposableArchitecture/TestSupport/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestSupport/TestStore.swift
@@ -220,6 +220,7 @@
       file: StaticString = #file,
       line: UInt = #line
     ) {
+      var stateAfterSend = self.state
       var receivedActions: [(action: Action, state: State)] = []
       var longLivingEffects: [String: Set<UUID>] = [:]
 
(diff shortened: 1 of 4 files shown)


Sources/ComposableArchitecture/TestSupport/TestStore.swift
```diff
@@ -168,17 +168,17 @@
     private let toLocalState: (State) -> LocalState
 
     private init(
+      environment: Environment,
```
stephencelis, line 171: Private so alphabetizing.

mbrandonw (approved):
awesome!

Issue #274, opened 2020-08-31
.debug() not working in custom configurations


gkaimakas:
**Describe the bug**
Not sure if this an issue of CA or something I might have missed but I am facing the following:

I have created a new project, renamed debug scheme to dev and copied it to create another one called staging. 
I have also edited each scheme to use the correct configuration.
I did not edit/delete/create any new Custom Flags / Active compilation conditions.

When I call `.debug()` in one of my reducers it does not print anything to the console.
I have also tried to use `#if DEBUG` directly on my project, which surprisingly, works.

**To Reproduce**
The behaviour I am seeing can be reproduced here: 
[CADebugBug.zip](https://github.com/pointfreeco/swift-composable-architecture/files/5155131/CADebugBug.zip)

**Expected behavior**
`.debug()` should print on the console.

**Environment**
 - Xcode [Version 11.6 (11E708)]
 - Swift [5.2.2]
 - OS: [iOS 13.6]
 - CA version: 0.7.0

mathieutozer:
I've been experiencing this too.

kaishin:
I have noticed this too in a similar setup.

AlexisQapa:
IIRC SPM doesn't support custom configurations for now (only debug and release) and Xcode tries to map your configuration name to either debug or release. If you rename your configuration to DevDebug Xcode might build your SPM dependencies with the debug configuration. You might want to differentiate configurations from schemes and use only two configurations and a scheme per environment.

kaishin:
I solved by renaming the configuration to Debug for now. Thanks for the pointer @AlexisQapa!

stephencelis:
This is a known issue being tracked by Apple

- https://forums.swift.org/t/swift-package-manager-and-custom-build-configurations/29181/5
- https://forums.swift.org/t/custom-build-configuration-names/29167/2

I'm going to convert this to a discussion in case folks would like to share their various workarounds.

PR #283, opened 2020-09-09
DEBUG flag defined in package

gkaimakas:
It appears that for a flag to be properly evaluated as a SWIFT_ACTIVE_COMPILATION_CONDITION it has to be explicitly defined in Package.swift.

More info: [docs.swift.org](https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html)

This resolves the issue I was having when I tried to use custom configurations at my project (#274). The same fix might have to be applied to `ComposableCoreLocation` & `ComposableCoreMotion` but since I haven't used them yet I decided not to apply it.  

To validate that the fix is working you can use the [repo](https://github.com/gkaimakas/CADebugBug) I created to demonstrate the issue. You have to use the TCA package as a local dependency ([instructions](https://developer.apple.com/documentation/swift_packages/editing_a_package_dependency_as_a_local_package)) and define the flag there.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -34,7 +34,8 @@ let package = Package(
       dependencies: [
         "CasePaths",
         "CombineSchedulers",
-      ]
+      ],
+      swiftSettings: [.define("DEBUG")]
     ),
     .testTarget(
       name: "ComposableArchitectureTests",


Package.swift
```diff
@@ -34,7 +34,8 @@ let package = Package(
       dependencies: [
         "CasePaths",
         "CombineSchedulers",
-      ]
+      ],
+      swiftSettings: [.define("DEBUG")]
```
stephencelis, line 38: Won't this set `DEBUG` for all builds, including release ones?

gkaimakas: It appears it does set it for all builds. I missed that

stephencelis:
@gkaimakas Thanks for looking into this! We're just not quite sure what the fix is for custom schemes. Is it possible to set the `DEBUG` flag in your project? Maybe in a downstream package that depends on TCA?

gkaimakas:
DEBUG is set in every configuration / scheme but it is not honoured by the package. After some more search it appears it is a limitation of SPM and not only found on TCA. Will look into it further.

As a workaround, I copied your debug code to my project and it works for now.

gkaimakas:
@stephencelis @mbrandonw would you consider removing the `#if DEBUG` statement from .debug() and leave it up to the user to correctly use it? 

Not sure how common is my use case though to justify a change as that.

Issue #279, opened 2020-09-05
`CaseStudy/SwiftUI/Alerts & ActionSheets` Exception on Increment


Mozahler:
This occurs using the zip downloaded from this site this week -- out of the box. 
On the device (iPhone XS Max iOS 13.6) and in simulator (iOS 13.6, iOS 13.7)
It also occurs using a zip downloaded at the beginning of August.

**To Reproduce**
Compile "Case Studies (SwiftUI)" scheme using the ComposableArchitecture.xcworkspace
run the example and select "Alert & ActionSheets"
Select Alert, click on the "Increment" button on the alert
This results in an exception.

**Expected behavior**
Alert to dismiss with counter incremented

This is called twice after tapping on "Increment", the exception occurs after the second .toSwiftUI(send:) call.
 ```
  WithViewStore(store, removeDuplicates: { $0?.id == $1?.id }) { viewStore in
      self.alert(item: viewStore.binding(send: dismiss)) { state in
        state.toSwiftUI(send: viewStore.send)
      }
    }
```
And these lines in the log stand out to me:

SwiftUI`SwiftUI.AlertBridge.preferencesDidChange(SwiftUI.PreferenceList) -> ():

; partial apply forwarder for closure #1 () -> () in SwiftUI.AlertBridge.preferencesDidChange(SwiftUI.PreferenceList) -> ()

` $defer #1 <A, B where A: SwiftUI.View, B: SwiftUI.AlertControllerConvertible, B.Key.Value == Swift.Optional<B>>() -> () in SwiftUI.AlertBridge.preferencesDidChange(SwiftUI.PreferenceList) -> ()

`**Screenshots**
<img width="130" alt="Screen Shot 2020-09-05 at 10 40 15 AM" src="https://user-images.githubusercontent.com/9641531/92307431-8677fd00-ef64-11ea-9a14-4765b697e927.png">

<img width="1402" alt="Screen Shot 2020-09-05 at 10 52 03 AM" src="https://user-images.githubusercontent.com/9641531/92307858-5e3dcd80-ef67-11ea-81c2-32250ea64723.png">

**Environment**
 - Xcode Version 11.7 (11E801a)
 - Apple Swift version 5.2.4 (swiftlang-1103.0.32.9 clang-1103.0.32.53)
Target: On the device (iPhone XS Max iOS 13.6) and in simulator (iPhone SE 2nd Generation, iOS 13.6, iOS 13.7)

[...]

PR #286, opened 2020-09-13
Fix crash in CaseStudies app when tapping on increment action from Alert

nsoojin:
**CaseStudies app crashes when tapping on "Increment" action from Alert.**

According to my investigation, we cannot reuse `var alert: AlertState<AlertAndSheetAction>?` property to trigger new alert from the action of its button. This also doesn't work in vanilla SwiftUI.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndActionSheets.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndActionSheets.swift
@@ -21,6 +21,7 @@ private let readMe = """
 struct AlertAndSheetState: Equatable {
   var actionSheet: ActionSheetState<AlertAndSheetAction>?
   var alert: AlertState<AlertAndSheetAction>?
+  var resultAlert: AlertState<AlertAndSheetAction>?
   var count = 0
 }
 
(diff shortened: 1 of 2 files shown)


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndActionSheets.swift
```diff
@@ -112,6 +118,10 @@ struct AlertAndSheetView: View {
             )
         }
       }
+      .alert(
```
stephencelis, line 121: We've found that nested alert and sheet view modifiers don't work, at least in most versions of SwiftUI: https://twitter.com/mbrandonw/status/1293909768927510529

Does the above `Alert` button still somehow show `alert` state? What version of Xcode and iOS simulator are you running?

stephencelis (commented):
Hey @nsoojin! Thanks for the PR. We actually had discussion about this problem here: https://github.com/pointfreeco/swift-composable-architecture/issues/279

We concluded that this is an Apple bug that only affects Xcode 11.7, so we didn't wanna complicate the case study when it works most places.

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndActionSheets.swift
nsoojin: I don't know what button you are referring to. But when resultAlert is not nil, this alert shows.
Currently I'm using Xcode 11.3 and I tested on real device(iPhone Xs Max)

nsoojin:
@stephencelis Thanks for the reply. I will close it. FYI I'm using Xcode 11.3 and it's happening. Could you tell me which Xcode version this works in?

PR #287, opened 2020-09-14
Buffer actions when sent recursively

mbrandonw:
Right now we do an `assertionFailure` when an action is sent to the store while it is processing an action. This is done for correctness reasons so that the 2nd action can't make state changes or return effects before the 1st action has had time to complete.

This situations arises mostly when one tries to execute an effect directly in a reducer (which is not an appropriate thing to do in general) or when multiple threads send actions to the store (also not appropriate). However, it also happens when simply presenting a UIKit view controller that uses `viewDidLoad` to send an action, which is a totally appropriate thing to do.

So, we want to get rid of the precondition, and instead properly buffer the actions to the store so that they can be processed in a well-defined order.

More info here:
https://forums.swift.org/t/uikit-and-racing-actions/40293/2


stephencelis (approved):
Sweet

kristojorg:
This is great! Am using a binding with a SwiftUI Map and it's sending me `coordinateRegion` changes on what must be a new thread and crashing my app. Glad I found this ticket : )

mbrandonw:
@kristojorg great to hear! Do keep in mind that this does not make `Store` thread safe. `send` should only ever be called on the main thread and effects should deliver their outputs on the main thread. So whatever Map API was being called on a background thread needs a `DispatchQueue.main.async` or `receive(on:)` to make it correct.

kevinrenskers:
Oh my god thank you so much for this PR and the new release. I was getting weird random crashes due to sending actions in `viewDidLoad` and was trying my best to work around it but it kept coming up. It seems all solved now! 🎉

Issue #289, opened 2020-09-15
TicTacToe Crash When Interrupting 2FA Login


nmccann:
**Describe the bug**
When running the TicTacToe application, it is possible to interrupt a long lived side effect on the Two Factor Authentication screen, which results in a crash due to an assertion:
`Fatal error: "LoginAction.twoFactorDismissed" was received by an optional reducer when its state was "nil".`

**To Reproduce**
* Run TicTacToe example project
* Select SwiftUI (though the same bug affects the UIKit version)
* Login with an email containing `2fa`
* Enter `1234` code
* Tap Submit
* Before request completes, press "< Login" in the top left corner
* App crashes due to fatal error

**Expected behavior**
App should return to login screen without crashing

**Environment**
- Xcode 11.7

**Additional context**
This sounds similar to #259 , but isn't quite the same error.
This also sounds similar to the issue [discussed in the forums here](https://forums.swift.org/t/ifletstore-and-effect-cancellation-on-view-disappear/38272), and might be able to be fixed by the solution described there.

This issue covers one of the biggest unknowns I have with this architecture - you're incentivized to break up your code into reducers that focus on a particular task and child reducers don't need to know about their parents, however if the parent has an optional child, then that child needs to expose some way of cancelling all active effects so that the parent can clean up after them (which, as a parent, that is pretty realistic but not super convenient). 

It'd be nice to see a canonical example of handling an optional child with long lived side effects that could get interrupted/cancelled. And if explicitly sending a "cancel" action to the child is the expected process, is there a better way to scale that? For example, if you have an optional child which itself has optional children with side effects that need to be canceled, it gets complicated.

stephencelis:
Good find! That's definitely a bug! And it's one that was reproducible pretty quickly in a test. The fix (and test) are for your perusal here: https://github.com/pointfreeco/swift-composable-architecture/pull/291

stephencelis:
> It'd be nice to see a canonical example of handling an optional child with long lived side effects that could get interrupted/cancelled. And if explicitly sending a "cancel" action to the child is the expected process, is there a better way to scale that? For example, if you have an optional child which itself has optional children with side effects that need to be canceled, it gets complicated.

Absolutely. We'll have a brainstorm on more complex examples like this. If you have any ideas feel free to chime in!

nmccann:
Thank you for the quick response and solution!

stephencelis:
[...]

PR #291, opened 2020-09-15
Fix Tic Tac Toe Optional Bug

stephencelis:
Addressing the good find reported in #289. Experiments with a fix that has the child 2FA module expose a "teardown cancel token" for its parent login module.

### DIFF
--- a/Examples/TicTacToe/Sources/Core/LoginCore.swift
+++ b/Examples/TicTacToe/Sources/Core/LoginCore.swift
@@ -94,7 +94,7 @@ public let loginReducer =
 
       case .twoFactorDismissed:
         state.twoFactor = nil
-        return .none
+        return .cancel(id: TwoFactorTearDownToken())
       }
     }
   )
--- a/Examples/TicTacToe/Sources/Core/TwoFactorCore.swift
+++ b/Examples/TicTacToe/Sources/Core/TwoFactorCore.swift
@@ -23,6 +23,10 @@ public enum TwoFactorAction: Equatable {
   case twoFactorResponse(Result<AuthenticationResponse, AuthenticationError>)
 }
 
+public struct TwoFactorTearDownToken: Hashable {
+  public init() {}
+}
+
 public struct TwoFactorEnvironment {
   public var authenticationClient: AuthenticationClient
   public var mainQueue: AnySchedulerOf<DispatchQueue>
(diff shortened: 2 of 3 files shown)


Examples/TicTacToe/Sources/Core/TwoFactorCore.swift
```diff
@@ -23,6 +23,10 @@ public enum TwoFactorAction: Equatable {
   case twoFactorResponse(Result<AuthenticationResponse, AuthenticationError>)
 }
 
+public struct TwoFactorTearDownToken: Hashable {
```
stephencelis, line 26: This is for simplicity, but I imagine if you want to enumerate over all the cancel tokens you could do something like:

```swift
public enum TwoFactorTearDownToken: CaseIterable, Hashable {
  case twoFactorRequest
  case someOtherLongLivingEffect
}
```

And then in the parent:

```swift
return .merge(TwoFactorTearDownToken.allCases.map(Effect.cancel(id:)))
```

nmccann: I like the approach of using an enum - even if you aren't using it for the purposes of tearing things down, that seems preferable to having a large number of empty structs to handle multiple cancellations.

Examples/TicTacToe/Sources/Core/LoginCore.swift
```diff
@@ -94,7 +94,7 @@ public let loginReducer =
 
       case .twoFactorDismissed:
         state.twoFactor = nil
-        return .none
+        return .cancel(id: TwoFactorTearDownToken())
```
nmccann, line 97: This approach works well for this particular case (that of a single long lived effect in an immediate child), but what about an alternative where you send an action to the child before nilling it out? Pseudocode:
```
case childDismissed:
send teardown action to child
then nil out state.child
```

This seems like it may be easier to extract to a higher order reducer (like the one that handles lifecycles). That also serves to encapsulate the teardown logic in the child. Also, if you had a deeply nested hierarchy it would allow the parent to tell it's child to teardown, which would in turn tell it's own child to teardown etc.

Examples/TicTacToe/Sources/Core/TwoFactorCore.swift
```diff
@@ -56,6 +60,7 @@ public let twoFactorReducer = Reducer<TwoFactorState, TwoFactorAction, TwoFactor
       .receive(on: environment.mainQueue)
       .catchToEffect()
       .map(TwoFactorAction.twoFactorResponse)
+      .cancellable(id: TwoFactorTearDownToken())
```
nmccann, line 63: Is it possible to have multiple cancellation tokens? For instance, one unique token that could be used for each type of effect in a reducer, and another "general" token that could be used to cancel everything? So for each effect you'd have to add a cancellable for both tokens.

That may not be much nicer than the enum approach you mentioned in another comment, but I'm just wondering if it's feasible.

mbrandonw (approved):

PR #294, opened 2020-09-23
Add App.swift file and darkmode compatible colors to VoiceMemos example

willisplummer:
I was playing around with this and noticed a couple quick enhancements


stephencelis:
Thanks again! Closing in favor of #316.

Issue #282, opened 2020-09-08
Optional reducer crash when onAppear used from TabView


maciejczechowski:
**Describe the bug**
We hit assertion on optional reducer if onAppear (from optional state) is called within tabView.
OnAppear is called immediately after state is set to nil, causing a crash due to assertion.

**To Reproduce**
The project file is here:
[ComposableTests.zip](https://github.com/pointfreeco/swift-composable-architecture/files/5189287/ComposableTests.zip)

In general, the structure is:

We have a root view, which can be either login or content.
Content has an option to log out.
 Right now this is done via struct, we'll be experimenting with enums soon (I don't think it has anything to do with that):

```swift
struct LoginState: Equatable {}
struct LoggedInState: Equatable{}

enum LoginAction: Equatable {  case login }
enum LoggedInAction: Equatable {
    case logout
    case onAppear
}

struct RootState: Equatable {
    var loginState: LoginState?
    var loggedInState: LoggedInState?
    
    init(){
        loginState = LoginState()
    }
}

enum RootAction: Equatable{
    case login(LoginAction)
    case loggedIn(LoggedInAction)
}
struct Environment {}

```

The reducer is a combination of optional pullbacks and switching main state:
```swift
let loginReducer = Reducer<LoginState, LoginAction, Environment>
{ state, action, environment in
    return .none
}

let loggedInReducer = Reducer<LoggedInState, LoggedInAction, Environment>
{ state, action, environment in
    return .none
}

let rootBaseReducer = Reducer<RootState, RootAction, Environment>{
    state, action, environment in
    switch action {
    
    case .login(.login):
        state.loggedInState = LoggedInState()
        state.loginState = nil
    case .loggedIn(.logout):
        state.loggedInState = nil
        state.loginState = LoginState()
    default:
        return .none
    }
    return .none
}

let rootReducer = Reducer<RootState,RootAction, Environment>.combine(
    loginReducer.optional().pullback(state: \.loginState,
                                     action: /RootAction.login,
                                     environment: { $0 }),
    
    loggedInReducer.optional().pullback(state: \.loggedInState,
                                        action: /RootAction.loggedIn,
                                        environment: { $0 }),
    rootBaseReducer
)
```

The content view is a simple tab:
```swift

struct ContentView: View {
    let store: Store<RootState, RootAction>
    var body: some View {
        IfLetStore(self.store.scope(state: {$0.loginState},
                                    action: RootAction.login),
                   then: { store in
[...]

PR #296, opened 2020-09-24
Breakpoint instead of assert in optional/forEach

stephencelis:
Based on feedback and earlier discussion.


Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -6,11 +6,11 @@ import Combine
 ///
 /// Reducers have 3 generics:
 ///
-/// * `State`: A type that holds the current state of the application
-/// * `Action`: A type that holds all possible actions that cause the state of the application to
-///   change.
-/// * `Environment`: A type that holds all dependencies needed in order to produce `Effect`s, such
-///   as API clients, analytics clients, random number generators, etc.
+///   * `State`: A type that holds the current state of the application.
```
stephencelis, line 9: Just being consistent here. I added some code examples embedded in bullet points later on in this PR, which requires an indentation level of 8, so it looks best to make sure the text of bullets are indented by 4.

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -266,9 +266,9 @@ public struct Reducer<State, Action, Environment> {
   /// Transforms a reducer that works on non-optional state into one that works on optional state by
   /// only running the non-optional reducer when state is non-nil.
   ///
-  /// Often used in tandem with `pullback` to transform a reducer on a non-optional local domain
-  /// into a reducer that can be combined with a reducer on a global domain that contains some
-  /// optional local domain:
+  /// Often used in tandem with `pullback` to transform a reducer on a non-optional child domain
+  /// into a reducer that can be combined with a reducer on a parent domain that contains some
+  /// optional child domain:
```
stephencelis, line 271: Any thoughts on updating our terminology from global-local to parent-child? It read better to me here. Global alone sounds like app-level domain, so you need to qualify as "more global" or "more local"...

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -290,39 +290,162 @@ public struct Reducer<State, Action, Environment> {
...
-        )
+        #if DEBUG
+          if breakpointOnNil {
+            fputs(
+              """
+              ---
+              Warning: Reducer.optional@\(file):\(line))
+
+              "\(debugCaseOutput(action))" was received by an optional reducer when its state was \
+              "nil". This is generally considered an application logic error, and can happen for a \
```
stephencelis, line 426: Qualified this with "generally" everywhere now and note that it is reasonable to ignore the result of some effects in a bullet point below.

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -290,39 +290,162 @@ public struct Reducer<State, Action, Environment> {
...
-          this optional reducer are canceled when optional state is set to "nil".
-
-          * This action was sent to the store while state was "nil". Make sure that actions for \
-          this reducer can only be sent to a view store when state is non-"nil". In SwiftUI \
-          applications, use "IfLetStore".
-          """,
-          file: file,
-          line: line
-        )
+        #if DEBUG
```
stephencelis, line 418: Won't check for breakpoints on release builds.

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -290,39 +290,162 @@ public struct Reducer<State, Action, Environment> {
...
+          if breakpointOnNil {
+            fputs(
```
stephencelis, line 420: Using `fputs` to print to stderr.

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -290,39 +290,162 @@ public struct Reducer<State, Action, Environment> {
...
+
+              * An in-flight effect emitted this action while state was "nil". While it may be \
+              perfectly reasonable to ignore this action, you may want to cancel the associated \
+              effect before state is set to "nil", especially if it is a long-living effect.
+
+              * This action was sent to the store while state was "nil". Make sure that actions \
+              for this reducer can only be sent to a view store when state is non-"nil". In \
+              SwiftUI applications, use "IfLetStore".
+              ---
+
```
stephencelis, line 442: This trailing line is needed since `fputs` doesn't print a trailing new line.

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -360,44 +483,59 @@ public struct Reducer<State, Action, Environment> {
   ///     `GlobalState`.
   ///   - toLocalAction: A case path that can extract/embed `(Int, Action)` from `GlobalAction`.
   ///   - toLocalEnvironment: A function that transforms `GlobalEnvironment` into `Environment`.
+  ///   - breakpointOnNil: Raises `SIGTRAP` signal when an action is sent to the reducer but the
+  ///     index is out of bounds. This is generally considered a logic error, as a child reducer
+  ///     cannot process a child action for unavailable child state.
```
stephencelis, line 488: I dunno if `breakpointOnNil` is the right name for the `forEach` reducers...maybe we can come up with something better.

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -290,39 +290,162 @@ public struct Reducer<State, Action, Environment> {
...
+  ///         IfLetStore(
+  ///           self.parentStore.scope(state: { $0.child }, action: { .child($0) }
+  ///         ) { childStore in
+  ///           // This destination only appears when child state is non-`nil`
+  ///           WithViewStore(childStore) { childViewStore in
+  ///             // So this action can only be sent when child state is non-`nil`
+  ///             Button("Child Action") { childViewStore.send(.action) }
+  ///           }
+  ///           ...
+  ///         }
```
stephencelis, line 398: If these additional docs look good I can create similar versions for `forEach`, or I can add a "see also" in `forEach` to read the `optional` documentation for more info (although I think the explanations are slightly different).

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -290,39 +290,162 @@ public struct Reducer<State, Action, Environment> {
...
   /// - See also: `Store.ifLet`, a UIKit helper for doing imperative work with a store on optional
   ///   state.
-  public func optional(_ file: StaticString = #file, _ line: UInt = #line) -> Reducer<
+  ///
+  /// - Parameter breakpointOnNil: Raises `SIGTRAP` signal when an action is sent to the reducer
+  ///   but state is `nil`. This is generally considered a logic error, as a child reducer cannot
+  ///   process a child action for unavailable child state.
+  /// - Returns: A reducer that works on optional state.
+  public func optional(
+    breakpointOnNil: Bool = true,
```
nmccann, line 410: Thank you for making this change - I definitely understand the reasons for the previous strict behaviour, but I think this is a nice balance. Default to strict, with the option to ignore if needed.

I've run into a particular case recently with a NavigationLink's binding that sends a "dismiss" action after the child state is nullified, and the only way I've found to work around it is to put in a delay to let the navigation complete before nulling out the child state. Ignoring the error would be preferable to the delay. Though I acknowledge that there may be something incorrect about how I'm using the navigation link that is causing the problem.

pteasima: As always, thanks for the great work 👏 .

Would you consider a setup where we take a `canIgnoreAction: (Action) -> Bool = { _ in false }` parameter? Id love to retain the strict behavior for most actions and explicitly list those that can come in post mortem.

jasdev:
(No rush of course, is there a targeted dot release for this? If not, I can totally work off `dont-assert-signal` and then switch back over to an official release when it’s ready. 👌🏽 Thanks for your work on adding this y’all. 🙏🏽)

nmccann:
After encountering a bug similar to [this one](https://github.com/pointfreeco/swift-composable-architecture/issues/282) I was only able to workaround it by switching to this branch - I'm hoping to further investigate and fix the underlying issue, but being able to fallback to this less strict behaviour is very valuable to help bypass some of the odd things that currently happen in SwiftUI.

I would appreciate if this could be merged, or if you could share the reasoning for it not being merged.

stephencelis:
@nmccann We plan on merging it but we want to make sure the docs and API is right. We'll try to get things across the finish line soon.

SteinerHannes:
Any progress on this?

mbrandonw (approved):

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -290,39 +290,162 @@ public struct Reducer<State, Action, Environment> {
...
   ///
   /// - See also: `IfLetStore`, a SwiftUI helper for transforming a store on optional state into a
   ///   store on non-optional state.
   /// - See also: `Store.ifLet`, a UIKit helper for doing imperative work with a store on optional
   ///   state.
-  public func optional(_ file: StaticString = #file, _ line: UInt = #line) -> Reducer<
+  ///
+  /// - Parameter breakpointOnNil: Raises `SIGTRAP` signal when an action is sent to the reducer
+  ///   but state is `nil`. This is generally considered a logic error, as a child reducer cannot
+  ///   process a child action for unavailable child state.
```
ferologics, line 407: NB: this almost exclusively happens because of erratic `NavigationLink` binding behavior. The action in the binding gets called multiple times, for mysterious reasons.

Sources/ComposableArchitecture/Reducer.swift
ferologics: @stephencelis In my experience this almost exclusively happens because of erratic NavigationLink binding behavior. The action in the binding gets called multiple times, for mysterious reasons. Maybe it's worthwhile to list it as one of the sub points and an Apple bug?

pteasima: I can confirm that NavigationLinks are weird in my CA apps too, and my I also blame SwiftUI 😅.

But @ferologics, doesn't something as simple as sending an action from `onDisappear` count?

PR #299, opened 2020-09-29
Fixes bug with LocalizedStringKeyTests

mluisbrown:
Currently, when you run the `LocalizedStringKeyTests` on a machine which is not using the `en_US` locale, they will fail:

`("You are 1 000th in line.") is not equal to ("You are 1,000th in line.")`

I currently have a custom locale where the number grouping symbol is a space, but many European locales use `.` for grouping and `,` for decimals. It would also fail for these locales.

The fix I made was to force the `en_US` locale for the `NumberFormatter` used in the test.

Whilst I was investigating this, I also found that the `formatted()` function in the `LocalizedStringKey` extension ignores the `locale` parameter passed, so I changed it to not ignore this parameter.


Tests/ComposableArchitectureTests/LocalizedStringKeyTests.swift
```diff
@@ -11,6 +11,7 @@ class LocalizedStringKeyTests: XCTestCase {
     )
 
     let formatter = NumberFormatter()
+    formatter.locale = Locale(identifier: "en_US")
```
ole, line 14: Nice change. You could even consider using `"en_US_POSIX"`, which is guaranteed not to change in the future. (Not that it's a big deal if the test breaks in the distant future if the US ever changes its number format.)

Sources/ComposableArchitecture/Internal/LocalizedStringKey.swift
```diff
@@ -21,7 +21,7 @@ extension LocalizedStringKey: CustomDebugOutputConvertible {
         return formatter?.string(for: value) ?? value as! CVarArg
       }
 
-    return String(format: key, locale: nil, arguments: arguments)
+    return String(format: key, locale: locale, arguments: arguments)
```
stephencelis, line 24: Ha good catch!

stephencelis (approved):
Looks good! Wanna make the small change @ole suggested?

mluisbrown:
Just made the change that @ole suggested 👍

PR #303, opened 2020-10-09
Implement heading methods for core location library

dannyhertz:
The start/stop heading methods were not implemented in the live core location manager. I added them!


stephencelis:
Good find, thanks!

Issue #306, opened 2020-10-14
Support for accuracyAuthorization API in ComposableCoreLocation


mackoj:
**Describe the bug**

Some part of `CLLocationManager` are not accessible in the the `LocationManager` like `accuracyAuthorization`.

**To Reproduce**

Try to call `accuracyAuthorization` on `LocationManager`.

**Expected behavior**

`LocationManager` should be a mirror of `CLLocationManager ` and propose the same feature.

**Environment**
 - Xcode 12
 - Swift 5.3
 - iOS 14

stephencelis:
Thanks for the report, @mackoj! Did you wanna try to PR? If not we'll try to get to this sometime soon.

mackoj:
I have a working patch I will submit a PR as soon as I can.

PR #307, opened 2020-10-14
Add support for AccuracyAuthorization

mackoj:
Fix for https://github.com/pointfreeco/swift-composable-architecture/issues/306

https://developer.apple.com/documentation/corelocation/claccuracyauthorization

- [x] Fix compiling issue with on macOS

### DIFF
--- a/Sources/ComposableCoreLocation/Interface.swift
+++ b/Sources/ComposableCoreLocation/Interface.swift
@@ -274,6 +274,10 @@ public struct LocationManager {
 
   var location: (AnyHashable) -> Location? = { _ in _unimplemented("location") }
 
+  var accuracyAuthorization: (AnyHashable) -> AccuracyAuthorization? = { _ in
+    _unimplemented("accuracyAuthorization")
+  }
+
   public var locationServicesEnabled: () -> Bool = { _unimplemented("locationServicesEnabled") }
 
   @available(tvOS, unavailable)
@@ -394,6 +398,10 @@ public struct LocationManager {
   public func heading(id: AnyHashable) -> Heading? { self.heading(id) }
 
   public func location(id: AnyHashable) -> Location? { self.location(id) }
+
+  public func accuracyAuthorization(id: AnyHashable) -> AccuracyAuthorization? {
+    self.accuracyAuthorization(id)
+  }
 
   @available(tvOS, unavailable)
   @available(watchOS, unavailable)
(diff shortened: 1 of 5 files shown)


mackoj:
If someone has an idea how to fix the macOS build I'm listening...

andreyz:
@mackoj I think you need to add 2 lines with compiler conditionals to hide CLAccuracyAuthorization from swift 5.2 compiler:, like so

``` swift
public enum AccuracyAuthorization {
  case fullAccuracy
  case reducedAccuracy

  #if compiler(>=5.3)
  @available(iOS 14, macCatalyst 14, macOS 11, tvOS 14, watchOS 7, *)
  init?(_ accuracyAuth: CLAccuracyAuthorization?) {
    switch accuracyAuth {
    case .fullAccuracy:
      self = .fullAccuracy
    case .reducedAccuracy:
      self = .reducedAccuracy
    default:
      return nil
    }
  }
  #endif
}
```

mackoj:
Thanks I will it

mackoj:
Hi @andreyz,

It don't seems to work either with this modification.

mackoj:
```swift
#if compiler(>=5.3)
```

or

```swift
#if swift(>=5.3)
```

Don't seems to work either.

stephencelis:
Thanks for the PR, @mackoj! We haven't had time to dig into the failure yet but we'll try to get to it soon.

andreyz:
> ```swift
> ```swift
> #if compiler(>=5.3)
> ```
> 
> 
> or
> ```swift
> #if swift(>=5.3)
> ```
> 
> 
> Don't seems to work either.
> ```

We were almost there, we needed `#if compiler(>=5.3.1)` based on https://en.wikipedia.org/wiki/Xcode#Toolchain_versions

mackoj:
@andreyz Thanks for the fix

Sources/ComposableCoreLocation/Interface.swift
```diff
@@ -395,6 +399,10 @@ public struct LocationManager {
 
   public func location(id: AnyHashable) -> Location? { self.location(id) }
 
+  public func accuracyAuthorization(id: AnyHashable) -> AccuracyAuthorization? {
```
stephencelis, line 402: One small change. Let's guard the client from calling code that is unavailable:
```suggestion
  @available(iOS 14.0, tvOS 14.0, watchOS 7.0, macOS 11.0, macCatalyst 14.0, *)
  public func accuracyAuthorization(id: AnyHashable) -> AccuracyAuthorization? {
```

stephencelis:
Looks great, thanks again!

PR #311, opened 2020-10-19
Don't depend on DispatchQueue's Equatable conformance

mluisbrown:
Among the changes I had to make in order to get the ReactiveSwift fork [building for Linux](https://github.com/trading-point/reactiveswift-composable-architecture/pull/27) was not depending on the `Equatable` conformance of `DispatchQueue`, and I realised that it probably isn't necessary at all. 

`Equatable` conformance allows the `switch self` and the matching on `.main`, however, in Linux `DispatchQueue` does not conform to `Equatable`, so this doesn't work.

Since for the other queues we are doing name matching we could do the same for `.main` whose name is always `"com.apple.main-thread"`


stephencelis (approved):
Ah nice, looks good!

PR #314, opened 2020-10-29
Rename publisher-transforming scope

stephencelis:
One small part in an effort to make views with stores easier to compile. We can test some of these overloads out before merging, but I think renaming this one is an improvement in general.


Sources/ComposableArchitecture/Store.swift
```diff
@@ -91,7 +91,7 @@ public final class Store<State, Action> {
   ///   - fromLocalAction: A function that transforms `LocalAction` into `Action`.
   /// - Returns: A publisher of stores with its domain (state and action) transformed.
   public func scope<P: Publisher, LocalState, LocalAction>(
-    state toLocalState: @escaping (AnyPublisher<State, Never>) -> P,
+    statePublisher toLocalState: @escaping (AnyPublisher<State, Never>) -> P,
```
mbrandonw, line 94: one other option to consider is renaming the function since it returns a publisher:

```swift
func publisherScope(...)
func scopePublisher(...)
```

mbrandonw: and should we do an `unavailable` release to help people out before fully removing?

stephencelis: Ah yeah I actually named it that first that but ended up changing to the argument label is since actions aren't scoped by publisher:

`scope(statePublisher:action:)` vs. `publisherScope(state:action:)`

Thoughts?

mbrandonw: I think `publisherScope` since it describes what's be returned, and it's nice to keep the shortness of `state:` and `action`:.

stephencelis: Done! Along with deprecations. I think it reads nicely.

mbrandonw (approved):

PR #318, opened 2020-11-02
Verbatim Alerts and Action Sheets

### LINKED ISSUE #293: [Breaking change,  0.8.0] AlertState expects LocalizedStringKey instead of String values
#275 introduced the ability to localise values based on `LocalizedStringKey`s. We pass the localized values directly into our AlertStates as we're using SwiftGen to generate type-safe string accessors. 

@stephencelis asked directly in #275 if TCA should still offer an initialiser that accepts `S: StringProtocol`. Is there anything that would speak against that? If not, I'm happy to open a PR for it. 🙂

ohitsdaniel:
Resolves #293.

@stephencelis introduced localizable Alerts and Actions sheets in #275. Unfortunately, this led to a breaking change for all library consumers that use localized, verbatim values directly. In order to avoid a breaking change, this PR introduces convenience initialisers for `AlertState<Action>` that allow to pass verbatim strings directly. Using `<S: StringProtocol>` (as mentioned in the original PR) instead of `String` directly was avoided as it would have lead to the generic type S bubbling up into the type definition (i.e. `AlertState<Action, S: StringProtocol>`). 

I'm not quite sure how we should handle UIKit interoperability, but this was already a problem with the previous approach as `NSLocalizedString` does not accept a `LocalizedStringKey` directly. The tic tac toe example fell back to using the formatted key as it's value, but this somehow doesn't feel right?

### DIFF
--- a/Examples/TicTacToe/Sources/Views-UIKit/LoginViewController.swift
+++ b/Examples/TicTacToe/Sources/Views-UIKit/LoginViewController.swift
@@ -136,14 +136,29 @@ class LoginViewController: UIViewController {
         guard let self = self else { return }
         guard let alert = alert else { return }
 
+        let title: String
+        switch alert.title {
+        case let .localized(key):
+          title = key.formatted()
+        case let .verbatim(content):
+          title = content
+        }
+
         let alertController = UIAlertController(
-          title: alert.title.formatted(), message: nil, preferredStyle: .alert)
+          title: title,
+          message: nil,
+          preferredStyle: .alert
+        )
+
         alertController.addAction(
           UIAlertAction(
             title: "Ok", style: .default,
             handler: { _ in
               self.viewStore.send(.alertDismissed)
-            }))
+            }
+          )
+        )
+
         self.present(alertController, animated: true, completion: nil)
       }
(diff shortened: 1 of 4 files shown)


Examples/TicTacToe/Sources/Views-UIKit/LoginViewController.swift
```diff
@@ -136,14 +136,29 @@ class LoginViewController: UIViewController {
...
+          title = content
+        }
+
         let alertController = UIAlertController(
-          title: alert.title.formatted(), message: nil, preferredStyle: .alert)
+          title: title,
+          message: nil,
+          preferredStyle: .alert
+        )
+
```
ohitsdaniel, line 152: -> UIKit interoperability problem.

stephencelis:
@ohitsdaniel Thanks for submitting this and sorry it took us so long to get around to looking into the problem. I've explored an alternate solution in #359, which has its own trade-offs. Gonna discuss to figure out how to unblock anyone affected by the issue.

stephencelis:
Closing in favor of #359. Thanks again for bringing this to our attention and for the solution! It helped push us to the finish line :smile:

PR #319, opened 2020-11-02
Remove LocationManager and MotionManager examples from README

mluisbrown:
Just removing these from the examples README 😄


stephencelis (approved):
🙈

PR #321, opened 2020-11-15
Fix docs for ViewStore.binding overload.

mbrandonw:


Sources/ComposableArchitecture/ViewStore.swift
```diff
@@ -199,7 +199,7 @@ public final class ViewStore<State, Action>: ObservableObject {
   ///
   /// For example, an alert binding can be dealt with like this:
   ///
-  ///     struct State { var alert: String? }
+  ///     typealias State = String
```
stephencelis, line 202: Ha I'm a lil confused about these examples in general now. Since we have `alert.message` below...

mbrandonw: Yeah this one has more things wrong with it than I saw at first. This probably got out of sync after the alert stuff.

mbrandonw:
Just pushed some updates to the docs. We should reconsider some of these overloads too.

Sources/ComposableArchitecture/ViewStore.swift
```diff
@@ -199,14 +199,14 @@ public final class ViewStore<State, Action>: ObservableObject {
...
   ///     enum Action { case alertDismissed }
   ///
   ///     .alert(
-  ///       item: self.store.binding(
+  ///       item: viewStorestore.binding(
```
stephencelis, line 206:
```suggestion
  ///       item: viewStore.binding(
```

stephencelis (approved):

PR #327, opened 2020-11-30
Ensure objectWillChange.send() is called from the main thread.

mycroftcanner:
> stephencelis > We try to note this in the README's FAQ, but TCA does not have an opinion on threading, and does no work to run on the main thread. The fact that it typically runs on the main thread in SwiftUI applications is due to the fact that most SwiftUI actions are sent on the main thread. If you want to run TCA on a background queue, you need to send actions to the store on that queue, and your reducer must receive effects on that queue.

If we are sending actions to the store from a background queue and receiving effects on that queue, we must also ensure that objectWillChange.send() will always be called from the main thread. 

>Publishing changes from background threads is not allowed.

### DIFF
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -68,7 +68,9 @@ public final class ViewStore<State, Action>: ObservableObject {
   /// The current state.
   public private(set) var state: State {
     willSet {
-      self.objectWillChange.send()
+      DispatchQueue.main.async { [weak self] in
+        self?.objectWillChange.send()
+      }
     }
   }
 


mbrandonw:
Hi @mycroftcanner, this change will have very far reaching and potentially dangerous ramifications. SwiftUI is very particular with how it gets notified of state changes, and I really have no idea how SwiftUI would handle this situation.

The best way to handle this is to make sure to always `.send` from the main thread in your UI code, and to make sure you effects always deliver on the main thread via `.receive(on:)`. This puts the responsibility on you rather than baking this notion directly into the library.

Also, it is not true that you can _never_ send actions to a store on a background thread. It's only true if you are using `ViewStore` in a SwiftUI view because updates to the `@ObservedObject` must be done on the main thread. If you are not using that then you are free to send actions on a background thread.

However, you must also be aware that the store does no synchronization whatsoever, so you must further make sure that _all_ actions are sent from the same thread. You should not send actions from multiple threads.

mycroftcanner:
>Hi @mycroftcanner, this change will have very far reaching and potentially dangerous ramifications. SwiftUI is very >particular with how it gets notified of state changes, and I really have no idea how SwiftUI would handle this situation.

I'll do some more testing but it seems that as long as objectWillChange.send() is called on the main thread, SwiftUI doesn't seem to mind.

> The best way to handle this is to make sure to always .send from the main thread in your UI code, and to make sure you effects always deliver on the main thread via .receive(on:). This puts the responsibility on you rather than baking this notion directly into the library.

I was doing just that but ran into some serious performance issues when my reducers. I have a complex hierarchy of recursive reducers driving some ASCollectionNodeController and it seems that when I navigate a few level down the hierarchy you can really feel some lag when actions are sent to the reducer while scrolling down (when you use Texture / AsyncDisplayKit you tend to notice the slightest lag as it is virtually frictionless as long as you stay away from the main thread.)

All my effects are already running in the background. Running the reducer on a background thread completely eliminated the lag. I'll create a sample project if you want to take a look.

> Also, it is not true that you can never send actions to a store on a background thread. It's only true if you are using ViewStore in a SwiftUI view because updates to the @ObservedObject must be done on the main thread. If you are not using that then you are free to send actions on a background thread. 

I didn't say you can never send actions to a store on a background thread. 

> However, you must also be aware that the store does no synchronization whatsoever, so you must further make sure that all actions are sent from the same thread. You should not send actions from multiple threads.

All my actions are sent from the same thread. It would be nice to be able to pass a queue to ViewStore.

mbrandonw:
Can you profile your code and see where the slow down is? There is a chance you are running into a performance problem with the reflection computation that case paths do. If that's the case then you may want to hand write the case paths to see if it improves performance. The reflection is many times slower than a hand-rolled case path, but the overall time is still so small that it isn't usually noticeable unless the code is being called many times.

> I didn't say you can never send actions to a store on a background thread.

Sorry, I was only commenting on this pull quote: "Publishing changes from background threads is not allowed". Not sure where that was stated, but it's not a claim we make as a universal truth. Only in the context of SwiftUI.

mycroftcanner:
> Can you profile your code and see where the slow down is? There is a chance you are running into a performance problem with the reflection computation that case paths do. If that's the case then you may want to hand write the case paths to see if it improves performance. The reflection is many times slower than a hand-rolled case path, but the overall time is still so small that it isn't usually noticeable unless the code is being called many times.
> 
> > I didn't say you can never send actions to a store on a background thread.
> 
> Sorry, I was only commenting on this pull quote: "Publishing changes from background threads is not allowed". Not sure where that was stated, but it's not a claim we make as a universal truth. Only in the context of SwiftUI.

Oh the quote is from the purple Xcode warning that appear when you call objectWillChange.send from a background thread.

I'll profile the code again tomorrow. Could you please provide an example of a hand-rolled case path?

mbrandonw:
Sure thing, if you have a nested enum action like this:

```swift
enum AppAction {
  case profile(ProfileAction)
}

enum ProfileAction {
  case buttonTapped
}
```

You can typically just do `/AppAction.profile` in order to construct a `CasePath<AppAction, ProfileAction>`. However, that is using reflection stuff under the hood to figure things out. Instead you can just construct one directly:

```swift
let profileCasePath = CasePath<AppAction, ProfileAction>(
  embed: AppAction.profile,
  extract: { 
    guard case let .profile(action) = $0 else { return nil }
    return action
  }
)
```

You can also house this value as a static in the `CasePath` type:

```swift
extension CasePath where Root == AppAction, Value == ProfileAction {
  static let profile = Self(
    embed: AppAction.profile,
    extract: { 
      guard case let .profile(action) = $0 else { return nil }
      return action
    }
  )
}
```

And then in your pullbacks you can even do:

```swift
reducer.pullback(state: ..., action: .profile, environment: ...)
```

Ideally Swift would be able to write this code for us like it does for key paths, and then we could eliminate the boilerplate and get the performance benefits. However, until then it can be sometimes necessary to do these hand rolled case paths in particularly hot code paths.

mycroftcanner:
This part is easy:

```swift
extension CasePath where Root == AppAction, Value == UserCollectionAction {
  static let search = Self(
    embed: AppAction.search, extract: {
      guard case let .search(action) = $0 else { return nil }
      return action
    }
  )
}

public let appReducer = Reducer<AppState, AppAction, AppEnvironment>.combine(
  campaignsReducer.pullback(
    state: \.campaigns,
    action: /AppAction.campaigns,
    environment: \.campaigns
  ),
  settingsReducer.pullback(
    state: \.settings,
    action: /AppAction.settings,
    environment: \.settings
  ),
  Reducer.empty.itemCollection(
    state: \.search,
    action: .search,
    environment: \.search,
    itemReducer: igUserReducer,
    done: IGUserAction.done
```

However in my high order reducer, I have difficulties creating a case path using this `let profileCasePath = CasePath<AppAction, ProfileAction>([...]` construct.

```swift
public extension Reducer {
  /// Enhances a reducer with collection logic.
  func itemCollection<ItemState, ItemAction, ItemEnvironment>(
    state: WritableKeyPath<State, ItemCollectionState<ItemState, ItemAction>>,
    action: CasePath<Action, ItemCollectionAction<ItemState, ItemAction>>,
    environment: @escaping (Environment) -> ItemCollectionEnvironment<ItemState, ItemEnvironment>,
    itemReducer: Reducer<ItemState, ItemAction, ItemEnvironment>,
    done: ItemAction? = nil
  ) -> Reducer where ItemState: Equatable & Identifiable, ItemAction: Equatable {
    Reducer<
      ItemCollectionState<ItemState, ItemAction>,
      ItemCollectionAction<ItemState, ItemAction>,
      ItemCollectionEnvironment<ItemState, ItemEnvironment>
    >.combine(
      itemReducer
        .forEach(
          state: \.items,
          action: /ItemCollectionAction<ItemState, ItemAction>.item(id:action:),
          environment: { $0.item }
        ),
      .init { state, action, environment in
```

mycroftcanner:
Thanks again for your suggestions. Using hand rolled CasePath improved the performance a lot. 

We still had some issues with performance... One of the API calls we were hitting was returning 1000 results instead of the 50 we expected. We removed all the extraneous struct copies we were making and got rid of most of the bottlenecks and implemented a workaround so that we wouldn't have to create a thousand states at once.

Interestingly, having the reducer running on a background had barely any impact on the user experience even under high load. 

I'll close the PR but it would be interesting to thing what changes would be necessary to easily run TCA in the background (dealing with SwiftUI, etc.)

PR #329, opened 2020-12-03
Adding some convenient initialisers

rkreutz:
Hey there, I believe this is a rather straightforward PR.

I've added a few convenience initialisers so I could combine different Reducers for different context (actions), them being:
- `init(forAction:handler)` & `init(forActions:handler)` which will execute the Reducer's `handler` if it receives the provided action(s). It handles actions with or without (`Void`) associated values.
- `init(bindingAction:to:handler)` & `init(bindingAction:to:handler)` which will automatically set an actions associated value to a key path in the state, optionally providing a handler to further customise state and/or trigger different side effects.

I thought this would be worth sharing :)

### DIFF
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -64,7 +64,7 @@ public struct Reducer<State, Action, Environment> {
   ///   - state: The current state of the reducer. This is mutable since it is an `inout` variable.
   ///   - actionResult: The associated type of the triggered action.
   ///   - environment: The environment stored in the reducer
-  init<ActionResult>(
+  public init<ActionResult>(
     forAction casePath: CasePath<Action, ActionResult>,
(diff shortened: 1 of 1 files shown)


stephencelis:
Hey @rkreutz! This one slipped by, sorry about that! 😄 Holiday season must've caught us off guard.

These are definitely fun! We have some other reducers that we'd like to bring into the core library, but think it'll be a matter of motivating them with examples and case studies.

If you have some interesting uses of these initializers we'd definitely take a PR for a case study first, and if we find folks are copying and pasting the logic into their code bases a lot we'd definitely consider moving them into the library itself!

Since you opened this PR I believe we opened the discussions tab on this repo. You might be interested in re-sharing your helpers there. Maybe with some examples that can be worked into a case study!

Okay to close this out in favor of those directions?

rkreutz:
hey @stephencelis,

yeah makes total sense, I'll open a thread on the discussion tab with these, no problem. Thanks for the clarification.

I'll close this now.

rkreutz:
https://github.com/pointfreeco/swift-composable-architecture/discussions/450

PR #332, opened 2020-12-14
Fix Reducer Docs

SteinerHannes:
Replace `struct` with `enum` for actions


stephencelis (approved):
Ha good catches! Sorry for the delay 😄

PR #335, opened 2020-12-18
Store.cancelAll()

hfossli:
Related to a discussion (coming in some minutes) on Swift Forums

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -6,6 +6,7 @@ import Foundation
 ///
 /// You will typically construct a single one of these at the root of your application, and then use
 /// the `scope` method to derive more focused stores that can be passed to subviews.
+
 public final class Store<State, Action> {
   var state: CurrentValueSubject<State, Never>
   var effectCancellables: [UUID: AnyCancellable] = [:]
@@ -14,7 +15,8 @@ public final class Store<State, Action> {
   private let reducer: (inout State, Action) -> Effect<Action, Never>
   private var synchronousActionsToSend: [Action] = []
   private var bufferedActions: [Action] = []
-
+  public private (set) var cancellationId: AnyHashable?
+  
   /// Initializes a store from an initial state, a reducer, and an environment.
   ///
   /// - Parameters:
@@ -31,7 +33,7 @@ public final class Store<State, Action> {
       reducer: { reducer.run(&$0, $1, environment) }
     )
   }
-
+  
   /// Scopes the store to one that exposes local state and actions.
   ///
   /// This can be useful for deriving new stores to hand to child views in an application. For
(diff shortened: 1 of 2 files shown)


hfossli:
Related to this discussion
https://forums.swift.org/t/how-can-the-composable-architecture-really-become-composable/43045/2

hfossli:
Also, have a look at
https://github.com/hfossli-agens/tca-ifletstore-effect-cancellation-demo/blob/nested-views-viewstore-cancelAll/Demo/App.swift#L79
and
https://github.com/hfossli-agens/tca-ifletstore-effect-cancellation-demo/blob/nested-views-viewstore-cancelAll/Demo/Detail.swift#L117

hfossli:
What do you think? Is this a good approach? If I write tests for this, can it be merged?

stephencelis:
Hey @hfossli! Sorry for not weighing in. Your solutions have definitely been on our radar, and we've been thinking about the problem and various solutions, but we haven't quite decided what direction to take yet. Thanks for sharing, though! We'll try to take a closer look soon.

hfossli-agens:
Thanks! That means a lot to hear :) It is definetely a tricky one. Looking forward to hear your thoughts. 🥰

ferologics:
Bump; I could really make use of something like this now. I assumed that when the state of the optional reducer is `nil` it's effects will be cancelled, but that's not the case for some reason.

stephencelis:
@hfossli :wave: We've been thinking about this problem for awhile and because we haven't had a solution we've been happy with, we've left all PRs open to think about.

Our [Concurrency Beta](https://github.com/pointfreeco/swift-composable-architecture/discussions/1186) provides one new solution, though, that we'd love your opinion on. Namely, it begins to return a task handle from `ViewStore.send`, which can be awaited and cancelled in the `View.task` modifier.

```swift
.task { await viewStore.send(.onAppear) }
```

When the view goes away, it cancels this task and all effects spun up by the associated action will be cancelled, as well.

johnrogers:
> @hfossli 👋 We've been thinking about this problem for awhile and because we haven't had a solution we've been happy with, we've left all PRs open to think about.
> 
> Our [Concurrency Beta](https://github.com/pointfreeco/swift-composable-architecture/discussions/1186) provides one new solution, though, that we'd love your opinion on. Namely, it begins to return a task handle from `ViewStore.send`, which can be awaited and cancelled in the `View.task` modifier.
> 
> ```swift
> .task { await viewStore.send(.onAppear) }
> ```
> 
> When the view goes away, it cancels this task and all effects spun up by the associated action will be cancelled, as well.

Glad to know there is a solution in the works. We are using TCA in a fairly large scale project at the moment (and loving it!) but have run into the same issues being documented here with long-living effects not being cancelled.

stephencelis:
@hfossli It's taken awhile but we've finally arrived at some cancellation helpers we're happy with, so we're going to close these related PRs out. I think we're a bit worried about `store.cancelAll()` being too blanketed, but are pretty happy with the following:

- Async cancellation via `await viewStore.send(.task).finish()`
- Cancellation tied to navigation APIs (this experiment is ongoing but we have some promising results so far)

Thanks again for taking the time to open these PRs and contribute to the conversation! If you have anything you'd like to discuss about our solutions, we'd love to hear about it. Please start a GitHub discussion!

PR #336, opened 2020-12-30
Add `Scene` conformance to `WithViewStore`

MaxDesiatov:
This would allow accessing `ViewStore` instances from a `body` definition of a type conforming to `Scene`. It could be useful for conditional rendering of scenes or sending actions from scene commands. Here's an example:

```swift
import ComposableArchitecture
import SwiftUI

@main
struct CommandsApp: App {
  private let store = Store(
    initialState: RootState(),
    reducer: rootReducer,
    environment: .live(rootEnvironment)
  )

  var body: some Scene {
      WindowGroup {
        WorkspaceView()
      }.commands {
        CommandGroup(after: CommandGroupPlacement.newItem) {
          Button("Open...") {
            viewStore.send(.open)
          }.keyboardShortcut("o", modifiers: [.command])
        }
      }
    }
  }
}
```


Sources/ComposableArchitecture/SwiftUI/SceneWithViewStore.swift
```diff
@@ -0,0 +1,86 @@
+#if compiler(>=5.3)
+
+import SwiftUI
+
+/// A structure that transforms a store into an observable view store in order to compute scenes from
+/// store state.
+@available(iOS 14.0, macOS 11.0, tvOS 14.0, watchOS 7.0, *)
+public struct SceneWithViewStore<State, Action, Content>: Scene where Content: Scene {
```
stephencelis, line 8: Interesting! Definitely seems like you should be able to use a view store with a scene. We've been holding onto a view store with an app delegate, but that's definitely not ideal.

I wonder, though: do you think it's possible to reuse `WithViewStore` in both places? We could probably unconstrain the base struct and use conditional conformances. Something like:

```swift
struct WithViewStore<State, Action, Content> {
  // ... properties
  var body: Content {

  }
}

extension WithViewStore: View where Content: View { ... }
extension WithViewStore: Scene where Content: Scene { ... }
```

Wanna try and see if that's possible?

And extensions with initializers that use `@ViewBuilder` and `@SceneBuilder` accordingly. And if that works I can definitely see other SwiftUI helpers like `IfLetStore` being useful to support both protocols as well!

MaxDesiatov: Yes, it does seem to work, I've pushed that in the latest commit.

stephencelis (approved):
Looks great, thanks!

PR #339, opened 2021-01-10
Replace deprecated optional with optional()

nspavlo:
Replace deprecated `optional` with `optional()`


stephencelis (approved):
Nice catch, thanks!

PR #342, opened 2021-01-11
Only include actions in TestStore debug output

stephencelis:
This broke a lil when we refactored the TestStore to use a real store under the hood.


mbrandonw (approved):
Nice, thanks

PR #343, opened 2021-01-12
Fix Tests

stephencelis:
This should get things building again. Command is only valid in macOS and iOS.


mbrandonw (approved):
nice, thanks!

PR #345, opened 2021-01-13
Adding CancellationBag

hfossli:
As per discussion on https://forums.swift.org/t/how-can-the-composable-architecture-really-become-composable/43045/2

This can be instead or in addition to #335 

Let me know if you approve of this direction and I'll spend time writing unit tests and documentation. ☀️

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -153,7 +153,7 @@ extension Effect {
     }
   }
   
-  public static func cancelAll(bag: CancellationBag = .global) -> Effect {
+  public static func cancelAll(bag: CancellationBag) -> Effect {
     return .fireAndForget {
       bag.cancelAll()
     }
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -1,5 +1,95 @@
+//
+//  Cancellation.swift
+//  Demo
+//
+//  Created by Håvard Fossli on 07/01/2021.
+//  Copyright © 2021 Darrarski. All rights reserved.
+//
+
 import Combine
 import Foundation
+import ComposableArchitecture
+
+fileprivate var bags: [AnyHashable: CancellationBag] = [:]
+fileprivate let bagsLock = NSRecursiveLock()
+fileprivate let cancellablesLock = NSRecursiveLock()
+
+public final class CancellationBag {
+  fileprivate var children: [CancellationBag] = []
+  fileprivate var cancellationCancellables: [AnyHashable: Set<AnyCancellable>] = [:]
+  fileprivate let id: AnyHashable
+  
+  public static var global: CancellationBag = {
+    struct GlobalID: Hashable {}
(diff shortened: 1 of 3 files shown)


hfossli:
This is of course an incremental non breaking change.

chrisballinger:
@hfossli Thank you for this! I think your solution is much more ergonomic and developer-friendly than the [`LifecycleAction`](https://github.com/pointfreeco/swift-composable-architecture/blob/3ab944e6aa69de3261a4d9035dc07b3d7e6111cc/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-Lifecycle.swift) approach.

In my case I was hitting [this issue](https://forums.swift.org/t/ifletstore-and-effect-cancellation-on-view-disappear/38272/11) (e.g. `Action was received by an optional reducer when its state was "nil"`) for what I think would be a relatively common scenario of:
* starting up a database observation, based on a database identifier in `State` during `onAppear`
* tearing down the observation during `onDisappear`

So far using your approach in this PR has fixed the crashes that I was seeing related to that workflow, and I like that the `cancelAll` approach ensures that any other dangling Cancellables don't get missed.

Sources/ComposableArchitecture/Effects/Cancellation.swift
```diff
@@ -26,53 +108,63 @@ extension Effect {
...
+  public static func cancelAll(bag: CancellationBag) -> Effect {
+    return .fireAndForget {
+      bag.cancelAll()
     }
   }
 }
 
-var cancellationCancellables: [AnyHashable: Set<AnyCancellable>] = [:]
-let cancellablesLock = NSRecursiveLock()
+fileprivate struct Lock {
```
gkaimakas, line 163: TCA already provides an extension on NSRecursiveLock for sync located in Internal/Locking.swift

hfossli: Yup, just wanted a simple standalone file that could be copied out and in to any project

andreyz:
@hfossli I think it's possible to extend Reducer with a method (that calls `optional` internally), that will simply cancel all local effects on nilling of the passed in reducer. To me this seems like a straightforward approach.

hfossli:
Interesting! I'm all ears. Do you have some example? I couldn't find any simple solutions.

Sources/ComposableArchitecture/Effects/Cancellation.swift
```diff
@@ -26,53 +108,63 @@ extension Effect {
...
+      
       return subject.handleEvents(
         receiveCompletion: { _ in cancellationCancellable.cancel() },
         receiveCancel: cancellationCancellable.cancel
       )
     }
     .eraseToEffect()
-
+    
     return cancelInFlight ? .concatenate(.cancel(id: id), effect) : effect
```
ddeneka, line 142: Shouldn't this be:
```swift
    return cancelInFlight ? .concatenate(.cancel(id: id, bag: bag), effect) : effect
```

stephencelis:
@hfossli I'm going to close this out in tandem with https://github.com/pointfreeco/swift-composable-architecture/pull/335#issuecomment-1212051805

Thanks again!

PR #348, opened 2021-01-15
Direct people to GH discussions instead of Swift forums

mbrandonw:
How do you feel about this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,7 +427,7 @@ The latest documentation for the Composable Architecture APIs is available [here
 
 ## Help
 
-If you want to discuss the Composable Architecture or have a question about how to use it to solve a particular problem, ask around on [its Swift forum](https://forums.swift.org/c/related-projects/swift-composable-architecture).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -427,7 +427,7 @@ The latest documentation for the Composable Architecture APIs is available [here
 
 ## Help
 
-If you want to discuss the Composable Architecture or have a question about how to use it to solve a particular problem, ask around on [its Swift forum](https://forums.swift.org/c/related-projects/swift-composable-architecture).
+If you want to discuss the Composable Architecture or have a question about how to use it to solve a particular problem, start a discussion [here](https://github.com/pointfreeco/swift-composable-architecture/discussions).
```
stephencelis, line 430: I'm down to switch, though we could also give both as choices:
```suggestion
If you want to discuss the Composable Architecture or have a question about how to use it to solve a particular problem, you can start a discussion [here](https://github.com/pointfreeco/swift-composable-architecture/discussions), or ask around on [its Swift forum](https://forums.swift.org/c/related-projects/swift-composable-architecture).
```

PR #349, opened 2021-01-16
Convenience initialised for stateless ViewStore

siejkowski:
Adding convenience initialiser for `ViewStore` when state is `Void`.

Rationale for change:
1. It simplifies the creation of `ViewStore` from `store.stateless` allowing for `ViewStore(store.stateless)`
2. It mirrors the available initialisers for `WithViewStore` SwiftUI helper view, so that you can both use `WithViewStore(store.stateless)` and `ViewStore(store.stateless)`

### DIFF
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -230,3 +230,9 @@ extension ViewStore where State: Equatable {
     self.init(store, removeDuplicates: ==)
   }
 }
+
+extension ViewStore where State == Void {
+  public convenience init(_ store: Store<Void, Action>) {
+    self.init(store, removeDuplicates: ==)
+  }
+}


mbrandonw:
Hey @siejkowski, thanks for the PR and sorry for the delayed response. 

I think this is a good addition. We have been wary of adding overloads to TCA APIs, such as `.binding` and `.scope`, because when used in complex SwiftUI views they can really blow up the type checker. However, the `ViewStore` initializer is usually used in much simpler situations, such as a view/view controller's `init`, and so I think this is safe. Also, soon Swift may get equatable tuples and so we could probably remove this someday in the future.

@stephencelis what do you think?

stephencelis:
@mbrandonw I think we can give this a shot. Seems safer than SwiftUI-related overloads.

mbrandonw:
Yeah let's do it! 👍

Issue #334, opened 2020-12-17
Case studies demos don't work on iOS 13.2


nsillik:
**Describe the bug**
The case studies demos don't work on iOS 13.2. This issue is similar, if not identical, to #238.

This is the case in both the latest release (`0.9.0`) and the main branch as of today (at commit `c1f88bd8608e8a26fd61ab46c9f11a63b0f4023d`)

**To Reproduce**

- Clone the repository.
- Open `ComposableArchitecture.xcworkspace`
- Select the "Case Studies (SwiftUI)" scheme
- Select a simulator running 13.2 (I specifically tested this on the "iPhone 11 (13.2.2)" simulator)
- Build & Run
- In the application tap the "Basics" row
- Tap the "+" button on the counter
- Note the number doesn't change from zero

**Expected behavior**
The counter number should increase to 1

**Environment**
 - Xcode Version 12.3 (12C33)
 - Swift 5.3
 - OS (if applicable): iOS 13.2, 13.1, and 13.0

RobertKoval:
iOS 13 itself has many bugs with SwiftUI and Combine until version 13.4

nsillik:
Sure, but if you look at the other issues this was fixed and regressed. This is also not related to SwiftUI at all, it reproduces in the UIKit case studies as well. It's likely a combine issue, since it was fixed before, I'm wondering if there's some `#if available` trickery that would make it work across all iOS 13 versions. Otherwise I'm probably going to have to drop using TCA and I _really_ don't want to, I'm loving working with it.

mluisbrown:
I have a suspicion that this issue might be related to this: https://forums.swift.org/t/question-about-valid-uses-of-observableobject-s-synthesized-objectwillchange/31141/2

In the ReactiveSwift fork of TCA, I added the following property in `ViewStore` to ensure that this wasn't a problem:

https://github.com/trading-point/reactiveswift-composable-architecture/blob/master/Sources/ComposableArchitecture/ViewStore.swift#L104

```swift
public lazy var objectWillChange = ObservableObjectPublisher()
```

[...]

Issue #238, opened 2020-07-29
WithViewStore or @ObservedObject var viewStore:.  doesn't update view [iOS 13.1]


romiroma:
**Describe the bug**
State updates, but view doesn't

**To Reproduce**
[Test.zip](https://github.com/pointfreeco/swift-composable-architecture/files/4993290/Test.zip)

**Expected behavior**
I expect view to be updated on state update inside WithViewStore

**Environment**
 - Xcode 11.6 (11E708)
 - Swift 5.2.4
 - iOS 13.1

** UPD:
- Doesn't work with TCA 0.6.0, but works fine with 0.5.0 version (If I set `Exact` version)

mbrandonw:
Hey @romiroma, thanks for the report! I was able to reproduce this locally, and it's definitely a problem with iOS 13.1, and 13.2, and then due to some Combine/SwiftUI fixes in 13.3 it works correctly.

The bug you are seeing is due to this PR https://github.com/pointfreeco/swift-composable-architecture/pull/199 where we changed some internals of `ViewStore` in order to close up an escape hatch that was created in some of the new iOS 14 APIs. However, Apple has since fixed that escape hatch, and so maybe we should revert this now, especially since it breaks older iOS.

Stephen and I will discuss today and come up with a plan. Thanks again!

---

As an aside, I came across some other weird behavior in just iOS 13.1. Our `TestScheduler` crashes because it seems the mere act of comparing two scheduler times results in a crash in Combine 😬:

```swift
DispatchQueue.SchedulerTimeType(.init(uptimeNanoseconds: 1)) <= DispatchQueue.SchedulerTimeType(.init(uptimeNanoseconds: 2)) // 🛑 crash
```

So even once we fix the issue you reported you may wanna upgrade to iOS 13.2 so that you can use the test scheduler.

romiroma:
Thank you, guys, for fast response! 💪👍

manmal:
@stephencelis For me, the problem persists with the current release (0.8.0) on iOS 13.2.2. This can be verified by using the test project created by @romiroma ([Test.zip](https://github.com/pointfreeco/swift-composable-architecture/files/4993290/Test.zip)), updating the TCA package to the current release, and running it on an iOS 13.2.2 simulator. 
<img src="https://user-images.githubusercontent.com/142797/93663835-daa5d580-fa6a-11ea-963d-60490f70bd81.gif" width="300">

malonehedges:
[...]

PR #353, opened 2021-01-18
Explicitly define `objectWillChange` on `ViewStore`.

nsillik:
Because the `ViewStore` no longer uses `@Published` variables, the conformance to `ObservableObject` no longer sythesizes the `objectWillChange` property on `ViewStore` in iOS 13.0.x, 13.1.x, and some 13.2 versions.

This caused `ViewStore` changes to not propagate, leading to views never updating on these OS versions. This fixes issues #334 and (the already closed) #238.


nsillik:
I'm also happy to revert the changes to the Case Studies project file, but it's nice being able to test and demo in all supported iOS versions.

Examples/CaseStudies/CaseStudies.xcodeproj/project.pbxproj
```diff
@@ -1030,6 +1030,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
+				IPHONEOS_DEPLOYMENT_TARGET = 13.0;
```
stephencelis, line 1033: Let's revert this in favor of running CI against iOS 13.2 (with Xcode 11.2.1).

```suggestion
```

Examples/CaseStudies/CaseStudies.xcodeproj/project.pbxproj
```diff
@@ -1083,6 +1084,7 @@
...
```
stephencelis, line 1087: This too.
```suggestion
```

Sources/ComposableArchitecture/ViewStore.swift
```diff
@@ -48,6 +48,10 @@ public final class ViewStore<State, Action>: ObservableObject {
 
   private var viewCancellable: AnyCancellable?
 
+  // N.B. `ViewStore` no longer contains any `@Published` variables, so `objectWillChange`
+  // won't be synthesized automatically. To work around issues on iOS 13 we explicitly declare it.
+  public lazy var objectWillChange = ObservableObjectPublisher()
```
stephencelis, line 53: Small suggestion:
```suggestion
  // N.B. `ViewStore` does not use a `@Published` property, so `objectWillChange`
  // won't be synthesized automatically. To work around issues on iOS 13 we explicitly declare it.
  public private(set) lazy var objectWillChange = ObservableObjectPublisher()
```

stephencelis (approved):
Thanks!

PR #359, opened 2021-01-21
Add TextState (for AlertState, ActionSheetState, etc.)

### LINKED ISSUE #293: [Breaking change,  0.8.0] AlertState expects LocalizedStringKey instead of String values

stephencelis:
Wanted to see if I could solve this quickly over a time boxed session this morning and seems to work well! Perhaps a bit scary to rely on reflection here (and we'll hopefully see if there are versions of SwiftUI in which the type layout has changed in CI).

Fix #293, #318.


Examples/TicTacToe/Sources/Views-UIKit/LoginViewController.swift
```diff
@@ -137,7 +137,7 @@ class LoginViewController: UIViewController {
         guard let alert = alert else { return }
 
         let alertController = UIAlertController(
-          title: alert.title.formatted(), message: nil, preferredStyle: .alert)
+          title: alert.title.debugOutput, message: nil, preferredStyle: .alert)
```
stephencelis, line 140: A couple reasons we may prefer #318:

1. Less reflection
2. UIKit interoperability (which we should maybe remove, example-wise). I've made `formatted()` internal in this PR since it seems weird to extend Apple's types with a method when it doesn't use any types we own.

Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -1,4 +1,78 @@
 import Combine
+import SwiftUI
+
+// NB: Deprecated after 0.10.0:
+
+extension ActionSheetState {
```
mbrandonw, line 6: I think this need to be made unavailable for macOS right?

ohitsdaniel:
Just wanted to add a couple of thoughts here:

I think, we should avoid spilling the View layer into the Model/State layer and therefore should avoid coupling TCA to SwiftUI by using any of its 'internal types'. I already mentioned in #318 that the switch to `LocalizedStringKey`s coupled people to using the SwiftUI way of localisation and therefore makes interoperability with UIKit a pain. 

`Text` has the benefit that it can be styled and but that again would rely on `Color`. This might open a can of worms and lead to more and more SwiftUI views spilling over into the state layer.

I think 'enforcing' the usage of SwiftUI types would be opinionated and we should give people the freedom to choose what works best for them and abstract on top of UIKit/SwiftUI. Just my two cents, both approaches work and do enable users to use both verbatim and localized strings. :)

mbrandonw:
Hey @ohitsdaniel, while it may seem weird to hold "view layer" types in state it _shouldn't_ be weird if they are simple value types. However, even though `Text` is a struct it turns out it is far from a value type 😕.

We just did some experimenting in a playground and found that `Text` and `LocalizedStringKey` are far stranger than we expected. Something as simple as this:

```swift
"Hello \("Blob")" as LocalizedStringKey == "Hello \("Blob")" as LocalizedStringKey
```

Will sometimes evaluate at `true` and sometimes as `false` 😧. Perhaps it's using some kind of hash value under the hood that changes with each process.

And even stranger, things like this _always_ return `false`:

```swift
Text("Hello \("Blob")") == Text("Hello \("Blob")")
```

It's a real bummer. `Text` and `LocalizedStringKey` are _almost_ proper value types, but the behavior above makes them a little too unstable. If Apple fixed these bugs (they may not even consider them bugs) then we would happily throw `Text` in our state and call it a day.

However, in the meantime we are sketching out an alternative to take matters into our own hands. There's just a small number of view modifiers that are overloaded for `Text -> Text` transformations, so we can easily list them out, and then instantly get equability on a complexly formatted text. Here's a rough gist of the work and soon we will whip this into shape for production:

https://gist.github.com/mbrandonw/c97aa866e0dfd9f883abca42d5a84a3c

ohitsdaniel:
Agree 100%. If `Text` and `LocalizedStringKey ` were proper value types, we wouldn't have to rebuild them. ;) 

Just another thought: Both Text and LocalizedStringKey are out of our control and might change their behavior at any point and cannot count on Apple informing us. Therefore, I would refrain from using them in states and only rely on simple data types + types that are under our control (which TextState would be).

stephencelis:
@ohitsdaniel That would be the plan! In an ideal world, if/when Apple fixes `LocalizedStringKey` (and `Text`)'s `Equatable` conformance, I think we'd be able drop `TextState` and rely on `Text` instead, but for now we think this would be an improvement that would fix the issue with verbatim strings _and_ allow us to test format strings.

stephencelis:
Just pushed a more fleshed out version of `TextState`.

Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-ReusableFavoriting.swift
```diff
@@ -78,7 +78,7 @@ extension Reducer {
             .cancellable(id: FavoriteCancelId(id: state.id), cancelInFlight: true)
 
         case let .response(.failure(error)):
-          state.alert = .init(title: .init(error.localizedDescription))
+          state.alert = .init(title: TextState(error.localizedDescription))
```
stephencelis, line 81: Even with `@_disfavoredOverload` on the deprecated initializer, `LocalizedStringKey` is favored here over `TextState`, so `.init` produces a deprecation warning. That warning will go away when we fully remove the deprecated interface.

mbrandonw (approved):
very nice. I think TextState could use some more tests, but I'm down to do a pass at those in a bit.

Sources/ComposableArchitecture/SwiftUI/TextState.swift
```diff
@@ -0,0 +1,304 @@
...
+}
+
+extension TextState: View {
+  public var body: some View {
+    Text(self)
+  }
+}
+
+extension String {
+  public init(state: TextState, locale: Locale? = nil) {
```
stephencelis, line 245: Added this just to keep our UIKit compatibility layer. It loses formatting. Dunno if it's possible to recover with attributed strings, but there are certain SwiftUI constructs that don't translate over (while `Color` can convert to `UIColor`, `Font` does not seem to convert to `UIFont`, etc.).

mbrandonw: yeah seems ok. eventually we should maybe create UIKit helpers like we have for `.alert(Store)`

stephencelis:
@ohitsdaniel We already rely on `LocalizedStringKey` so going to merge this as implemented for now. In the future we could abstract this type away with our own data type as well, but that's a bit more of a task 😄

ohitsdaniel:
Go for it! Happy to see this issue resolved. 🙌

Sources/ComposableArchitecture/SwiftUI/TextState.swift
andreyz: I did write above mentioned helpers for UIKit, unfortunately this PR has made this extension impossible, due to Button's `type` becoming internal... Was that strictly necessary?

``` swift
extension AlertState.Button {
  func toUIAlertAction(send: @escaping (Action) -> Void) -> UIAlertAction {
    let action = { if let action = self.action { send(action) } }
    switch self.type {
    case let .cancel(.some(title)):
      return UIAlertAction(title: title.formatted(), style: .cancel, handler: { _ in action() })
    case .cancel(.none):
      return UIAlertAction(title: nil, style: .cancel, handler: { _ in action() })
    case let .default(title):
      return UIAlertAction(title: title.formatted(), style: .default, handler: { _ in action() })
    case let .destructive(title):
      return UIAlertAction(
        title: title.formatted(), style: .destructive, handler: { _ in action() })
    }
  }
}
```

andreyz: And I know that `formatted()` is no more, just pasted the snippet as is, before updating/deleting it due to 0.11.0

stephencelis: > Was that strictly necessary?

Ah, definitely not. I had made the (poor) assumption that this API didn't need to be public, but now see how it can be depended upon. I'll revert this part of the change.

andreyz: Thanks!

Issue #357, opened 2021-01-20
TestStore incorrectly scopes `.receive` steps


lukeredpath:
**Describe the bug**
When using a scoped `TestStore`, `.send` steps are correctly scoped to `LocalAction` but `.receive` steps are scoped to `Action`.

This seems like an easy thing to fix but I wasn't entirely sure if this was a bug or the expected behaviour.

**To Reproduce**

```swift
        struct ExampleState: Equatable {
            var nested: NestedState = .init()
        }
        struct NestedState: Equatable {}

        enum ExampleAction: Equatable {
            case nested(NestedAction)
        }
        enum NestedAction: Equatable {
            case example
        }

        let store = TestStore(
            initialState: ExampleState(),
            reducer: Reducer<ExampleState, ExampleAction, Void>.empty,
            environment: ()
        )
        .scope(state: \.nested, action: ExampleAction.nested)

        store.assert(
            .send(.example),
            .receive(.example) // <<< Type 'ExampleAction' has no member 'example'
        )
```

Changing the `.receive` to `.receive(.nested(.example))` works.

**Expected behavior**
Send and Receive should both be scoped to the locally scoped action.

**Environment**
 - Xcode 12.3
 - Swift 5.3.2

stephencelis:
@lukeredpath This is by design, though it may be a little surprising and could definitely be better documented. Scoped test stores are generally intended to test view state and view actions (and I think view actions are only ever _sent_ by the user, never received by an effect). These user actions could also lead to side effects, though, where more general actions are received by the reducer, and in order to assert against these actions, the test store needs to be able to call `receive` with those actions and _not_ the same action type it sends.

Does that make sense? Any concerns with this approach?

lukeredpath:
@stephencelis thanks for the explanation. I did have a suspicion it might be deliberate, but I wasn't sure why. Your explanation makes complete sense, I do think it would be helpful if this was a bit better documented - perhaps even renaming it to something like `viewScope` might make it more obvious?

lukeredpath:
I think this can be closed, thanks for the updated docs.

PR #360, opened 2021-01-22
Document `scope` a bit more deeply

stephencelis:
Doing a lil evening writing and wanted to address #357. Got a little side-stepped realizing we don't really document scoping a bit more deeply. Will try to flesh things out a bit more before merging so leaving as a draft for now, but didn't wanna lose track of it.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -42,15 +42,119 @@ public final class Store<State, Action> {
   ///     struct AppAction { case login(LoginAction), ... }
   ///
   ///     // A store that runs the entire application.
-  ///     let store = Store(initialState: AppState(), reducer: appReducer, environment: ())
+  ///     let store = Store(
+  ///       initialState: AppState(),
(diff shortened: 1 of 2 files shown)


mbrandonw (approved):
nice

lukeredpath:
This is really helpful, thank you. While you're looking at this I'd love to know a bit more about the use cases for publisherScope - has this been covered anywhere?

stephencelis:
> This is really helpful, thank you. While you're looking at this I'd love to know a bit more about the use cases for publisherScope - has this been covered anywhere?

@lukeredpath The library uses this helper internally for the non-SwiftUI version of `store.ifLet`:

https://github.com/pointfreeco/swift-composable-architecture/blob/5ddbcdd21a638cc9119cd624663129787a8331d5/Sources/ComposableArchitecture/UIKit/IfLetUIKit.swift#L48-L53

It's not a common operation, though, and can introduce some tricky cycles, so I wonder if we remove it in the future to reduce API surface area.

PR #377, opened 2021-02-04
Add debug printing for key paths and binding actions

stephencelis:
Based on the stuff you found earlier!


stephencelis:
Gonna close this for now till we have time to revisit.

PR #387, opened 2021-02-16
Adds `StepType.sequence(Step)`

heyltsjay:
This is a sketch towards making the testing suite a touch more composable.
Common user-journey steps can be abstracted into a `Step.sequence` and can be vended by helper functions in the test.

example usage:
```swift
testStore.assert([
   login2fa(),
   enter2faCode(),
   dismiss(),
])
```

where squirreled away elsewhere:
```swift
func login2fa() -> Step {
  .sequence([
     .send(.emailTextChanged) { ... },
     .send(.passwordTextChanged) { ... },
     .send(.loginButtonTapped) { ... },
  ])
}
```

### DIFF
--- a/Sources/ComposableArchitecture/TestSupport/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestSupport/TestStore.swift
@@ -259,7 +259,7 @@
         store.scope(state: self.toLocalState, action: TestAction.send)
       )
 
-      for step in steps {
+      func assert(step: Step) {
         var expectedState = toLocalState(snapshotState)
 
         func expectedStateShouldMatch(actualState: LocalState) {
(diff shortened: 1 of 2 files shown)


stephencelis (commented):
@heyltsjay Thanks for submitting this PR! @mbrandonw and I think that this would be a good addition! Just a coupla requests:

- Can you revert the changes to TicTacToe? We'd like to think a bit more about the feature before changing sample code.
- Can you add an overload version of `Step.sequence()` that takes a variadic list of steps?

Once you make those changes I think we're happy to merge and start experimenting with this!

heyltsjay:
@stephencelis all set

- [x] revert the changes to TicTacToe
- [x] overload version that takes a variadic list of steps

stephencelis (approved):
Thanks!

PR #413, opened 2021-03-05
Tie long-living effect assertions to line of originating action

stephencelis:
Provides a nice ergonomic upgrade for debugging tests.

Before:
image

After:
image


mbrandonw (approved):
Very nice

PR #414, opened 2021-03-06
Addresses a couple of documentation issues for IdentifiedArray

Steven0351:
1. The AppState example is missing a generic. Based on the context of the documentation, I filled in what I thought would be appropriate
2. The ForEachStore example was missing a colon and comma.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
@@ -14,7 +14,7 @@ import Foundation
 ///
 /// This domain can be pulled back to a larger domain with the `forEach` method:
 ///
-///     struct AppState { var counters = IdentifiedArray<UUID, CounterState>() }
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
```diff
@@ -14,7 +14,7 @@ import Foundation
 ///
 /// This domain can be pulled back to a larger domain with the `forEach` method:
 ///
-///     struct AppState { var counters = IdentifiedArray<Int>(id: \.self) }
+///     struct AppState { var counters = IdentifiedArray<UUID, CounterState>() }
```
stephencelis, line 17: Good catch! I think we could even use the type alias:
```suggestion
///     struct AppState { var counters = IdentifiedArrayOf<CounterState>() }
```

stephencelis (approved):
Good catches, thanks!

PR #418, opened 2021-03-08
Add a ViewStore convenience initializer for stateless stores

RabugenTom:
Because `Void` can't be `Equatable`, we cannot use the other convenience initializer.
This new initializer makes `removeDuplicate` to always returns `true`, and allows to quickly instantiate stateless `ViewStore`s as `ViewStore(store.stateless)`.

### DIFF
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -230,3 +230,9 @@ extension ViewStore where State: Equatable {
     self.init(store, removeDuplicates: ==)
   }
 }
+
+extension ViewStore where State == Void {
+  public convenience init(_ store: Store<State, Action>) {
+    self.init(store, removeDuplicates: { _, _ in true })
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hey @RabugenTom, thanks for the PR! There's actually a PR with this open currently https://github.com/pointfreeco/swift-composable-architecture/pull/349.

We're still debating whether or not we want to bring in another overload into TCA. The more overloads the more strain on the Swift compiler :/

RabugenTom:
Ah, sorry, I didn't see #349.
Maybe something like
```swift
extension ViewStore where State == Void {
  public convenience init<S>(stateless store: Store<S, Action>) {
    self.init(store.stateless, removeDuplicates: { _, _ in true })
  }
}
```
wouldn't cause more strain on the compiler. But it adds complexity to the API thought. You can close this PR if you think it's not worth it otherwise, as #349 is keeping the discussion open.

PR #386, opened 2021-02-12
Improve performance of ForEachStore

stephencelis:
This improves upon the issue brought up in #381 and active discussion of #382.

PR #419, opened 2021-03-08
Display a correct error when the ForEachStore is empty

mackoj:
When we move to 0.15.0 from  0.14.0 our codebase broke because of https://github.com/pointfreeco/swift-composable-architecture/pull/386 the `!` at [line 143](https://github.com/pointfreeco/swift-composable-architecture/blob/65a0c6321fbb2884234f69df08f7ffa8c4cf0de8/Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift#L143).

This PR add a simple error message to prevent this issue.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
@@ -84,7 +84,7 @@ where Data: Collection, ID: Hashable, Content: View {
     self.data = data
     self.content = {
       WithViewStore(store.scope(state: { $0.map { $0[keyPath: id] } })) { viewStore in
-        if viewStore.state.isEmpty { fatalError("viewStore.state should not be empty") }
+        if viewStore.state.isEmpty { fatalError("State should not be empty") }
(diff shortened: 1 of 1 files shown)


stephencelis:
Hey @mackoj, thanks for the report. We think that this error is more due to a bug in SwiftUI collection caching, and in fact has been reported separately in #434. A `ForEachStore` should render just fine with an empty collection (just like `ForEach` renders fine), so I think we wanna figure out a different workaround here.

mackoj:
ok

Sajjon:
I get this too, using Example add "Todos", here are reproduction steps:

```sh
received action:
  AppAction.addTodoButtonTapped

received action:
  AppAction.todo(id:, action: .textFieldChanged)

received action:
  AppAction.todo(id:, action: .checkBoxToggled)

received action:
  AppAction.sortCompletedTodos

received action:
  AppAction.addTodoButtonTapped

received action:
  AppAction.todo(id:, action: .textFieldChanged)

Fatal error: Unexpectedly found nil while unwrapping an Optional value: file ComposableArchitecture/ForEachStore.swift, line 143
```

(I use TCA commit: 1ce105f8bf7457d94385056a1946c6dc6f338556)

PR #421, opened 2021-03-09
TextState: DebugOutputConvertible

stephencelis:
TextState currently dumps a lot of info when fed through TCA's debug output system. This shortens things up to make it more readable in tests.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/TextState.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/TextState.swift
@@ -321,11 +321,24 @@ extension TextState: CustomDebugOutputConvertible {
           output = "<baseline-offset=\(baselineOffset)>\(output)</baseline-offset>"
         case .bold, .fontWeight(.some(.bold)):
           output = "**\(output)**"
-        case let .font(.some):
+        case .font(.some):
           // TODO: Better describe fonts?
           output = "<font>\(output)</font>"
         case let .fontWeight(.some(weight)):
-          output = "<font-weight=\(weight)>\(output)</font-weight>"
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/SwiftUI/TextState.swift
```diff
@@ -302,3 +302,51 @@ extension LocalizedStringKey: CustomDebugOutputConvertible {
...
+      }
+      for modifier in textState.modifiers {
+        switch modifier {
+        case .bold,
+             // TODO: Distinguish
+             .fontWeight(.some(.semibold)),
+             .fontWeight(.some(.bold)),
+             .fontWeight(.some(.heavy)),
+             .fontWeight(.some(.black)):
+          output = "**\(output)**"
```
stephencelis, line 326: Currently doing some very basic Markdown-like formatting, but would probably be better to fully capture the text modifiers. Any ideas?

Sources/ComposableArchitecture/SwiftUI/TextState.swift
```diff
@@ -302,3 +302,51 @@ extension LocalizedStringKey: CustomDebugOutputConvertible {
...
+        case .italic:
+          output = "_\(output)_"
+        case .strikethrough:
+          output = "~~\(output)~~"
+
+        // TODO:
```
stephencelis, line 332: These too should prob be captured somehow.

Sources/ComposableArchitecture/SwiftUI/TextState.swift
mbrandonw: only thing I can think of is XML tags 😬

PR #427, opened 2021-03-12
Add more permissive fire-and-forget effect

stephencelis:
We've experimented with this and it is a bit nicer to iterate on features without having to do the extra dance of ignoring output and failure all the time.

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -333,4 +333,23 @@ extension Publisher {
       .catch { Just(.failure($0)) }
       .eraseToEffect()
   }
+
+  /// Turns any publisher into an `Effect<T, Never>` for any type `T` by ignoring all output and
+  /// failures.
+  ///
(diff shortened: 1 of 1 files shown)


stephencelis:
(I've kept the previous effect in since I don't think they conflict, though we should consider deprecating it in the future.)

Sources/ComposableArchitecture/Effect.swift
```diff
@@ -333,4 +333,23 @@ extension Publisher {
...
+  /// This is useful for times you want to fire off an effect but don't want to feed any data back
+  /// into the system.
+  ///
+  ///     case .buttonTapped:
+  ///       return analyticsClient.track("Button Tapped")
+  ///         .fireAndForget()
+  ///
+  /// - Returns: An effect that never produces output or errors.
+  public func fireAndForget<T>() -> Effect<T, Never> {
+    func absurd<A>(_ never: Never) -> A {}
```
jasdev, line 349: (Might be able to nix this func def., since it isn’t used. 😄)

mbrandonw (approved):

PR #436, opened 2021-03-16
fireAndCatch - For ignoring Output, while catching Failures

rhysm94:
`fireAndCatch` is a way to perform an `Effect` which may return some output, but the user only needs to handle the cases where the `Effect` fails.

An example of where this might be desirable is when subscribing to a `Publisher` where the `Output` type is `Void`. There's nothing to send back to the store, no data to persist in the `State`, but the publisher might fail. It's useful to be able to catch the error. A useful - but verbose - pattern to follow is to flatMap the `Void` publisher to an `Empty<Never, Failure>` to ensure the `Output` type is `Never`, and catch this to an `Effect`, before mapping into an `Action` which takes a `Result<Never, Failure>`. With `fireAndCatch`, the ignoring is all handled, and the `Failure` is routed to the `Output` of the publisher, allowing it to map to an `Action` which just takes a `Failure`.

As discussed here: https://github.com/pointfreeco/swift-composable-architecture/discussions/435

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -338,4 +338,24 @@ extension Publisher {
       .catch { _ in Empty() }
       .eraseToEffect()
   }
+
+  /// Turns any publisher into an `Effect` for any output and failure type by ignoring all output,
+  /// but retaining any failure.
+  ///
+  ///     case let .saveUser(user):
+  ///       return environment.saveUser(user)
+  ///         .fireAndCatch()
+  ///         .map(UserAction.saveFailure)
+  ///
+  /// This is useful for times you want to fire off an effect, and only want to feed data back into the
+  /// system if the effect fails.
+  ///
+  /// - Returns: An `Effect` which completes immediately upon receiving successful output, and only publishes errors
+  public func fireAndCatch() -> Effect<Failure, Never> {
+    flatMap { _ in
+      Empty<Failure, Failure>()
+    }
+    .catch(Just.init)
+    .eraseToEffect()
+  }
 }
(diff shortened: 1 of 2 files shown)


rhysm94:
Hey @stephencelis @mbrandonw - just wondering if this is something you'd like to add to TCA, or if you'd rather it go into a community TCA project?

stephencelis:
Hey @rhysm94! Sorry for not replying earlier. We've had a busy month of releases :)

This helper definitely looks handy! We're still figuring out how we want to let the library's design and surface area evolve, and also how to name new additions (without getting too bikesheddy 😂). Typically our guide has been well-worn experience in our projects. For things we're still experimenting with we sometimes introduce them in a case study that can be copied and pasted into other folks' projects, and with enough time/feedback we may introduce them back into the library proper.

So I think we may take a large project like @isowords and see where such a helper could be help us out. In the meantime, if this is code you want to share, feel free to ship a microlibrary and advertise it here!

rhysm94:
> Hey @rhysm94! Sorry for not replying earlier. We've had a busy month of releases :)
> 
> This helper definitely looks handy! We're still figuring out how we want to let the library's design and surface area evolve, and also how to name new additions (without getting too bikesheddy 😂). Typically our guide has been well-worn experience in our projects. For things we're still experimenting with we sometimes introduce them in a case study that can be copied and pasted into other folks' projects, and with enough time/feedback we may introduce them back into the library proper.
> 
> So I think we may take a large project like [isowords](https://github.com/pointfreeco/isowords) and see where such a helper could be help us out. In the meantime, if this is code you want to share, feel free to ship a microlibrary and advertise it here!

Sorry for the late reply! Congrats on the release of isowords - it’s a great game, and an invaluable resource for developing TCA-based apps!

I think that’s a very reasonable approach! If/when I (can) make something open source that uses `fireAndCatch`, I’ll re-make this PR again!

stephencelis:
Thanks! And feel free to open a discussion, in the future, too!

simme:
Fwiw I solved this by using a `.catchToEffect()` and a `.compactMap()` using CasePaths!
```swift
    return environment.databaseClient.saveProduct(product)
      .catchToEffect()
      .compactMap(/Result.failure)
      .map(BoostrapAction.boostrapFailed)
      .eraseToEffect()
```
Still more code than the proposed solution, and it's a little annoying you have to do `...ToEffect()` twice.

PR #440, opened 2021-03-18
feat(Reducer.pullback): introducing lens for pullbacks

maximkrouk:
Hi, there is no ability to share state between child modules, but it's a great feature and I implemented it using custom lens

Btw i need to create something like this
```swift
struct ParentState {
    var items: [ChildState]
    var collectionViewConfig: CollectionViewConfiguration
}

struct CollectionViewConfiguration {
    var isPaginationEnabled: Bool
    var ignoresSafeAreaInsets: Bool
}

struct CollectionViewState<Element> {
    var items: [Element]
    var config: CollectionViewConfiguration
}

parentReducer = .combine(
    childReducer.forEach(state: \.items)
    collectionViewReducer.pullback(state: Lens(
        extract: { state in 
            CollectionViewState(
                items: state.items
                config: state.collectionViewConfig
            )
        },
        embed: { collection, state
            state.items = collection.items
            state.collectionViewConfig = collection.config
        }
    ))
)
```

- I had to implement inout behavior using subscript to avoid escapes of inout globalState
- I'm not sure about the performance
- I implemented it before I saw reply for my message here https://forums.swift.org/t/how-can-the-composable-architecture-really-become-composable/43045/17

So maybe discuss if TCA needs these pullbacks, my job was just to make it 🌚

### DIFF
--- a/Sources/ComposableArchitecture/Lens.swift
+++ b/Sources/ComposableArchitecture/Lens.swift
@@ -1,0 +1,60 @@
+import CasePaths
+
+public struct Lens<Root, Value> {
+  public init(
+    extract: @escaping (Root) -> Value,
+    embed: @escaping (Value, inout Root) -> Void
+  ) {
+    self._extract = extract
+    self._embed = embed
+  }
+  
+  public init(_ keyPath: WritableKeyPath<Root, Value>) {
+    self.init(
+      extract: { $0[keyPath: keyPath] },
+      embed: { $1[keyPath: keyPath] = $0 }
+    )
+  }
+  
+  public static func readonly(_ keyPath: KeyPath<Root, Value>) -> Self {
+    self.init(
+      extract: { $0[keyPath: keyPath] },
+      embed: { _, _ in }
+    )
+  }
+  
+  private var _extract: (Root) -> Value
+  private var _embed: (Value, inout Root) -> Void
+  
+  public func extract(from root: Root) -> Value { _extract(root) }
+  public func embed(_ value: Value, in root: inout Root) { _embed(value, &root) }
+  
+  public subscript(
+    provider: ((inout Root) -> Void) -> Void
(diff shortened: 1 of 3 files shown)


stephencelis:
Hi @maximkrouk! Thanks for taking the time to open this. The current recommendation for these kinds of transformations is to leverage writable key paths using computed properties, which would contain the same kind of code you feed to provide the lens. In your example:

```swift
struct ParentState {
  var items: [ChildState]
  var collectionViewConfig: CollectionViewConfiguration

  var collectionViewState: CollectionViewState<ChildState> {
    get { .init(items: self.items, config: self.collectionViewConfig) }
    set { (self.items, self.collectionViewConfig) = (newValue.items, newValue.collectionViewConfig) }
  }
}

struct CollectionViewConfiguration {
    var isPaginationEnabled: Bool
    var ignoresSafeAreaInsets: Bool
}

struct CollectionViewState<Element> {
    var items: [Element]
    var config: CollectionViewConfiguration
}

parentReducer = .combine(
    childReducer.forEach(state: \.items, ...)
    collectionViewReducer.pullback(state: \.collectionViewState, ...)
)
```

We have an example of this in the case studies demo called "shared state," though we should probably better document this feature. Another benefit of using properties is the compiler can make optimizations there that it cannot for lens closures.

I think this should hopefully solve the issue for you today! But please let us know if we're missing anything.

maximkrouk:
I used the same approach somewhere, but I forgot 'bout it, that's why I opened the PR 🙂
Anyway maybe it would be useful to have a documentation with code samples, examples are great, but it's not always convenient to search there 🌚
Also I agree, that defining shared state accessors in state feels better, than Lens pullbacks

stephencelis:
Thanks for taking the time all the same! We'll continue to work on documentation as time goes on :smile:

PR #441, opened 2021-03-18
State-driven URL opening reducer

lukeredpath:
# State-driven URL opening

I’d like to present a reusable component for opening external URLs in your app, driven by state. I think it would make a good candidate for inclusion within TCA and would complement existing re-usable components like `Alert` and `ActionSheet`. This component has been designed for use with SwiftUI but a UIKit version could be implemented as a future enhancement.

## What’s the problem this solves?

There are a number of ways to open external URLs. UIKit provides `UIApplication.shared.open` and iOS 14/macOS provides `OpenURLAction` which can be accessed from the SwiftUI Environment using the `\.openURL` environment key. In addition to this, on macOS 11/iOS 14 etc. there is the SwiftUI `Link` component.

For very simple use cases where you just want to display a link to some external URL then `Link` (or a `Button` if you need to support older platforms) is absolutely fine and probably what you need in a lot of cases. You can’t easily write an automated test for it but that might be an acceptable trade-off if you’re literally just opening some fixed external URL.

Sometimes however, you want to be able to trigger the opening of a URL as a result of some store action - this could be an action you explicitly send from your view which triggers the URL opening (perhaps alongside some other behaviour), or it could be an action sent from a button in an alert or action sheet (using TCA’s `AlertState` and `ActionSheetState` components). It could also be an action sent as the result of an `Effect`. The URL could be dynamic, or computed based on some other part of your state. It could just be the case that you want to test the behaviour.

In these cases it would be really useful to trigger an external URL opening from your reducer itself.

## Alternatives considered - an effect-based approach

My first thought on how to implement this was to treat the opening of an external URL as an `Effect`. This is probably a reasonable approach as it does feel like a side-effect. You could probably implement it something like this:

```swift
struct FeatureEnvironment {
  var openURL: (URL) -> Effect<Never, Never>
}

enum FeatureAction {
  case tappedOpenURLButton
}

let featureReducer = Reducer<FeatureState, FeatureAction, FeatureEnvironment> { state, action environment in
  switch action {
  case .tappedOpenURLButton:
    return environment
      .openURL("http://example.com")
      .fireAndForget()  
  }
}
```

Whilst this fairly straightforward and not a lot of code, it does have some downsides:

* You need to pass around the `openURL` dependency to every feature that needs to be able to open a URL (you could potentially address this using the `SystemEnvironment` idea in the TCA examples folder but its still a fair amount of boilerplate).
* Testing fire-and-forget effects is not the most ergonomic and often requires some kind of mock dependency with some mutable local state that you assert on in a `.do` block, e.g.:

```swift
var openedURL: URL?

let store = TestStore(
  initialState: ...,
  reducer: ...,
  environment: FeatureEnvironment(
    openURL: { url in openedURL = url }
  )
)

store.assert(
  .send(.tappedOpenURLButton),
  .do { _ in
    XCTAssertEqual(.some("http://example.com"), openedURL)
  }
)
```

For these reasons, inspired by the existing `TextState`, `AlertState` and `ActionSheetState` components I took a more state-base approach.

## State-based URL opening

The way this component works is around a feature domain based on a single value of type `URL?` - the idea is that you have some URL property in your feature domain that you set to a value you need opening and it just opens. Conceptually, our feature is saying “this is the URL that should be opened” and the actual effect of opening it in whatever external application should handle it is handled entirely in the view layer, using a SwiftUI view modifier.

There are a number of advantages to this approach:

* Minimal boilerplate - just three lines of code to integrate the URL opening domain into your feature domain, and a single SwiftUI view modifier to attach the URL opening behaviour to your view.
* Opening a URL is a one-line state mutation and you don’t even need to take care of setting it back to `nil` again once the URL has been opened as the component handles it for you.
* Easy to test - its just a state mutation so you can test this like any other state mutation using `TestStore`.

So with all this said, what does it actually look like? Lets adapt our previous example to use the new component.

Firstly we need to embed the URL opening domain in our feature domain:

```swift
struct FeatureState {
  var urlToOpen: URL? // 1. An optional URL property
}

enum FeatureAction {
  case tappedOpenURLButton
  case openURL(OpenURLViewAction) // 2. Embed the components domain actions
}

let featureReducer = Reducer<FeatureState, FeatureAction, Void> { state, action, _ in
  switch action {
  case .tappedOpenURLButton:
    state.urlToOpen = URL(string: "http://example.com") // 3. Set the URL when you want to open it
  }
}
.opensURL( // 4. Attach the component's high-level reducer
  \FeatureState.urlToOpen,
  action: /FeatureAction.openURL
)
```

Next, we need to attach the view modifier to our view:

```swift
struct FeatureView: View {
  let store: Store<FeatureState, FeatureAction>
  
  var body: some View {
      Button("Open URL") {
        viewStore.send(.tappedOpenURLButton)
      }
    }
    .opensURL(
      store.scope(
        state: \.urlToOpen,
        action: FeatureAction.openURL
      )
    )
  }
}
```

And that’s it!

We can test this behaviour, including simulating the URL actually being opened, without the need for any mocks:

```swift
class FeatureTests: XCTestCase {
  func testOpeningURL() {
    let store = TestStore(
      initialState: FeatureState(),
      reducer: featureReducer,
      environment: ()
    )
    
    store.assert(
      .send(.tappedOpenURLButton) {
        $0.urlToOpen = "http://example.com"
      },
      .send(.openURL(.openedURL)) {
        $0.urlToOpen = nil
      }
    )
  }
}
```

No mocks, no dependencies, no mutable local state and no raw assertions in `.do` blocks.

## Caveats

This is designed for use with SwiftUI but it does have a dependency on UIKit and specifically `UIApplication` - I'm not sure there is a SwiftUI only API for checking if URLs can be opened.

## Copyright and License

This was developed out of work on our app here at Community.com and the code in this PR is Copyright 2021 Community.com, Inc. and licensed for use under the same MIT terms as the rest of this repo.

@stephencelis @mbrandonw do you have any preferred way of showing copyright notices for contributed code? I could add an MIT License header to each file but that seems a bit overkill - could something be added to the LICENSE file in the repo perhaps?

### DIFF
--- a/Examples/CaseStudies/CaseStudies.xcodeproj/project.pbxproj
+++ b/Examples/CaseStudies/CaseStudies.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		4F5AC11F24ECC7E4009DC50B /* 01-GettingStarted-SharedStateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4F5AC11E24ECC7E4009DC50B /* 01-GettingStarted-SharedStateTests.swift */; };
+		A370BB3D25F94F7B00EBF637 /* 01-GettingStarted-OpeningURLs.swift in Sources */ = {isa = PBXBuildFile; fileRef = A370BB3C25F94F7B00EBF637 /* 01-GettingStarted-OpeningURLs.swift */; };
 		CA0C0C4724B89BEC00CBDD8A /* 04-HigherOrderReducers-LifecycleTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = CA0C0C4624B89BEC00CBDD8A /* 04-HigherOrderReducers-LifecycleTests.swift */; };
 		CA0C51FB245389CC00A04EAB /* 04-HigherOrderReducers-ReusableOfflineDownloadsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = CA0C51FA245389CC00A04EAB /* 04-HigherOrderReducers-ReusableOfflineDownloadsTests.swift */; };
 		CA25E5D224463AD700DA666A /* 01-GettingStarted-Bindings-Basics.swift in Sources */ = {isa = PBXBuildFile; fileRef = CA25E5D124463AD700DA666A /* 01-GettingStarted-Bindings-Basics.swift */; };
@@ -148,6 +149,7 @@
 
 /* Begin PBXFileReference section */
 		4F5AC11E24ECC7E4009DC50B /* 01-GettingStarted-SharedStateTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "01-GettingStarted-SharedStateTests.swift"; sourceTree = "<group>"; };
+		A370BB3C25F94F7B00EBF637 /* 01-GettingStarted-OpeningURLs.swift */ = {isa = PBXFileReference; indentWidth = 2; lastKnownFileType = sourcecode.swift; path = "01-GettingStarted-OpeningURLs.swift"; sourceTree = "<group>"; tabWidth = 2; };
 		CA0C0C4624B89BEC00CBDD8A /* 04-HigherOrderReducers-LifecycleTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "04-HigherOrderReducers-LifecycleTests.swift"; sourceTree = "<group>"; };
 		CA0C51FA245389CC00A04EAB /* 04-HigherOrderReducers-ReusableOfflineDownloadsTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "04-HigherOrderReducers-ReusableOfflineDownloadsTests.swift"; sourceTree = "<group>"; };
 		CA25E5D124463AD700DA666A /* 01-GettingStarted-Bindings-Basics.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "01-GettingStarted-Bindings-Basics.swift"; sourceTree = "<group>"; };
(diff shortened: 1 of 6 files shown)


lukeredpath:
I'm afraid I'm struggling with CI a bit here - its a bit odd that its trying to compile with an old version of Xcode - I don't even have a copy of Xcode 11 on this machine to try and fix this. Shouldn't it only need to compile on the latest version of Xcode?

stephencelis:
Hey @lukeredpath!

Thanks for opening this PR and sharing. It's very interesting stuff! In general we agree with the approach of leveraging state over effects for this kind of thing, and you've come up with a really nice solution to the problem.

I understand wanting to merge this directly into the core library. Here's our initial thinking, though:

First, we really don't want to modify the core license. We welcome contributions from anyone and via any company, but we would like to keep things simple in this repository and not worry about the legal logistics this kind of thing introduces.

Second, we've tried to keep TCA very small and try to only ship helpers that map directly to the corresponding SwiftUI interface. Your addition is really cool but is perhaps just far enough outside of the realm of SwiftUI's existing interface that it feels more like its own library. While the helpers we offer for sheet/alert/binding/etc. map directly to the SwiftUI interface, your `opensURL` is something that works differently than SwiftUI's `onOpenURL` and `@Environment(\.openURL)`. It's cool in its own right, don't get us wrong! But we usually try to demonstrate these kinds of additions with dedicated case studies and demo applications.

So with that said we think there are a couple paths forward:

- We're definitely down to take this code as an isolated case study that folks will have visibility into and be able to copy and paste into their code bases (like `SystemEnvironment`), but we're reluctant to do so if they need to retain your company's license.
- We'd also like to get a TCA community organization going in the near future (we've had a lot on our plate, but it'll happen!), and we'd be happy to have the library hosted there when the org is ready.

How does all that sound?

lukeredpath:
Hey Stephen, thanks for the reply. It's totally cool if you think this is a bit outside of the core scope to include in the main library. I'm quite happy to publish this as a separate helper library.

Perhaps I wasn't clear enough about the licensing, but the license this for this code is MIT, the same as the rest of the repo. It's not any particular variation or custom license. It's just that I need to make it clear that the copyright is still owned by Community - my understanding is that contributions to open source repos do not transfer copyright in any way, so the copyright statement would just make it clear.

If I were to release publish the library code as a dedicated library in a repo on our own organisation, would you still be willing to accept the case study in this PR, with our own package just added as a dependency for the case studies app?

stephencelis:
> If I were to release publish the library code as a dedicated library in a repo on our own organisation, would you still be willing to accept the case study in this PR, with our own package just added as a dependency for the case studies app?

I think we probably want to avoid third-party dependencies since if they break we have no way of fixing them. We're down to have a fully isolated one, but maybe it makes more sense to have the example demos in your library, similar to how we ship demos in [ComposableCoreLocation](https://github.com/pointfreeco/composable-core-location) [ComposableCoreMotion](https://github.com/pointfreeco/composable-core-motion).

lukeredpath:
👍🏻 in that case, I will close this and re-post the link in Discussions once I've got it all set up.

stephencelis:
Awesome! Look forward to seeing it 😄 And thanks for being understanding!

lukeredpath:
Just dropping a link here in case anyone comes across this PR. https://github.com/Shimmur/composable-open-url

PR #446, opened 2021-03-19
Remove SPM instructions

stephencelis:
I think this is now obsolete. In fact we should probably migrate TicTacToe to use a Swift package, as we've done in isowords.


stephencelis:
- [ ] Let's instead leave/deprecate this notice when Xcode 12.5 is released

PR #448, opened 2021-03-19
Stop using `Form` in Tic-Tac-Toe example

### LINKED ISSUE #395: 2FA SwiftUI crashes in TicTacToe example
**Describe the bug**
In the TicTacToe example application, when you choose the SwiftUI application flow and opt into the Two Factor Auth. section, when you tap "Next" after typing in `1234`, the app with crash with the error:

```
Fatal error: "LoginAction.twoFactorDismissed" was received by an optional reducer when its state was "nil". This can happen for a few reasons:
```

with a long list of reasons as to why it's potentially crashed.

This crash appears to be caused by `LoginSwiftView.swift` line num. 71. where the app sends `.twoFactorDismissed` to the now nil'd out `LoginState`, as the `NavigationLink` receives `false` for the `isActive` binding.

**To Reproduce**
 - Open the TicTacToe sample application
 - Select SwiftUI version
 - Enter `2fa@test.com` in the Email field
 - Enter `password` in the Password field
 - Tap Log in
 - Enter `1234` and tap Submit
 - App navigates to the New Game screen, then immediately crashes

**Expected behavior**
I expected the app to progress to the New Game screen and not crash 😅

**Screenshots**
Screenshot 2021-02-24 at 09 25 43

**Environment**
 - Xcode 12.4
 - Swift 5.3
 - Composable Architecture version 0.15.0

**Additional context**
If you login with the 2FA flow through the UIKit version, it doesn't crash, and retains the login if you navigate backwards and switch to the SwiftUI version.

### LINKED ISSUE #406: Password in TicTacToe seems to be "password" rather than "PASSWORD"
**Describe the bug**
Testing the TicTacToe example, it seems to me that the password is "password" rather than "PASSWORD"?

**To Reproduce**
Try to Login to the TicTacToe example.  Note that the instructions suggest the password is "PASSWORD" but this doesn't seem to work.  "password" does work.  I suppose this is the result of using all caps in the instructions?

**Screenshots**
Screen Shot 2021-03-02 at 3 00 13 PM

stephencelis:
1. It has a bug where it writes to a binding at the wrong time (fixes #395)
2. Its UI changed to uppercase section titles in iOS 14, causing confusion (fixes #406)

### DIFF
--- a/Examples/TicTacToe/Sources/Views-SwiftUI/LoginSwiftView.swift
+++ b/Examples/TicTacToe/Sources/Views-SwiftUI/LoginSwiftView.swift
@@ -33,19 +33,18 @@ public struct LoginView: View {
 
   public var body: some View {
     WithViewStore(self.store.scope(state: { $0.view }, action: LoginAction.view)) { viewStore in
-      VStack {
-        Form {
-          Section(
-            header: Text(
-              """
+      ScrollView {
+        VStack(spacing: 16) {
+          Text(
+            """
               To login use any email and "password" for the password. If your email contains the \
               characters "2fa" you will be taken to a two-factor flow, and on that screen you can \
               use "1234" for the code.
               """
-            )
-          ) { EmptyView() }
+          )
 
-          Section(header: Text("Email")) {
+          VStack(alignment: .leading) {
+            Text("Email")
             TextField(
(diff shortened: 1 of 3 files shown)


mbrandonw (approved):
nice

PR #451, opened 2021-03-20
Shared state management with StateBinding

RabugenTom:
Hello!,

Including a substate with shared values in a parent state is currently inconvenient and error-prone:
Each construction is bespoke, and one has to make sure that getters and setters are coherent. Changing the order of the child properties changes the generated initializer, forcing to update the parent's call site. Adding a property to the child state forces to inform all the parents of its existence or to resort to custom strategies.
Furthermore, avoiding deduplication in the setter has to be implemented explicitly each time it's required.

This PR attempts to improve the situation on many points. One has only to describe once the connection between the parent state and its child state, and one can specify if and how values should be deduplicated at this point. 
The description is mainly realized by providing couples of `KeyPaths`.

Two types are added to this end: `PropertyBinding`, which describe a binding between a property `V` of the parent and the child, and `StateBinding` which describe how a list of `PropertyBinding` is applied, such as the parent's properties and the child instance are in sync. From the `StateBinding`, one can generate *bona fide* accessors to a child state instance.

These types were added in the form of a new SwiftUI case-study called `01-GettingStarted-SharedState-StateBindings` (albeit it would work the same in UIKit). The case study present many use-cases: substates with internal properties, optional substates, fine-grained deduplication, etc. It also shows that the syntax is very similar in each configuration.

This is purely additional and only concerns the `State` part of TCA. I've already presented the idea in #409, but this is a much more refined version with only two types and a simplified syntax <del>thanks to a `@resultBuilder`</del>(Edit: [no more builder](https://github.com/pointfreeco/swift-composable-architecture/pull/451#issuecomment-803663539)).

Please tell me what you think.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-SharedState-StateBindings.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-SharedState-StateBindings.swift
@@ -14,6 +14,7 @@ struct SharedStateWithBinding: Equatable {
   struct FeatureState: Equatable {
     var name: String = "" // Used for this case study presentation
     var isCountInternal: Bool = true // Used for this case study presentation
+    var isSelected: Bool? = nil
 
     var text: String = ""
     var count: Int = 0
@@ -115,7 +116,7 @@ struct SharedStateWithBinding: Equatable {
       get: { src, dest in dest.text = src.content },
       set: { src, dest in src.content = dest.text }
     )
-
+  
   var feature5: FeatureState? {
     get { Self._feature5.get(self) }
     set { Self._feature5.set(&self, newValue) }
@@ -134,10 +135,40 @@ struct SharedStateWithBinding: Equatable {
     get { Self._feature6.get(self) }
     set { Self._feature6.set(&self, newValue) }
   }
+  
+  // MARK: "Collection of features" -
+  // In this example, we have an `IdentifiedArray` of `FeatureState` and we use `map` to transform
+  // `FeatureState` bindings into `IdentifiedArrays<_,FeatureState>` bindings. This allows to
+  // update each feature of the collection from `self`'s values on the fly. Because the way
+  // back is undecided (how to update `self` from the collection newValue), we request to
+  // update `self` using the element with `selectedID` as a base. This is purely optional and we
+  // can return `nil` if we don't want to update `self`, rendering the binding read-only.
+  let selectedID = "Id:1"
+  fileprivate var _features7: IdentifiedArray<String, FeatureState> = IdentifiedArray<String, FeatureState>( [
+      .init(name: "Id:1", isSelected: true),
+      .init(name: "Id:2", isSelected: false),
+      .init(name: "Id:3", isSelected: false)
+    ], id: \.name)
+  
+  // Since we work on an IdentifiedArray, we need to map the binding so we can directly
+  // use `FeatureState` keyPaths. We update only `Feature.text` from `self.content`. The
+  // `count` in `FeatureState` states are thus internal, stored by the `_feature7` array.
+  fileprivate static let _feature7 = StateBinding(\Self._features7)
+    .map(
+      .rw(\.content, \.text),
+      // The feature with `SelectedID` will be used to update values in `SharedStateWithBinding`.
+      // We return a `FeatureState` instance that will go upflow when setting the value.
+      // If we return `nil`, the binding would be read-only for its properties.
+      getBack: { $1[id: $0.selectedID] }
+    )
+  
+  var feature7: IdentifiedArray<String, FeatureState> {
+    get { Self._feature7.get(self) }
+    set { Self._feature7.set(&self, newValue) }
+  }
 }
 
 // MARK: Internal
-
 private extension SharedStateWithBinding {
   // Used for this case study presentation
   var feature1Name: String { "Manual binding" }
@@ -161,6 +192,7 @@ enum SharedStateWithBindingAction {
   case feature4(FeatureAction)
   case feature5(FeatureAction)
   case feature6(FeatureAction)
+  case feature7(String, FeatureAction)
 
   case toggleFeature3
   case toggleFeature5
@@ -198,6 +230,10 @@ let sharedStateWithBindingReducer =
       boundFeatureReducer.pullback(state: \.feature6,
                                    action: /SharedStateWithBindingAction.feature6,
                                    environment: { _ in () }),
+      
+      boundFeatureReducer.forEach(state: \.feature7,
+                                  action: /SharedStateWithBindingAction.feature7,
+                                  environment: { _ in () }),
 
       Reducer<SharedStateWithBinding, SharedStateWithBindingAction, Void> {
         state, action, _ in
@@ -272,6 +308,13 @@ struct SharedStateWithBindingView: View {
                                               action: SharedStateWithBindingAction.feature6)
           )
         }
+
+        Section(header: Text("Features array, selected: \(viewStore.selectedID)"),
+                footer: Text("These three feature are stored as an array in State. Only the first one is fully responsive, with its values updating the whole app. The two other ones have an internal \"count\" value, but their textField is read-only from the state.")) {
+          ForEachStore(self.store.scope(state: { $0.feature7 },
+                                       action: SharedStateWithBindingAction.feature7),
+                       content: FeatureView.init(store:))
+        }
       }
     }
     .listStyle(GroupedListStyle())
@@ -289,7 +332,7 @@ struct SharedStateWithBindingView: View {
             TextField(viewStore.name,
                       text: viewStore.binding(keyPath: \.text, send: FeatureAction.binding))
               .textFieldStyle(RoundedBorderTextFieldStyle())
-
+ 
             Stepper("\(viewStore.count)",
                     value: viewStore.binding(keyPath: \.count, send: FeatureAction.binding),
                     in: 0 ... 9)
@@ -301,7 +344,21 @@ struct SharedStateWithBindingView: View {
             .font(.footnote)
             .foregroundColor(viewStore.isCountInternal ? .red : .green)
             .frame(maxWidth: .infinity, alignment: .trailing)
+          if viewStore.isSelected == false {
+            Text("The text field is updated by the global state but it can't update it back")
+              .font(.footnote)
+              .foregroundColor(.orange)
+              .lineLimit(nil)
+              .frame(maxWidth: .infinity, alignment: .leading)
+          }
         }
+        .listRowBackground(
+          (viewStore.isSelected == false)
+            ? Color.yellow.opacity(0.1)
+            : (viewStore.isSelected == true )
+            ? Color.green.opacity(0.1)
+            : Color.clear
+        )
         .id(viewStore.name)
         .listRowInsets(EdgeInsets(top: 11, leading: 8, bottom: 8, trailing: 8))
       }
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-SharedState-StateBindings.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-SharedState-StateBindings.swift
@@ -86,7 +86,7 @@ struct SharedStateWithBinding: Equatable {
   // All the properties of `Feature` are bound to some property of `self`. A new instance is created
   // from the provided `with:` argument. We use `Self.self` because we need to tie the first generic
   // parameter of `StateBinding` to `Self`.
-  fileprivate static let _feature4 = StateBinding(Self.self, with: FeatureState.init)
+  fileprivate static let _feature4 = StateBinding<Self, FeatureState> { .init() }
     .ro(\.feature4Name, \.name)
     .ro(\.isCountInternal, \.isCountInternal)
     .rw(\.content, \.text)
@@ -102,7 +102,11 @@ struct SharedStateWithBinding: Equatable {
   // instance is created from a function of `Self`. We use the `shouldShowFeature5` flag to decide if the
   // property is nil or not. Because it has no private storage, the `feature5` instance is completly set
   // when accessed and the flag is enough to condition its existence and content without ambiguity.
-  fileprivate static let _feature5 = StateBinding<SharedStateWithBinding, FeatureState?>(with: { $0.shouldShowFeature5 ? .init() : nil })
+  fileprivate static let _feature5 = StateBinding<SharedStateWithBinding, FeatureState?> {
+      $0.shouldShowFeature5
+      ? .init()
+      : nil
+    }
     .ro(\.feature5Name, \.name)
     .ro(\.isCountInternal, \.isCountInternal)
     .rw(\.count, \.count)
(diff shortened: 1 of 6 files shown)


RabugenTom:
It looks like from CI that older compilers have difficulties resolving types.

RabugenTom:
The whole PR now works on all supported Swift versions of TCA.
The `@PropertyBindingsBuilder` requires Swift 5.4, as using `@_functionBuilder` instead of `@resultBuilder` triggers a seemingly unavoidable warning once Swift 5.4 is used.

The usage of the builder is optional, and it can handle the "classical" Array syntax, so its adoption is automatic without intervention when the user switches to Swift 5.4+

The case study has been updated to show both usages (as the `@PropertyBindingsBuilder` allows to lighten the syntax even more), but this clutters a little the example with `#if swift(>=5.4)` statements. There is of course a lot of room for improvement.

Because I wanted to maintain syntax compatibility between Swift 5.1 and 5.4, and because functions using the result builder can't have the same arguments of functions not using it, I had to duplicate a bunch of code. This can be partially solved by using internal common functions which are called by both public versions. I'll implement this if this PR goes farther.

Please tell me what you think.

RabugenTom:
I've completely refactored the PR. The `@resultBuilder` is gone, and the same code is used for all supported compiler versions. The new API is as short as with the `@resultBuilder`, but here is far less type inference, and we also get better completion at call site. 

Here is an example with an optional sub-state with internal properties and deduplication:
```swift
struct Feature {
  var id: UUID
  var text: String
  var count: Int
  var observedText: String
  // The parent shouldn't have to manage this:
  internal var draggedPosition: CGPoint?
}

struct State {
  var id: UUID
  var title: String
  var count: Int
  var textObservedByEveryone: String
  
  // StateBinding Approach
  private var _feature: Feature?
  static var _feature = StateBinding(\Self._feature)
    .ro(\.id, \.id)
    .rw(\.title, \.text)
    .rw(\.count, \.count)
    .rw(\.textObservedByEveryone, \.observedText, removeDuplicates: ==)
  
  var feature: Feature? { // Always the same shape
    get { Self._feature.get(self) }
    set { Self._feature.set(&self, newValue) }
  }
  
  // Current Approach
  private var _otherFeature: Feature?
  var otherFeature: Feature? {
    get {
      var feature = _otherFeature
      feature?.id = id
      feature?.text = title
      feature?.count = count
      feature?.observedText = textObservedByEveryone
      return feature
    }
    set {
      _feature = newValue
      guard let feature = newValue else { return }
      title = feature.text
      count = feature.count
      if textObservedByEveryone != feature.observedText {
        textObservedByEveryone = feature.observedText
      }
    }
  }
}
```
Please let me know what you think.

RabugenTom:
Optionality is now automatically managed from the returned `Child` type, and there are only two branches: 

1) `Child` state with some private storage on `Parent`'s side
Initialize the `StateBinding` from the `WritableKeyPath<Parent,Child>` which correspond to the storage. 
```swift
let stateBiding = StateBinding(\Parent._child)
```

2) `Child` state completely supported by the `Parent`'s properties
Initialize the `StateBinding` by providing a `Child` instance as 
    - a function `() -> Child` (like `Child.init` if compatible) **or** 
    - a function `(Parent) -> Child` (where one can condition the nullity of the `Child` instance from the `Parent`'s state)
```swift
let stateBiding = StateBinding<Parent, Child>(with: Child.init)
```

Once the base `StateBinding` is built, one can compose the bindings by chaining with:
  - `.rw(\Parent.value, \Child.value)` for readwrite binding between `Parent` and `Child`;
  - `.rw(\Parent.value, \Child.value, removeDuplicates: ==)` for readwrite binding between `Parent` and `Child` with deduplication on the setter side;
  -  `.ro(\Parent.value, \Child.value)` for readonly binding from `Parent` to `Child`;
  -  `with(get: (parent, child in ...), set: (parent, child in ...))` where one has complete control over the binding between the entities.
```swift
let stateBiding = StateBinding<Parent, Child> { .init() }
  .rw(\.text, \.title)
  .ro(\.count, \.count)
  .rw(\.sharedWithOtherFeatures, \.sharedText, removeDuplicates: ==)
  .with( get: { parent, child in child.isOdd = parent.count.isMultiple(of:2) })
  ...
```

Accessors for the exposed `Child` state can then be implemented as 
```swift
var child: Child {
  get { stateBinding.get(self) }
  set { stateBinding.set(&self, newValue) }
}
```

This `child` will always be in sync with the `Parent` instance and its siblings. The cognitive burden happens only when defining the `StateBinding` and the process is standardized.

The `PropertyBinding` type is not exposed anymore. It may be removed at some point, but I think it provides a convenient "unit of work" abstraction.

Please let me know what you think.

AlexisQapa:
I didn't have the chance to try it yet but this seems really nice ! I wonder if there is a way to unify the binding and the accessor in one single decl with a property wrapper of something.

RabugenTom:
Thanks! I didn't manage to make it work with property wrappers unfortunately: in order to pull-out values from `Parent`, you need to access it from the wrapper instance, which is not possible with current versions of Swift, especially when `Parent` is a value type ([See more on the topic in Swift Forums](https://forums.swift.org/t/property-wrappers-access-to-both-enclosing-self-and-wrapper-instance/32526/23)).

One more compact alternative would be to host the internal storage inside the binding (when applies). In this case, the binding should live in the `Parent`'s instance (it is static right now to conveniently refer to `Self`, but it can be defined anywhere). We would need to forward `Equatable` & co conformance, and I'm a little wary about corner cases. I feel it should work though.

In this case, one could avoid to refer to `Self` in the accessors, and one would also avoid to name the internal storage (when  applies), like I did in my previous comment where I presented the DSL on a property without internal storage.

RabugenTom:
It turns out hosting the storage instance inside the `StateBinding` isn't straightforward without using a syntax somewhat as long as the current one.
Furthermore, it would cause issue with `Equatable`, as we can't compare free form `PropertyBinding`'s (I think those defined from `KeyPath`'s could be compared, but it may be a landmine).

RabugenTom:
*The last CI job seems to have failed on TicTacToe, which I have not touched (all the code from the PR is in CaseStudiesSwiftUI). I'm not expert in CI though, but it looks like Xcode 12.4 has crashed when trying to build TicTacToes.*

Anyway. I've added a `map` function to `StateBinding` and `PropertyBinding` (which is now exposed again). It allows to work on generic types such as `Array`, `IdentifiableArray`, `Dictionary`, and `Optional` of `ChildState`, using directly `KeyPath< ChildState,*>`. This renders the whole PR even more convenient as it now provide way to update an array of features with some parent's state property when accessed for example. An optional `getBack` function can be specified in case one wants to pullback some content from the `Array<Feature>`'s newValue in the setter. This function is by default noop for all types but the `Optional`, for which it returns the optional value itself by default.

For example
```swift
struct Book {
  struct Chapter {
    var bookTitle: String = ""
    var content: String = ""
  }
  var title: String = "Alice in Wonderland"

  var _chapters: [Chapter] = [.init(), .init(), .init()]
  static var chapters = StateBinding(\Self._chapters) // (*)
    .map(
      .ro(\.title, \.bookTitle)  // (**)
    )

  var chapters: [Chapter] {
    get { Self.chapters.get(self) }
    set { Self.chapters.set(&self, newValue) }
  }

  // (*) StateBinding<Book, [Chapter]> with internal properties storage at `Book._chapters`
  // (**) PropertyBinding<Book, Chapter> (read-only), so each chapter is updated with `Book.title`.
}
```

When accessing `\Book.chapters`, each one will have its `bookTitle` up to date with `\Book.title`. Some specific overrides were implemented for `Optional`, so the user don't have to call `map` and can work directly with unwrapped `KeyPath`'s. If there were only one `Chapter` instead of an `Array`, the syntax would be exactly the same without the `map`.

The PR now provides a systematic way to manage state synchronization between a parent and a child feature. It allows to replace custom code by composable (*sic*) elements,  handle uni- and bi-directional synchronization, or containers of the child state (via the `map` operation).

I'm sure it can be improved in many ways. I don't know if such feature could have its place in TCA (first as a Case Study), or if it should go the "additional library" way (with much less visibility and adoption of course). I'd be glad if @mbrandonw or @stephencelis could pitch in on the matter.

RabugenTom:
It turns out `StateBinding` are some kind of programmable `WritableKeypath`, and boilerplate accessors are not required anymore! 

If the parent state conforms to the marker protocol `StateContainer`, we can directly pullback reducers using the StateBinding:
```swift
featureReducer.pullback(binding: ParentState.feature, action:...)
```
Stores can be scoped using the `get` method on `StateBinding` (which is a (Parent) -> Feature function):
```swift
store.scope(binding: ParentState.feature.get, action:...)
```

To use this optional feature, the marker protocol is unfortunately necessary because one can't use `inout` in subscript arguments. Otherwise, one would be able to install the subscript in the StateBinding directly and drop the marker Protocol completely. [This may come some day hopefully](https://forums.swift.org/t/inout-subscript-parameters/31429).

I did not overload existing pullback and scoping methods, and `StateBinding` variant are using the keyword "binding" instead of "state", so it shouldn't cause an additional strain on the compiler.

`StateBinding` now guarantee child-state coherence in a shorter way than most manual implementations, providing a DSL to describe the `Parent<->Child` states relation in a declarative way.

Please let me know what you think.

stephencelis:
@RabugenTom This definitely looks very cool 😄 Brandon and I will dive a bit deeper and discuss it soon! It's a bummer the property wrapper exploration you had didn't work with today's Swift, but very cool to see your workaround!

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-SharedState-StateBindings.swift
```diff
@@ -0,0 +1,350 @@
...
+    set {
+      self.content = newValue.text
+      self.count = newValue.count
+    }
+  }
+
+  // MARK: State Binding with internal storage -
+  // A private instance of FeatureState is stored.
+  // This instance will participate to state comparisons if any. In the binding, we only declare
+  // the relevant property `name`, bound with `self.content`. The private instance stores all the
```
RabugenTom, line 51: This should be `text`, bound with `self.content`

RabugenTom:
Thanks @stephencelis! I've updated the documentation to match latests developments.
I'm not touching this PR anymore until you have the time to look at it. I'm feeling there is something fundamental laying around, and you'll probably spot things I'm not even aware of.
Forgive me the terminology if some prior art exists, as I named things quite naively!
Thanks!

RabugenTom:
I spent the last days reimplementing this with protocols instead of functions. Implementing all the generics was very annoying, but it's super nice on call-site now, with a leaner syntax and fast code-completion.

I've benchmarked this on a 20 strings struct with a similar bound child and unfortunately, the results are not looking good:
```
name                            time         std        iterations
------------------------------------------------------------------
LargeValues.Ad-hoc - Direct       383.000 ns ± 125.87 %    1000000
LargeValues.Ad-hoc - KeyPaths     380.000 ns ±  97.97 %    1000000
LargeValues.Binding - Protocols 21841.000 ns ±  13.06 %      62175
LargeValues.Binding - Functions 40716.000 ns ±  17.69 %      32616
SmallValues.Ad-hoc - Direct        26.000 ns ± 661.27 %    1000000
SmallValues.Binding - Protocols  1155.000 ns ±  97.01 %    1000000
SmallValues.Binding - Functions  1059.000 ns ±  86.16 %    1000000
```
`LargeValues` is the 20-strings struct,` SmallValues` is a 3-strings struct (I didn't checked deep nesting yet). The protocol-based approach is around 50 times slower than an ad-hoc implementation (either using "." or keyPaths subscripts), which makes it very hard to sell.

I've carefully checked the code and when inlined, it should be very similar to the ad-hoc implementation. I'll try to profile it to see if something is taking specifically a lot of time.

In the meantime, I think it's safer to close this PR, as it will go nowhere in its current shape.

Too bad the performance is not there yet, because I think the UX is way better for the developer. In the app I'm developing, I spent almost 1 hour last week tracking a bug that would have been prevented by such bindings.
Maybe the more structured protocol version would allow to generate ad-hoc code in some way. I'll think about it as it would bring the best of both worlds.

RabugenTom:
After investigation, the main bottleneck turned out to be `KeyPath`'s. I've stripped everything down to its bare minimum to locate where so much time was lost, and I've ended isolating a single `KeyPath` subscript responsable for 90% of the call's duration (unfortunately unavoidable). The ad-hoc implementation using `KeyPath` to explicitly sync values is even faster than the direct one using `.`. I guess the compiler is able to correctly optimize `KeyPath`'s in this explicit configuration (they're generics in the binding), and this looks confirmed by Joe Groff [in this bug report](https://bugs.swift.org/browse/SR-9323?focusedCommentId=41666&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-41666)(it may have evolved since).

Bindings with protocols are also accepting blocks instead of `KeyPath`. If I'm using these less convenient closures instead of `KeyPath` to set the values , the binding approach falls down from 50x to 3x slower than the ad-hoc implementation:
```
name                                      time         std        iterations
----------------------------------------------------------------------------
LargeValues.Ad-hoc - Direct                  392.000 ns ± 104.99 %    1000000
LargeValues.Ad-hoc - KeyPaths                377.000 ns ±  71.77 %    1000000
LargeValues.Binding - Protocols - KeyPaths 23021.000 ns ±  17.57 %      57643 
LargeValues.Binding - Protocols - Blocks    1085.000 ns ±  37.63 %    1000000 *
LargeValues.Binding - Functions            40499.000 ns ±  12.31 %      34021
```
`KeyPath`'s are widely used in TCA, and bindings are providing many other features (like optional state deduplication to avoid updating the parent's if the child's value is the same), so I'll continue to work in this direction and open-source the implementation on my side when it will be a little more polished. I don't know if @mbrandonw or @stephencelis had time to look at the PR. If so, I would be greatly interested in their comments if they have some.

I'm also hoping that `KeyPath`'s subscript's speed will improve in the future. There are some open bug reports on the topic (like [SR-9323](https://bugs.swift.org/browse/SR-9323) or [SR-11983](https://bugs.swift.org/browse/SR-11983) for example).

JackYoustra:
I'm running into a problem with lots of shared state, have you made any progress on this?

tgrapperon:
Hey @JackYoustra! This is a PR I made with my legacy account.
I had another shot at it a couple months ago while generalizing #1227, and I managed to get a very nice syntax thanks to a smart trick (but I didn't pushed the generalization).
Unfortunately, it was going to inherit generic `KeyPath` slowness, but I think this latest experiment had enough information to potentially generate the explicit code from the generic derivation using some Xcode/SwiftPM plugin. I didn't push farther and parked the idea as I had to focus on personal projects unfortunately. Feel free to open a discussion with a sample case if you want, so I can provide more precise feedback and maybe tailor the API to fix real-life issues.

JackYoustra:
@tgrapperon thanks for responding so promptly 🤗. I'm looking at your repo that you mention in the pr https://github.com/tgrapperon/swift-derived-state and it looks pretty good, can I use this directly? Anything that you would caution me about with using it besides speed?

tgrapperon:
@JackYoustra I wouldn't advise to do so. I didn't checked this code for a long time, and I'm not sure it'll work out of the box with `ReducerProtocol`. The solution I recently experimented with should allow to opt-out easily by baking the transforms as explicit code via codegen. In the meantime, I would recommend to use ad-hoc solutions. Are you stuck with a specific configuration? If so, feel free to open a discussion so we can dive deeper.

JackYoustra:
@tgrapperon it seems tricky! I've put my thoughts at https://github.com/pointfreeco/swift-composable-architecture/discussions/1445, if you have any ideas / think my dynamic solution is acceptable, please let me know!

PR #454, opened 2021-03-22
Updated TestStore DSL

stephencelis:
This PR introduces a newer, flatter way of asserting with the test store. Instead of calling `store.assert(...)` with a series of steps, one can now invoke `store.send` and `store.received` directly, improving how/where XCTest failures are surfaced.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-AlertsAndActionSheetsTests.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-AlertsAndActionSheetsTests.swift
@@ -13,23 +13,21 @@ class AlertsAndActionSheetsTests: XCTestCase {
       environment: AlertAndSheetEnvironment()
     )
 
-    store.assert(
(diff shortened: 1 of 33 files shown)


mbrandonw (approved):

nmccann:
Very minor (and I almost managed to comment on this before it was merged), but there are a number of usages of `self.scheduler` that could be simplified to `scheduler`. I could create a PR for that, assuming there wasn't a reason for keeping them that way.

stephencelis:
@nmccann We may be in the minority here, but we prefer to always be explicit about `self` even where it can be implicit. We think it makes it clearer when a value is a property vs. a local variable or constant, especially when reading outside of Xcode, like in PRs. We'd take submissions to add `self.` if we've missed it anywhere, though 😁

nmccann:
Fair enough - I'll gladly admit that it does help readability in PRs. Even if you're in the minority - it's your project and you raise some compelling points, so I'm happy to defer to your preferences :) I've long ago realized there is no "correct" choice in regards to such things - internal consistency within a project is the more important thing.

PR #469, opened 2021-04-03
Fixes #434 by only using the newly-scoped state in case it returned a non-nil value.

### LINKED ISSUE #434: Fatal error: Unexpectedly found nil while unwrapping an Optional value: file ComposableArchitecture/ForEachStore.swift, line 143
**Describe the bug**
When filtering a ForEachStore we sometimes get a Fatal error: Unexpectedly found nil while unwrapping an Optional value: file ComposableArchitecture/ForEachStore.swift, line 143
The bug seems to appear in the Todos example as well.
We did not see this behavior before the ForEachStore-fixes.

**To Reproduce**
Tested on device iPhone 7 and simulator iPhone 11:
In the Todos project:
Add 20 Todos
mark 2 of them completed
press the Active segment and directly after the Completed segment.

**Expected behavior**
No Fatal error

**Screenshots**

https://user-images.githubusercontent.com/23285231/111308354-dd206b00-865a-11eb-8597-070421171394.mov

**Environment**
 - Xcode 12.4
 - OS (if applicable): iOS 14.4.1

fabstu:
`ForEachStore` has issues with first inserting and afterwards updates I think, which is shown in https://gist.github.com/fabstu/17ecfff9f327654eb69fb1ae92b2a0fb, I think #434 is similar. What deletion and the Todos-example have in common is, that the scoping does not have a value it can obtain through `$0[id: id]`. This fix simply ignores the parent-state change and waits for a real value to pass on the state change to the scoped store.

I do not know whether this design is sane, but it avoids crashes with the Todos example in #434 and for my example too. Is this testable?

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -176,6 +176,32 @@ public final class Store<State, Action> {
       .sink { [weak localStore] newValue in localStore?.state.value = toLocalState(newValue) }
     return localStore
   }
+  
+  /// opnalScope is the same as .scopes but does pass on state changes for which toLocalState returns a nil value.
+  public func optionalScope<LocalState, LocalAction>(
+    state toLocalState: @escaping (State) -> LocalState?,
+    action fromLocalAction: @escaping (LocalAction) -> Action
+  ) -> Store<LocalState, LocalAction> {
+    let localStore = Store<LocalState, LocalAction>(
+      initialState: toLocalState(self.state.value)!,
+      reducer: { localState, localAction in
+      self.send(fromLocalAction(localAction))
+      guard let newLocalState = toLocalState(self.state.value) else {
+        return .none
+      }
+      localState = newLocalState
+      return .none
+      }
+    )
+    localStore.parentCancellable = self.state
+      .sink { [weak localStore] newValue in
+        guard let newLocalState = toLocalState(newValue) else {
+          return
+        }
+        localStore?.state.value = newLocalState
+      }
+    return localStore
+  }
 
   /// Scopes the store to one that exposes local state.
   ///
--- a/Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
@@ -139,8 +139,8 @@ where Data: Collection, ID: Hashable, Content: View {
       WithViewStore(store.scope(state: { $0.ids })) { viewStore in
         ForEach(viewStore.state, id: \.self) { id in
           content(
-            store.scope(
-              state: { $0[id: id] ?? data[id: id]! },
+            store.optionalScope(
+              state: { $0[id: id] ?? data[id: id] },
               action: { (id, $0) }
             )
           )


stephencelis:
Hi @fabstu! Thanks for digging a bit deeper into this and for taking the time to submit a fix. Because this is a workaround for a SwiftUI bug we thought it might be best to contain any fix within the `ForEachStore` view, and we actually found that we could use `IfLetStore` inside and get things working relatively quickly:

https://github.com/pointfreeco/swift-composable-architecture/pull/471

Would you give this alternative a try to make sure it fixes things for you?

fabstu:
This also seems to fix the crash issue. With box solutions the Todos example then has an inconsistency where fast switching between filters fails to update the Todo list correctly, see the video below. Before it was crashing so it is an improvement and it only happens in case you switch between filters fast, so its definitely an improvement.

https://user-images.githubusercontent.com/20703451/113596891-374d9400-963b-11eb-908f-853756c429c8.mp4

I guess you would normally use the `@ObservedObject` pattern for viewStore to fix such a problem, right? But I do not see how to apply that here, might be my lack of experience though. Anyway here is where I add a little log output to my solution to see why we observe this, but the log output is exactly the same.

First I created 1 empty todo and selected the `All` filter. The following output is the result of switching to `Completed` and back again, first fast then slowly.

Fast, afterwards wrongly populated.

```log
Refreshing local state2 []
Ignoring new parent state because unwrapping returned nil2
received action:
  AppAction.filterPicked(.completed)

Refreshing local state2 [Todos.Todo(description: "", id: 39149CEA-7F7A-4C69-BB97-81B8190953D7, isComplete: false)]
received action:
  AppAction.filterPicked(.all)

Refreshing local state2 [Todos.Todo(description: "", id: 39149CEA-7F7A-4C69-BB97-81B8190953D7, isComplete: false)]
```

Slow, afterwards correctly populated. Same output.

```log
Refreshing local state2 []
Ignoring new parent state because unwrapping returned nil2
received action:
  AppAction.filterPicked(.completed)

Refreshing local state2 [Todos.Todo(description: "", id: 39149CEA-7F7A-4C69-BB97-81B8190953D7, isComplete: false)]
received action:
  AppAction.filterPicked(.all)

Refreshing local state2 [Todos.Todo(description: "", id: 39149CEA-7F7A-4C69-BB97-81B8190953D7, isComplete: false)]
```

My device is an iPhone XR with iOS 14.4.1 (I lowered minimum iOS version for this in the Todos app).

stephencelis:
@fabstu I think this is a bug that affects vanilla SwiftUI apps, as well. Was able to reproduce some strange behavior (filed with Apple) here: https://gist.github.com/stephencelis/bb9418e4cb0bccb0e3b58cd58949e419

fabstu:
@stephencelis  Removing `.animation()` from the `Picker` fixes that inconsistency, nice. I hope Apple someday fixes all those bugs, having such inconsistencies present gives me quite a headache each time. :-)

Removing the view instead of keeping it until it is automatically removed by the next refresh cycle seems the right thing to do, so gonna close this. Thanks for the quick update!

PR #482, opened 2021-04-08
Fix background colors in UIKit demos to work in light and dark mode

hungtruong:
I noticed while playing around with the demos that the UIKit demos are setting the background color to `white`, which breaks some of the views in dark mode (the counter label ends up being white on a white background). Setting them to `systemBackground` fixes this.

I tested all of the views that I changed manually and they work in both light and dark mode now. #teamdarkmode


mbrandonw (approved):
Amazing, thanks so much @hungtruong!

PR #497, opened 2021-04-19
Clarify docs for UIKit `ifLet` function

mluisbrown:
Make it explicit in the documentation for `ifLet` that either the `else` or `unwrap` closure will be called based on the *initial* state of the optional state (`nil` or non-`nil`).


stephencelis (approved):
Thanks! We'll make a few small adjustments after we merge.

PR #501, opened 2021-04-20
Add @ViewBuilder to ForEachStore and IfLetStore

stephencelis:
We noticed this awhile back but must have forgotten about it.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/IfLetStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/IfLetStore.swift
@@ -42,7 +42,7 @@ public struct IfLetStore<State, Action, Content>: View where Content: View {
   ///   - elseContent: A view that is only visible when the optional state is `nil`.
   public init<IfContent, ElseContent>(
     _ store: Store<State?, Action>,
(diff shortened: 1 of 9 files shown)


Sources/ComposableArchitecture/SwiftUI/IfLetStore.swift
```diff
@@ -42,7 +42,7 @@ public struct IfLetStore<State, Action, Content>: View where Content: View {
   ///   - elseContent: A view that is only visible when the optional state is `nil`.
   public init<IfContent, ElseContent>(
     _ store: Store<State?, Action>,
-    then ifContent: @escaping (Store<State, Action>) -> IfContent,
+    @ViewBuilder then ifContent: @escaping (Store<State, Action>) -> IfContent,
     else elseContent: @escaping @autoclosure () -> ElseContent
```
stephencelis, line 46: I also wonder if this should be changed to:

```suggestion
    @ViewBuilder else elseContent: @escaping () -> ElseContent
```

Would be a bigger change and require deprecations, though.

mbrandonw (approved):

PR #502, opened 2021-04-20
Add failing iflet store test

dannyhertz:


stephencelis:
@dannyhertz A lot has changed upstream, so we decided to test this out again and the extra emission has gone away! Note that the test does not pass currently because we now squash synchronous mutations down, but if you want it to pass you can add a `.receive(on:)` the synchronous effect and introduce a test scheduler that advances a tick after sending the child action.

Going to close this out as the bug appears fixed, but let us know if you have further issues!

PR #525, opened 2021-04-30
Modernize examples

mbrandonw:
Some of the example code was looking a little old, so updating everything with our more modern practices, e.g. key paths for functions, `ViewState.init` and `.main`/`.test`/`.immediate` scheduler helpers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
       - '*'
 
 jobs:
-  tests:
+  library:
     runs-on: macos-10.15
     strategy:
       matrix:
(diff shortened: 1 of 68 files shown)


.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ on:
       - '*'
 
 jobs:
-  tests:
+  library:
```
mbrandonw, line 12: this updates CI to run library tests on 11.3/11.7/12.4 and run example app tests only on 11.7/12.4 so that we can use key path functions.

Examples/TicTacToe/Sources/Core/LoginCore.swift
```diff
@@ -38,63 +38,63 @@ public struct LoginEnvironment {
   }
 }
 
-public let loginReducer =
+public let loginReducer = Reducer<LoginState, LoginAction, LoginEnvironment>.combine(
   twoFactorReducer
-  .optional()
```
mbrandonw, line 43: this is mostly just a re-indentation so best to view this PR with whitespace off.

Examples/TicTacToe/Sources/Views-SwiftUI/LoginSwiftView.swift
```diff
@@ -15,6 +17,16 @@ public struct LoginView: View {
     var isLoginButtonDisabled: Bool
     var password: String
     var isTwoFactorActive: Bool
+
+    init(state: LoginState) {
```
stephencelis, line 21: Should this be in an extension? Just thinking that in isowords it's convenient to be able to create ViewState independent of having domain State first, and might be nice for examples to encourage that as well.

mbrandonw: down to do that, but in practice we seem to always have the `init` up in the `struct` and I think that makes it easier to make changes in the future.

stephencelis (approved):

Examples/TicTacToe/Sources/Views-SwiftUI/LoginSwiftView.swift
stephencelis: I just seem to remember a recent refactor we did made GameOverView previews a lot easier to mock out by having the default init around. Or do you mean define the member wise one too?

PR #528, opened 2021-05-04
Fix ForEachStore

stephencelis:
The fix in #471 introduced overhead in which every `ForEachStore` content block was being called eagerly, which means the creation of unnecessary stores. PRed is an alternative fix that still avoids the original crash reported, and seems safe in practice. The force unwrap is a little scary, but maybe unavoidable...

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
@@ -131,21 +131,22 @@ where Data: Collection, ID: Hashable, Content: View {
   where
     EachContent: View,
     Data == IdentifiedArray<ID, EachState>,
-    Content == WithViewStore<
-      [ID], (ID, EachAction), ForEach<[ID], ID, IfLetStore<EachState, EachAction, EachContent?>>
-    >
+    Content == WithViewStore<[ID], (ID, EachAction), ForEach<[ID], ID, EachContent>>
   {
     self.data = store.state.value
     self.content = {
       WithViewStore(store.scope(state: { $0.ids })) { viewStore in
-        ForEach(viewStore.state, id: \.self) { id in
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
```diff
@@ -131,21 +131,22 @@ where Data: Collection, ID: Hashable, Content: View {
...
-          //     re-evaluate views for elements no longer in the collection.
+        ForEach(viewStore.state, id: \.self) { id -> EachContent in
+          // NB: We cache elements here to avoid a potential crash where SwiftUI may re-evaluate
+          //     views for elements no longer in the collection.
           //
           // Feedback filed: https://gist.github.com/stephencelis/cdf85ae8dab437adc998fb0204ed9a6b
-          IfLetStore(
-            store.scope(state: { $0[id: id] }, action: { (id, $0) }),
-            then: content
+          let element = store.state.value[id: id]!
```
stephencelis, line 144: This force unwrap seems to be safe, it's the scope's `state` closure in which the underlying value may no longer reflect our expectation (of the element still being around).

fabstu: I don't really understand when this is recomputed. Is it updated on each viewStore id-refresh? Then it would be a valid item and fine, because it can only have been deleted if its not in the current scope anymore.

stephencelis: This is merely a cached value for invalid views to fall back on if the SwiftUI runtime tries to evaluate the body of one of these for-eaches after an element is removed from the dictionary. This is to work around a SwiftUI bug that we reported way back during the first beta and sadly isn't yet fixed, but ideally this initial cached value is never rendered in the view itself. If it is, though, please let us know!

mbrandonw (approved):

PR #533, opened 2021-05-07
Add UIKit `DiffableListsOfState` example

p4checo:
## Motivation

A new `DiffableListsOfState` UIKit example was created to demonstrate the usage of TCA with a list of (collection view) cells that have a store and are backed by a diffable data source.

It uses a `Store.child` helper to return a single scoped store for a particular child (cell), from its identifier. This is useful to
avoid re-scoping all children when only a subset of children needs to be created (e.g. in a diffing set up, only insertion stores need to be created).

## Changes

- Create new `DiffableListsOfState` UIKit example
  + Create new DiffableCounterList TCA "stack" and`DiffableCountersTableViewController` which uses Diffing in a `UICollectionView` via `UICollectionViewDiffableDataSource`.
  + Create new `CounterCollectionViewCell` which wraps a counter, to demonstrate injecting child stores in cells.
  + Enrich `CounterState` with a `id: UUID` property for diffing.

## `DiffableListsOfState` Demo

https://user-images.githubusercontent.com/1391324/116435915-c701ef00-a843-11eb-9a67-b9f3007745fb.mov

### DIFF
--- a/Examples/CaseStudies/CaseStudies.xcodeproj/project.pbxproj
+++ b/Examples/CaseStudies/CaseStudies.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0AAA31292644B32A0091B0B1 /* CounterCollectionViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0AAA31262644B3290091B0B1 /* CounterCollectionViewCell.swift */; };
+		0AAA312B2644B32A0091B0B1 /* DiffableListsOfState.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0AAA31282644B3290091B0B1 /* DiffableListsOfState.swift */; };
 		4F5AC11F24ECC7E4009DC50B /* 01-GettingStarted-SharedStateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4F5AC11E24ECC7E4009DC50B /* 01-GettingStarted-SharedStateTests.swift */; };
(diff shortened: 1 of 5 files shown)


andreyz:
@simme maybe this will help with marrying Collection View and TCA?

stephencelis:
@p4checo Cool stuff, thanks for sharing! We'll take a deeper look next week when we have a moment.

p4checo:
@stephencelis @mbrandonw did this work by any chance inspire you for [episode 148](https://www.pointfree.co/episodes/ep148-derived-behavior-collections)? It certainly has some similar ideas! 😅 

If so, does that mean you'll be taking a look at this soon? 😇🙏🏼

p4checo:
@stephencelis @mbrandonw  hi there! 👋🏼  

Made the `Store.child` a private helper on the example itself, to avoid increasing public API surface (which I think would be the deal breaker in this PR 😇)

That means this PR is now _only_ a UIKit example on how to use TCA with a diffable data source backed VC. Please let me know if you think it adds value and if it's in line with what you want for the case studies.

p4checo:
Hi @stephencelis @mbrandonw 👋🏼 

Sorry for pushing, but this PR has sadly been sitting here waiting for your review for over a year now. Do you think these changes are worth it, or should I just close this? Thanks!

stephencelis:
Hey @p4checo! Sorry again for the looong delay. It becomes very easy to lose track of open PRs after some time. I think what we've come to understand is that we should be very comfortable with all the case studies in the repo to be able to maintain them well, so we're trying to be quite careful in adding surface area here. That said, whenever you solve a problem like this we think the GitHub discussions are the perfect place to share and for folks to discover more and more ways of solving problems with TCA. I know it's been some time, but if you do want to share your diffable demo there, it'd be the perfect place!

PR #536, opened 2021-05-09
Cancel in flight timers with the same id

dannyhertz:
As discussed in [an issue](https://github.com/pointfreeco/swift-composable-architecture/discussions/532#discussioncomment-711916) *not* cancelling in flight timer with an existing ID is fairly unusual/unexpected behavior. This PR just sets the cancelInFlight parameter to true for the underlying timer effect.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Timer.swift
+++ b/Sources/ComposableArchitecture/Effects/Timer.swift
@@ -103,6 +103,6 @@ extension Effect where Failure == Never {
       .autoconnect()
       .setFailureType(to: Failure.self)
       .eraseToEffect()
-      .cancellable(id: id)
+      .cancellable(id: id, cancelInFlight: true)
   }
 }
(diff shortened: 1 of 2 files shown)


andreyz:
My only concern with the change that it's somewhat breaking behavior altering change. Maybe there's a softer, more gradual (with deprecation warning first) introduction of such change? I don't have any code that could subtly change behavior due to this, but some developers might.

dannyhertz:
> My only concern with the change that it's somewhat breaking behavior altering change. Maybe there's a softer, more gradual (with deprecation warning first) introduction of such change? I don't have any code that could subtly change behavior due to this, but some developers might.

Definitely would be less potentially disruptive to pass cancelInFlight as an explicit param and default it to false.

stephencelis:
While it might be a breaking change, it seems unlikely that anyone relies on this behavior (though if there are use cases folks can think of or _are_ taking advantage of, please let us know). We'll think on it over the weekend, though!

PR #550, opened 2021-05-19
Deferring Side Effect

wendyliga:
AFAIK, the only way we can delay side effect is to put chain delay on the side effect itself.

```swift
case let .locationTapped(location):
    struct SearchWeatherId: Hashable {}

    state.locationWeatherRequestInFlight = location

    return environment.weatherClient
      .weather(location.id)
      .delay(for: 0.5, scheduler: environment.mainQueue)
      .catchToEffect()
      .map(SearchAction.locationWeatherResponse)
      .cancellable(id: SearchWeatherId(), cancelInFlight: true)
```

but because of declarative nature, the delay is not on `environment.weatherClient.weather(location.id)` execution itself, but rather the `.catchToEffect()`, `.map(SearchAction.locationWeatherResponse)`, and below operation.

so i purpose a new Effect API named `deffered`, that delay the execution of the side effect.

```swift
case let .locationTapped(location):
    struct SearchWeatherId: Hashable {}

    state.locationWeatherRequestInFlight = location

    return environment.weatherClient
            .weather(location.id)
            .catchToEffect()
            .map(SearchAction.locationWeatherResponse)
            .deferred(for: 3, scheduler: environment.mainQueue)
```

with this new API, the execution of fetching weather info on weather client will happend based on given `dueTime`.

## Update
- rename the API into `deferred`


wendyliga:
just realize can utilize `debounce` https://github.com/pointfreeco/swift-composable-architecture/blob/main/Sources/ComposableArchitecture/Effects/Debouncing.swift on effect.

will close the PR, thank you

mbrandonw:
Hi @wendyliga, thanks for the PR!

We actually have a version of this operator in isowords that we wanted to first test out over there before bringing into TCA: https://github.com/pointfreeco/isowords/pull/94

It may be time to bring it over to this repo. We will discuss and let you know.

Sources/ComposableArchitecture/Effects/Deferring.swift
```diff
@@ -0,0 +1,29 @@
+import Combine
+
+extension Effect {
+  /// Returns an effect that will be executed after given `dueTime`.
+  ///
+  ///     case let .textChanged(text):
+  ///       return environment.search(text)
+  ///         .map(Action.searchResponse)
+  ///         .deferred(for: 0.5, scheduler: environment.mainQueue)
```
stephencelis, line 9:
```suggestion
  /// ```swift
  /// case let .textChanged(text):
  ///   return environment.search(text)
  ///     .map(Action.searchResponse)
  ///     .deferred(for: 0.5, scheduler: environment.mainQueue)
  /// ```
```

stephencelis (approved):
Cool! This looks great, thanks! Will merge when green :smile:

PR #559, opened 2021-05-21
Add Stateless

jeffersonsetiawan:
Let said we have a AppState that need an optional stateless Child (`Store<Void, ChildAction>`) , that will shown/hide using `IfLetStore`
```swift
struct AppState: Equatable {
  	var childState: Void?
}
```
We can't make that Void because Void is not `Equatable`.
So This PR is adding a helper `Stateless` struct that contains no property but it is Equatable
```swift
struct AppState: Equatable {
  	var childState: Stateless?
}
```

So you can do this now:
```swift
IfLetStore(
  store.scope(state: \.childState, action: AppAction.child).stateless,
  then: ChildView.init(store:),
  else: { Text("No Child") }
)
```
Update:
I used it in the UIKit world (actually my project is using Texture, which similar api as UIKit)

Open for any input 🙏🏻

### DIFF
--- a/Sources/ComposableArchitecture/Stateless.swift
+++ b/Sources/ComposableArchitecture/Stateless.swift
@@ -1,0 +1,23 @@
+import Foundation
+
+/// A helper to using Stateless in IfLetStore
+/// Example case:
+/// struct State {
+///   var childState: Stateless?
+/// }
+///
+/// enum AppAction {
+///   case child(ChildAction)
+/// }
+///
+/// Type of ChildView store is Store<Void, ChildAction>
+///
+///     IfLetStore(
+///       store.scope(state: \.childState, action: AppAction.child),
+///       then: ChildView.init(store:),
+///       else: { Text("No Child") }
+///     )
+///
+public struct Stateless: Equatable {
+  public init() {}
+}


tgrapperon:
Using a simple `Bool` for `AppState` wouldn't be a simpler option in this case?

jeffersonsetiawan:
> Using a simple `Bool` for `AppState` wouldn't be a simpler option in this case?

Isn't `ifLet` differentiate only from `nil` and `not-nil`?

tgrapperon:
Yes, but you can use the `Bool` directly like
```swift
if viewStore.childState {
  ChildView(store: store.scope(state: { _ in () }, action: AppAction.child))
} else {
  Text("No Child")
}
```

jeffersonsetiawan:
> Yes, but you can use the `Bool` directly like
> 
> ```swift
> if viewStore.childState {
>   ChildView(store: store.scope(state: { _ in () }, action: AppAction.child))
> } else {
>   Text("No Child")
> }
> ```

oops I forget we can do that in SwiftUI..
Thanks @tgrapperon 

Actually, I create this helper for the UIKit world, because I need to listen every change of that state, that's why in UIKit I need the `IfLet`

jeffersonsetiawan:
Here is the simplified code of the usage.

```swift
class ParentVC: UIViewController {
	private var myChildView: ChildView?

	private let store: Store<AppState, AppAction
	

	init(store: Store<AppState, AppAction) {
		self.store = store
		setup()
	}

	func setup() {
		store.scope(
			state: \.childState,
			action: AppAction.child
		).ifLet(then: {
			// add ChildView
			self.myChildView = ChildView(store: $0.stateless)
			self.view.addSubview(self.myChildView)
		}, else: {
			// remove ChildView
			self?.myChildView = nil
			self?.setNeedsLayout()
		})
	}
}


class ChildView: UIView {
	private let store: Store<Void, ChildAction>

	init(store: Store<Void, ChildAction>) {
		self.store = store
		super.init()
	}
}
```

stephencelis:
@jeffersonsetiawan Thanks for the discussion and sharing your idea! Given that you've solved it using @tgrapperon's suggestion I'm going to close this, but if you have more you'd like to share since first opening this PR, please feel free to start a GitHub discussion!

PR #563, opened 2021-05-21
UIKit helpers generating UIAlertController from AlertState & ActionSheetState

andreyz:
Sharing production code used to bridge AlertState and AlertSheetState with UIKit presentation logic.

Should address requests like #561 and others.


larryonoff:
What about lighten os requirements to iOS 11 ?

andreyz:
@larryonoff TCA package has following requirements
```
  platforms: [
    .iOS(.v13),
    .macOS(.v10_15),
    .tvOS(.v13),
    .watchOS(.v6),
  ],
```

larryonoff:
> @larryonoff TCA package has following requirements
> 
> ```
>   platforms: [
>     .iOS(.v13),
>     .macOS(.v10_15),
>     .tvOS(.v13),
>     .watchOS(.v6),
>   ],
> ```

No questions than :)

Sources/ComposableArchitecture/UIKit/ActionSheetStateUIKit.swift
```diff
@@ -0,0 +1,27 @@
...
+@available(iOS 13, *)
+@available(macCatalyst 13, *)
+@available(macOS, unavailable)
+@available(tvOS 13, *)
+@available(watchOS, unavailable)
+extension ActionSheetState {
+  /// UIKit helper generating UIAlertController from corresponding ActionSheetState
+  /// - Parameter send: A function that wraps a alert action in the view store's action type.
+  /// - Returns: UIAlertController ready to presented by UIViewController.
+  public func toUIAlertController(send: @escaping (Action) -> Void) -> UIAlertController {
```
stephencelis, line 13: The main change we think is API. This looks a lot like the `toSwiftUI` stuff we define [here](https://github.com/pointfreeco/swift-composable-architecture/blob/a4f544e892f04822def9f5934ff4d7ecfdf1329e/Sources/ComposableArchitecture/SwiftUI/Alert.swift#L244), but that naming is an internal detail and not reflected in the public-facing API.

What do you think of adding this functionality as a convenience initializer on `UIAlertController` instead? Something like:

```swift
UIAlertController(state: AlertState, send: (Action) -> Void)
```

Sources/ComposableArchitecture/UIKit/AlertStateUIKit.swift
```diff
@@ -0,0 +1,80 @@
...
+  ///        func viewDidLoad() {
+  ///          ...
+  ///          viewStore.publisher
+  ///            .settingsAlert
+  ///            .sink { [weak self] alert in
+  ///              guard let self = self else { return }
+  ///              if let alert = alert {
+  ///                let alertView = alert.toUIAlertController(send: {
+  ///                  self.viewStore.send(.settings($0))
+  ///                })
```
stephencelis, line 27: This would then change to:
```suggestion
  ///                let alertView = UIAlertController(state: alert, send: {
  ///                  self.viewStore.send(.settings($0))
  ///                })
```

stephencelis (commented):
Hey @andreyz! Thanks for submitting this and sorry for taking so long to respond :smile:

This looks like a good improvement to us! Just one small request for the public-facing API.

andreyz:
Fun! Thanks for comments. I'll update soon.

andreyz:
@stephencelis please have another look when you get a moment 🙌

stephencelis (approved):

stephencelis:
Thanks!

PR #572, opened 2021-05-26
Remove convenience initializer on Store

stephencelis:
Like we discussed, this simplifies the internals a lil bit and shouldn't have much of measurable cost. If so we could maybe inline them away?


mbrandonw (approved):
nice, I like this much better.

PR #573, opened 2021-05-28
fix: compile tvOS

byMohamedali:
The project was not compiling for tvOS.

image

I fixed it by making the popover function unavailable for tvOS 👍


stephencelis:
Ha, nice catch, thanks!

We'll merge this but please note that the `iso` branch is not mainline and is used for experimenting in our game [isowords](https://github.com/pointfreeco/isowords). We definitely don't recommend depending on `iso` in your own projects, and instead copy/paste the functionality that is useful to you, so that the branch doesn't render your application broken in the future.

PR #576, opened 2021-05-31
Fix closure parameter.

yimajo:
Is this a typo? 😆


mbrandonw (approved):
Haha it is! That’s a funny one. Thanks!

PR #579, opened 2021-06-01
Only breakpoint when debugger is attached

stephencelis:
Currently, debug builds will crash when the breakpoint signal is raised. Since we consider these breakpoints to be non-fatal (and skip them in release mode), let's simply ignore them when a debugger isn't attached.


mbrandonw (approved):
nice

PR #596, opened 2021-06-14
Keep the StoreObservableObject instance in SwitchStore

tgrapperon:
The `StoreObservableObject` instance is now the same for the lifecycle of the view. This should avoid unwanted side-effects when vending a new `ObservableObject` instance each time `body` is evaluated.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/SwitchStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/SwitchStore.swift
@@ -48,7 +48,7 @@ public struct SwitchStore<State, Action, Content>: View where Content: View {
   public let store: Store<State, Action>
   public let content: () -> Content
   private let storeObservableObject: StoreObservableObject<State, Action>
-  public init(
+  init(
     store: Store<State, Action>,
     @ViewBuilder content: @escaping () -> Content
   ) {
--- a/Sources/ComposableArchitecture/SwiftUI/SwitchStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/SwitchStore.swift
@@ -47,18 +47,19 @@ import SwiftUI
 public struct SwitchStore<State, Action, Content>: View where Content: View {
   public let store: Store<State, Action>
   public let content: () -> Content
-
-  init(
+  private let storeObservableObject: StoreObservableObject<State, Action>
+  public init(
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hi @tgrapperon, thanks for the PR.

From what I understand how SwiftUI works, the code before and after this change should be equivalent. Since `SwitchStore` doesn't hold onto `@State`, `@ObservedObject`, or anything that observes changes, it will never re-compute its body unless the entire struct is recreated. So it seems that whether we create `StoreObservableObject` in the initializer or in the body should be equivalent.

So, I don't think we need to do this change, but if you did want to make the initializer `public` I think that would be a good change because it would allow people to build up their own `SwitchStore`s in ways we don't currently support.

mbrandonw:
> but if you did want to make the initializer `public` I think that would be a good change because it would allow people to build up their own `SwitchStore`s in ways we don't currently support.

Oops spoke too soon. We do _not_ want to make `SwitchStore`'s initializer public. We make it internal because we want to control the ways it can be created by allowing only `CaseLet` views to appear in the body.

tgrapperon:
Thanks for the feedback @mbrandonw. I understand your point about the two cases being equivalent (another case where local reasoning fails). Even worse, the initializer may be called more than `body` so the PR was worse than the current implementation.
I also agree with the initializer being internal. I updated the PR after seeing today's video.

The only thing I'm not at ease with TCA and SwiftUI is the constant reinstantiation of Stores and ViewStores. I have not assessed the impact on performance, but I feel a lot of time is spent recreating things that are the same. I don't know if this is something you explored with @stephencelis, but I think it would be great to be able to reuse `Store` and `ViewStore` instances when scoping/viewing with identifiable parameters (like keypaths or enum's casepaths). When scoping a store with the same keypath and casepath, TCA would return the same `Store` instance. Furthermore, each store that supports it can store a `ViewStore` instance that can be returned when using `WithViewStore`. This looks very complex though, and I don't know if it's worth it performance-wise.

That being said, I'll close the PR.

mbrandonw:
You definitely bring up some good points, and we're glad you opened the PR and expressed your concerns. We do have some performance improvements coming to TCA very soon. It doesn't directly address the issue you bring up (that of re-creating stores/view stores needlessly), but it does make their operations more efficient.

PR #597, opened 2021-06-14
Improve some docs.

mbrandonw:
This is just part 1 of what I'm sure will by many, many parts to update docs for taking full advantage of DocC. In this PR I've added the info about thread safety from the readme directly on the `Store` type.

I've also removed the leading spaces on code fences for `Store` and `ViewStore`, but there's a lot more to do.


Sources/ComposableArchitecture/Store.swift
```diff
@@ -5,7 +5,81 @@ import Foundation
...
+///     }
+///   }
+/// }
+/// ```
+///
+/// ### Scoping
+///
+/// The most important operation defined on ``Store`` is the ``scope(state:action:)-9iai9`` method, which allows you to transform a store into one that deals with local state and actions. This is necessary for passing stores to subviews that only care about a small portion of the entire application's domain.
+///
+/// For example, if an application has a tab view at its root with a tab for activity, search and profile, then we can model the domain like this:
```
stephencelis, line 33:
```suggestion
/// For example, if an application has a tab view at its root with tabs for activity, search, and profile, then we can model the domain like this:
```

Sources/ComposableArchitecture/Store.swift
```diff
@@ -5,7 +5,81 @@ import Foundation
...
+/// }
+///
+/// enum AppAction {
+///   case activity(ActivityState)
+///   case profile(ProfileState)
+///   case search(SearchState)
+/// }
+/// ```
+///
+/// And we can construct a view for each of these domains by applying ``scope(state:action:)-9iai9`` to a store that holds onto the full app domain in order to transform it into a store for each of these sub-domains:
```
stephencelis, line 49:
```suggestion
/// We can construct a view for each of these domains by applying ``scope(state:action:)-9iai9`` to a store that holds onto the full app domain in order to transform it into a store for each sub-domain:
```

Sources/ComposableArchitecture/Store.swift
```diff
@@ -5,7 +5,81 @@ import Foundation
...
+///
+/// This approach makes the fewest number of assumptions about how effects are created and transformed, and prevents unnecessary thread hops and re-dispatching. It also provides some testing benefits. If your effects are not responsible for their own scheduling, then in tests all of the effects would run synchronously and immediately. You would not be able to test how multiple in-flight effects interleave with each other and affect the state of your application. However, by leaving scheduling out of the ``Store`` we get to test these aspects of our effects if we so desire, or we can ignore if we prefer. We have that flexibility.
+///
+/// See also: ``ViewStore`` to understand how once observes changes to the state in a ``Store`` and sends user actions.
```
stephencelis, line 82:
```suggestion
/// See also: ``ViewStore`` to understand how one observes changes to the state in a ``Store`` and sends user actions.
```

Sources/ComposableArchitecture/Store.swift
```diff
@@ -5,7 +5,81 @@ import Foundation
...
+///
+///       ProfileView(store: self.store.scope(state: \.profile, action: AppAction.profile))
+///         .tabItem { Text("Profile") }
+///     }
+///   }
+/// ```
+///
+/// ### Thread safety
+///
+/// The `Store` class is not thread safe, and so all interactions with an instance of ``Store`` (including all of its scopes and derived ``ViewStore``s) must be done on the same thread. Further, if the store is powering a SwiftUI or UIKit view, as is customary, then all interactions must further be done on the _main_ thread.
```
stephencelis, line 71:
```suggestion
/// The `Store` class is not thread-safe, and so all interactions with an instance of ``Store`` (including all of its scopes and derived ``ViewStore``s) must be done on the same thread. Further, if the store is powering a SwiftUI or UIKit view, as is customary, then all interactions must be done on the _main_ thread.
```

Sources/ComposableArchitecture/Store.swift
```diff
@@ -5,7 +5,81 @@ import Foundation
...
+///
+/// The reason stores are not thread safe is due to the fact that when an action is sent to a store, a reducer is run on the current state, and this process cannot be done from multiple threads. It is possible to make this process thread safe by introducing locks or queues, but this introduces new complications:
```
stephencelis, line 73:
```suggestion
/// The reason stores are not thread-safe is due to the fact that when an action is sent to a store, a reducer is run on the current state, and this process cannot be done from multiple threads. It is possible to make this process thread-safe by introducing locks or queues, but this introduces new complications:
```

Sources/ComposableArchitecture/Store.swift
```diff
@@ -5,7 +5,81 @@ import Foundation
...
+///         )
+///       )
+///     }
+///   }
+/// }
+/// ```
+///
+/// ### Scoping
+///
+/// The most important operation defined on ``Store`` is the ``scope(state:action:)-9iai9`` method, which allows you to transform a store into one that deals with local state and actions. This is necessary for passing stores to subviews that only care about a small portion of the entire application's domain.
```
stephencelis, line 31: Did we wanna keep docs at 110 cols before merging?

Sources/ComposableArchitecture/Store.swift
```diff
@@ -290,7 +374,7 @@ public final class Store<State, Action> {
     self.scope(state: { _ in () })
   }
 
-  /// Returns an "actionless" store by erasing action to `Never`.
+  ///  Returns an "actionless" store by erasing action to `Never`.
```
stephencelis, line 377:
```suggestion
  /// Returns an "actionless" store by erasing action to `Never`.
```

Sources/ComposableArchitecture/ViewStore.swift
```diff
@@ -9,41 +9,45 @@ import SwiftUI
...
-///     }
-///    ```
+/// @objc func incrementButtonTapped() {
+///   self.viewStore.send(.incrementButtonTapped)
+/// }
+/// ```
+///
+/// ### Thread safety
 ///
+/// The ``ViewStore`` class is not thread safe, and all interactions with it must happen on the main thread. See the documentation of the ``Store`` class for more information why this decision was made.
```
stephencelis, line 50:
```suggestion
/// The ``ViewStore`` class is not thread-safe, and all interactions with it must happen on the main thread. See the documentation of the ``Store`` class for more information why this decision was made.
```

stephencelis (approved):
Nice nice, just a few small tweaks.

Sources/ComposableArchitecture/Store.swift
mbrandonw: won't swift-format take care of this for us? if not I will fix it.

PR #598, opened 2021-06-15
Add Indonesian Translation

wendyliga:
i've added indonesian translation, also add `Translations` section to `Table of contents`
feel free for feedback


mbrandonw (approved):
Awesome, thanks @wendyliga!

PR #601, opened 2021-06-18
ForEachStore Dictionary support

mashe:
Hello community!

I am pretty new to SCA, so, please, point if i am missing something.
I have noticed that `Reducer` has `forEach` function that supports `Dictionary<Key, LocalState>`, but there is no such corresponding `ForEachStore` initialisation.
Is it made on some purpose? If not, here is PR with Dictionary support for ForEachStore.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
@@ -166,6 +166,41 @@ where Data: Collection, ID: Hashable, Content: View {
       }
     }
   }
+  
+  /// Initializes a structure that computes views on demand from a store on a collection of data and
+  /// an identified action.
+  ///
+  /// - Parameters:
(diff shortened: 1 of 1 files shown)


stephencelis (commented):
Very cool! The one thing that I think we'd want to get a handle on is the fact that dictionaries are unordered, so looping over `.keys` is nondeterministic. I'm not sure what the "correct" way of handling this is, but we'd probably at least want the initializer to be able to describe a sorting strategy. We'll think on this a bit.

mashe:
@stephencelis thanks for pointing me that.
Not sure, if passing the sorting strategy to initialiser is a good idea: `ForEachStore` is expected to be called within `View` and, as for me, the sorting strategy is something out of `View` responsibility. So it looks like that data has to be sorted within `State` and passed being already prepared. And it means that `Dictionary` is not a proper tool for the case.
If you have no objections, I close the issue as a conceptually wrong one.

PR #603, opened 2021-06-19
Avoid writing CasePath’s LocalState into GlobalState when it cannot have changed

tgrapperon:
If I’m not mistaken, with the current version of the reducer for `CasePath` states, the local state is written back into the global state each time the reducer runs, whether it has changed or not. This may needlessly trigger observers upstream.

This workaround uses a simple wrapper with a `didSet` flag that is set as soon as the reducer writes the state (it does not check if the state was effectively modified though). The local state is then written into the global one only if this flag was set.

This wrapper is used only for `CasePath`’s pullbacks, as other variants are using in place access.

### DIFF
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -469,7 +469,7 @@ public struct Reducer<State, Action, Environment> {
     .init { globalState, globalAction, globalEnvironment in
       guard let localAction = toLocalAction.extract(from: globalAction) else { return .none }
 
-      guard var localState = toLocalState.extract(from: globalState) else {
+      guard let localState = toLocalState.extract(from: globalState) else {
         if breakpointOnNil {
           breakpoint(
             """
@@ -499,10 +499,17 @@ public struct Reducer<State, Action, Environment> {
         }
         return .none
       }
-      defer { globalState = toLocalState.embed(localState) }
+      
+      var wrappedState = StateWrapper(localState)
+      
+      defer {
+        if wrappedState.stateDidChange {
+          globalState = toLocalState.embed(wrappedState.state)
+        }
+      }
 
       let effects = self.run(
-        &localState,
+        &wrappedState.state,
         localAction,
         toLocalEnvironment(globalEnvironment)
       )
@@ -511,7 +518,7 @@ public struct Reducer<State, Action, Environment> {
       return effects
     }
   }
-
+  
   /// Transforms a reducer that works on non-optional state into one that works on optional state by
   /// only running the non-optional reducer when state is non-nil.
   ///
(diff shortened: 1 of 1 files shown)


lukeredpath:
Are you actually seeing this trigger observers? Setting the state if it hasn't changed should behave like a no-op because the ViewStore removed duplicate changes automatically. This seems redundant to me.

tgrapperon:
The purpose of this PR is to avoid needless work for the `ViewStore`’s, or other observers like `Store.parentCancellable`. 

> ultimately the store's state will be set to a new copy of the state after every action is processed once it reaches the root

Are you sure? I'm not seing this in the code. All reductions or pullbacks are using inout accessors to the state (excepted the one with `CasePath`’s). I don't see where the state is written outside of a reduction. As I'm seeing it, all reducers are ultimately working on the global state, and if this state changes, it will trigger the `parentObservable` of its scoped children, which will cascade to do the same with their own scoped children. If the reducer doesn’t write the state, the `CurrentValueSubject` doesn’t fire, and the whole chain stays put. No ViewStore has to check if something has changed (which is never free, and almost certainly heavier than a unique boolean check at the beginning of the chain).

This PR aims to preserve this property for `CasePath` pullbacks, but maybe I’m reading TCA wrong.

lukeredpath:
Right, I see what you're saying, it seems like this optimises for a case where an action results in no state mutations at all in your entire app reducer chain - but how often would that happen? If any part of the state changes after it's passed through the root reducer, it will trigger a cascade of observer updates.

Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -499,10 +499,17 @@ public struct Reducer<State, Action, Environment> {
...
+      
+      defer {
+        if wrappedState.stateDidChange {
+          globalState = toLocalState.embed(wrappedState.state)
+        }
+      }
 
       let effects = self.run(
-        &localState,
+        &wrappedState.state,
```
stephencelis, line 512: @tgrapperon I'm not quite I understand how this is intended to work. Shouldn't this line always set `stateDidChange` to true, thus updating `globalState`? Or do you have provide an example (or better, test case) that demonstrates how `stateDidChange` can remain false?

tgrapperon: TIL inout arguments are reassigned at the end of the function execution.
https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID545

This is quite a surprise to me, as I was expecting some kind of reference semantic, but it makes sense now. 
I don't think the optimisation they're mentioning kick's in in reducers. Maybe that's something to explore.

I'll need to adjust my mental model though.
All of this makes the PR irrelevant and I'll close it. 
Thanks for reviewing it.

tgrapperon:
This PR was made under the false assumption that inout argument were not mutated if the function was not mutating them explicitly in its body. This is not true as the simple fact of calling `f<T>(value: inout T) {}` mutates `value`.

This makes the workaround useless, as it will always detect the mutation caused by being passed as an `inout` argument to the reducer.

It is possible that the optimization mentioned in https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID545 kicks in in Release builds, but all of this shows that I'm not expert enough to have an opinion on the matter.

PR #608, opened 2021-06-21
Relax ViewStore.binding constraint

stephencelis:
Must've been accidental.


stephencelis:
Ah yeah, the constraint is inherited. Was hoping it would be possible to relax for types that can't be equatable but not sure it's possible. Will close for now.

PR #624, opened 2021-06-29
Added multicast to ViewStore publisher to reduce equality checks

iampatbrown:
I don't think there is too much overhead using multicast and creating an additional CurrentValueSubject. It might be possible to implement similar behaviour inside StorePublisher as an alternative though.


iampatbrown:
Could consider storing the CurrentValueSubject inside the ViewStore and just call objectWillChange inside the sink... 
```swift
public init(
    _ store: Store<State, Action>,
    removeDuplicates isDuplicate: @escaping (State, State) -> Bool
  ) {
    self._state = CurrentValueSubject(store.state.value)
    let publisher = store.state
      .removeDuplicates(by: isDuplicate)
      .multicast(subject: self._state)
      .autoconnect()
    
    self.publisher = StorePublisher(publisher)
    self._send = store.send
    self.viewCancellable = publisher
      .sink { [weak self] _ in self?.objectWillChange.send() }
  }

  /// The current state.
  public var state: State {
    self._state.value
  }
  ```

stephencelis:
@iampatbrown Thanks for this! @mbrandonw and I took inspiration from #618 and wrote a very naive benchmark that showed that multicasting with a new subject does add a bit of overhead. We'd like to write a beefier benchmark with state that's a bit more real-world, though, to weigh this overhead against the equality check it's preventing. We're going to keep this open while we investigate further, but wanted to let you know where we're at 😄

iampatbrown:
Great idea :) Was looking at that discussion as well. I'll definitely investigate further using the benchmark branch.

iampatbrown:
@stephencelis and @mbrandonw storing the state in a CurrentValueSubject and forwarding changes seems to work better on my end.

stephencelis:
@iampatbrown Looks like it! We're seeing a much more minimal cost with those changes. We'd like to flesh out some benchmarks before merging this, but assume that it will be a favorable change :smile: Thanks again for taking the time to dig into the problem and submit this PR!

iampatbrown:
No worries. Thanks for pointing me in the right direction. 

I noticed with CurrentValueSubject `state.send(newValue)` benchmarked better than `state.value = newValue` and also `guard let self = self` worked better than `self?.` when weak self was referenced more than once inside a closure. It was fairly insignificant but interesting to learn more about.

iampatbrown:
@stephencelis Okay for me to close this one? Pretty straight forward for you guys to implement on your end when ready.

stephencelis:
We're happy to leave it open and merge when we've weighed the performance. If being open is getting in the way for you, let us know. We can track and link to it in another issue.

iampatbrown:
Cool. We can leave it open. I'll just fix up the formatting. At some point I would have used swift format which made a change to the convenience initialisers.

stephencelis:
@iampatbrown No need to worry about formatting 😄 We automatically run `swift-format` on `main` when branches are merged.

mbrandonw:
Hi @iampatbrown, I just merged main into this branch and did a small clean up pass here https://github.com/pointfreeco/swift-composable-architecture/pull/624/commits/48845a9bc0edc17209118ab6f53ec8b54b6a6d37. We will be merging soon!

mbrandonw (approved):
Going to merge this now. We benchmarked this with your latest changes the times even went down by about ~5%.

Thanks for the work @iampatbrown!

PR #632, opened 2021-07-06
Don't use CasePath in .binding

mbrandonw:
Deprecating the `.binding` helper that takes a `CasePath` directly and adding a new one that only takes an extraction.


stephencelis (approved):

stephencelis:
Getting rid of the deprecated version gets 11.7 building. We could also leave it in with a `#if compiler(>=5.3)` check. I think it's highly unlikely that there are many cases in the wild of non-literal case paths being passed to `Reducer.binding`, but hopefully this covers whatever instances are out there.

Issue #625, opened 2021-06-30
`ViewStore.binding` firing when state is nil


ferologics:
When a view is dismissed, and the parent nils the child's state, an `optional()` reducer will receive actions from child's view binding even if the child state is `nil`, thus we get an `assertionFailure`. How to prevent that? 

Could the `IfLetStore` somehow remove the view immediately when it's optional state is set to nil (to not leave room for any child state action emission)?

stephencelis:
:wave: @ferologics Thanks for the report. We'd love to help but need a bit more info. When you tap the "new issue" button we have a "bug report" template that tries to collect a bunch of info that make it easier for us to get a handle on issues. The most helpful things would be:

- Could you attach a project demonstrating the problem?
- Could you let us know what version of swift-composable-architecture, Xcode, OS, etc., the problem is occurring in? OS behavior changes over time (SwiftUI 1 vs. 2 vs. 3), so it'd be helpful to know if this is a long-standing issue, new, etc.

stephencelis:
Hi @ferologics, I'm going to close this now, since there's not enough information to reproduce. If either of the other issues that mention this one, #631 or #634, are the same issue, please let us know! And if it's an entirely different issue, please attach a project that captures it here. We'd love to figure out a fix! I'll reopen this if you do.

stephencelis:
(This might also be a related issue: https://github.com/pointfreeco/swift-composable-architecture/issues/282)

ferologics:
@stephencelis I'm apologize for the lack of documentation provided around the issue. During my debugging I've found that the `assertionFailure was triggered from within a `GeometryReader` of the child view (who's state was nil at the time). The `GeometryReader` updates a binding passed from parent to child, which would re-evaluate upon the child's view dismissal.

At the end of the day we still had to use a `[ID:State]` dictionary (which kept the state "alive" in case any actions would be sent to it). Perhaps some may use this as an alternative to `optional(breakpointOnNil: false)` (although I would not encourage it).

PR #634, opened 2021-07-07
Change ViewStore state update behavior

maximkrouk:
I had an issue with a ComposableCollectionView, so that fix resolves the issue.
- `ComposableCollectionView` is a dataSource of self and it returns `store.items.count` from `numberOfItemsInSection` method
- `ComposableCollectionView` subscribes to `state.items.count` changes and updates itself
- After loading data, published `state.items.count` is equal to 10, but `viewStore.items.count` is 0
- So state update is received before the `ViewStore.state` acually changed

The fix is that `ViewStore.state` is taken directly from the `Store` now, so we don't even have to assign a new `Store.state` to the `ViewStore.state` anymore, but instead of this we should send `ViewStore.objectWillChange` event from the subscription

### DIFF
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -72,17 +72,17 @@ public final class ViewStore<State, Action>: ObservableObject {
     removeDuplicates isDuplicate: @escaping (State, State) -> Bool
   ) {
     self.publisher = StorePublisher(store.state, removeDuplicates: isDuplicate)
-    self._state =  { store.state.value }
+    self._state = { store.state.value }
     self._send = store.send
     self.viewCancellable = store.state
       .dropFirst()
       .removeDuplicates(by: isDuplicate)
       .sink { [weak self] _ in self?.objectWillChange.send() }
   }
-  
+
   let _state: () -> State
   let _send: (Action) -> Void
-  
+
   /// The current state.
   public var state: State { _state() }
 
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -72,22 +72,19 @@ public final class ViewStore<State, Action>: ObservableObject {
     removeDuplicates isDuplicate: @escaping (State, State) -> Bool
   ) {
     self.publisher = StorePublisher(store.state, removeDuplicates: isDuplicate)
-    self.state = store.state.value
+    self._state =  { store.state.value }
     self._send = store.send
     self.viewCancellable = store.state
       .dropFirst()
       .removeDuplicates(by: isDuplicate)
-      .sink { [weak self] in self?.state = $0 }
+      .sink { [weak self] _ in self?.objectWillChange.send() }
   }
-
-  /// The current state.
-  public private(set) var state: State {
-    willSet {
-      self.objectWillChange.send()
-    }
-  }
-
+  
+  let _state: () -> State
   let _send: (Action) -> Void
+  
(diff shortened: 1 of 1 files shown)


maximkrouk:
Probably that should fix this issue https://github.com/pointfreeco/swift-composable-architecture/issues/625

stephencelis:
@maximkrouk I think that issue may be unrelated. This seems like a reasonable change of behavior, though! @mbrandonw and I will discuss soon!

maximkrouk:
As an addition I'd like to mention, that I tried to delay an update for 0.1-0.2s, but it wasn't enough for `viewStore.state` to update, so inability to get a real state of the system is an unexpected behavior for me and it also disables the benefit from using `CurrentValueSubject` instead of `@Published` for Store.state

iampatbrown:
@maximkrouk can you try with the changes from [pull/#624](https://github.com/pointfreeco/swift-composable-architecture/pull/624) and let me know if that works on your end.

stephencelis:
@iampatbrown We're gonna merge this, since it's been a long-standing complaint in behavior. We'll tweak it in another PR so that it should be compatible with the work on your branch. Thanks again, @maximkrouk!

stephencelis (approved):

clayellis-cricut:
This change is setting off red flags for me and I just want to make sure that the reasoning of the `willSet` semantics of `ObservableObject` were considered before making this change (I'm sure they have, just being cautious) as this could potentially have an impact on view update performance. The entire thread is worth reading, but this post in particular: https://forums.swift.org/t/is-this-a-bug-in-published/31292/39

mbrandonw:
@clayellis-cricut Thanks for brining this up! You are totally right. We moved a little too fast on this 😬.

We were able to get a failing test that proves that we are not calling `objectWillChange` at the right time:

```swift
func testWillSet() {
  let reducer = Reducer<Int, Void, Void> { count, _, _ in
    count += 1
    return .none
  }

  let store = Store(initialState: 0, reducer: reducer, environment: ())
  let viewStore = ViewStore(store)

  var results: [Int] = []

  viewStore.objectWillChange
    .sink { _ in results.append(viewStore.state) }

  viewStore.send(())
  viewStore.send(())
  viewStore.send(())

  // ❌ XCTAssertEqual failed: ("[0, 1, 2]") is not equal to ("[1, 2, 3]")
  XCTAssertEqual([0, 1, 2], results) 
}
```

However, coincidentally, the changes in @iampatbrown's PR https://github.com/pointfreeco/swift-composable-architecture/pull/624 actually fix this issue in addition to what that PR aims to solve (that of too many equality checks). So we are going to get that PR fixed up and merged into main soon.

Luckily we haven't cut a release yet so no one should have pulled in this code (though to be fair we aren't sure how to make SwiftUI do something wrong from flipping `willSet` with `didSet`, but I'm sure there is some way).

tgrapperon:
> we aren't sure how to make SwiftUI do something wrong from flipping willSet with didSet

I don't know if it answers your question, but in one of this year's WWDC videos ([Discover concurrency in SwiftUI](https://developer.apple.com/videos/play/wwdc2021/10019/) I guess), they explained how `willSet` is used to trigger a snapshot of the state that will be compared with the current state on the next tick to decide if the view should update.
 If `didSet` is used to trigger `objectWillChange`, snapshot may occur too late and SwiftUI may not spot state changes.

mbrandonw:
@tgrapperon yeah, I definitely know we are not supposed to do it like that, but I don't know of a sequence of steps to show an actual problem in a SwiftUI application.

PR #639, opened 2021-07-08
Fix websocket case study and FactClient.

mbrandonw:
Two small things I found while testing some of the new case path stuff:

* The web socket case study has an effect that needs to a `.recieve(on:)`
* The failing `FactClient` should be wrapped in `#if DEBUG`.


Examples/CaseStudies/SwiftUICaseStudies/FactClient.swift
```diff
@@ -29,12 +29,14 @@ extension FactClient {
     })
 }
 
+#if DEBUG
 extension FactClient {
   // This is the "unimplemented" fact dependency that is useful to plug into tests that you want
```
stephencelis, line 34:
```suggestion
  // This is the "failing" fact dependency that is useful to plug into tests that you want
```

stephencelis (approved):

PR #641, opened 2021-07-08
Deprecate array-based forEach/ForEachStore

stephencelis:
The array-based versions of `forEach` and `ForEachStore` are pretty problematic in general. They can lead to a lot of bugs around long-living effects if elements are moved or removed from the collection. But perhaps worse, we've found that `ForEach` can evaluate when it shouldn't, which can lead to crashes when indexing into an array at a stale offset.

Let's finally deprecate these APIs in favor of identified array.


mbrandonw (approved):
The time has come...

PR #643, opened 2021-07-08
Add Benchmarking Target

stephencelis:
This adds a barebones benchmarking target, which has been helpful to run, but being off on its own branch made it less convenient. There's plenty to decide in the future as to how we want to be benchmarking things, and @tgrapperon has some promising experiments here: https://github.com/pointfreeco/swift-composable-architecture/discussions/618

In the interest of making a lil progress, though, let's get a baseline into `main`, and we can discuss the above more soon.


mbrandonw (approved):
Nice.

What do you think about adding a new CI step too to make sure benchmarks compile?

PR #644, opened 2021-07-09
Added custom swapAt to IdentifiedArray

### LINKED ISSUE #606: swapAt breaks IdentifiedArray's state invariant
**Describe the bug**

`swapAt` breaks` IdentifiedArray`'s state invariant, so that two "identical" `IdentifiedArray` are no longer equal

**To Reproduce**

Copy-paste the provided test to `IdentifiedArrayTests.swift` file:

```swift
    func testSwap() {
        struct User: Equatable, Identifiable {
            let id: Int
            var name: String
        }

        var swapped = IdentifiedArrayOf<User>([
            User(id: 0, name: "Blob"),
            User(id: 1, name: "Blob Jr."),
        ])

        swapped.swapAt(0, 1)

        let recreated = IdentifiedArrayOf<User>([
            User(id: 1, name: "Blob Jr."),
            User(id: 0, name: "Blob"),
        ])

        XCTAssertEqual(swapped, recreated)
        XCTAssertEqual(swapped.ids, recreated.ids)
    }
```

**Expected behavior**
`IdentifiedArray`s should be equal

**Environment**
 - Xcode 13 / 12.5
 - Swift 5.5 / 5.4
 - OS iOS 15 / macOS 11.4

**Additional context**

`XCTAssertEqual failed: ("[0, 1]") is not equal to ("[1, 0]")` in logs states that `ids` are out of order.

Internal [`func swapAt` | swift/MutableCollection.swift](https://github.com/apple/swift/blob/main/stdlib/public/core/MutableCollection.swift#L264-L270) updates values via index subscripts:

```swift
@inlinable
public mutating func swapAt(_ i: Index, _ j: Index) {
  guard i != j else { return }
  let tmp = self[i]
  self[i] = self[j]
  self[j] = tmp
}
```

And the [current subscript implementation | IdentifiedArray.swift](https://github.com/pointfreeco/swift-composable-architecture/blob/main/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift#L93-L97) never updates the `ids` after modification: 

```swift
  public subscript(position: Int) -> Element {
    // NB: `_read` crashes Xcode Preview compilation.
    get { self.dictionary[self.ids[position]]! }
    _modify { yield &self.dictionary[self.ids[position]]! }
  }
```

I wonder what is the proper fix here? Possibilities that instantly come to mind are:

- provide a custom `swapAt` function
- update the `ids` field after each `_modify`

Maybe it would be nice to migrate to `OrderedDictionary` under the hood in the long term run?

iampatbrown:
Fixes #606

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift
@@ -218,6 +218,10 @@ where ID: Hashable {
     self.shuffle(using: &rng)
   }
 
+  public mutating func swapAt(_ i: Index, _ j: Index) {
+    self.ids.swapAt(i, j)
+  }
+
   public mutating func reverse() {
     ids.reverse()
   }
--- a/Tests/ComposableArchitectureTests/IdentifiedArrayTests.swift
+++ b/Tests/ComposableArchitectureTests/IdentifiedArrayTests.swift
@@ -227,6 +227,28 @@ final class IdentifiedArrayTests: XCTestCase {
       XCTAssertEqual([1, 3, 5, 4, 2], array.ids)
     }
   #endif
+  
+  func testSwapAt() {
+    struct User: Equatable, Identifiable {
+      let id: Int
+      var name: String
+    }
+
+    var swapped = IdentifiedArrayOf<User>([
+      User(id: 0, name: "Blob"),
+      User(id: 1, name: "Blob Jr."),
+    ])
+
+    swapped.swapAt(0, 1)
+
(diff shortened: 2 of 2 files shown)


stephencelis:
@iampatbrown Thanks for taking the time to PR. We were in the middle of prepping https://github.com/pointfreeco/swift-identified-collections to solve this problem (and more), so we're going to close in favor of that. One thing we were struggling with with conformances like these is that even though they work when called directly, they break when used in a completely generic context 😞 Hopefully the new approach avoids that, though!

iampatbrown:
Identified Collections looks like a great addition :)

PR #652, opened 2021-07-12
Store publisher should retain its view store

### LINKED ISSUE #651: ViewStore.publisher doesn't seem to work as expected in 0.21.0
**Describe the bug**
I can see from the release notes that 0.21.0 changes the behaviour of ViewStore.publisher but I can't see how it is causing the issue I'm seeing. For some reason it seems to publish just a single change even though I know the state is changing.

We have some integration tests that need to log the user in to our app - we do this by creating a `ViewStore` around a real store with some live dependencies and sending actions that simulate the process of a user logging in. We then sink on the `ViewStore.publisher` around our root App Store to wait for the necessary state change that would indicate that the user is logged in. This worked fine with 0.20.0 and no longer works with 0.21.0.

**To Reproduce**
A simplified representation of our app state:

```swift
struct AppState: Equatable {
  var session: SessionState
}

enum SessionState: Equatable {
  case loggedOut(LoggedOutState)
  case loggedIn(LoggedInState)
}
```

The login functionality is managed inside the `LoggedOut` domain. Our integration test code does something similar to this:

```swift
let loggedOutStore = app.appStore.scope(
  state: { extract(case: SessionState.loggedOut, from: $0.session)! },
  action: { AppAction.session(.loggedOut($0)) }
)

let viewStore = ViewStore(loggedOutStore)

// send a bunch of actions to viewStore that simulate the user entering their login details and
// hitting the login button. Tapping the login button will start a bunch of async effects that make
// various API requests.

// Now wait for the `AppState.session` to flip from `.loggedOut` to `.loggedIn`:

ViewStore(store)
  .publisher
  .compactMap { extract(case: SessionState.loggedIn, from: $0.session) }
  .prefix(1)
  .sink(receiveValue: callback) // callback is passed into the helper function that encapsulates this behaviour
  .store(in: &cancellables)
```

**Expected behavior**
In 0.20.0, when the login completes the session state changes to `.loggedIn` and this triggers the `callback` function above to be executed. In 0.21.0 this never happens. In fact, if I simplify the above to:

```swift
ViewStore(store)
  .publisher
  .sink { print($0) }
  .store(in: &cancellables)
```

Then the `.sink` is only called once - I don't see any intermediate state changes (and there should be quite a few changes within the `loggedOutState` while the login process happens).

**Environment**
  - TCA 0.21.0
  - Xcode 12.5.1
  - OS: iOS 14.

stephencelis:
Fixes #651.

CC: @lukeredpath


mbrandonw (approved):

lukeredpath:
I'm happy to try this out on my app/0.21.0 integration branch if you want before merging?

mbrandonw:
@lukeredpath That'd be great, let us know how it goes!

lukeredpath:
I've just updated my branch to use this bug-fix branch and reverted my workaround and my tests are still passing. 👍🏻 🚢

stephencelis:
Awesome! We'll merge and assuming nothing else comes up in the next day or so, cut another release soon.

lukeredpath:
@stephencelis any plans for a 0.21.1 release with this bug fix? We're shipping our first app release with TCA next week and it would be great to get this version update in before we ship. Thanks.

stephencelis:
Hot off the presses: https://github.com/pointfreeco/swift-composable-architecture/releases/0.22.0 😄

PR #653, opened 2021-07-12
Adds convenience pullback

ohitsdaniel:
We mix and match enum and struct states in our codebase. 

In order to extract a enum state from a struct state, we currently have to combine to pullbacks:

```swift
enumStateReducer.pullback(
  state: /EnumState.aCase,
  action: /GlobalAction.enumAction,
  environment: \.localEnvironment
)
.pullback(
  state: \.enumState, 
  action: .`self`, 
  environment: { $0 }
)
```

This PR adds a convenience pullback that allows to combine a WritableKeyPath and a CasePath to first extract a field of EnumState and then extract LocalState from it using a CasePath.

```swift
enumStateReducer.pullback(
    state: \.enumState,
    matching: /EnumState.aCase,
    action: /GlobalAction.enumAction,
    environment: \.localEnvironment
)
```

### DIFF
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -512,6 +512,109 @@ public struct Reducer<State, Action, Environment> {
     }
   }
 
+  /// Transforms a reducer that works on local state, action, and environment into one that works on
+  /// global state, action and environment. It accomplishes this by providing 4 transformations to
+  /// the method:
+  ///
+  ///   * A writable key path that can get/set a piece of enum state from the global state.
+  ///   * A case path that can extract/embed a local state into a enum state.
+  ///   * A case path that can extract/embed a local action into a global action.
+  ///   * A function that can transform the global environment into a local environment.
+  ///
+  /// This operation is important for breaking down large reducers into small ones. When used with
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hey @ohitsdaniel, glad you are getting use out of case path pullbacks! 🙂

Rather than defining another overload for `pullback` we actually have plans to once and for all unite key paths and case paths into a single package. We've been long working on the API for this thing, and while we have more work to do, its getting close. Once that is done you will be able to accomplish what you want quite easily with a single pullback.

So, I would say in the meantime you should define this overload in your own projects, and hopefully we will work out the last remaining issues on our solution soon.

ohitsdaniel:
Hey @mbrandonw, thanks for the quick response!

Looking forward to what you'll have in store for Key / CasePaths. 

If there's anything I can help with, feel free to [send me a message](https://twitter.com/Oh_Its_Daniel).

I'll close this PR for now as we probably won't need it long-term.

mbrandonw:
Hi @gtranchedone, unfortunately nothing officially supported right now, but we are working to refine the APIs to make them ready for the public. We just need more time.

In the meantime I say continue using the extension you created because our final version probably won't look _that_ much different, and so should be easy to port over if you want to later.

PR #655, opened 2021-07-13
Use in-house benchmarking framework instead of swift-benchmark

tgrapperon:
Hello!

I was unhappy with the the Google's `swift-benchmark` library:
- It doesn't allow automatic comparison of benchmarks, and the results are not exposed within the benchmark app.
- It uses `DispatchTime.now` to timestamp events, which can be unreliable according to Stack Overflow.
- It looks semi-archived, with a lonely PR sitting for more than 6 months without any acknowledgement from the maintainers.

I've drafted this library today to try to fix these issues.
Firstly, this library is source compatible with `swift-benchmark`. In `swift-parsing`, one can remove the `swift-benchmark` dependency and create a `Benchmark` target with this code and it "just works" (there is only one change to make: annotate one of the internal function helper as `mutating` as `BenchmarkSuite` is a now a struct)

The library introduces an intermediate group called `BenchmarkStudy` which compares some alternative cases to a baseline one. The hierarchy is `Suite`>`Study`>`Case`. Each study is independent from another. 

Each case is instantiated with a block receiving a `(MeasureDelimiter) -> Void` argument. You can use this function with `MeasureDelimiter`'s cases `.start` or `.stop` to optionally delimit the portion of code that is measured. There is no setup or tear-down as the delimiters allow to execute setup, benchmarks and tear-down in the same block. The use of the delimiters is optional and the library uses the block boundaries if they're absent.

As an example, the updated code that take advantage of the studies for TCA is:
```swift
benchmarkSuite("Scoping") { suite in
  
  suite.addStudy("Deep scoping") { study in
    
    study.setBaseline("No scope") { measure in
      let store1 = Store(initialState: 0, reducer: counterReducer, environment: ())
      let viewStore = ViewStore(store1)
      measure(.start)
      viewStore.send(true)
    }
    
    study.addCase("Scope (1)") { measure in
      let store1 = Store(initialState: 0, reducer: counterReducer, environment: ())
      let store2 = store1.scope { $0 }

      let viewStore = ViewStore(store2)
      measure(.start)
      viewStore.send(true)
    }
    
    study.addCase("Scope (2)") { measure in
      let store1 = Store(initialState: 0, reducer: counterReducer, environment: ())
      let store2 = store1.scope { $0 }
      let store3 = store2.scope { $0 }

      let viewStore = ViewStore(store3)
      measure(.start)
      viewStore.send(true)
    }
    
    study.addCase("Scope (3)") { measure in
      let store1 = Store(initialState: 0, reducer: counterReducer, environment: ())
      let store2 = store1.scope { $0 }
      let store3 = store2.scope { $0 }
      let store4 = store3.scope { $0 }

      let viewStore = ViewStore(store4)
      measure(.start)
      viewStore.send(true)
    }
  }
}
```
Each benchmark case is run by default 1,000,000 times or during 5s, whichever comes first. Parameters are always shared in a study.

The library runs all the cases of all the studies of all the suites and present the results in a human friendly way. As an example, the current benchmark for TCA shows:
```
== Scoping ==================================================================
               Duration Difference Variation Performance Std.Error Iterations
-- Deep scoping -------------------------------------------------------------
No scope  (*) 12.738 µs                                    ±0.12 %     276000
Scope (1)     22.478 µs  +9.740 µs  +76.46 %       x0.57   ±0.11 %     159270
Scope (2)     31.876 µs +19.138 µs +150.24 %       x0.40   ±0.13 %     113743
Scope (3)     42.199 µs +29.461 µs +231.29 %       x0.30   ±0.15 %      86687
=============================================================================
```
The asterisk `(*)` denotes the fastest performer. The first row is empty for Difference, Variation (in %), and Performance (in "times faster") because it's the baseline. The time unit is automatically determined according to the data to be displayed (it is the same for all the cases of a study).

On the technical side, the library uses `clock_gettime_nsec_np` to retrieve a timestamp, which is the Apple's recommendation. It also uses `Accelerate` if available to compute the mean and standard deviation of the results.

More importantly, the library is already set to execute `transversal` benchmarks in the spirit `swift-collections-benchmark`. You can parametrize the execution block with a generic parameter (let say the number of pullbacks), and the library will execute the same sequence described above, but also varying the parameter according to your prescription. This is not really exposed right now as I don't now how to present the results in ASCII, but I guess it should be possible to draw graph images at some point. The currently exposed non-parametric API is hiding the generic parameter behind a typealias.

I guess this module should be extracted as an independent library (it also misses `@inlinable` annotations for now). If there is some interest, I can update `swift-parsing` benchmarks to take advantage of the studies. I've tested it at home and it gives similar results as `swift-benchmark`. In source compatibility mode, it runs all the benchmarks as baseline-only studies, but it can be way nicer if some proper studies are defined. For example, here is a partial result running `Comma separated` benchmarks in the `Numeric` suite, both as a proper study and in source compatibility:

```
== Numerics ============================================================================================
                                          Duration Difference Variation Performance Std.Error Iterations
-- Comma separated -------------------------------------------------------------------------------------
Double.parser                       (*)  13.772 ms                                    ±1.37 %        360
Scanner.scanDouble                      111.553 ms +97.781 ms +710.02 %       x0.12   ±0.71 %         45
String.split                             36.388 ms +22.616 ms +164.22 %       x0.38   ±0.92 %        137
--------------------------------------------------------------------------------------------------------
Comma separated: Double.parser           13.898 ms                                    ±1.39 %        357
Comma separated: Scanner.scanDouble     113.646 ms                                    ±0.57 %         44
Comma separated: String.split            36.877 ms                                    ±0.89 %        136
========================================================================================================
```
and running using `swift-benchmark`:
```
name                                         time             std        iterations
-----------------------------------------------------------------------------------
Numerics.Comma separated: Double.parser       12137402.000 ns ±  30.29 %        107
Numerics.Comma separated: Scanner.scanDouble 113536294.000 ns ±   4.45 %         13
Numerics.Comma separated: String.split        36109335.000 ns ±  10.55 %         39
```

Most importantly, I guess this library will provide a more flexible base to benchmark TCA. I have no issue with @mbrandonw  and @stephencelis taking ownership of the code. This is a very rough draft and I'm sure there is many ways to improve it.

### DIFF
--- a/.swiftpm/xcode/xcshareddata/xcschemes/swift-composable-architecture-benchmark.xcscheme
+++ b/.swiftpm/xcode/xcshareddata/xcschemes/swift-composable-architecture-benchmark.xcscheme
@@ -28,6 +28,16 @@
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
       shouldUseLaunchSchemeArgsEnv = "YES">
       <Testables>
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "ComposableArchitectureTests"
+               BuildableName = "ComposableArchitectureTests"
+               BlueprintName = "ComposableArchitectureTests"
+               ReferencedContainer = "container:">
+            </BuildableReference>
+         </TestableReference>
       </Testables>
    </TestAction>
    <LaunchAction
--- a/Package.swift
+++ b/Package.swift
@@ -17,7 +17,6 @@ let package = Package(
     )
   ],
   dependencies: [
-    .package(name: "Benchmark", url: "https://github.com/google/swift-benchmark", from: "0.1.0"),
     .package(url: "https://github.com/pointfreeco/combine-schedulers", from: "0.5.0"),
     .package(url: "https://github.com/pointfreeco/swift-case-paths", from: "0.4.0"),
     .package(url: "https://github.com/pointfreeco/swift-identified-collections", from: "0.1.0"),
@@ -42,8 +41,7 @@ let package = Package(
     .target(
       name: "swift-composable-architecture-benchmark",
       dependencies: [
-        "ComposableArchitecture",
-        .product(name: "Benchmark", package: "Benchmark"),
+        "ComposableArchitecture"
       ]
     ),
   ]
--- a/Sources/swift-composable-architecture-benchmark/Benchmark/Benchmark+Compat.swift
+++ b/Sources/swift-composable-architecture-benchmark/Benchmark/Benchmark+Compat.swift
@@ -1,0 +1,46 @@
+import Foundation
+
+public func benchmark(_ label: String, block: @escaping () -> Void) {
+  let benchmark = BenchmarkCase(label: label,
+                                block: BenchmarkBlock(block: { _ in block() }))
+  let study = BenchmarkStudy(name: label, baseline: benchmark, cases: [], configuration: .init())
+  defaultBenchmarkSuite.studies.append(study.erased)
+}
+
+public var defaultBenchmarkSuite = BenchmarkSuite(name: "Default Suite", studies: [])
+
+public extension BenchmarkSuite {
+  mutating func benchmark(_ label: String, block: @escaping () -> Void) {
+    let benchmark = BenchmarkCase(label: label,
+                                  block: BenchmarkBlock(block: { _ in block() }))
+    let study = BenchmarkStudy(name: label, baseline: benchmark, cases: [], configuration: .init())
+    studies.append(study.erased)
+  }
+}
+
+public extension BenchmarkSuite {
+  mutating func register(benchmark: AnyBenchmark) {
+    let benchmarkCase = BenchmarkCase(label: benchmark.name, block: .init(block: { measure in
+      benchmark.setUp()
+      var state = BenchmarkState()
+      measure(.start)
+      try! benchmark.run(&state)
+      measure(.stop)
+      benchmark.tearDown()
+
+    }))
+    studies.append(
+      BenchmarkStudy(name: benchmark.name, baseline: benchmarkCase, cases: [], configuration: .init()).erased
+    )
+  }
+}
+
+public struct BenchmarkSetting {}
+public struct BenchmarkState {}
+public protocol AnyBenchmark {
+  var name: String { get }
+  var settings: [BenchmarkSetting] { get }
+  func setUp()
+  func run(_ state: inout BenchmarkState) throws
+  func tearDown()
+}
--- a/Sources/swift-composable-architecture-benchmark/Benchmark/Benchmark.swift
+++ b/Sources/swift-composable-architecture-benchmark/Benchmark/Benchmark.swift
@@ -1,0 +1,14 @@
+import Foundation
+
+public enum Benchmark {
+  public static var suites: [BenchmarkSuite] = [defaultBenchmarkSuite]
+  public static func main(_ suites: [BenchmarkSuite] = Self.suites) {
+    var reports = String()
+    for suite in suites {
+      let report = suite.execute()
+      reports.append(report.formattedString() + "\n")
+    }
+    print("Benchmark Results:")
+    print(reports)
+  }
+}
--- a/Sources/swift-composable-architecture-benchmark/Benchmark/BenchmarkBlock.swift
+++ b/Sources/swift-composable-architecture-benchmark/Benchmark/BenchmarkBlock.swift
@@ -1,0 +1,50 @@
+import Foundation
+
+public enum MeasureDelimiter {
+  case start
+  case stop
+}
+
+public struct ParametricBenchmarkBlock<Parameter> {
+  public init(block: @escaping (_ measure: (MeasureDelimiter) -> Void,
+                                _ parameter: Parameter) -> Void)
+  {
+    self.block = block
+  }
+
+  let block: (_ measure: (MeasureDelimiter) -> Void, _ parameter: Parameter) -> Void
+
+  func execute(parameter: Parameter) -> TimeInterval {
+    var explicitStartTime: UInt64?
+    var explicitEndTime: UInt64?
+    let measure: (MeasureDelimiter) -> Void = {
+      switch $0 {
+      case .start:
+        guard explicitStartTime == nil else {
+          fatalError("`measure(.start)` was called more than once.")
+        }
+        explicitStartTime = timestampInNanoseconds()
+      case .stop:
+        guard explicitEndTime == nil else {
+          fatalError("`measure(.stop)` was called more than once.")
+        }
+        explicitEndTime = timestampInNanoseconds()
+      }
+    }
+    let implicitStartTime = timestampInNanoseconds()
+    block(measure, parameter)
+    let implicitEndTime = timestampInNanoseconds()
+
+    let startTime = explicitStartTime ?? implicitStartTime
+    let endTime = explicitEndTime ?? implicitEndTime
+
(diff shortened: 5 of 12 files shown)


stephencelis:
Whoa! This is really impressive stuff, @tgrapperon! Thanks for sharing 😄

You bring up a lot of good points. swift-benchmark is definitely less-than-ideal in many ways, and it does appear to be a shelved project for now. You've certainly improved upon a lot of its deficiencies.

We share your hunch that it probably doesn't make sense to hide this code away in the TCA repo, as we benchmark several of our libraries with the same tool (CasePaths, Parsing), so it would make sense to have a single package that each library depends on.

After discussing with @mbrandonw I think we're not quite ready to take on the work associated with refining and maintaining our own benchmarking library. We may in the future, though, and we'd certainly return to this draft as a starting point! Even though swift-benchmark is far from perfect, I think it serves our needs in the meantime, but if another library surfaces, we will be open to making a switch when we have the bandwidth.

If you decide to open source this code as your own package, we'll try to offer feedback, and we'll keep our eyes on it!

tgrapperon:
Thanks @stephencelis!

I'll think again about the topic and I'll open source a draft library on my side then. I don't plan to maintain a fully fledged oss benchmarking library neither, but I can shelter it until it reaches a point where it can be incorporated into your projects. I'll keep you posted.

I think one can abstract the kind of result of each benchmark case to something more general than elapsed time, and it would be quite powerful to measure arbitrary values (like the number of time a ViewStore publisher emits for example). 

Do you want me to close the PR now, or do we wait for my hosted version to be published?

stephencelis:
Since there's no plan to merge it at this time I think closing is fine. If you'd like more eyes on it, though, definitely feel free to open a GitHub discussion with a link here!

tgrapperon:
Hey @mbrandonw and @stephencelis! I've extracted the benchmarking library and just published [`swift-measure`](https://github.com/tgrapperon/swift-measure).

I completely refactored the code and the library can now benchmark other things than the time spent. Many things were abstracted and it should be ultimately possible to measure the number of time a store does something on device at some point, and present the result in the form of a SwiftUI sheet for example. 

Benchmarks (or measures) have a chain based API. For example, the standard benchmark block that computes the time spent to execute something is built as:
```swift
measureTime(label, tag: tag, block) // Measure the time spent for one execution, outputs TimeInterval
  .repeat(iterations: iterations, duration: duration) // repeat the same operation, outputs [TimeInterval]
  .map { $0.extractMeasure() } // Extract the mean and standard error from [TimeInterval], outputs Measure<TimeInterval>? (the array can be empty
  .unwrap(default: 0.0) // Unwraps the optional measure, producing a final Measure<TimeInterval>
```
There are other higher-order blocks like `forEach` which repeat a measure over a sequence of inputs (this can be useful to generate reports like `swift-collections-benchmark` for example), and the user can build its own setups. Most of them are defined at https://github.com/tgrapperon/swift-measure/blob/main/Sources/Measure/Blocks/Block.swift
There is an example of a custom measure at https://github.com/tgrapperon/swift-measure/blob/92770892d1c794ccf04f289fdc1a9d00572d0353/Tests/BenchmarkTests/BenchmarkTests.swift#L44 and below.

I've also updated the benchmarks of [`swift-case-paths`](https://github.com/tgrapperon/swift-case-paths/tree/swift-measure) and  [`swift-parsing`](https://github.com/tgrapperon/swift-parsing/tree/swift-measure)  to take advantage of the library. I'm not doing any PR right now, but you can have a look if you want.

Here is the kind of output generated by the library when it outputs strings to the console:
```
swift-case-paths
== Extraction =======================================================================
                          Value  Difference Variation  Performance  Error  Iterations
-- Success --------------------------------------------------------------------------
Manual                (*)  53 ns                                   ±0.20 %    1000000
Reflection                127 ns     +75 ns  +141.58 %       x0.41 ±0.09 %    1000000
Reflection (uncached)     703 ns    +650 ns +1233.12 %       x0.08 ±0.05 %    1000000
-- Failure --------------------------------------------------------------------------
Manual                (*)  57 ns                                   ±0.26 %    1000000
Reflection                 95 ns     +38 ns   +66.08 %       x0.60 ±0.14 %    1000000
Reflection (uncached)     658 ns    +600 ns +1048.15 %       x0.09 ±0.07 %    1000000
=====================================================================================
```
or 
```
swift-parsing
== Numerics ===================================================================
                   Value    Difference Variation Performance  Error  Iterations
-- Int ------------------------------------------------------------------------
Int.init     (*)      73 ns                                  ±0.75 %    1000000
Parser                78 ns      +5 ns   +6.54 %       x0.94 ±0.26 %    1000000
-- Comma separated - Int ------------------------------------------------------
String.split     105.182 ms                                  ±0.46 %         48
Parser       (*)   5.278 ms -99.904 ms  -94.98 %      x19.93 ±0.16 %        937
Scanner           73.178 ms -32.004 ms  -30.43 %       x1.44 ±0.32 %         69
-- Double ---------------------------------------------------------------------
Double.init  (*)     101 ns                                  ±0.18 %    1000000
Parser               140 ns     +39 ns  +38.25 %       x0.72 ±0.15 %    1000000
Scanner              527 ns    +426 ns +422.26 %       x0.19 ±3.56 %    1000000
-- Comma separated - Double ---------------------------------------------------
String.split      29.038 ms                                  ±1.35 %        172
Parser       (*)   9.768 ms -19.270 ms  -66.36 %       x2.97 ±0.27 %        508
Scanner           88.261 ms +59.223 ms +203.95 %       x0.33 ±0.54 %         57
===============================================================================

== README Example =================================================================
                        Value   Difference Variation Performance  Error  Iterations
-- README Example -----------------------------------------------------------------
Adhoc                  7.498 µs                                  ±0.03 %     624168
Parser: Substring      3.831 µs  -3.667 µs  -48.91 %       x1.96 ±0.04 %    1000000
Parser: UTF8      (*)  1.484 µs  -6.014 µs  -80.20 %       x5.05 ±0.04 %    1000000
Scanner               18.708 µs +11.210 µs +149.50 %       x0.40 ±0.03 %     258062
===================================================================================

== String Abstractions ========================================================
                 Value     Difference  Variation Performance  Error  Iterations
-- String Abstractions --------------------------------------------------------
Substring     1194.212 µs                                    ±0.21 %       4185
UTF8      (*)  146.747 µs -1047.464 µs  -87.71 %       x8.14 ±0.05 %      33968
===============================================================================
```

If you have the time to have a look, please let me know what you think!
Thanks!

tgrapperon:
I've just updated the library to show how to build as study that count the number of time a comparison is realized when sorting an array using two different algorithms. The study shows the results for a range of array's sizes, and the time spent could have been measured as easily. Everything was setup from the client side, with the tools provided by the library.

It's quite astonishing what a functional architecture can achieve in a 4 days solo side project. 
I think there's room for a leaner and more powerful API though.

The report generated is in the form of 
```
== Sorting algorithms ========================================================================================
                   Value      Difference    Variation  Performance  Error  Iterations
-- Time elapsed ----------------------------------------------------------------------------------------------
Swift sort  (*)   573.868 µs                                       ±0.11 %       8683
Bubble sort     98905.897 µs +98332.029 µs +17134.96 %       x0.01 ±0.47 %         51
--------------------------------------------------------------------------------------------------------------
                Comparisons 10 100   250   500    1000   2500     5000    10000    12500     15000     20000  
-- Number of comparisons -------------------------------------------------------------------------------------
Swift Sort  (*)      17.5×N 36 1419  4521  10123  20070   41262    88330   186401   259974    351550    392616
Bubble Sort        3038.6×N 45 4950 31125 124750 499500 3123750 12497500 49995000 78118750 112492500 199990000
==============================================================================================================
```
and the code to set it up can be found at https://github.com/tgrapperon/swift-measure/blob/main/Tests/BenchmarkTests/BenchmarkTests.swift

Issue #661, opened 2021-07-16
Bug: `BufferedActions` hanging action


heyltsjay:
`BufferedActions` will queue, but not handle, actions that come in while it's in the process of sending. This results in the action dangling in the buffer indefinitely until some other action comes through. This is a regression from 0.19.0

**To Reproduce**
[Buffer Bug.zip](https://github.com/pointfreeco/swift-composable-architecture/files/6831026/Buffer.Bug.zip)

1. Run the example project
1. Press the button on `ViewController`
1. Note that `viewStore.send(.didLoad)` is executed as expected in ViewControllerB
1. Note that `.didLoad` is not handled by either reducer
1. Dismiss ViewControllerB
1. Note that `.didLoad` and `didDismiss` are handled by reducer in succession.

**Expected behavior**
`.didLoad` should be processed by the reducer on the runloop it was sent (or at least in a timely manner)

**Environment**
  - swift-composable-architecture version >= 0.20.0 (not an issue in 0.19.0)
  - Xcode 12.5

mbrandonw:
This must be a consequence of https://github.com/pointfreeco/swift-composable-architecture/pull/619 where we combined the notions of "synchronous actions" and "buffered actions". We thought it was logically equivalent, but perhaps not :/

Would you be able to get us a failing test so that we can look into a fix?

heyltsjay:
Sure, added in #662

stephencelis:
Fixed in #662, thanks!

PR #662, opened 2021-07-16
Buffered Actions Failing Test

heyltsjay:
- Adds a failing test demonstrating #661


stephencelis:
Wow, that was a subtle bug! I think I tracked down a fix, though, and pushed it here. Thanks for the test! If things are green I'll merge 😄

PR #663, opened 2021-07-16
Alternate `ViewStore.binding` helper

stephencelis:
It occurred to me we could just leverage dynamic member lookup and derive view store bindings given simply a binding action...there's a caveat, though.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Forms.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Forms.swift
@@ -56,7 +56,7 @@ struct BindingFormView: View {
           HStack {
             TextField(
               "Type here",
-              text: viewStore.binding(keyPath: \.text, send: BindingFormAction.binding)
+              text: viewStore.binding(BindingFormAction.binding).text
             )
             .disableAutocorrection(true)
             .foregroundColor(viewStore.toggleIsOn ? .gray : .primary)
@@ -64,12 +64,12 @@ struct BindingFormView: View {
           }
           .disabled(viewStore.toggleIsOn)
 
-          Toggle(isOn: viewStore.binding(keyPath: \.toggleIsOn, send: BindingFormAction.binding)) {
+          Toggle(isOn: viewStore.binding(BindingFormAction.binding).toggleIsOn) {
             Text("Disable other controls")
           }
 
           Stepper(
-            value: viewStore.binding(keyPath: \.stepCount, send: BindingFormAction.binding),
+            value: viewStore.binding(BindingFormAction.binding).stepCount,
             in: 0...100
           ) {
             Text("Max slider value: \(viewStore.stepCount)")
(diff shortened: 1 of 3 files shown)


Sources/ComposableArchitecture/SwiftUI/Binding.swift
```diff
@@ -258,4 +258,29 @@ extension ViewStore {
...
+  /// TextField(
+  ///   "Enter text",
+  ///   text: viewStore.binding(Action.binding).text
+  /// )
+  /// ```
+  ///
+  /// - Parameter action: A function that wraps a binding action in the view store's action type.
+  /// - Returns: A binding.
+  public func binding(_ action: @escaping (BindingAction<State>) -> Action) -> Binding<State>
+  where State: Equatable {
```
stephencelis, line 280: All of state needs to be equatable to return a `Binding` here. This is a common enough practice in TCA apps, but an obstacle nonetheless.

We could overcome that by having a `ViewStore.Binding` helper that has dynamic member lookup over equatable sub-state, instead, but wanted to start a convo first to see if it's worth introducing that type just to clean up call sites a lil bit.

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Forms.swift
```diff
@@ -56,20 +56,20 @@ struct BindingFormView: View {
           HStack {
             TextField(
               "Type here",
-              text: viewStore.binding(keyPath: \.text, send: BindingFormAction.binding)
+              text: viewStore.binding(BindingFormAction.binding).text
```
stephencelis, line 59: It's decently shorter, so might be worth it. It also reads a bit jumbly right now: a lot of `binding.binding` noise. I wonder if we can come up with a less repetitive call site.

Sources/ComposableArchitecture/SwiftUI/Binding.swift
mbrandonw: This is already the case with the current `.binding(keyPath:send:)` right?

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Forms.swift
mbrandonw: yeah I like this!

Sources/ComposableArchitecture/SwiftUI/Binding.swift
stephencelis: Nah that's only equatable on the `Value`. This requires equatable on `Root`.

iampatbrown:
Nice addition. I've always played around with the binding call site and never been able to come up with something that feels right. Keen to see where this leads :)

I tried something like this at one point but didn't really like the idea of using a protocol.
```swift
public protocol Bindable {
  associatedtype State
  static func binding(_: BindingAction<State>) -> Self
}

extension ViewStore where State: Equatable, Action: Bindable, Action.State == State {
  public var binding: Binding<State> { self.binding(Action.binding) }
}

extension Reducer where Action: Bindable, Action.State == State {
  public var binding: Self {  self.binding(action: /Action.binding) }
}
```
leading to 
```swift
viewStore.binding.text
```

stephencelis:
@iampatbrown Yup! @junebash brought that up that great idea in https://github.com/pointfreeco/swift-composable-architecture/discussions/370

I'm going to close this draft for now as we kick around a few more ideas, though :smile:

Issue #496, opened 2021-04-17
Nilling out child state in response to child's action replays old state in child store


dannyhertz:
**Describe the bug**
In a UIKit based app (have not tried SwiftUI but have a feeling its related to the UIKit IfLet functionality) if you you nil out a child state (in parent reducer) in response to a child's actions (which also changes some state) then child ViewStore publishers seem to emit older state values (after emitting the correct values first).

**To Reproduce**

[TcaCombineApp.zip](https://github.com/pointfreeco/swift-composable-architecture/files/6330158/TcaCombineApp.zip)

```swift
import UIKit
import ComposableArchitecture
import Combine

struct Environment {}

struct ChildState: Equatable {
    var alertText: String?
}

enum ChildAction: Equatable {
    case alertDismissed
    case dismissalRequested
    case didAppear
}

let childReducer = Reducer<ChildState, ChildAction, Environment> { state, action, environment in
    switch action {
    case .didAppear:
        state.alertText = "Hey!"
        return .none
    case .alertDismissed:
        state.alertText = nil
        return Effect(value: .dismissalRequested)
    case .dismissalRequested:
        return .none
    }
}

struct AppState: Equatable {
    var child: ChildState?
}

enum AppAction {
    case child(ChildAction)
    case buttonSelected
}

let appReducer: Reducer<AppState, AppAction, Environment> = .combine(
    childReducer.optional()
        .pullback(state: \.child, action: /AppAction.child, environment: { $0 }),
    Reducer<AppState, AppAction, Environment> { state, action, _ in
        switch action {
        case .child(.dismissalRequested):
            state.child = nil
            return .none
        case .buttonSelected:
            state.child = .init()
            return .none
        default:
            return .none
        }
    }
)

class ChildViewController: UIViewController {

    private var cancellables = Set<AnyCancellable>()
    private let store: Store<ChildState, ChildAction>
    private let viewStore: ViewStore<ChildState, ChildAction>

    init(store: Store<ChildState, ChildAction>) {
        self.store = store
        self.viewStore = ViewStore(store)
        super.init(nibName: nil, bundle: nil)
    }

    required init?(coder: NSCoder) {
        fatalError("init(coder:) has not been implemented")
    }

    override func viewDidLoad() {
        super.viewDidLoad()
        view.backgroundColor = .blue

        viewStore.publisher.alertText
            .sink(receiveValue: { [weak self] title in
                guard let self = self, let title = title else { return }
[...]

PR #665, opened 2021-07-19
Introduce `compactScope` operation for IfLetStore

ollitapa:
As discussed in #496, when using `IfLetStore`, an action from scoped child store to nil out local state might result in local state being replayed back to earlier value right before the view is removed. The issue can be seen in 

https://github.com/pointfreeco/swift-composable-architecture/blob/ba2b9e08e2776bbb02f27ab7123dae767f106fad/Sources/ComposableArchitecture/SwiftUI/IfLetStore.swift#L50-L54

where initial state is copied and used later when the scope transformation returns nil.

A `compactScope` operation to avoids this by not forwarding non-nil values to the child store. 

It's a bit awkward solution, because the implementation is nearly the same as regular `scope`. I did not want to sacrifice performance of regular `scope` by having it call `compactScope` internally.  Maybe a better implementation could be found that could combine both scoping methods.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -365,6 +365,63 @@ public final class Store<State, Action> {
     self.publisherScope(state: toLocalState, action: { $0 })
   }
 
+  /// Scopes the store to one that exposes optional local state, but only forwards non-nil values.
+  ///
+  /// - Parameters:
+  ///   - initialLocalState: Initial local state must be provided because `toLocalState` might produce a nil value.
+  ///   - toLocalState: A function that transforms `State` into optional `LocalState`.
+  ///   - fromLocalAction: A function that transforms `LocalAction` into `Action`.
+  /// - Returns: A new store with its domain (state and action) transformed. Only forwards non-nil states.
+  func compactScope<LocalState, LocalAction>(
+    initialLocalState: LocalState,
+    state toLocalState: @escaping (State) -> LocalState?,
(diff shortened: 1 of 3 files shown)


stephencelis:
@ollitapa Thanks for the PR and discussion! This is definitely an annoying pain point, currently. And members of the community have done some good experimentation here. @darrarski wrote about navigation in general awhile ago here, and specifically brought up the issue:

https://github.com/darrarski/darrarski/blob/main/2021/04/Thoughts-on-SwiftUI-navigation/README.md

In it @Thomvis also provides an elegant solution, which I'll reproduce here. You can write a helper transform function that caches the last `some` case of an optional:

```swift
extension Optional {
  static var lastWrappedValue: (Self) -> Self {
    var lastWrapped: Wrapped?
    return {
      lastWrapped = $0 ?? lastWrapped
      return lastWrapped
    }
  }
}
```

And you can then use this helper when you scope to `IfLetStore`:

```swift
IfLetStore(
  store.scope(state: Optional.lastWrappedValue),
  then: content
)
```

This should ensure that when you transition away from this view it won't flash blank or even back to the initial value.

We're experimenting with more ergonomic navigation tools to address this issue, but they're not quite there yet. They are, however, powered by this same helper.

ollitapa:
Ah, good! That seems a much nicer solution. I added another PR #667 that looks a lot cleaner. Maybe it can be added for the time being, while you work on better tools for navigation :+1:

sadiklar:
19 Tem 2021 Pzt 19:16 tarihinde Olli Tapaninen ***@***.***>
şunu yazdı:

PR #667, opened 2021-07-21
Fix IfLetStore so that it does not replay first seen state on nil-value

ollitapa:
A more elegant fix to problem in #496 suggested by @stephencelis in #665.

When using IfLetStore, an action from scoped child store to nil out local state might result in local state being replayed back to earlier value right before the view is removed. The issue can be seen in

https://github.com/pointfreeco/swift-composable-architecture/blob/ba2b9e08e2776bbb02f27ab7123dae767f106fad/Sources/ComposableArchitecture/SwiftUI/IfLetStore.swift#L50-L54

where initial state is copied and used later when the scope transformation returns nil.

This solution ensures that the last nil-value sent to the store will play back the last seen non-nil value.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/IfLetStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/IfLetStore.swift
@@ -51,8 +51,11 @@ public struct IfLetStore<State, Action, Content>: View where Content: View {
   ) where Content == _ConditionalContent<IfContent, ElseContent> {
     self.store = store
     self.content = { viewStore in
-      if let state = viewStore.state {
-        return ViewBuilder.buildEither(first: ifContent(store.scope(state: { $0 ?? state })))
+      if viewStore.state != nil {
+        let unwrapper = Optional<State>.lastWrappedValue
+        // Force unwrap is safe here because first value from scope is non-nil and scoped store
+        // is dismanteled after last nil value.
+        return ViewBuilder.buildEither(first: ifContent(store.scope(state: { unwrapper($0)! })))
       } else {
         return ViewBuilder.buildEither(second: elseContent())
       }
@@ -72,8 +75,11 @@ public struct IfLetStore<State, Action, Content>: View where Content: View {
   ) where Content == IfContent? {
     self.store = store
     self.content = { viewStore in
-      viewStore.state.map { state in
-        ifContent(store.scope(state: { $0 ?? state }))
+      viewStore.state.map { _ in
+        let unwrapper = Optional<State>.lastWrappedValue
+        // Force unwrap is safe here because first value from scope is non-nil and scoped store
+        // is dismanteled after last nil value.
+        return ifContent(store.scope(state: { unwrapper($0)! }))
       }
     }
   }
@@ -84,5 +90,15 @@ public struct IfLetStore<State, Action, Content>: View where Content: View {
       removeDuplicates: { ($0 != nil) == ($1 != nil) },
       content: self.content
     )
+  }
+}
+
+extension Optional {
+  static var lastWrappedValue: (Self) -> Self {
+    var lastWrapped: Wrapped?
+    return {
+      lastWrapped = $0 ?? lastWrapped
+      return lastWrapped
+    }
   }
 }


stephencelis:
@ollitapa Unfortunately this won't work for `IfLetStore` used outside of navigation. If you conditionally render a component inside a screen, for example ☹️

We're not quite ready to address this problem in the library due to subtle things like this, but please do use the `Optional.lastWrappedValue` in your own projects where it helps!

darrarski:
Hi! Some time ago, [I wrote](https://github.com/darrarski/darrarski/blob/main/2021/04/Thoughts-on-SwiftUI-navigation/README.md) about navigation issues when working with SwiftUI and ComposableArchitecture. I've even created [an extension](https://github.com/darrarski/tca-swiftui-navigation-demo/blob/2711bc8c1c5d35d166a9b722d9db9479c460e07c/Demo/Sources/Utils/NavigationLink%2BStore.swift) to `NavigationLink` that does a similar thing as the proposed changes. It's nice to see such an addition to ComposableArchitecture.

However, I would like to notice that it's not always an ideal solution to the problem. In one of the apps I was working on, I've experienced a case where it caused unwanted behavior. The problem was on iPad when using `NavigationView` with default `ColumnNavigationViewStyle` style. The view presented in the navigation, on the right side of the screen, once presented, was not removed when it should be. The reason was simple - when the state was `nil`, the `lastWrappedValue` was used to render it anyway (when it should not be rendered). While it makes sense to prolong the rendering of the view while it's being dismissed (like in standard navigation on iPhone), it's not always what we want.

With this being said, I don't think it's a good idea to change the current behavior of `IfLetStore`. I would rather allow users to decide if the last-known-non-nil state should be used to render a view when the actual state is `nil`.

ollitapa:
Hmm, I'm not sure I see the issue 😅. At least this sort of view works just fine

```swift
struct TestState: Equatable {
    var number: Int?
}

let reducer = Reducer { (state: inout TestState, action: Void, env: Void) -> Effect in

    state.number = state.number == nil ? 1 : nil

    return .none
}

let store = Store(initialState: TestState(number: nil), reducer: reducer, environment: ())

struct TestView: View {

    var body: some View {
        VStack {

            IfLetStore(store.scope(state: \.number)) { numberStore in
                Button(action: { store.view.send(()) }) {
                    Text("Set to nil")
                }
                Text("Number: \(numberStore.view.state)")
            } else: {
                Button(action: { store.view.send(()) }) {
                    Text("Set to 1")
                }
                Text("No number")
            }
        }
    }
}
```

I think it's actually the correct behaviour for `IfLetStore` in it's original use-case (since ifContent should not ever receive the replay of the first value). 

Now that I did a bit more experimentations, this does not fix an issue when it is used with `NavigationLink` where the `else` block will still get called before the animation to dismiss starts. It looks to me like the `IfLetStore` should not be used as the destination, but as you said, an entirely new view to handle navigation should be done (maybe a some sort of `CompactStoreView` 😉 ). 

It's a bit of a bummer that we can't write unit test to showcase the incorrect/correct behaviours for SwiftUI 😓 

But yeah, I can close this if folks prefer to keep the current behaviour.

stephencelis:
@ollitapa It's subtle, but the `viewStore != nil` branch will evaluate correctly, still, so it'll empty out the view on a programmatic sheet dismissal before the sheet is dismissed.

You can see the problem on your branch (and `mail`) if you swap out this case study's line:

https://github.com/pointfreeco/swift-composable-architecture/blob/main/Examples/CaseStudies/SwiftUICaseStudies/03-Navigation-Sheet-LoadThenPresent.swift#L92

With:

```swift
          then: { counterStore in
            NavigationView {
              CounterView(store: counterStore)
                .navigationBarItems(
                  trailing: Button("Dismiss") {
                    viewStore.send(.setSheet(isPresented: false))
                  }
                )
            }
          }
```

Tapping the "Dismiss" button will dismiss the sheet but the sheet will immediately go blank (you can turn on slow animations to make it very clear).

This is why scoping what's sent to `IfLetStore` is what should manage the caching for presented views, and why `IfLetStore` unfortunately can't manage it itself.

We're exploring how navigation helpers might hide away this detail, but @darrarski brings up even more subtle points to consider!

stephencelis:
> However, I would like to notice that it's not always an ideal solution to the problem. In one of the apps I was working on, I've experienced a case where it caused unwanted behavior. The problem was on iPad when using `NavigationView` with default `ColumnNavigationViewStyle` style. The view presented in the navigation, on the right side of the screen, once presented, was not removed when it should be. The reason was simple - when the state was `nil`, the `lastWrappedValue` was used to render it anyway (when it should not be rendered). While it makes sense to prolong the rendering of the view while it's being dismissed (like in standard navigation on iPhone), it's not always what we want when using `IfLetStore`.

In our experience, the caching is only necessary in presented contexts like sheets and full screen covers. For whatever reason, popping navigation seems to cache the views. Do you have any code we could look at where caching the last wrapped value is necessary for `NavigationView`?

darrarski:
@ollitapa oh, I am sorry, but I've got confused. You are right, the solution you are proposing should not affect us in the way I described.

@stephencelis The issue occurs when programmatically dismissing a view by setting its state to `nil`. The `IfLetStore` immediately removes the presented view from the hierarchy, before the dismiss animation ends. Here is an example of how does it looks:

https://user-images.githubusercontent.com/1384684/126568915-0c50167e-ba72-47fd-a095-24910032d5cd.mp4

Notice the presented view disappears before it's fully dismissed. Here is the [source code](https://gist.github.com/darrarski/5e7b3c66c3879f75f542b06143d9aed8) of the above example.

stephencelis:
Ah yeah, that's basically the same as the sheet example from my comment above. Sounds like care needs to be taken depending on the stack navigation style.

ollitapa:
> It's subtle, but the `viewStore != nil` branch will evaluate correctly, still, so it'll empty out the view on a programmatic sheet dismissal before the sheet is dismissed.

I think we are mixing up two issues here. I'm sorry, that's my fault since I thought that they were the same issue and fixed by the same solution, but thanks for proving me wrong. 

To my knowledge the issues are:

1. `IfLetStore` gets replayed the first value
2. Navigation calls the empty `else` block before dismiss

The first issue is fixed here by the `Optional.lastWrappedValue` inside the `IfLetStore` and can not be fixed outside. I made a sample project to demonstrate the issue here: https://github.com/ollitapa/IfLetStoreTest. The upper `Text` is done with current IfLetStore and the lower one is done with this PR fix applied.



As for the second issue, seen in the code sample by @darrarski, could be fixed temporarily (until more ergonomic navigation helpers are available) by applying this PR together with the outside scoping:

```swift
          NavigationLink(
              destination: IfLetStore(
                store.scope(
                  state: \.detail,
                  action: AppAction.detail
                )
                .sope(state: Optional.lastWrappedValue),

                then: DetailView.init(store:)
              ),
             ...
         )
```

Key here is to not apply the scoping to the `isActive` Binding that drives the navigation.

stephencelis (approved):
Ah, I understand now! Thanks for clarifying. This is a nice fix 😄

PR #669, opened 2021-07-22
Fix `Effect.throttle` data races and scheduler value return

p4checo:
The `Effect.throttle` operator had multiple data races when updating `throttleTimes` and `throttleValues` shared state, because:

1. `Effect.throttle` can be called from any scheduler.
2. The internal `flatMap` runs on the current chain scheduler, but the
throttled (delayed) value runs on the passed in `scheduler` parameter
(which can be different).

By protecting shared state with a lock (similar to the cancellables'), these data races should be addressed.

Additionally, the `Effect.throttle` should return all values in the `scheduler` passed in, so that the API contract is honored and values come from where callers expect them to.

## Changes

- Add new `throttleLock` `recursive lock to protect `throttleTimes` and `throttleValues` shared state in `Effect.throttle` operator.

- Ensure all values in `Effect.throttle` come from the passed in `scheduler`.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Throttling.swift
+++ b/Sources/ComposableArchitecture/Effects/Throttling.swift
@@ -44,12 +44,7 @@ extension Effect {
             for: scheduler.now.distance(to: throttleTime.advanced(by: interval)), scheduler: scheduler
           )
           .handleEvents(
-            receiveOutput: { _ in
-              throttleLock.lock()
-              defer { throttleLock.unlock() }
-
-              throttleTimes[id] = scheduler.now
-            }
+            receiveOutput: { _ in throttleLock.sync { throttleTimes[id] = scheduler.now } }
           )
           .setFailureType(to: Failure.self)
           .eraseToAnyPublisher()
--- a/Sources/ComposableArchitecture/Effects/Throttling.swift
+++ b/Sources/ComposableArchitecture/Effects/Throttling.swift
@@ -19,34 +19,46 @@ extension Effect {
     scheduler: S,
     latest: Bool
   ) -> Effect where S: Scheduler {
-    self.flatMap { value -> AnyPublisher<Output, Failure> in
-      guard let throttleTime = throttleTimes[id] as! S.SchedulerTimeType? else {
-        throttleTimes[id] = scheduler.now
-        throttleValues[id] = nil
-        return Just(value).setFailureType(to: Failure.self).eraseToAnyPublisher()
+    self.receive(on: scheduler)
+      .flatMap { value -> AnyPublisher<Output, Failure> in
+        throttleLock.lock()
+        defer { throttleLock.unlock() }
+
+        guard let throttleTime = throttleTimes[id] as! S.SchedulerTimeType? else {
+          throttleTimes[id] = scheduler.now
+          throttleValues[id] = nil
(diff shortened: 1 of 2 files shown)


stephencelis:
@p4checo Thanks for the PR! Looks like the changes are breaking our test expectations. Can you look into what may have changed?

Sources/ComposableArchitecture/Effects/Throttling.swift
```diff
@@ -19,34 +19,46 @@ extension Effect {
...
-    self.flatMap { value -> AnyPublisher<Output, Failure> in
-      guard let throttleTime = throttleTimes[id] as! S.SchedulerTimeType? else {
-        throttleTimes[id] = scheduler.now
-        throttleValues[id] = nil
-        return Just(value).setFailureType(to: Failure.self).eraseToAnyPublisher()
-      }
+    self.receive(on: scheduler)
+      .flatMap { value -> AnyPublisher<Output, Failure> in
+        throttleLock.lock()
+        defer { throttleLock.unlock() }
```
stephencelis, line 25: Oh, and one other note, the library has an internal `sync` helper for `NSRecursiveLock` (we use a lock for managing cancellation elsewhere), so you could potentially employ it throughout:

```swift
throttleLock.sync {
  ...
}
```

`defer` is probably straightforward enough, but gives us less granular control.

p4checo: Thanks for the heads up, had missed that nice helper! 🙈  I'll change it so use `sync` 👍🏼

p4checo: However, in the `flatMap` "main" body I think it's perhaps better to leave it as is, because it's easier to ensure no escapes from the critical section.

In the `handleEvents` below the `.sync` makes complete sense and makes it much terser.

What do you say?

Sources/ComposableArchitecture/Effects/Throttling.swift
```diff
@@ -19,34 +19,41 @@ extension Effect {
     scheduler: S,
     latest: Bool
   ) -> Effect where S: Scheduler {
-    self.flatMap { value -> AnyPublisher<Output, Failure> in
-      guard let throttleTime = throttleTimes[id] as! S.SchedulerTimeType? else {
-        throttleTimes[id] = scheduler.now
-        throttleValues[id] = nil
-        return Just(value).setFailureType(to: Failure.self).eraseToAnyPublisher()
-      }
+    self.receive(on: scheduler)
```
p4checo, line 22: An alternative solution would be to add the `receive(on:)` to each of the first the two exit paths (`Just(value)` calls that return immediately). 

The current approach ensures any value returned from the `flatMap` is always delivered on `scheduler` (unless explicitly scheduled elsewhere). However, we incur in an extra dispatch (and minor performance penalty) on the exit path where the value is delayed, because it is then dispatched by `delay` once again. My reasoning was tothat it was worth it to keep things simple, but can definitely change it if you prefer.

Sources/ComposableArchitecture/Effects/Throttling.swift
stephencelis: I think the way you have it is fine! We can always reorganize if we feel any reason to later. Thanks :smile:

stephencelis (approved):

PR #673, opened 2021-07-26
Add an async/await aware .send method to View Store.

mbrandonw:
Based on the work in the [most recent](https://www.pointfree.co/episodes/ep154-async-refreshable-composable-architecture) Point-Free episode, we are introducing _beta_ support for an async/await aware `.send` method. This API could change in the future with future Xcode betas, so use with caution 🙂.

This method allows you to send an action into the store and the suspend while some piece of boolean state is `true`. This can be particularly handy for interfacing with SwiftUI's new `.refreshable` API:

```swift
.refreshable { 
  await viewStore.send(.pulledToRefresh, while: \.isLoading)
}
```

### DIFF
--- a/Sources/ComposableArchitecture/Beta/Concurrency.swift
+++ b/Sources/ComposableArchitecture/Beta/Concurrency.swift
@@ -108,7 +108,8 @@ extension ViewStore {
 
   /// Suspends while a predicate on state is `true`.
   ///
-  ///   - predicate: A predicate on `State` that determines for how long this method should suspend.
+  /// - Parameter predicate: A predicate on `State` that determines for how long this method should
+  /// suspend.
   public func suspend(while predicate: @escaping (State) -> Bool) async {
     var cancellable: Cancellable?
     try? await withTaskCancellationHandler(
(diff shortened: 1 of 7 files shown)


stephencelis (approved):

Sources/ComposableArchitecture/Beta/Concurrency.swift
```diff
@@ -0,0 +1,136 @@
...
+          guard !Task.isCancelled else {
+            continuation.resume(throwing: CancellationError())
+            return
+          }
+          cancellable = self.publisher
+            .filter { !predicate($0) }
+            .prefix(1)
+            .sink { _ in
+              continuation.resume()
+              _ = cancellable
```
tgrapperon, line 129: Out of curiosity, is this different from capturing `cancellable` with brackets like you do on l.116? Wouldn't
```swift
.sink { [cancellable] _ in continuation.resume() }
```
be equivalent?

mbrandonw: Good question. If we don't use the `cancellable` in the closure then we get an unused warning. So I think we gotta do the `_ = cancellable` dance.

PR #676, opened 2021-07-26
Cancel multiple effects with a variadic list.

mbrandonw:


Sources/ComposableArchitecture/Effects/Cancellation.swift
```diff
@@ -74,6 +74,15 @@ extension Effect {
...
   }
+
+  /// An effect that will cancel multiple currently in-flight effects with the given identifiers.
+  ///
+  /// - Parameter ids: A variadic list of effect identifiers.
+  /// - Returns: A new effect that will cancel any currently in-flight effects with the given
+  ///   identifiers.
+  public static func cancel(ids: AnyHashable...) -> Effect {
+    .merge(ids.map(Effect.cancel(id:)))
+  }
```
stephencelis, line 85:
```suggestion
  /// An effect that will cancel multiple currently in-flight effects with the given identifiers.
  ///
  /// - Parameter ids: A variadic list of effect identifiers.
  /// - Returns: A new effect that will cancel any currently in-flight effects with the given
  ///   identifiers.
  public static func cancel(ids: AnyHashable...) -> Effect {
    .cancel(ids: ids)
  }

  /// An effect that will cancel multiple currently in-flight effects with the given identifiers.
  ///
  /// - Parameter ids: An array of effect identifiers.
  /// - Returns: A new effect that will cancel any currently in-flight effects with the given
  ///   identifiers.
  public static func cancel(ids: [AnyHashable]) -> Effect {
    .merge(ids.map(Effect.cancel(id:)))
  }
```

stephencelis: (I was hoping it'd be better about the diff. First function should be the same other than calling to the second.

stephencelis (approved):

Sources/ComposableArchitecture/Effects/Cancellation.swift
mbrandonw: ah yeah, I guess also in that case might as well genericize to a `Sequence`?

I'll make the change locally

PR #678, opened 2021-07-28
Allow multiple strategies in `Effect.throttle`

p4checo:
The `Effect.throttle` operator allowed only two strategies (first, latest) via the `latest` parameter. While this covers most frequent scenarios, it doesn't allow any other strategy which is necessary on some use cases.

One such use case is when we need to throttle values but are interested in *all* intermediate values, because they can come from different sources and we can't afford to discard any and rely on just the first or latest. One concrete example is an e-commerce app where we can update multiple item quantities' via UI which are then merged and mapped into rate limited (throttled) network requests to an API, that contain multiple updates each.

To allow this and other scenarios, a new `Effect.ThrottleStrategy` was created that's essentially a witness type around a closure responsible for determining the final value to return in each throttle interval (essentially a working as a `reduce`).

Backwards compatibility was preserved by keeping the old API, which forwards to the new one.

## Changes

- Add new `Effect.ThrottleStrategy` witness type.

- Update `Effect.throttle` to receive a new `ThrottleStrategy`.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Throttling.swift
+++ b/Sources/ComposableArchitecture/Effects/Throttling.swift
@@ -19,18 +19,39 @@ extension Effect {
     scheduler: S,
     latest: Bool
   ) -> Effect where S: Scheduler {
+    self.throttle(id: id, for: interval, scheduler: scheduler, strategy: latest ? .latest : .first)
+  }
+
+  /// Throttles an effect so that it only publishes one output per given interval.
+  ///
+  /// - Parameters:
+  ///   - id: The effect's identifier.
+  ///   - interval: The interval at which to find and emit each element, expressed in the time
+  ///   system of the scheduler.
+  ///   - scheduler: The scheduler you want to deliver the throttled output to.
+  ///   - strategy: The strategy used do determine the final element to emit from all elements
+  ///   received during each interval.
+  /// - Returns: An effect that only emits one element during the specified interval according to
+  /// the defined strategy.
+  public func throttle<S, T>(
+    id: AnyHashable,
+    for interval: S.SchedulerTimeType.Stride,
+    scheduler: S,
+    strategy: ThrottleStrategy<T, Output>
+  ) -> Effect<T, Failure> where S: Scheduler {
     self.receive(on: scheduler)
-      .flatMap { value -> AnyPublisher<Output, Failure> in
+      .flatMap { value -> AnyPublisher<T, Failure> in
         throttleLock.lock()
         defer { throttleLock.unlock() }
 
         guard let throttleTime = throttleTimes[id] as! S.SchedulerTimeType? else {
           throttleTimes[id] = scheduler.now
           throttleValues[id] = nil
+          let value = strategy.nextValue(nil, value)
           return Just(value).setFailureType(to: Failure.self).eraseToAnyPublisher()
         }
 
-        let value = latest ? value : (throttleValues[id] as! Output? ?? value)
+        let value = strategy.nextValue(throttleValues[id] as! T?, value)
         throttleValues[id] = value
 
         guard throttleTime.distance(to: scheduler.now) < interval else {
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/Effects/Throttling.swift
```diff
@@ -45,7 +47,12 @@ extension Effect {
             scheduler: scheduler
           )
           .handleEvents(
-            receiveOutput: { _ in throttleLock.sync { throttleTimes[id] = scheduler.now } }
+            receiveOutput: { _ in
+                throttleLock.sync {
+                    throttleTimes[id] = scheduler.now
+                    throttleValues[id] = nil
```
p4checo, line 53: This fixes the bug (when using a strategy different than `latest`), while ensuring we clean up shared state

Tests/ComposableArchitectureTests/EffectThrottleTests.swift
```diff
@@ -104,13 +112,96 @@ final class EffectThrottleTests: XCTestCase {
...
+
+    // A third value is throttled.
     XCTAssertEqual(values, [1, 2])
+
+    runThrottledEffect(value: 7)
+
+    scheduler.advance(by: 0.25)
+
+    // The third (throttled) value emits.
+    XCTAssertEqual(values, [1, 2, 6])
```
p4checo, line 134: This would previously fail as values would contain `[1, 2, 2]` (the previous interval's first wasn't cleared)

stephencelis:
Hi @p4checo! Thanks for the PR. This looks really cool! We had a couple questions and thoughts.

> The `Effect.throttle` operator allowed only two strategies (first, latest) via the `latest` parameter. While this covers most frequent scenarios, it doesn't allow any other strategy which is necessary on some use cases.
> 
> One such use case is when we need to throttle values but are interested in _all_ intermediate values, because they can come from different sources and we can't afford to discard any and rely on just the first or latest. One concrete example is an e-commerce app where we can update multiple item quantities' via UI which are then merged and mapped into rate limited (throttled) network requests to an API, that contain multiple updates each.
> 
> To allow this and other scenarios, a new `Effect.ThrottleStrategy` was created that's essentially a witness type around a closure responsible for determining the final value to return in each throttle interval (essentially a working as a `reduce`).

This makes sense and there's definitely value in introducing more configurability here. Our main apprehension in changing the current API, though, is that it's currently designed to match [the corresponding Combine API](https://developer.apple.com/documentation/combine/fail/throttle(for:scheduler:latest:)) as closely as possible. Throughout the library (as well as [composable-core-motion](https://github.com/pointfreeco/composable-core-motion), etc.) we take this approach: we try our best not to stray from any original APIs that our APIs are based off of, except for parameters that act as "glue" code between TCA and other frameworks, like the `id` parameters that effects have. We feel that this lends a level of familiarity to our APIs for folks that are already familiar with Combine's `Publishers.Throttle`, etc. So this is the main reason we're a little wary of adding new API surface area here that goes beyond the original.

One thought would be to still ship your effect as its own library. @mbrandonw and I have working on getting a "pointfreecommunity" of projects started after seeing a lot of interesting TCA libraries come out of the community. We haven't quite gotten around to launching the endeavor just yet, but would you consider submitting this super-powered version of throttle as a library that could eventually live there instead? Something like effect-throttle-strategy, or effect-strategic-throttle? Or there's probably a much better naming convention to adopt 😅

> Additionally, a bug was also fixed in Effect.throttle that didn't clean up `throttleValues` shared state on the delayed values path, e.g. leading to the previous first strategy (`latest` = false) returning old (incorrect) values on some scenarios.

Would you be interested in shipping this bug fix in a separate PR? If not, we'll try to take a look soon.

p4checo:
Hi @stephencelis and @mbrandonw 👋🏼! Thanks!

> Our main apprehension in changing the current API, though, is that it's currently designed to match the corresponding Combine API as closely as possible. 

I knew that modifying this API would be a bit risky, but I had to try anyway as I thought it was an worthy improvement 😅. 

> We feel that this lends a level of familiarity to our APIs for folks that are already familiar with Combine's Publishers.Throttle, etc. So this is the main reason we're a little wary of adding new API surface area here that goes beyond the original.

I understand your argument of trying to match Combine's API as much as possible, but I also think that it can in a way limit  more advanced usages as it effectively restricts TCA's own API evolution. I'm totally in favor of reducing friction for folks using the framework, and familiarity definitely helps. However, TCA **is** open source (unlike Combine), so IMHO folks can always dive into the actual code and understand what's going on if they find something _unfamiliar_ 😇 (I know many sadly don't though).

> One thought would be to still ship your effect as its own library. 

That's always a possibility, however in this specific case it can also cause unwanted side effects, because of the `Effect.throttle`'s reliance on a static shared state (e.g. `throttleValues`). If there are two `Effect.throttle` implementations in essentially two different modules, then each will have its own shared state. This means that if a user somehow uses the two `throttle` APIs with the same `id` expecting them to interoperate it won't have the expected outcome. I know this is a far fetched example, but it's technically possible 😆 

> @mbrandonw and I have working on getting a "pointfreecommunity" of projects started after seeing a lot of interesting TCA libraries come out of the community. 

This is great! I definitely think TCA will benefit from something like this 👏🏼 

--

I do have a 3rd option for this though, which had already occurred to me when it hit me that I hadn't kept backwards compatibility and/or transition warning (via `@available` annotation) 🙈. Since this new variant is completely backwards compatible with the previous, Combine-friendly API, how about if we added an overload preserving the old API but using this more powerful one underneath? Something like:

```swift
public func throttle<S>(
  id: AnyHashable,
  for interval: S.SchedulerTimeType.Stride,
  scheduler: S,
  latest: Bool
) -> Effect where S: Scheduler {
  self.throttle(id: id, for: interval, scheduler: scheduler, strategy: latest ? .latest : .first)
}
```

IMO this strikes a good balance, and would be the best of both worlds. What do you say?

>> Additionally, a bug was also fixed in Effect.throttle that didn't clean up `throttleValues` shared state on the delayed values path, e.g. leading to the previous first strategy (latest = false) returning old (incorrect) values on some scenarios.
>
> Would you be interested in shipping this bug fix in a separate PR? If not, we'll try to take a look soon.

Sure, I can open a separate PR fixing the bug if this PR ends up being closed 👍🏼 Or would you prefer it to be a separate PR anyway?

stephencelis:
> I do have a 3rd option for this though, which had already occurred to me when it hit me that I hadn't kept backwards compatibility and/or transition warning (via `@available` annotation) 🙈. Since this new variant is completely backwards compatible with the previous, Combine-friendly API, how about if we added an overload preserving the old API but using this more powerful one underneath? Something like:

Hm, this _is_ a tempting idea, though I think we'll want to let the idea marinate for a bit longer before we decide to make an exception.

> Sure, I can open a separate PR fixing the bug if this PR ends up being closed 👍🏼 Or would you prefer it to be a separate PR anyway?

Let's open a separate PR anyway just so we can unblock the bug fix while this PR is blocked and we think on it.

stephencelis:
@p4checo Chatted with @mbrandonw again about this and while we definitely see its usefulness and understand your points, we want to avoid adding the complexity of its API surface area to the repo at this time. Luckily you can still define this more powerful `throttle` in your own projects, caveats aside.

Thanks, though, for opening this PR (and others!). We appreciate the idea and discussion around it :smile:

PR #680, opened 2021-07-28
Enables animations for alert button actions

fonkadelic:
I've noticed that currently there seems no way to perform animations when alert actions are sent.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
@@ -39,14 +39,18 @@ extension Effect where Failure == Never {
 }
 
 struct AnimationsState: Equatable {
-  var circleCenter = CGPoint.zero
+  var alert: AlertState<AnimationsAction>? = nil
+  var circleCenter = CGPoint(x: 50, y: 50)
   var circleColor = Color.white
   var isCircleScaled = false
 }
 
 enum AnimationsAction: Equatable {
+  case alertButtonTapped
   case circleScaleToggleChanged(Bool)
+  case dismissAlert
   case rainbowButtonTapped
+  case resetButtonTapped
   case setColor(Color)
   case tapped(CGPoint)
 }
@@ -59,8 +63,16 @@ let animationsReducer = Reducer<AnimationsState, AnimationsAction, AnimationsEnv
   state, action, environment in
 
   switch action {
+  case .alertButtonTapped:
+    state = .init()
+    return .none
+    
   case let .circleScaleToggleChanged(isScaled):
     state.isCircleScaled = isScaled
+    return .none
+    
+  case .dismissAlert:
+    state.alert = nil
     return .none
 
   case .rainbowButtonTapped:
@@ -69,6 +81,15 @@ let animationsReducer = Reducer<AnimationsState, AnimationsAction, AnimationsEnv
         .map { (output: .setColor($0), duration: 1) },
       scheduler: environment.mainQueue.animation(.linear)
     )
+    
+  case .resetButtonTapped:
+    state.alert = .init(
+      title: .init("Reset state?"),
+      primaryButton: .destructive(.init("Reset"))
+        .send(.alertButtonTapped, withAnimation: .default),
+      secondaryButton: .cancel()
+    )
+    return .none
 
   case let .setColor(color):
     state.circleColor = color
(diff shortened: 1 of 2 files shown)


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
```diff
@@ -39,14 +39,18 @@ extension Effect where Failure == Never {
...
 struct AnimationsState: Equatable {
-  var circleCenter = CGPoint.zero
+  var alert: AlertState<AnimationsAction>? = nil
+  var circleCenter = CGPoint(x: 50, y: 50)
   var circleColor = Color.white
   var isCircleScaled = false
 }
 
 enum AnimationsAction: Equatable {
+  case alertButtonTapped
```
stephencelis, line 49: What do you think of changing the name to something a little more specific and re-alphabetizing?
```suggestion
  case resetConfirmationButtonTapped
```

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
```diff
@@ -59,16 +63,33 @@ let animationsReducer = Reducer<AnimationsState, AnimationsAction, AnimationsEnv
...
       values: [Color.red, .blue, .green, .orange, .pink, .purple, .yellow, .white]
         .map { (output: .setColor($0), duration: 1) },
       scheduler: environment.mainQueue.animation(.linear)
     )
+    
+  case .resetButtonTapped:
+    state.alert = .init(
+      title: .init("Reset state?"),
+      primaryButton: .destructive(.init("Reset"))
+        .send(.alertButtonTapped, withAnimation: .default),
```
stephencelis, line 89: Instead of this "fluent" style, what do you think of adding `animation` as a parameter to these button helpers?
```suggestion
      primaryButton: .destructive(.init("Reset"), animation: .default),
```

Sources/ComposableArchitecture/SwiftUI/Alert.swift
```diff
@@ -129,7 +129,15 @@ public struct AlertState<Action> {
...
     public var action: Action?
+    public var animation: Animation??
     public var type: `Type`
+    
+    public func send(_ action: Action, withAnimation animation: Animation?) -> Self {
+      var button = self
+      button.action = action
+      button.animation = animation
+      return button
+    }
```
stephencelis, line 140: Instead of this single `send`, we could overload each of the helpers below with a version that takes an `Animation?`. This way the API would match how `ViewStore.send` has an overload that takes an `animation` parameter. It'll be a bunch of overloads, but shouldn't be so so bad.

stephencelis (commented):
@fonkadelic Thanks for the PR! This is something that @mbrandonw and I have been thinking about for awhile, and even came up while we were working on isowords. I think your PR is basically good to go, just a couple small things.

- I think we want to tweak the API design a little bit. Rather than the fluent style of chaining `.send` onto an alert button, we think it might make more sense for this API to look a little more like `ViewStore.send`, by using overloads (on `.destructive`, etc.) to specify animation instead.
- And a very small change to the case study you updated :smile:

Comments below to highlight these things a bit more!

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
fonkadelic: Sounds good. Already considered this as an alternative.

stephencelis (approved):
Thanks! I think we want to mess with the APIs a bit more before release, but this looks good to merge!

Sources/ComposableArchitecture/SwiftUI/Alert.swift
```diff
@@ -129,33 +129,38 @@ public struct AlertState<Action> {
...
   public struct Button {
     public var action: Action?
+    public var animation: Animation??
     public var type: `Type`
 
     public static func cancel(
       _ label: TextState,
-      send action: Action? = nil
+      send action: Action? = nil,
+      animation: Animation? = nil
```
stephencelis, line 138: Ha in hindsight these arguments introduce some strange states, like `send: nil, animation: .default`. We have an idea on how to fix this but will do so in a follow-up PR.

fonkadelic: Yeah. This was something i've struggled with. Looking forward to the follow-up PR.

PR #681, opened 2021-07-28
Update AlertState.Button APIs for animation

stephencelis:
This PR refines the domain modeling I suggested in #680.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndActionSheets.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndActionSheets.swift
@@ -48,8 +48,8 @@ let alertAndSheetReducer = Reducer<
       message: .init("This is an action sheet."),
       buttons: [
         .cancel(),
-        .default(.init("Increment"), send: .incrementButtonTapped),
-        .default(.init("Decrement"), send: .decrementButtonTapped),
+        .default(.init("Increment"), action: .send(.incrementButtonTapped)),
+        .default(.init("Decrement"), action: .send(.decrementButtonTapped)),
       ]
     )
     return .none
@@ -66,7 +66,7 @@ let alertAndSheetReducer = Reducer<
       title: .init("Alert!"),
       message: .init("This is an alert"),
       primaryButton: .cancel(),
-      secondaryButton: .default(.init("Increment"), send: .incrementButtonTapped)
+      secondaryButton: .default(.init("Increment"), action: .send(.incrementButtonTapped))
     )
     return .none
 
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
@@ -83,8 +83,7 @@ let animationsReducer = Reducer<AnimationsState, AnimationsAction, AnimationsEnv
       title: .init("Reset state?"),
       primaryButton: .destructive(
         .init("Reset"),
-        send: .resetConfirmationButtonTapped,
-        animation: .default
+        action: .send(.resetConfirmationButtonTapped, animation: .default)
       ),
       secondaryButton: .cancel()
     )
--- a/Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
@@ -165,6 +165,19 @@ extension ActionSheetState: Equatable where Action: Equatable {
 @available(macOS, unavailable)
 @available(tvOS 13, *)
 @available(watchOS 6, *)
+extension ActionSheetState: Hashable where Action: Hashable {
+  public func hash(into hasher: inout Hasher) {
+    hasher.combine(self.title)
+    hasher.combine(self.message)
+    hasher.combine(self.buttons)
+  }
+}
+
+@available(iOS 13, *)
+@available(macCatalyst 13, *)
+@available(macOS, unavailable)
+@available(tvOS 13, *)
+@available(watchOS 6, *)
 extension ActionSheetState: Identifiable {}
 
 extension View {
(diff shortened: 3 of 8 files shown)


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndActionSheets.swift
```diff
@@ -48,8 +48,8 @@ let alertAndSheetReducer = Reducer<
       message: .init("This is an action sheet."),
       buttons: [
         .cancel(),
-        .default(.init("Increment"), send: .incrementButtonTapped),
-        .default(.init("Decrement"), send: .decrementButtonTapped),
+        .default(.init("Increment"), action: .send(.incrementButtonTapped)),
+        .default(.init("Decrement"), action: .send(.decrementButtonTapped)),
```
stephencelis, line 52: `action` is the argument name that `SwiftUI.Alert.Button` constructors take, so this will already more closely model the upstream API.

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
```diff
@@ -83,8 +83,7 @@ let animationsReducer = Reducer<AnimationsState, AnimationsAction, AnimationsEnv
       title: .init("Reset state?"),
       primaryButton: .destructive(
         .init("Reset"),
-        send: .resetConfirmationButtonTapped,
-        animation: .default
+        action: .send(.resetConfirmationButtonTapped, animation: .default)
```
stephencelis, line 86: There's an overload that takes an animation now.

Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
```diff
@@ -160,19 +160,6 @@ extension ActionSheetState: Equatable where Action: Equatable {
   }
 }
 
-@available(iOS 13, *)
-@available(macCatalyst 13, *)
-@available(macOS, unavailable)
-@available(tvOS 13, *)
-@available(watchOS 6, *)
-extension ActionSheetState: Hashable where Action: Hashable {
```
stephencelis, line 168: `Animation` is not hashable, so we lose this conformance...unless you have ideas?

Sources/ComposableArchitecture/SwiftUI/Alert.swift
```diff
@@ -128,47 +128,86 @@ public struct AlertState<Action> {
...
+      Self(action: action, type: .destructive(label: label))
     }
 
-    public enum `Type` {
-      case cancel(label: TextState?)
-      case `default`(label: TextState)
-      case destructive(label: TextState)
+    public static func cancel(
+      _ label: TextState,
+      send action: Action?
```
stephencelis, line 163: Should I also deprecate these old ones? The one issue is we can't do a simple `rename` here...

Sources/ComposableArchitecture/SwiftUI/Alert.swift
```diff
@@ -128,47 +128,86 @@ public struct AlertState<Action> {
...
+    }
+
+    public static func send(_ action: Action, animation: Animation?) -> Self {
+      .init(type: .animatedSend(action, animation: animation))
+    }
+
+    enum ActionType {
+      case send(Action)
+      case animatedSend(Action, animation: Animation?)
+    }
```
stephencelis, line 203: Using an internal enum here to work around https://bugs.swift.org/browse/SR-10077, and so we can exhaustively interpret things...but should it be `public` so people can write their own code that plucks out the values inside?

stephencelis: (Also, this enum leaves us open to supporting UIKit animations in the future, should someone submit a PR.)

Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
mbrandonw: I think we could just leave `Animation` out of the hashing. it means a slightly less performant hash but still semantically correct.

Sources/ComposableArchitecture/SwiftUI/Alert.swift
mbrandonw: yeah I think deprecate. bummer we can't rename, but I think that's ok.

mbrandonw (approved):

Sources/ComposableArchitecture/SwiftUI/Alert.swift
andreyz: @stephencelis re: supporting UIKit animations, would calling something like this in #563 potentially work?

``` swift
import struct SwiftUI.Animation
func withUIKitAnimation(_ animation: Animation?, body: @escaping () -> Void) -> Void {
  switch animation {
  case .default?:
    UIView.animate(
      withDuration: 0.35,
      delay: 0,
      options: [.curveEaseInOut, .allowUserInteraction],
      animations: body
    )
  ...
  default:
    body()
  }
}
```

stephencelis: @andreyz I think we would add brand new enum cases that takes the arguments sent to `UIView.animate`. That would seem easier to get right than trying to convert a SwiftUI animation over to UIKit.

andreyz: @stephencelis do you mean hypothetically an extra `.animatedSend` -like case with explicit parameters for duration, options, etc? If so, I'll do it, after my earlier PR lands.

stephencelis: I think so...the only complication is that animated Combine schedulers can at least nest, so you could perform UIKit and SwiftUI actions on the same action. It may sound unlikely, but we do just this in isowords for screens that show both UIKit and SwiftUI...probably less likely for alerts, though, so definitely feel free to start exploring a solution there!

andreyz: Thanks for encouragement and feedback @stephencelis. I'll check isowords and noodle on it.

PR #683, opened 2021-07-28
Fix `Effect.throttle` delayed value shared state cleanup

p4checo:
Effect.throttle didn't clean up `throttleValues` shared state on the delayed values path, leading the first value strategy (`latest: false`) to return incorrect (old) values on some scenarios.

Opened as separate PR following discussion on #678. 

## Changes

- Add missing `throttleValues[id]` clean up in `Effect.throttle` delayed value path.


stephencelis (approved):
Thanks!

PR #686, opened 2021-07-29
Fix build error when minimum target is macOS 11.0

aroben:
macOS 11.0 introduced an overload of Publisher.flatMap(maxPublishers:_:)
for when the failure type is Never, documented here:
https://developer.apple.com/documentation/combine/publisher/flatmap(maxpublishers:_:)-qxf

This causes a compiler error in Effect.fireAndForget when the minimum
deployment target is 11.0 or later:

    Sources/ComposableArchitecture/Effect.swift:364:7: error: ambiguous use of 'flatMap(maxPublishers:_:)'
          self
          ^
    Combine.Publisher:3:17: note: found this candidate
        public func flatMap<T, P>(maxPublishers: Subscribers.Demand = .unlimited, _ transform: @escaping (Self.Output) -> P) -> Publishers.FlatMap<P, Self> where T == P.Output, P : Publisher, Self.Failure == P.Failure
                    ^
    Combine.Publisher:3:17: note: found this candidate
        public func flatMap<P>(maxPublishers: Subscribers.Demand = .unlimited, _ transform: @escaping (Self.Output) -> P) -> Publishers.FlatMap<Publishers.SetFailureType<P, Self.Failure>, Self> where P : Publisher, P.Failure == Never
                    ^

We can fix this by adding explicit type parameters to the Empty
publisher used in the flatMap().

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -362,7 +362,7 @@ extension Publisher {
   ) -> Effect<NewOutput, NewFailure> {
     return
       self
-      .flatMap { _ in Empty() }
+      .flatMap { _ in Empty<NewOutput, Failure>() }
       .catch { _ in Empty() }
       .eraseToEffect()
   }


stephencelis:
@aroben Hm, I just created a fresh macOS project targeting 11 and CombineSchedulers appears to build just fine as is. Do you have a project that reproduces the issue on 0.5.1? And can you let us know how you're building things? (Version of Xcode, OS, etc.)

aroben:
I'm using Xcode 13 Beta 4, running on macOS 11.5.

I can reproduce like this:

```shellsession
$ git clone https://github.com/pointfreeco/swift-composable-architecture
$ cd swift-composable-architecture
$ xcodebuild -workspace ComposableArchitecture.xcworkspace -scheme ComposableArchitecture build MACOSX_DEPLOYMENT_TARGET=11.0
```

aroben:
I know this project is targeting 10.15, but my app is targeting 11.0 and builds in an environment where all dependencies get targeted to 11.0 as well by overriding `MACOSX_DEPLOYMENT_TARGET`. So this fix isn't strictly needed for swift-composable-architecture itself, though is needed for my build environment. It seems like a bit of harmless future-proofing for when swift-composable-architecture bumps its deployment target to 11.0+.

stephencelis (approved):
Huh! I can repro from the command line, but for whatever reason couldn't in a project. Thanks for submitting a fix!

aroben:
In normal usage, if you have an Xcode project targeting 11.0 that uses swift-composable-architecture, Xcode will still build swift-composable-architecture itself with a deployment target of 10.15, so the build error doesn't occur. It's only any issue when you change swift-composable-architecture's deployment target like in that `xcodebuild` invocation.

PR #689, opened 2021-07-30
Expose `Store`'s current state snapshot

p4checo:
## Motivation

Unlike `Reducer.run` which allows deriving custom third party reducers from existing ones, `Store` doesn't give as much flexibility to create custom third party operators based on `scope` that often require access to the current (initial/snapshot) state.

This is caused by `Store` not exposing its current state publicly as a read-only property. To work around this, devs resort to wrapping the store in a `ViewStore` to have access to its current (initial/snapshot) state first, to then build upon `scope` and use it as needed. This is obviously not ideal.

By creating a `Store.currentState` read-only computed property that wraps `Store.state`, this restriction is lifted. This way, wrapping `ViewStore` is no longer necessary, which makes creating third party scoping operators more straightforward and ergonomic. In terms of `Store.state` encapsulation, I think it should be essentially equivalent to `ViewStore(self).state`, so I don't foresee any concerning differences unless I'm missing something.

**PS:**

For consistency's sake, I think that the `Store`'s public state property should be `state`, and it's `private`/`internal` one `_state`. I didn't make that change because I wanted the "idea" to be reviewed first before increasing changes. Will gladly make that change if you agree with this addition (and the rename).

## Examples

This change will reduce friction when creating custom scoping operators, which is especially frequent for those of us still using TCA with UIKit 😭. Furthermore, it avoids having to increase API surface in TCA, which understandably you want to be very careful with.

Such examples include:

`Store.child` (from #533)
```swift
public func child<LocalState, LocalAction, ID>(
  id: ID,
  state toLocalStateContainer: @escaping (State) -> IdentifiedArray<ID, LocalState>,
  action fromLocalAction: @escaping (ID, LocalAction) -> Action
) -> Store<LocalState, LocalAction> {
  let scopedStore = scope(state: toLocalStateContainer)

  // NB: We cache current element here to avoid a crash where UIKit may re-evaluate the store
  //     following item deletion in table/collection views.
  let element = scopedStore.state.value[id: id]! // <-- can now use `scopedStore.currentState`
  
  return scopedStore.scope(
    state: { $0[id: id] ?? element },
    action: { fromLocalAction(id, $0) }
  )
}
```

`Store.forEach` (from this [comment](https://github.com/pointfreeco/swift-composable-architecture/discussions/351#discussioncomment-312688) in #351)
```swift
public func forEach<LocalState: Equatable, LocalAction>(
  state: @escaping (State) -> [LocalState],
  action: @escaping ((Int, LocalAction)) -> Action
) -> [Store<LocalState, LocalAction>] {
  let scopedStore = scope(state: state)
  let viewStore = ViewStore(scopedStore) // <-- can now use `scopedStore.currentState`
  let localStores = zip(viewStore.indices, viewStore.state).map { index, element in
    scopedStore.scope(
      state: { index < $0.endIndex ? $0[index] : element },
      action: { action((index, $0)) }
    )
  }
  return localStores
}
```

Unfortunately, a similar scenario happens almost every time when trying to implement custom scoping operators, because initial state is often required. 

## Changes

- Add new `Store.currentState` computed property that wraps `Store.state`.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -121,6 +121,9 @@ public final class Store<State, Action> {
   private let reducer: (inout State, Action) -> Effect<Action, Never>
   private var bufferedActions: [Action] = []
 
+  /// The current state.
+  public var currentState: State { state.value }
+
   /// Initializes a store from an initial state, a reducer, and an environment.
   ///
   /// - Parameters:
--- a/Tests/ComposableArchitectureTests/StoreTests.swift
+++ b/Tests/ComposableArchitectureTests/StoreTests.swift
@@ -6,6 +6,15 @@ import XCTest
 final class StoreTests: XCTestCase {
   var cancellables: Set<AnyCancellable> = []
 
+  func testCurrentStateEqualsState() {
+    let reducer = Reducer<Double, Void, Void> { _, _, _ in .none }
+    let store = Store(initialState: .pi, reducer: reducer, environment: ())
+
+    XCTAssertEqual(store.currentState, store.state.value)
+    store.state.value = 2 * .pi
+    XCTAssertEqual(store.currentState, store.state.value)
+  }
+
   func testCancellableIsRemovedOnImmediatelyCompletingEffect() {
     let reducer = Reducer<Void, Void, Void> { _, _, _ in .none }
     let store = Store(initialState: (), reducer: reducer, environment: ())


stephencelis:
Hey @p4checo, thanks for starting the discussion!

We think the explicit `ViewStore` call as a good thing, since it means you must consider the implications of peeking inside the store to look at its state. We even think it might be a good idea to adopt this style internally in the library, assuming there are no serious performance implications.

One thing to consider in your examples above (and when using `ViewStore.init` in general), is that they potentially miss out on updates to store state over time, especially if it is called in a code path that is not being driven by the store's publisher. `ViewStore(store)` signals quite loudly that this code needs to be careful and either understand that it is being used within the context of a store subscription (like a SwiftUI view or `store.publisher` sink), or that it is aware that it doesn't need to be.

Given that context, do you still have concerns with requiring `ViewStore.init`?

p4checo:
Hi @stephencelis 👋🏼  Thanks for the explanation!

> We think the explicit ViewStore call as a good thing, since it means you must consider the implications of peeking inside the store to look at its state. We even think it might be a good idea to adopt this style internally in the library, assuming there are no serious performance implications.

I see your point, and it's a fair one. It does indeed make you think a bit harder _how_ you can pluck out the store's current state, which in turn _can_ make you think a bit more about any implications 😆. In my opinion, people should think about the implications of their code regardless of the level of "friction" required for a particular action 🤷🏼

> One thing to consider in your examples above (and when using `ViewStore.init` in general), is that they potentially miss out on updates to store state over time, especially if it is called in a code path that is not being driven by the store's publisher. ViewStore(store) signals quite loudly that this code needs to be careful and either understand that it is being used within the context of a store subscription (like a SwiftUI view or store.publisher sink), or that it is aware that it doesn't need to be.

Do you mean that accessing the `Store.state` should only happen in the `Store`'s context/queue? Regardless of the API used for this (i.e. via `ViewStore` or `Store`), shouldn't these caveats/notes/warnings/assumptions simply be documented? (If they aren't already 😅)

Regarding missing updates over time, I think devs should be aware that by accessing `ViewStore.state` or `Store.state` you're essentially accessing the _current_ snapshot and not observing values.

If I missed the point, could you please elaborate? 😇 

> Given that context, do you still have concerns with requiring ViewStore.init?

There're a couple of additional things we should consider though: 

1. `ViewStore` removes duplicates from its parent `Store`'s observed values. This could potentially cause unintended and hard to track issues if custom equality functions (or `Equatable` conformances) are used, because `ViewStore.state` may not always be the same as `Store.state`.
2. Probably not as problematic as 1, but `ViewStore` documentation states it should only be accessed from the main thread. Even though technically in these situations (store scoping operators) we create "dedicated" `ViewStores` only to pluck out the state (and not drive UI), these can be created from a thread which isn't main, possibly leading to some confusion as to if it is correct to do so.

stephencelis:
> I see your point, and it's a fair one. It does indeed make you think a bit harder _how_ you can pluck out the store's current state, which in turn _can_ make you think a bit more about any implications 😆. In my opinion, people should think about the implications of their code regardless of the level of "friction" required for a particular action 🤷🏼

We think that `ViewStore(store).state` isn't too much friction

> Do you mean that accessing the `Store.state` should only happen in the `Store`'s context/queue? Regardless of the API used for this (i.e. via `ViewStore` or `Store`), shouldn't these caveats/notes/warnings/assumptions simply be documented? (If they aren't already 😅)
> 
> Regarding missing updates over time, I think devs should be aware that by accessing `ViewStore.state` or `Store.state` you're essentially accessing the _current_ snapshot and not observing values.
> 
> If I missed the point, could you please elaborate? 😇

Ah sorry, I didn't mean thread-safety or anything like that 😅 What I meant was that a store's state should only be accessed from a context that is observing it over time. In SwiftUI you get this behavior automatically because the view observes `ViewStore` as an `ObservableObject`. In UIKit you must take a bit more care and ideally only access state in `.publisher.sink { … }` blocks.

Unfortunately, because `ViewStore` is a concept shared between SwiftUI and UIKit, this means UIKit can access a view store's state even outside `.publisher.sink { … }` blocks. We've discussed potentially introducing separate view store types for SwiftUI and UIKit in the future to help steer folks away from this kind of access, as it can lead to accidentally creating views that don't react to changes over time.

> There're a couple of additional things we should consider though:
> 
> 1) `ViewStore` removes duplicates from its parent `Store`'s observed values. This could potentially cause unintended and hard to track issues if custom equality functions (or `Equatable` conformances) are used, because `ViewStore.state` may not always be the same as `Store.state`.

Sorry, could you elaborate? I'm not quite sure I understand your point.

> 2) Probably not as problematic as 1, but `ViewStore` documentation states it should only be accessed from the main thread. Even though technically in these situations (store scoping operators) we create "dedicated" `ViewStores` only to pluck out the state (and not drive UI), these can be created from a thread which isn't main, possibly leading to some confusion as to if it is correct to do so.

Hopefully this was clarified by #692, but in general all interactions with view stores should happen on the same thread, or rather, all `send`s should happen on the same thread. Scoping can invoke `send` on the parent, so you must be careful here, as well.

p4checo:
> We think that ViewStore(store).state isn't too much friction

Fair enough 😄 (And I agree that it's not that bad 😆)

> Ah sorry, I didn't mean thread-safety or anything like that 😅 What I meant was that a store's state should only be accessed from a context that is observing it over time. In SwiftUI you get this behavior automatically because the view observes ViewStore as an ObservableObject. In UIKit you must take a bit more care and ideally only access state in .publisher.sink { … } blocks.

Got it. It was in line what I had understood. Thanks!

> Unfortunately, because ViewStore is a concept shared between SwiftUI and UIKit, this means UIKit can access a view store's state even outside .publisher.sink { … } blocks. We've discussed potentially introducing separate view store types for SwiftUI and UIKit in the future to help steer folks away from this kind of access, as it can lead to accidentally creating views that don't react to changes over time.

That would be nice. Unfortunately the UIKit "part" of TCA hasn't been receiving much love lately 💔😇 

> > There're a couple of additional things we should consider though:
> >
> > ViewStore removes duplicates from its parent Store's observed values. This could potentially cause unintended and hard to track issues if custom equality functions (or Equatable conformances) are used, because ViewStore.state may not always be the same as Store.state.
>
> Sorry, could you elaborate? I'm not quite sure I understand your point.

Sure, my point is that if we're _really_ strict, `ViewStore.state` is not always equal to `Store.state`, precisely because `ViewStore` applies an extra transformation to remove duplicate values from its parent `Store`. This means that using 
`ViewStore.state` to capture the current state of any `Store` **assuming it's the same** doesn't hold in every situation.

In most real-life scenarios, `State`'s `Equatable` conformance will probably be auto-generated so this shouldn't pose any problems as all properties are considered. However, it's possible that a custom `Equatable` conformance on `State` (or equality function) is used that doesn't use all properties (e.g. to not consider all properties of a cell's state for diffing purposes). In such a scenario, accessing `ViewStore.state` could lead an outdated value of state being returned, meaning that any scoped store would for instance be initialized with an outdated initial state as well (if we relied on `ViewStore.state` for it). 

Hope this made it clearer, and sorry if I'm being a bit pedantic 🙏🏼 

> > Probably not as problematic as 1, but ViewStore documentation states it should only be accessed from the main thread. Even though technically in these situations (store scoping operators) we create "dedicated" ViewStores only to pluck out the state (and not drive UI), these can be created from a thread which isn't main, possibly leading to some confusion as to if it is correct to do so.
> 
> Hopefully this was clarified by #692, but in general all interactions with view stores should happen on the same thread, or rather, all sends should happen on the same thread. Scoping can invoke send on the parent, so you must be careful here, as well.

Yes it was 😄 

So far we have always ensured both custom scoping operators and any `send` calls are invoked on the main thread (i.e. from either inside the `ViewStore`'s sink or some UIKit call like Table/Collection View delegate methods, button taps, ...). However in non UI environments extra care should certainly be taken.

stephencelis:
> That would be nice. Unfortunately the UIKit "part" of TCA hasn't been receiving much love lately 💔😇

There are a few UIKit-focused improvements that recently landed, including UIAlertController contributions from the community, as well as improvements to view store publisher behavior/performance. We try to keep the balance of the broader Apple ecosystem/community, though.

> Sure, my point is that if we're _really_ strict, `ViewStore.state` is not always equal to `Store.state`, precisely because `ViewStore` applies an extra transformation to remove duplicate values from its parent `Store`. This means that using
> `ViewStore.state` to capture the current state of any `Store` **assuming it's the same** doesn't hold in every situation.
> 
> In most real-life scenarios, `State`'s `Equatable` conformance will probably be auto-generated so this shouldn't pose any problems as all properties are considered. However, it's possible that a custom `Equatable` conformance on `State` (or equality function) is used that doesn't use all properties (e.g. to not consider all properties of a cell's state for diffing purposes). In such a scenario, accessing `ViewStore.state` could lead an outdated value of state being returned, meaning that any scoped store would for instance be initialized with an outdated initial state as well (if we relied on `ViewStore.state` for it).
> 
> Hope this made it clearer, and sorry if I'm being a bit pedantic 🙏🏼

Sorry, I really am having trouble understanding, and I'm afraid maybe we're talking about different things still. In our opinion, state should ideally only be accessed through the view store and through a subscription to state updates. I'm not quite sure when store state would be accessed and different than view store state, how this would pose a problem, and how exposing state on plain ole stores would be a solution to the problem. I understand that custom equatable conformances could lead to missing updates on sub-state, but isn't that's an existing thing one must be aware of when using `removeDuplicates` in general? Does the proposed PR solve this problem, and how?

Is it OK to close this PR for now and then, if you'd like to continue discussing publicizing `Store` state, start a discussion instead? FWIW it is quite easy for an app to publicize state to be accessible:

```swift
extension Store {
  public var currentState: State { ViewStore(self, removeDuplicates: { _, _ in false }).state }
}
```

However we do _not_ recommend this for the reasons previously brought up in this thread.

p4checo:
> Sorry, I really am having trouble understanding, and I'm afraid maybe we're talking about different things still. In our opinion, state should ideally only be accessed through the view store and through a subscription to state updates. I'm not quite sure when store state would be accessed and different than view store state, how this would pose a problem, and how exposing state on plain ole stores would be a solution to the problem.

No worries. I'm probably looking at this wrong, and it's a far-fetched scenario anyway, which shouldn't be an issue in most situations.

> I understand that custom equatable conformances could lead to missing updates on sub-state, but isn't that an existing thing one must be aware of when using removeDuplicates in general? Does the proposed PR solve this problem, and how?

Yes, we should be aware of that. My point was that `ViewStore.state` is not equivalent to `Store.state`, precisely because it can lead to missing updates on sub-state. The proposed PR "solves" this problem by exposing the `Store.state` which doesn't skip updates on sub-state based on removeDuplicates - making it easier to reason about. But as mentioned before, this shouldn't be an issue in most scenarios, and you already made it clear you only want state to be accessed via view store, so no point walking in circles 😆  

> Is it OK to close this PR for now and then, if you'd like to continue discussing publicizing Store state, start a discussion instead? 
Sure, I can close this PR. I don't think we need to continue discussing this, though 😄 

Thanks for your patience! 🙏🏼

PR #690, opened 2021-08-02
@FocusState case study.

mbrandonw:
Now that the episode is out we can have a case study about `@FocusState`.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-FocusState.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-FocusState.swift
@@ -43,6 +43,7 @@ let focusDemoReducer = Reducer<
 }
 .binding(action: /FocusDemoAction.binding)
 
+#if compiler(>=5.5)
 struct FocusDemoView: View {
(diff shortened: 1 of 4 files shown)


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-FocusState.swift
```diff
@@ -0,0 +1,108 @@
...
+      state.focusedField = .password
+    }
+    return .none
+  }
+}
+.binding(action: /FocusDemoAction.binding)
+
+struct FocusDemoView: View {
+  let store: Store<FocusDemoState, FocusDemoAction>
+  @FocusState var focusedField: FocusDemoState.Field?
```
stephencelis, line 48: Do we need some availability checks for Xcode 12.5?

stephencelis (approved):
Looks good! Just the one comment.

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-FocusState.swift
mbrandonw: good catch, added just now

PR #699, opened 2021-08-06
Add some docs about ViewStore publisher.

mbrandonw:
#698 brought up some interesting points that make it clear we need to flesh out the docs for `viewStore.publisher`.

### DIFF
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -104,10 +104,10 @@ public final class ViewStore<State, Action>: ObservableObject {
   /// This means you can either use the value passed to the closure or you can reach into
   /// `viewStore.state` directly.
   ///
-  /// - Note: that due to a bug in Combine (or feature?), the order you `.sink` on a publisher has
-  /// no bearing on the order the `.sink` closures are called. This means the work performed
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/ViewStore.swift
```diff
@@ -83,7 +83,31 @@ public final class ViewStore<State, Action>: ObservableObject {
...
+  ///   .sink { precondition($0 == viewStore.state) }
+  /// ```
+  ///
+  /// This means you can either use the value passed to the closure or you can reach into
+  /// `viewStore.state` directly.
+  ///
+  /// - Note: that due to a bug in Combine (or feature?), the order you `.sink` on a publisher has
+  /// no bearing on the order the `.sink` closures are called. This means the work performed
+  /// inside `viewStore.publisher.sink` closures should be completely independent of each other.
+  /// Later closures cannot assume that earlier ones have already run.
```
stephencelis, line 110:
```suggestion
  ///   no bearing on the order the `.sink` closures are called. This means the work performed
  ///   inside `viewStore.publisher.sink` closures should be completely independent of each other.
  ///.  Later closures cannot assume that earlier ones have already run.
```

stephencelis (approved):

PR #705, opened 2021-08-10
Add `catchToEffect()` with mapping function

eimantas:
This PR adds a `catchToEffect<T>(_ f: @escaping (Result<Output,Failure>) -> T)` which basically is a `catchToEffect` followed by a map function:

given:

```swift
enum StateAction {
  case opSuccess(OpValue)
  case opFailure(Error)
}
```

then you can do this:

```swift
case .doSomething:
  return env.doingSomething()
    .catchToEffect {
      switch $0 {
      case let .success(value):
        return .opSuccess(value)
      case let .failure(error):
        return .opFailure(error)
      }
    }
```

Since `catchToEffect()` with map is (usually?) used to map result to state actions, there is a place for a shorter operator that accepts action enum cases with associated values that map directly to `Output` and `Failure` types:

```swift
...
case .doSomething:
  return env.doingSomething()
    .catchToEffect(StateAction.opSuccess, StateAction.opFailure)
```

But this is not included in this PR. Given the need - new one can be created.

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -340,6 +340,34 @@ extension Publisher {
       .eraseToEffect()
   }
 
+  /// Turns any publisher into an ``Effect`` that cannot fail by wrapping its output and failure into
+  /// result and then applying passed in function to it.
+  ///
+  /// This is a convenience operator for writing `catchToEffect()` followed by a `map()` .
+  ///
+  /// ```swift
+  /// case .buttonTapped:
+  ///   return fetchUser(id: 1)
+  ///     .catchToEffect {
+  ///       switch $0 {
+  ///       case let .success(response):
+  ///         return ProfileAction.updatedUser(response)
+  ///       case let .failure(error):
+  ///         return ProfileAction.failedUserUpdate(error)
+  ///       }
+  ///     }
+  /// ```
+  ///
+  /// - Parameters:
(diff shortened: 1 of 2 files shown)


AlexisQapa:
while catchToEffect has a nice symmetry to eraseToEffect, I quite like catch(to: Action.result). 
Using default generic parameter could we get to catch(to: .result) ?

mbrandonw:
Thanks for this @eimantas!

One thing to note is that it is more common to bundle the success/failure into a `Result` that is delivered to an action rather than holding onto two separate actions:

```swift
enum Action {
  case buttonTapped
  case response(Result<Int, Error>)
}
...

case .buttonTapped:
  return environment.fetch(id: 1)
    .catchToEffect(Action.response)
```

So, for this reason, I do not think we need another overload of `catchToEffect` that independently maps the success and failure cases.

> while catchToEffect has a nice symmetry to eraseToEffect, I quite like catch(to: Action.result).
> Using default generic parameter could we get to catch(to: .result) ?

This is definitely interesting, but one thing to be wary of is that Combine already has a `.catch` operator, and it also takes a trailing closure. That may make it easier to run into ambiguous operator errors.

We are definitely down to find a better name for this operator, so if y'all wanted to start a new GH discussion to explore ideas we'd be open to it.

mbrandonw (approved):
Going to merge this and make a few small follow up changes.

Thanks again @eimantas!

PR #718, opened 2021-08-17
Add French translation of README.md

nikitamounier:
Hello there! After seeing that an Italian translation of the README.md was added in the latest release, I felt compelled to put my bilingual proficiency to the test. Here's the link if you want to test your French skills too 😁: 

[https://gist.github.com/nikitamounier/0e93eb832cf389db12f9a69da030a2dc](https://gist.github.com/nikitamounier/0e93eb832cf389db12f9a69da030a2dc)

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,10 +437,10 @@ If you want to discuss the Composable Architecture or have a question about how 
 
 The following translations of this README have been contributed by members of the community:
 
+* [French](https://gist.github.com/nikitamounier/0e93eb832cf389db12f9a69da030a2dc)
 * [Korean](https://gist.github.com/pilgwon/ea05e2207ab68bdd1f49dff97b293b17)
 * [Indonesian](https://gist.github.com/wendyliga/792ea9ac5cc887f59de70a9e39cc7343)
 * [Italian](https://gist.github.com/Bellaposa/5114e6d4d55fdb1388e8186886d48958)
-* [French](https://gist.github.com/nikitamounier/0e93eb832cf389db12f9a69da030a2dc)
 
 If you'd like to contribute a translation, please [open a PR](https://github.com/pointfreeco/swift-composable-architecture/edit/main/README.md) with a link to a [Gist](https://gist.github.com)!
 
--- a/README.md
+++ b/README.md
@@ -440,6 +440,7 @@ The following translations of this README have been contributed by members of th
 * [Korean](https://gist.github.com/pilgwon/ea05e2207ab68bdd1f49dff97b293b17)
 * [Indonesian](https://gist.github.com/wendyliga/792ea9ac5cc887f59de70a9e39cc7343)
 * [Italian](https://gist.github.com/Bellaposa/5114e6d4d55fdb1388e8186886d48958)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -440,6 +440,7 @@ The following translations of this README have been contributed by members of th
 * [Korean](https://gist.github.com/pilgwon/ea05e2207ab68bdd1f49dff97b293b17)
 * [Indonesian](https://gist.github.com/wendyliga/792ea9ac5cc887f59de70a9e39cc7343)
 * [Italian](https://gist.github.com/Bellaposa/5114e6d4d55fdb1388e8186886d48958)
+* [French](https://gist.github.com/nikitamounier/0e93eb832cf389db12f9a69da030a2dc)
```
stephencelis, line 443:
```suggestion
* [French](https://gist.github.com/nikitamounier/0e93eb832cf389db12f9a69da030a2dc)
* [Korean](https://gist.github.com/pilgwon/ea05e2207ab68bdd1f49dff97b293b17)
* [Indonesian](https://gist.github.com/wendyliga/792ea9ac5cc887f59de70a9e39cc7343)
* [Italian](https://gist.github.com/Bellaposa/5114e6d4d55fdb1388e8186886d48958)
```

stephencelis (approved):
Thanks so much! Going to make one small change to keep things in order, then will merge!

tgrapperon:
@nikitamounier This is great! 
Please note that one should use "bibliothèque" instead of "librairie" in French (this is a false friend).

nikitamounier:
> @nikitamounier This is great! 
> 
> Please note that one should use "bibliothèque" instead of "librairie" in French (this is a false friend).

I wasn't sure myself, so I did some research, and I found "librairies Swift" as the translation for Swift libraries - for example, [this video](https://youtu.be/QgD31hyHORg) calls Alamofire "une librairie open source". Nevertheless, I don't work on Swift in a french environment, but if you do, I'll take your word for it. 

It seems like there are a couple of other translation errors I made: for instance, I called side effects "effets dérivés", however Vincent Pradeilles helpfully pointed out that the correct term is "effets de bord". If anyone notices any other errors, make sure to tell me and then I'll package up all the corrections in one big PR.

tgrapperon:
I'm not working in a French environment, but I'm almost certain that "Bibliothèque" is the correct term. See https://fr.wikipedia.org/wiki/Bibliothèque_logicielle for example. "Librairie" is the place where you sell books.
In my head, I associate side effects with "effets secondaires" which seems closer to the English meaning. The coined term seems to be ["effet de bord"](https://fr.wikipedia.org/wiki/Effet_de_bord_(informatique)) (unfortunately, as "effet de bord" has a totally different meaning in mathematics and physics).
Using Wikipedia to reach the French version from the English page is a great way to make sure you're keeping the correct context.
I used the PR because I didn't want to pollute the gist with comments that would be outdated as soon as you fix them, but if one can delete them without traces, I can continue there without issues!

nikitamounier:
I see, thanks for the help and the advice – I updated the gist. Let me know if there's anything else!

PR #721, opened 2021-08-19
🆔 Pass the id of the environment in a `forEach` reducer

ferologics:
The change updates the parameters with an `ID`, but perhaps it's better to introduce new `forEach` overload instead of modifying the existing API? I think there's a pretty good case for having `ID` specific environments, similar to the state/action.


iampatbrown:
Hey Fero :) 

Interesting idea. Any chance you could provide a use case for this? I'm not sure when I would pass the ID to the environment? 

An additional overload would be needed so it doesn't conflict with existing code.

ferologics:
@iampatbrown our use case was with the [MatrixSDK](https://github.com/matrix-org/matrix-ios-sdk). It is one of our main `Environment` dependencies of our app. For each chat we construct a [`Timeline`](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/MatrixSDK/Data/MXEventTimeline.h). Each timeline is keyed by an `ID`. Instead of passing the whole SDK object, we only wanted to pass the `Timeline` for a particular chat `ID` to the `ChatEnvironment` (`HomeEnvironment (list of chats) -> ChatEnvironment (particular chat ID)`). This was not possible with the `forEach` at that moment, thus we extended it.

All in all, `Environment`s couldn't be scoped well enough for our use case, and they can benefit from additional information about the `ID`. With this change, we can scope the `GlobalEnvironment`s dependencies in advance to better model the child reducer's `Environment` domain.

stephencelis:
@ferologics Thanks for starting a discussion about this!

After some thought, we were wondering why the dependency doesn't take the id as a parameter, and why instead the environment is dynamically derived from a given id. Generally we like to think of the environment as a "static" set of dependencies, so we're a bit worried about supporting more dynamic behavior like this.

ferologics:
Correct me if I'm wrong @stephencelis, but there's no way for the dependency "to take the id as a parameter" if no id is passed in to the `forEach` environment constructor.

```swift
struct HomeEnvironment {
  let matrix: Matrix
}

struct ChatEnvironment {
  let timeline: Timeline
}

let homeReducer = Reducer.combine...
  chatReducer.forEach( ...
    environment: { homeEnv in
      ChatEnvironment(
        timeline: homeEnv.matrix.timeline(for: id) // 👈 the missing id
      )
    }
  )
...
```

stephencelis:
@ferologics It would mean slightly more boilerplate in the `chatReducer`:

```swift
  return environment.timeline(for: state.id)
    ...
}
```

ferologics:
@stephencelis sure, while passing the id in state is possible, it doesn't seem very ergonomic.

All "pulledback" child reducers of `chatReducer` potentially need to replicate this boilerplate in order to access the appropriate timeline.

And in general, passing in more than is needed defies dependency injection and the notion of **only passing the dependencies that you actually depend on**.

Could you elaborate on why you think static dependencies are a good thing in the first place? as opposed to dynamically scoped.

ferologics:
Closing in favor of the extension -- the relevant changes can be found on [env-id](https://github.com/globekeeper/swift-composable-architecture/tree/env-id) branch.

PR #726, opened 2021-08-19
Clean up `suspend(while:)`

stephencelis:
Realizing we can just lean on the new `Publisher.values` bridge to simplify things.

### DIFF
--- a/Sources/ComposableArchitecture/Beta/Concurrency.swift
+++ b/Sources/ComposableArchitecture/Beta/Concurrency.swift
@@ -218,35 +218,10 @@ import SwiftUI
     /// - Parameter predicate: A predicate on `State` that determines for how long this method
     ///   should suspend.
     public func suspend(while predicate: @escaping (State) -> Bool) async {
-      let cancellable = Box<AnyCancellable?>(wrappedValue: nil)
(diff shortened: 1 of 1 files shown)


mbrandonw (approved):
lol wow. much better

Sources/ComposableArchitecture/Beta/Concurrency.swift
```diff
@@ -218,35 +218,10 @@ import SwiftUI
...
-  private class Box<Value> {
-    var wrappedValue: Value
-
-    init(wrappedValue: Value) {
-      self.wrappedValue = wrappedValue
+      _ = await self.publisher
+        .values
+        .filter { !predicate($0) }
+        .first(where: { _ in true })
     }
```
iampatbrown, line 225: Nice thinking! Could also skip the filter if you wanted.
```swift
 _ = await self.publisher
        .values
        .first(where: { !predicate($0) })
```

stephencelis: Ha good call! Wanna do a quick PR?

iampatbrown: No worries #730

PR #735, opened 2021-08-22
Add convenience overloads of `Reducer.pullback`

roanutil:
I often have reducers composed together that may share state, actions, environment, or a combination of those. It's not a real problem to scope those values in pullback but I prefer not to.

I've placed the overloads in a separate file since they would clutter up the main reducer file. As a result, access to the private `reducer` property isn't allowed.

# Changes
- Add convenience overloads of `Reducer.pullback`

### DIFF
--- a/Sources/ComposableArchitecture/Reducer+Convenience.swift
+++ b/Sources/ComposableArchitecture/Reducer+Convenience.swift
@@ -1,0 +1,90 @@
+import Foundation
+
+extension Reducer {
+    /// - Parameters:
+    ///   - toLocalState: A key path that can get/set `State` inside `GlobalState`.
+    ///   - toLocalAction: A case path that can extract/embed `Action` from `GlobalAction`.
+    /// - Returns: A reducer that works on `GlobalState`, `GlobalAction`, `GlobalEnvironment`.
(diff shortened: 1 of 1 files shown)


stephencelis:
@roanutil Thanks for taking the time to PR!

Unfortunately we have found that overloads in Swift come with a significant cost, not only in compile times but also in mental burden of figuring out which one to pick from autocomplete. For example, SwiftUI's `ForEach` has 10 overloads, and so autocomplete for that API isn't super useful. We are even considering removing a number of overloads that we already have in the library, such as the `.binding` method on `ViewStore`.

Adding overloads to `pullback` also would bring about questions as to adding overloads to other `pullback`-like methods, like `forEach`, etc.

Because of this we're a bit hesitant to add overloads to the library at this time, and would recommend always passing `state`, `action`, and `environment`, even if some of these parameters don't change. Note that you can use `\.self`, `/.self`, and `{ $0 }` to keep things unchanged in a relatively succinct fashion:

```swift
reducer.pullback(state: \.self, action: /.self, environment: { $0 })
```

How's that sound?

roanutil:
That makes sense and it seemed like a change that was likely to not be merged. Thanks for the fast response.

PR #750, opened 2021-08-25
Add CI for DocC warnings

stephencelis:


Makefile
```diff
@@ -21,6 +21,19 @@ test-library:
...
 		-destination platform="$(PLATFORM_WATCHOS)"
 
+DOC_WARNINGS := $(shell xcodebuild clean docbuild \
+	-scheme ComposableArchitecture \
+	-destination platform="$(PLATFORM_MACOS)" \
+	-quiet \
+	2>&1 \
+	| grep 'warning: ' \
+	| sed 's|$(PWD)|.|g' \
+	| tr '\n' '\1')
```
stephencelis, line 31: I couldn't figure out how to make `xcodebuild docbuild` fail on warnings, so doing some gnarly Makefile stuff here to hold onto the warnings and print them. Doing a `tr` dance because of this: https://stackoverflow.com/questions/54067438/how-to-preserve-whitespace-when-saving-command-output-to-a-makefile-variable

PR #754, opened 2021-08-27
Clean up `Store.ifLet`

iampatbrown:
I was playing around with `.publisherScope` and noticed you might be able to shorten `Store.ifLet`. The tests pass but I don't really know the implications of this change.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -331,23 +331,15 @@ public final class Store<State, Action> {
     }
 
     return toLocalState(self.state.eraseToAnyPublisher())
-      .map { localState in
-        let localStore = Store<LocalState, LocalAction>(
-          initialState: localState,
-          reducer: .init { localState, localAction, _ in
-            self.send(fromLocalAction(localAction))
-            localState = extractLocalState(self.state.value) ?? localState
-            return .none
+      .map { initialState in
+        var localState = initialState
+        return self.scope(
+          state: { state in
+            localState = extractLocalState(state) ?? localState
+            return localState
           },
-          environment: ()
+          action: fromLocalAction
         )
-
-        localStore.parentCancellable = self.state
-          .sink { [weak localStore] state in
-            guard let localStore = localStore else { return }
-            localStore.state.value = extractLocalState(state) ?? localStore.state.value
-          }
-        return localStore
       }
       .eraseToAnyPublisher()
   }
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -331,23 +331,11 @@ public final class Store<State, Action> {
     }
 
     return toLocalState(self.state.eraseToAnyPublisher())
-      .map { localState in
-        let localStore = Store<LocalState, LocalAction>(
-          initialState: localState,
-          reducer: .init { localState, localAction, _ in
-            self.send(fromLocalAction(localAction))
-            localState = extractLocalState(self.state.value) ?? localState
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -322,32 +322,17 @@ public final class Store<State, Action> {
     action fromLocalAction: @escaping (LocalAction) -> Action
   ) -> AnyPublisher<Store<LocalState, LocalAction>, Never>
   where P.Output == LocalState, P.Failure == Never {
-
```
iampatbrown, line 325: What are the chances of someone doing something like `.publisherScope(state: { $0.collect(2).compactMap(\.last) })`? Do people use this or is it mainly for `.ifLet`

Sources/ComposableArchitecture/UIKit/IfLetUIKit.swift
```diff
@@ -49,32 +49,19 @@ extension Store {
     then unwrap: @escaping (Store<Wrapped, Action>) -> Void,
     else: @escaping () -> Void = {}
   ) -> Cancellable where State == Wrapped? {
-    let elseCancellable =
-      self
-      .publisherScope(
-        state: { state in
-          state
-            .removeDuplicates(by: { ($0 != nil) == ($1 != nil) })
+    return self.state
```
stephencelis, line 52: Oh nice! So you were able to remove `publisherScope` entirely?

stephencelis:
@iampatbrown The diff looks great to us! One small request would be to revert the `publisherScope` changes, since they are no longer used by `ifLet`. We'll probably deprecate `publisherScope` once this is merged, as we think it's probably not used outside of the current implementation of `ifLet`, is a pretty complicated operation, and we've never found another use for it in a project that wasn't later refactored away.

Sources/ComposableArchitecture/UIKit/IfLetUIKit.swift
iampatbrown: That was the plan. Wasn't exactly sure what the unwrapped store's expected behaviour was though. If it looks okay to you though, that works for me.

stephencelis: The main place we utilize this is the UIKit part of Tic-Tac-Toe demo, and it seemed to perform well there!

iampatbrown:
> revert the `publisherScope` changes, since they are no longer used by `ifLet`. We'll probably deprecate `publisherScope`
No worries :) That was more me experimenting. I tried to find other repos using `.publisherScope` to see  how it gets used with now luck. I wasn't sure if people ever did something like this `.publisherScope(state: { $0.collect(2).compactMap(\.last) })`. Which wouldn't really work atm. It makes more sense now knowing it was mainly used for `ifLet`

Sources/ComposableArchitecture/UIKit/IfLetUIKit.swift
iampatbrown: Yeah I wasn't sure if the `childStores` from `publisherScope` are meant to receive state updates at the same rate as `publisherScope` receives `childStores`. It that makes sense :P

stephencelis:
@iampatbrown We pinged a couple heavy UIKit TCA users to try out this PR branch. Assuming everything works for them we'll merge! Hopefully they'll get around to trying it out soon.

iampatbrown:
Sounds good.

stephencelis (approved):

stephencelis:
Thanks @diederich for testing this in a large app!

PR #755, opened 2021-08-29
Alternative to `CurrentValueSubject` in `ViewStore`

iampatbrown:
This is a proof of concept as a way to provide consistent ordering for StorePublisher subscribers (discussed in #698). @mbrandonw mentioned:

> It would be great if it could be fixed at the library level

so I thought I'd give it a try. Pretty sure it will need to be made thread safe and have some extensive tests/benchmarks added.

I can see the benefits of sticking with CurrentValueSubject and I haven't thoroughly tested this... just checking to see if it's something worth exploring further. I used [CombineExt/CurrentValueRelay](https://github.com/CombineCommunity/CombineExt/blob/main/Sources/Relays/CurrentValueRelay.swift) as a starting point.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/Create.swift
+++ b/Sources/ComposableArchitecture/Internal/Create.swift
@@ -23,7 +23,7 @@
 import Combine
 import Darwin
 
-class DemandBuffer<S: Subscriber> {
+final class DemandBuffer<S: Subscriber> {
   private var buffer = [S.Input]()
   private let subscriber: S
   private var completion: Subscribers.Completion<S.Failure>?
--- a/Sources/ComposableArchitecture/Internal/Create.swift
+++ b/Sources/ComposableArchitecture/Internal/Create.swift
@@ -23,7 +23,7 @@
 import Combine
 import Darwin
 
-private class DemandBuffer<S: Subscriber> {
+class DemandBuffer<S: Subscriber> {
   private var buffer = [S.Input]()
   private let subscriber: S
   private var completion: Subscribers.Completion<S.Failure>?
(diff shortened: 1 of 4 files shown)


Sources/ComposableArchitecture/Internal/Create.swift
```diff
@@ -23,7 +23,7 @@
 import Combine
 import Darwin
 
-private class DemandBuffer<S: Subscriber> {
+class DemandBuffer<S: Subscriber> {
```
iampatbrown, line 26: This was only so I could use the existing `DemandBuffer` and would have to be changed back or marked as final

stephencelis:
@iampatbrown Thanks for the PR! @mbrandonw and I just chatted about it and think it's a simple enough addition that we think it's worth merging, especially if it makes things a bit more deterministic than the tools Apple ships. The existing benchmarks (simple as they are) don't appear to be negatively affected.

We also think that thread safety is probably not necessary due to the fact that it is only used in the store, which is already designed to not be thread safe.

If you think this is ready, please feel free to remove the draft status and we'll look to merge!

iampatbrown:
> We also think that thread safety is probably not necessary due to the fact that it is only used in the store, which is already designed to not be thread safe.

I thought this might have been the case as well. My main concern was if people were using `viewStore.publisher.subscribe(on:)`. Had a feeling this might be impacted. But yeah, the docs are pretty clear about not being thread safe though. I guess we can revert back fairly easily if there are any problems. 

Any preference on the file structure and naming?

stephencelis:
Ha didn't mean to leave you hanging here!

> Any preference on the file structure and naming?

Nope! What you have looks great. If you think this is ready, please move it into review and we'll get it merged :smile:

iampatbrown:
@stephencelis Hahaha. No worries. Did you give it a test drive? Pretty sure it was all good when we originally looked at it.

stephencelis:
@iampatbrown Yup! All looked good to us and seems to benchmark about the same 😄

stephencelis (approved):

PR #757, opened 2021-09-01
NavigateAndLoad: Cancel loading on dismiss

filblue:


Examples/CaseStudies/SwiftUICaseStudies/03-Navigation-NavigateAndLoad.swift
```diff
@@ -37,17 +37,18 @@ let navigateAndLoadReducer =
...
       NavigateAndLoadState, NavigateAndLoadAction, NavigateAndLoadEnvironment
     > { state, action, environment in
+      struct CancelId: Hashable {}
       switch action {
       case .setNavigation(isActive: true):
         state.isNavigationActive = true
         return Effect(value: .setNavigationIsActiveDelayCompleted)
-          .delay(for: 1, scheduler: environment.mainQueue)
-          .eraseToEffect()
+          .deferred(for: 1, scheduler: environment.mainQueue)
```
stephencelis, line 45: Any reason for this change? I think we can stick with Combine's `delay` to keep this example simple since it should behave the same.
```suggestion
          .delay(for: 1, scheduler: environment.mainQueue)
          .eraseToEffect()
```

Examples/CaseStudies/SwiftUICaseStudies/03-Navigation-Sheet-PresentAndLoad.swift
```diff
@@ -35,17 +35,18 @@ let presentAndLoadReducer =
...
       PresentAndLoadState, PresentAndLoadAction, PresentAndLoadEnvironment
     > { state, action, environment in
+      struct CancelId: Hashable {}
       switch action {
       case .setSheet(isPresented: true):
         state.isSheetPresented = true
         return Effect(value: .setSheetIsPresentedDelayCompleted)
-          .delay(for: 1, scheduler: environment.mainQueue)
-          .eraseToEffect()
+          .deferred(for: 1, scheduler: environment.mainQueue)
```
stephencelis, line 43: Same here:
```suggestion
          .delay(for: 1, scheduler: environment.mainQueue)
          .eraseToEffect()
```

stephencelis (commented):
Thanks for the PR! We're definitely down to adding cancellation to case studies that fire off effects like this. Just one note on preferring `delay` to `deferred` in these cases.

Would you like to add cancellation to the other loadable studies as well?

Examples/CaseStudies/SwiftUICaseStudies/03-Navigation-NavigateAndLoad.swift
filblue: Just tried to save on the `.eraseToEffect()` line. )

filblue:
@stephencelis I've also added cancellation to `UIKitCaseStudies/NavigateAndLoad.swift`. I believe LoadThenNavigate ones don't need this, right?

stephencelis:
> @stephencelis I've also added cancellation to `UIKitCaseStudies/NavigateAndLoad.swift`. I believe LoadThenNavigate ones don't need this, right?

I think they do. If you back out of those studies while the request is in flight, they'll still write to state, which means if you drill back in it'll immediately double-drill you to the loaded state.

filblue:
@stephencelis thank you for clarification, updated.

stephencelis (approved):
Thanks!

PR #761, opened 2021-09-03
Custom ObservableObject to fix animation

iampatbrown:
Do you guys mind testing this out. Got it working on my end. Not sure if it's a good idea though. There might be a better way to achieve the same thing as well.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Binding.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Binding.swift
@@ -245,6 +245,7 @@ extension ViewStore {
     dynamicMember keyPath: WritableKeyPath<State, BindableState<Value>>
   ) -> Binding<Value>
   where Action: BindableAction, Action.State == State, Value: Equatable {
+    self.observedViewStore?.projectedBinding(for: keyPath) ??
     self.binding(
       get: { $0[keyPath: keyPath].wrappedValue },
       send: { .binding(.set(keyPath, $0)) }
--- a/Sources/ComposableArchitecture/SwiftUI/Binding.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Binding.swift
@@ -245,7 +245,7 @@ extension ViewStore {
     dynamicMember keyPath: WritableKeyPath<State, BindableState<Value>>
   ) -> Binding<Value>
   where Action: BindableAction, Action.State == State, Value: Equatable {
-    self.observedViewStore?.projectedBinding(for: keyPath) ??
+    self.projectedBinding(for: keyPath) ??
     self.binding(
       get: { $0[keyPath: keyPath].wrappedValue },
       send: { .binding(.set(keyPath, $0)) }
--- a/Sources/ComposableArchitecture/SwiftUI/ObservedViewStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/ObservedViewStore.swift
@@ -14,13 +14,17 @@ public struct ObservedViewStore<State, Action>: DynamicProperty {
     $wrappedValue
   }
 
-  func projectedBinding<Value>(for keyPath: WritableKeyPath<State, BindableState<Value>>) -> Binding<Value> where Action: BindableAction, Action.State == State, Value: Equatable {
+  func projectedBinding<Value>(for keyPath: WritableKeyPath<State, BindableState<Value>>) -> Binding<Value>
+    where Action: BindableAction, Action.State == State, Value: Equatable
+  {
     return $wrappedValue[binding: keyPath]
   }
 }
 
-extension ViewStore  {
-  func projectedBinding<Value>(for keyPath: WritableKeyPath<State, BindableState<Value>>) -> Binding<Value>? where Action: BindableAction, Action.State == State, Value: Equatable {
+extension ViewStore {
+  func projectedBinding<Value>(for keyPath: WritableKeyPath<State, BindableState<Value>>) -> Binding<Value>?
+    where Action: BindableAction, Action.State == State, Value: Equatable
+  {
     self.observedViewStore?.projectedBinding(for: keyPath)
   }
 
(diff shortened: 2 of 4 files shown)


iampatbrown:
@mbrandonw @stephencelis marking as draft for now. Will test this out more over the weekend.

iampatbrown:
@mbrandonw @stephencelis have you been able to get bindings working properly inside `TextField`. I can't seem to do things like `state.text = String(state.text.prefix(5))`. I know this is a bit of SwiftUI issue though... was just curious...

stephencelis:
Hm, this was auto-closed when the upstream branch merged. Wanna reopen against `main`?

> @mbrandonw @stephencelis have you been able to get bindings working properly inside `TextField`. I can't seem to do things like `state.text = String(state.text.prefix(5))`. I know this is a bit of SwiftUI issue though... was just curious...

Once upon a time doing `WithViewStore(self.store, removeDuplicates: { _, _ in false }) { … }` helped here, though I think even that doesn't work anymore. I think it's probably better to validate on de-focus these days.

iampatbrown:
There's no Composable Formatters in the pipeline is there?... 
It's just bugging me... there's no real way for us to detect this is there? As in the Binding State ends up out of sync with the store.

stephencelis:
@iampatbrown Can you illustrate the problem? We're not sure it's the binding getting out of sync with the store so much as the `TextField`'s state, which could be modeling itself internally in a different way. We haven't dug too deep into it.

As for composable formatters, you could say that ;) Bidirectional parsing is a formatter in disguise, so one could be passed to the formatter-based `TextField` initializer. That works differently, though, as it allows any input and only does the parsing/printing when the value is committed on de-focus.

iampatbrown:
Yeah it's definitely `TextField`'s internal state. For example in [0159-safer-conciser-forms-pt2](https://github.com/pointfreeco/episode-code-samples/tree/main/0159-safer-conciser-forms-pt2) there's
```swift
case .binding(\.$displayName):
    state.displayName = String(state.displayName.prefix(16))
    return .none
```
I've tweaked the code slightly to get it to work as `.binding()` at the end of the reducer is going to override this state change... but you end up in a situation like this

*Profile*
```
My Really Long Display Nam 
```

And then I press `e` which changes the view to

*Profile*
```
My Really Long Display Name
```
and the debug output is
```
received action:
  SettingsAction.binding(
    BindingAction.set(
      WritableKeyPath<SettingsState, BindableState<String>>,
      "My Really Long Display Name"
    )
  )
(No state changes)
```
It just seems like a fairly common thing people might want to do. So I was thinking it would be good to have a recomended way of handling it.

iampatbrown:
Actually the following seems to work fine. 

```swift
struct SettingsState: Equatable {
  @BindableState var displayName = ""
}

enum SettingsAction: BindableAction, Equatable {
  case binding(BindingAction<SettingsState>)
  case validate(PartialKeyPath<SettingsState>)
  case resetButtonTapped
}

let settingsReducer = Reducer<SettingsState, SettingsAction, Void> { state, action, _ in
  switch action {
  case .validate(\.displayName):
    state.displayName = String(state.displayName.prefix(6))
    return .none
  case .binding:
    return .none
  case .validate:
    return .none
  case .resetButtonTapped:
    state = .init()
    return .none
  }
}.binding()

struct ContentView: View {
  let store: Store<SettingsState, SettingsAction>

  var body: some View {
      Form {
        Section(header: Text("Profile")) {
          TextField("Display name", text: viewStore.$displayName)
            .onChange(of: viewStore, keyPath: \.displayName, send: .validate(\.displayName))
        }

        Button("Reset") {
          viewStore.send(.resetButtonTapped)
        }
      }
      .navigationTitle("Settings")
    }
  }
}
```

```swift
struct OnViewStoreChangeModifier<State, Action, LocalState>: ViewModifier where LocalState: Equatable {
  let viewStore: ViewStore<State, Action>
  let toLocalState: (State) -> LocalState
  let action: Action

  func body(content: Content) -> some View {
    content.onChange(of: toLocalState(viewStore.state)) { _ in
      viewStore.send(action)
    }
  }
}

extension View {
  func onChange<State, Action, LocalState>(
    of viewStore: ViewStore<State, Action>,
    state toLocalState: @escaping (State) -> LocalState,
    send action: Action
  ) -> some View where LocalState: Equatable {
    self.modifier(
      OnViewStoreChangeModifier(
        viewStore: viewStore,
        toLocalState: toLocalState,
        action: action
      )
    )
  }

  func onChange<State, Action, LocalState>(
    of viewStore: ViewStore<State, Action>,
    keyPath: WritableKeyPath<State, LocalState>,
    send action: Action
  ) -> some View where LocalState: Equatable {
    self.modifier(
      OnViewStoreChangeModifier(
        viewStore: viewStore,
        toLocalState: { $0[keyPath: keyPath] },
        action: action
      )
    )
  }
}
```

Hadn't thought about using 
```swift
case validate(PartialKeyPath<SettingsState>)
```

iampatbrown:
Might be better to pass the `Store` though...
```swift
TextField("Display name", text: viewStore.$displayName)
  .onChange(of: self.store.scope(state: \.displayName), send: .validate(\.displayName))
```
```swift
struct OnViewStoreChangeModifier<State, Action>: ViewModifier where State: Equatable {
  let store: Store<State, Action>
  let action: Action

  func body(content: Content) -> some View {
      content.onChange(of: viewStore.state) { _ in
        viewStore.send(action)
      }
    }
  }
}

extension View {
  func onChange<State, Action>(
    of store: Store<State, Action>,
    send action: Action
  ) -> some View where State: Equatable {
    self.modifier(OnViewStoreChangeModifier(store: store, action: action))
  }
}
```

iampatbrown:
Actually.... I'd probably go for something like this...
```swift
WithCheckedBinding(viewStore.$displayName) {
  TextField("Display name", text: $0)
}
```

```swift
public struct WithCheckedBinding<Value, Content>: View where Value: Equatable, Content: View {
  @Binding var base: Value
  @State var expected: Value
  let content: (Binding<Value>) -> Content

  public init(_ base: Binding<Value>, @ViewBuilder content: @escaping (Binding<Value>) -> Content) {
    self._base = base
    self._expected = State(initialValue: base.wrappedValue)
    self.content = content
  }

  var checkedBinding: Binding<Value> {
    Binding { expected }
    set: { newValue, transaction in
      expected = newValue
      $base.transaction(transaction).wrappedValue = newValue
    }
  }

  var isValid: Bool { base == expected }
  
  public var body: some View {
    content(checkedBinding)
      .onChange(of: isValid) { isValid in
        if !isValid { expected = base }
      }
  }
}
```
That way the logic can stay inside the reducer.

PR #770, opened 2021-09-06
ObservedViewStore for projectedBindings

iampatbrown:
Wraps the ObservedObject for the ViewStore so the projectedValue can be accessed for use with bindings. This improves the behaviour of bindings with animations.


iampatbrown:
@stephencelis @mbrandonw turns out this could be made a lot simpler... I haven't profiled this yet. Can't image it's too bad though.

Sources/ComposableArchitecture/SwiftUI/Binding.swift
```diff
@@ -265,11 +265,18 @@ import SwiftUI
...
       dynamicMember keyPath: WritableKeyPath<State, BindableState<Value>>
     ) -> Binding<Value>
-    where Action: BindableAction, Action.State == State, Value: Equatable {
-      self.binding(
-        get: { $0[keyPath: keyPath].wrappedValue },
-        send: { .binding(.set(keyPath, $0)) }
-      )
+      where Action: BindableAction, Action.State == State, Value: Equatable
+    {
+      ObservedObject(wrappedValue: self).projectedValue[bindable: keyPath]
```
stephencelis, line 270: Could this trick be employed in `ViewStore.binding(get:send:)` instead? Then we could get it everywhere for free I think...

stephencelis (commented):
Thanks for looking into a workaround for this! Sorry we haven't had a chance to dive in yet, but looks like you've made some improvements along the way :smile:

Just one question about generalizing:

Sources/ComposableArchitecture/SwiftUI/Binding.swift
iampatbrown: Should be able to. I'll test it out.

iampatbrown: Hmmm... I think `subscript(bindable:)` is playing a large part here. Making it difficult with closures. Really want a simple solution for `ViewStore.binding(get:send:)`... back to the drawing board...

iampatbrown: @stephencelis thoughts on something like this?
```swift
final class WrappedState<LocalState>: ObservableObject {
  let get: () -> LocalState
  let set: (LocalState) -> Void

  init<State, Action>(
    viewStore: ViewStore<State, Action>,
    get toLocalState: @escaping (State) -> LocalState,
    send localStateToViewAction: @escaping (LocalState) -> Action
  ) {
    let initialState = toLocalState(viewStore.state)
    self.get = { [weak viewStore] in
      viewStore.flatMap { toLocalState($0.state) } ?? initialState
    }
    self.set = { [weak viewStore] newValue in
      viewStore?.send(localStateToViewAction(newValue))
    }
  }

  var state: LocalState {
    get { self.get() }
    set { self.set(newValue) }
  }

  var binding: Binding<LocalState> {
    ObservedObject(wrappedValue: self).projectedValue.state
  }
}
```
making `ViewStore.binding(get:send:)` 
```swift
public func binding<LocalState>(
  get: @escaping (State) -> LocalState,
  send localStateToViewAction: @escaping (LocalState) -> Action
) -> Binding<LocalState> {
  WrappedState(viewStore: self, get: get, send: localStateToViewAction).binding
}
```

stephencelis: @iampatbrown Looks cool! The only thing that pops out is ownership, like when is `viewStore` unavailable to this observable object and when would it fall back to initial state. Did you encounter a situation that ran that code path?

iampatbrown: I don't think I'd end up in a situation where the binding would exist without the `viewStore`. I was just being over cautious. Would you recommend using a strong reference here?

iampatbrown: I don't overly love the idea of wrapper class like this... so far it's all I've got that has the expected binding behaviour

stephencelis: Ha I have the same knee-jerk reaction, but it's a lean class, and it's the best way so far to reproduce the behavior (thanks!). Ideally we benchmark it, but I doubt it adds much overhead.

I think the ideal would be for SwiftUI to document and offer APIs to derive custom bindings, but in lieu of that, reliable/predictable workarounds are definitely valuable.

iampatbrown: Cool. I'd be interested to learn how you'd go about bench marking something like this.

So far I've just isolated the implementation and compared some basic tasks using `Time Profiler` and `swift-benchmark`. I repeated the process with the TCA implementations as well. 

The current `Binding.init` seems to perform the best, but that's mainly because it isn't accessing the `ObservableObjectLocation`. The `WrappedState` method preforms pretty similar to using the native `ObservedObject` bindings. In fact, `store.binding(get: { $0.text }, send: { .textChanged($0) })` appeared better than `observedStore.projectedValue.state.text`. Probably to do with keyPath performance vs closures. 

Anyway, I'll leave this one for now. If you need me to do anything, just let me know.

stephencelis:
@iampatbrown Just pushed a branch that I think can further simplify things by leveraging a key path subscript:

https://github.com/pointfreeco/swift-composable-architecture/compare/binding-fix

Key path subscripts require their parameters be hashable, but we can use a private wrapper to work around the issue, since we don't care about hashability.

If these changes look good to you, feel free to bring them into this PR!

iampatbrown:
@stephencelis Awesome. I wasn't sure if this was an okay thing to do :P Looks nice!

stephencelis:
For posterity, this feedback captures the context of the issue this PR resolves: https://gist.github.com/stephencelis/fddee6ef2420a525e23fe9c07cbcaefb

While this PR gets view store bindings to behave as consistently as bindings derived from simple observable objects, that doesn't mean it isn't without its consequences. Observable object bindings (as of iOS 15) don't always behave as well as state bindings, or as well as how we vended bindings prior to merging this PR. For example: https://gist.github.com/stephencelis/bd7051b736bdab02798e670930d3ac30

iampatbrown:
The quest continues.... @stephencelis think it's worth rolling this back? Is this also related to the double write (#800) issue ie. double writes are with `@ObservedObject/@StateObject` but `@State`?

stephencelis:
@iampatbrown I think we’re good to keep the change. The behavior is the same as observable objects, so hopefully Apple fixes things upstream. In the meantime people can restore these kinds of animations manually.

iampatbrown:
> The behavior is the same as observable objects.

Good point.

stephencelis:
> Is this also related to the double write (#800)

Oh also, sorry to not answer this. I think it's only related so much as both are Apple bugs 😇

iampatbrown:
Hahaha. All good. Curious if your bug reports have been looked into much....

PR #772, opened 2021-09-06
Add threading warning to Store.send

IanKeen:
Any async Effect (including fireAndForget) could result in a runtime crash when updating internal state like `effectCancellables`
This adds a debug warning to let users know


mbrandonw (approved):
Awesome addition @IanKeen!

We have a few tweaks we want to make to this idea, but we'll merge your PR and then open a new PR with the changes.

Thanks!

PR #777, opened 2021-09-07
Conform BindableState to CustomDebugStringConvertible

### LINKED ISSUE #776: BindableState properties dump output in snapshots is not ideal
**Describe the bug**
I've just updated our app to use 0.26.0 so we can take advantage of the new Binding features. Its great, I love it. 👍🏻

Unfortunately, there is one minor nit - if you have snapshot tests that take a snapshot of your app state using `swift-snapshot-testing` and the `.dump` format, there has been a regression in usefulness of the output.

Before:

```
AppState
   - someString: ""
```

After:

```
AppState
   - _someString: Swift.String
```

**Expected behavior**
The underscored property name is understandable because of the way property wrappers work however I would expect the value to be displayed as it was before.

I was able to work around this by defining a local extension that conforms `BindableState` to `CustomDebugStringConvertible`:

```swift
extension BindableState: CustomDebugStringConvertible where Value: CustomDebugStringConvertible {
    public var debugDescription: String {
        wrappedValue.debugDescription
    }
}
```

It looks like adding this conformance to the library should fix this - I'm happy to open a PR.

**Environment**
  - swift-composable-architecture version [e.g. 0.26.0]
  - Xcode 12.5

lukeredpath:
This ensures that bindable state properties are displayed in a useful way when using `.dump` snapshot tests on your app state.

Fixes #776

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Binding.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Binding.swift
@@ -226,6 +226,12 @@ import SwiftUI
     }
   }
 
+  extension BindableState: CustomDebugStringConvertible where Value: CustomDebugStringConvertible {
+    public var debugDescription: String {
+        self.wrappedValue.debugDescription
+    }
+  }
+
   /// An action type that exposes a `binding` case for the purpose of reducing.
   ///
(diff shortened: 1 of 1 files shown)


stephencelis (approved):
Nice catch! Thanks!

PR #779, opened 2021-09-08
Fix unfunctional cancel button handlers in AlertsAndActionSheets case study

filblue:
… study


stephencelis:
@filblue Thanks for the PR! I think because this case study doesn't attach any logic to cancellation we could probably simplify by removing those actions entirely. Wanna give that a shot instead?

filblue:
@stephencelis I did consider removing the actions, but was not sure if they intentionally exist there for educational purposes. Updated.

stephencelis:
Thanks! Yeah I think we can add them back if they demonstrably do something in the future, but right now it seems like we're encouraging an action per button even when they do nothing.

PR #781, opened 2021-09-09
Fix "BindableState: Codable" strategies

stephencelis:
[This forum post](https://forums.swift.org/t/propery-wrapper-decoding-difference-between-singlevaluecontainer-and-init-from-decoder/51918) reminded me that we had to work around a bug in Tagged:

https://github.com/pointfreeco/swift-tagged/blob/73f93e1fb534b88ba467ae90f8730167e804a32d/Sources/Tagged/Tagged.swift#L67-L75

I'm not sure if there's a bug filed for this behavior or not.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Binding.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Binding.swift
@@ -204,13 +204,23 @@ import SwiftUI
 
   extension BindableState: Decodable where Value: Decodable {
     public init(from decoder: Decoder) throws {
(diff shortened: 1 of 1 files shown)


mbrandonw (approved):
Nice, good catch!

Should we also make `BindableState` conform to `Comparable` when `Value: Comparable`? Any other protocols to synthesize?

stephencelis:
> Any other protocols to synthesize?

I think _maybe_ `Sendable`, but can't think of others. I'm sure someone will let us know soon 😅

PR #784, opened 2021-09-11
Update binding examples

iampatbrown:
Was going through the examples and updated what I could. Saw you were doing tic-tac-toe so left that out.

### DIFF
--- a/Examples/Todos/Todos/Todos.swift
+++ b/Examples/Todos/Todos/Todos.swift
@@ -8,8 +8,8 @@ enum Filter: LocalizedStringKey, CaseIterable, Hashable {
 }
 
 struct AppState: Equatable {
-  var editMode: EditMode = .inactive
-  var filter: Filter = .all
+  @BindableState var editMode: EditMode = .inactive
+  @BindableState var filter: Filter = .all
   var todos: IdentifiedArrayOf<Todo> = []
 
   var filteredTodos: IdentifiedArrayOf<Todo> {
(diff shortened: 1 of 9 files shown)


stephencelis:
Also, please share the existing diff on the discussions if you think it would be interesting for folks!

iampatbrown:
I thought that might have been the case :) that's why I kept the commits seperate... just in case :P Good call.

stephencelis:
> With the todo example, I opted to leave `.send(.checkBoxToggled)` over using `wrappedValue.toggle()`. Do you have a preference with that one?

That's the right decision I think!

iampatbrown:
Reverted all except Todos example

Examples/Todos/Todos/Todos.swift
```diff
@@ -142,15 +142,40 @@ struct AppView: View {
...
+      self = .binding(bindingAction.pullback(\AppState.view))
+    case .clearCompletedButtonTapped:
+      self = .clearCompletedButtonTapped
+    case let .delete(indexSet):
+      self = .delete(indexSet)
+    case let .move(source, destination):
+      self = .move(source, destination)
+    }
+  }
+}
```
stephencelis, line 177: You know what, I completely scanned over this bit when reviewing earlier. Let us think on it a bit more. View state throws a bit of a wrench into something that's otherwise simple...I think we'd probably want to either forego view state entirely in the Todos app (it's simple enough that the scoping we're doing is doubtfully having much of an affect on performance), or maybe avoid introducing binding helpers here... 😔

iampatbrown: hahahah abort mission!!! All good. I'll close this for now. There's always the option to introduce a seperate example using pullback if needed. Thanks for the feedback.

stephencelis: Thanks again for exploring! Hopefully we come up with a better way to work with TCA bindings and view domains in the future that eliminates the extra boilerplate.

PR #786, opened 2021-09-13
Remove #if DEBUG/#endif surrounding `Effect.failing`

andreyz:
Deletes unnecessary preprocessor macros around `Effect.failing`, obsoleted when `XCTFail` got a Release build stub in XCTDynamicOverlay.

It'll also help resolve [related issue in Composable-Core-Location package](https://github.com/pointfreeco/composable-core-location/issues/14)


mbrandonw (approved):
Thanks!

PR #787, opened 2021-09-13
Convenience to allow better re-use of the AlertState<Action>.ButtonAction

kdeda:
I'm developing a Mac application using TCA.
I also have a custom SheetView that works in conjunction with AlertState.Button
Upon upgrading to 0.27.0 I noticed that I was unable to extend
```
extension AlertState.Button {
    public func toSwiftUI(send: @escaping (Action) -> Void) -> SwiftUI.Button<Text> {
``` 
due to the fact that AlertState<Action>.ButtonAction.ActionType is private

I chose to provide access to the buttonAction() rather than make all those inner types public.


stephencelis:
> I'm developing a Mac application using TCA.
> I also have a custom SheetView that works in conjunction with AlertState.Button

@kdeda Thanks for the PR! Can you share a bit more about your use case? Maybe a before/after of your helper or pseudocode that looks similar to it? Would love to know how these APIs are being used in practice, and it could inform their design.

kdeda:
@stephencelis I'v added a new commit that contains a demo macOS app. 
../swift-composable-architecture/Examples/ModalSheet/DemoApp/DemoApp.xcodeproj

The simple app will show the partially completed SheetView which makes use of the new buttonAction() to create a fancy macOS modal sheet.
The SheetView on this demo is tailored to be used by my existing app WhatSize and so it still in flux.

kdeda:
I have rebased with the latest code changes you guys put up.
Refactored.
And the demo app is running.
The desired change at this point is on Sources/ComposableArchitecture/SwiftUI/Alert.swift
```
    /// Convenience to allow third parties to build on top of this type
     /// For example a macOS app could use this to build a sheet and normal SwiftUI.Button<Text> instead
     public func buttonAction(send: @escaping (Action) -> Void) -> () -> Void {

```

PR #794, opened 2021-09-14
Update AlertState to use new APIs

stephencelis:
Assuming the new APIs are locked-in, this updates our existing APIs to consume the new ones and not the deprecated old ones. In the process I introduced the ability to do 3+ buttons on alerts.

Some food for thought:

  * Should `ActionSheetState` be renamed to `ConfirmationDialogState`? Pretty wordy...
  * Or should it all be `AlertState` and we could introduce something like:
    ```swift
    enum AlertStateMode {
      case alert
      case confirmationDialog(titleVisibility: Visibility)
    }
    ```


Sources/ComposableArchitecture/SwiftUI/Alert.swift
```diff
@@ -177,10 +181,19 @@ public struct AlertState<Action> {
     }
   }
 
-  public enum ButtonType {
-    case cancel(label: TextState?)
-    case `default`(label: TextState)
-    case destructive(label: TextState)
+  public enum ButtonRole {
```
stephencelis, line 184: This is breaking...but the rename seemed worth it to unify with Apple's APIs...

Sources/ComposableArchitecture/Internal/Binding+IsPresent.swift
```diff
@@ -0,0 +1,16 @@
...
+
+extension Binding {
+  func isPresent<Wrapped>() -> Binding<Bool> where Value == Wrapped? {
+    .init(
+      get: { self.wrappedValue != nil },
+      set: { isPresent in
+        if !isPresent {
+          withTransaction(self.transaction) {
+            self.wrappedValue = nil
+          }
```
iampatbrown, line 11: `withTransaction` might not be needed here because `self.transaction` should already be applied to `self.wrappedValue`. It might even override other `withTransaction/withAnimation` calls higher up.
In the event you wanted to add support for `.transaction()/.animation()` to this binding, I've found the best way to maintain the animation precedence would be
```swift
extension Binding {
  func isPresent<Wrapped>() -> Binding<Bool> where Value == Wrapped? {
    .init(
      get: { self.wrappedValue != nil },
      set: { isPresent, transaction in
        guard !isPresent else { return }
        self.transaction(transaction).wrappedValue = nil
      }
    )
  }
}
```

stephencelis: Interesting! Good to know, thanks @iampatbrown!

Sources/ComposableArchitecture/SwiftUI/ActionSheet.swift
```diff
@@ -118,18 +118,41 @@ public struct ActionSheetState<Action> {
   public var buttons: [Button]
   public var message: TextState?
   public var title: TextState
+  public var titleVisibility: Visibility
 
   public init(
     title: TextState,
+    titleVisibility: Visibility = .automatic,
```
mbrandonw, line 125: should we do another `init` that doesn't take `titleVisibility` and mark its availability?

Sources/ComposableArchitecture/SwiftUI/Alert.swift
```diff
@@ -213,8 +243,7 @@ extension AlertState: CustomDumpReflectable {
       children: [
         "title": self.title,
         "message": self.message as Any,
-        "primaryButton": self.primaryButton as Any,
-        "secondaryButton": self.secondaryButton as Any,
+        "buttons": self.buttons,
```
mbrandonw, line 246: let's add tests for the custom dump of `AlertState` and `ConfirmationDialogState`

mbrandonw (approved):

Issue #788, opened 2021-09-13
Main thread checks on iOS 15 - unexpected result


lukeredpath:
**Describe the bug**
I'm currently doing some smoke testing of our app on iOS 15 using the latest Xcode beta (beta 5) in preparation for the forthcoming final release.

We're using TCA 0.27 and it looks like the recently added main thread checks aren't working as expected, I'm getting output like this:

```
Warning:

The store was interacted with on a thread that is different from the thread the store was created on:

"ViewStore.send([redacted action])" was called on the wrong thread. Make sure that "ViewStore.send" is always called on the initial thread.

  Initial thread: <NSThread: 0x600000162680>{number = 29, name = main}
  Current thread: <_NSMainThread: 0x6000001b40b0>{number = 1, name = main}

The "Store" class is not thread-safe, and so all interactions with an instance of "Store" (including all of its scopes and derived view stores) must be done on the same thread.
---

Caught debug breakpoint. Type "continue" ("c") to resume execution.
```

The weird thing here is I know that the first action sent to my store would have been on the main thread because we fire an action in `application(_:didFinishLaunching)` so I don't know how the initial thread is _not_ the main thread (although strangely it has the name "main").

**To Reproduce**
If I can put together a reproducer I will uploaded it here.

**Expected behavior**
When the first action in the app is sent on the main thread, other actions sent from the main thread should not trigger this breakpoint.

**Environment**
  - TCA 0..27
  - Xcode 13b5
  - iOS 15.

lukeredpath:
Hmm, interesting. So our root store is created by a coordinator object that is initialised as a singleton, (e.g. `static let liveCoordinator = Coordinator()`) - the store is created in the `Coordinator`'s init.

We first reference this live coordinator in the `application(_:didFinishLaunching)` method although due to the static I'm not 100% of the lifecycle.

IIRC, "thread 29" appeared to be owned by `NSURLConnection` - I don't know if that provides a clue. We do reference our app coordinator in `application(_:handleEventsForBackgroundSession:completionHandler:)` too but I'd be surprised if this method was even called in my very limited simulator smoke testing (I hadn't even backgrounded the app). Very odd that this thread had the name "main" though. I've always assumed the main thread is *always* Thread 1.

I will keep an eye on this and see if it happens again as we continue to test. It doesn't help that we are still running a more than month old beta. 🤞🏻 for an updated Xcode this week.

mbrandonw:
[...]

PR #803, opened 2021-09-16
Thread messaging

stephencelis:
Should avoid confusion like that of #788.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -121,6 +121,7 @@ public final class Store<State, Action> {
   private let reducer: (inout State, Action) -> Effect<Action, Never>
   private var bufferedActions: [Action] = []
   #if DEBUG
+    private let id = UUID()
     private let initialThread = Thread.current
   #endif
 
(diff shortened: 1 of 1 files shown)


mbrandonw (approved):

Sources/ComposableArchitecture/Store.swift
```diff
@@ -429,22 +431,28 @@ public final class Store<State, Action> {
...
-          An effect returned from the action "\(debugCaseOutput(action))" completed on the \
-          wrong thread. Make sure to use ".receive(on:)" on any effects that execute on background \
+          An effect returned from the action "\(debugCaseOutput(action))" completed on the wrong \
+          thread. Make sure to use ".receive(on:)" on any effects that execute on background \
           threads to receive their output on the same thread the store was created on.
           """
 
+      case .scope:
+        message = """
+          "ViewStore.scope" was called on the wrong thread. Make sure that "ViewStore.scope" is \
```
lukeredpath, line 441: Shouldn't this be `Store.scope`?

mbrandonw: whoops, good catch! just pushed a fix.

lukeredpath:
Thank you for this! :)

PR #805, opened 2021-09-16
Restrict thread checks to the main thread

tgrapperon:
This is some implementation for the middle ground solution I've proposed in #802 .
`Store(initialState:...` should always run on the main thread and thread safety is checked at runtime.
`Store.uncheckedThreadStore(initialState:...` are stores opting out from thread safety checks. This initializer should be used for background stores.


tgrapperon:
Maybe `uncheckedThreadStore` should be renamed `withUncheckedThread`.

mluisbrown:
This looks like it would definitely solve the problems [we are seeing](https://github.com/pointfreeco/swift-composable-architecture/discussions/802#discussioncomment-1343246) 👍

tgrapperon:
This probably should be closed in favor of #807.

PR #807, opened 2021-09-17
Use dispatch queue specific key to check thread safety

tgrapperon:
This is an alternative way to handle thread checks, as proposed by @mluisbrown in #802, and an alternative to #805 that should work for arbitrary queues.

The store's initializer gets an additional `dispatchQueue` argument (`.main` by default), and some `DispatchSpecificKey` is used to store some store's metadata in the `DispatchQueue`. 

Thread safety is delegated to GCD if I understand correctly.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -124,23 +124,48 @@ public final class Store<State, Action> {
     private let dispatchToken: DispatchToken?
   #endif
 
-  /// Initializes a store from an initial state, a reducer, an environment, and a dispatch queue.
+  /// Initializes a store from an initial state, a reducer, and an environment.
   ///
   /// - Parameters:
   ///   - initialState: The state to start the application in.
   ///   - reducer: The reducer that powers the business logic of the application.
   ///   - environment: The environment of dependencies for the application.
-  ///   - boundTo: Some dispatch queue onto which the store is assumed to process
-  ///   operations. If no queue is specified, the store should run on the main queue.
   ///
-  /// - Note: You are responsible for dispatching store's operations on the correct queue.
+  /// - Note: This store should operate on the main queue. You can use the factory
+  /// ``boundTo(dispatchQueue:initialState:reducer:environment:)`` to instantiate
+  /// a store you intend to run on some arbitrary serial queue.
   public convenience init<Environment>(
     initialState: State,
     reducer: Reducer<State, Action, Environment>,
-    environment: Environment,
-    boundTo dispatchQueue: DispatchQueue = .main
-  ) {
-    
+    environment: Environment
+  ){
+    self.init(
+      initialState: initialState,
+      reducer: reducer,
+      environment: environment,
+      dispatchToken: nil
+    )
+  }
+  
+  /// Initializes a store from an initial state, a reducer, and an environment, that operate
+  /// on some dispatch queue.
+  ///
+  /// - Parameters:
+  ///   - dispatchQueue: Some dispatchQueue the store interact on.
+  ///   - initialState: The state to start the application in.
+  ///   - reducer: The reducer that powers the business logic of the application.
+  ///   - environment: The environment of dependencies for the application.
+  ///
+  /// - Note: This factory is only a declaration of intent to operate on a dispatch queue.
+  /// This only enables runtime checks that interactions with the store happen on the specified
+  /// queue for debug builds, and you are responsible for dispatching these interactions on the
+  /// correct queue.
+  public static func boundTo<Environment>(
+    dispatchQueue: DispatchQueue,
+    initialState: State,
+    reducer: Reducer<State, Action, Environment>,
+    environment: Environment
+  ) -> Self {
     #if DEBUG
     let token = DispatchToken(dispatchQueue: dispatchQueue)
     if let token = token {
@@ -151,8 +176,7 @@ public final class Store<State, Action> {
     #else
     let token: DispatchToken? = nil
     #endif
-    
-    self.init(
+    return self.init(
       initialState: initialState,
       reducer: reducer,
       environment: environment,
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -470,11 +470,15 @@ public final class Store<State, Action> {
         return
       }
     
-      let queueDescription = DispatchQueue.getSpecific(key: DispatchToken.specificKey)?
-                                          .first?
-                                          .queueDescription
-                             ?? "\(DispatchQueue.main)"
-    
+      let queueDescription: String
+      if let token = DispatchQueue.getSpecific(key: DispatchToken.specificKey)?.first {
+        queueDescription = token.queueDescription
+      } else if Thread.isMainThread {
+        queueDescription = "\(DispatchQueue.main)"
+      } else {
+        queueDescription = "Unknown queue"
+      }
+      
       let message: String
       switch status {
       case let .effectCompletion(action):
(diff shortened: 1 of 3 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -121,20 +121,53 @@ public final class Store<State, Action> {
...
-  public init<Environment>(
+  ///   - dispatchQueue: Some dispatch queue onto which the store is assumed to process
+  ///   operations. If no queue is specified, the store should run on the main queue.
+  ///
+  /// - Note: You are responsible for dispatching store's operations on the correct queue.
+  public convenience init<Environment>(
+    initialState: State,
+    reducer: Reducer<State, Action, Environment>,
+    environment: Environment,
+    dispatchQueue: DispatchQueue = .main
```
lukeredpath, line 141: What do you think about naming this `bindToDispatchQueue`?

tgrapperon:
A few comments on the implementation:

When a root store is instantiated, some `UUID?` token is generated. For pseudo-performance reasons, this token is always `nil` for the `main` queue, and some random `UUID` otherwise. When this token is non-`nil`, it is registered with the dispatch queue using a `DispatchSpecificKey<Set<UUID>>`. This token is passed to scoped stores of this root store.

When checking that operations are dispatched on the correct queue, the store checks if:
- The thread is the main tread if the token is nil,
  *or* 
- The current dispatch queue tokens contains the store's token otherwise.

I've also added a test with a store using an arbitrary dispatch queue.

The initializer gains an additional argument that will be visible when using code completion. If this is not desired, some factory can be implemented instead.

Sources/ComposableArchitecture/Store.swift
tgrapperon: This is indeed possible. My initial implementation used a factory `.attachedTo(dispatchQueue: DispatchQueue, initialState:…`. I'm still not sure if a factory is preferable to the initializer in order to preserve leanness of code completion.
Going back to `bindTo`, I find it a little awkward to have an active action in an initializer (which is already bearing some action itself). I would have thought to use the passive `boundTo` instead. The other small issue is the collision with `Binding` which can be confusing for users discovering the library.
However, I'm not a native speaker, so I'm not in the best position to evaluate this. What do you think?

mluisbrown: I'm fine with either `dispatchQueue` or `boundTo`.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -424,9 +459,16 @@ public final class Store<State, Action> {
   @inline(__always)
   private func threadCheck(status: ThreadCheckStatus) {
```
mluisbrown, line 460: Strictly speaking these should probably be renamed to `queueCheck` and `QueueCheckStatus`.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -466,8 +508,7 @@ public final class Store<State, Action> {
 
         \(message)
 
-          Store created on: \(self.initialThread)
-          Action sent on: \(Thread.current)
+          Action sent on: \(Thread.current) instead of \(Thread.main)
```
mluisbrown, line 511: For Stores not using the main queue, this message is misleading. For that case, since it's not possible to obtain the current queue (nor do we have the queue on which it was created), the message could be changed to something like:

```
Action sent on a different DispatchQueue to the one the Store was created on
```

Or, if we were to hold onto the creation `DispatchQueue` in `Store`, we could use:

```
Store was created on \(queue.name), but the Action was sent on a different queue
```

tgrapperon: Indeed, `Thread.main` a carry on from #805 where checks were restricted to it. I forgot to update the message. 

I think storing the name of the dispatch queue (or even its description while we're at it) is preferable. I'll update the PR.

Sources/ComposableArchitecture/Store.swift
tgrapperon: I went with boundTo with the last commit, as there are many references to the store being bound to some queue in the breakpoint message now.

mluisbrown (approved):

tgrapperon:
The UUID was replaced by some DispatchToken type that wraps it and also stores the queue's description (only the `UUID` is used for comparisons).
This allows to display messages with queues' descriptions for the bound queue (stored) and the calling queue (retrieved from the dispatch queue).
I've also replaced all "thread" mentions by "queue".

tgrapperon:
I converted the overloaded initializer to a factory called `.boundTo(dispatchQueue: DispatchQueue, initialState:...`. This brings back the initializer to its original signature, and code completion should be only affected when using `.init(` calls, as it may also propose `.boundTo(` after entering the dot. This configuration happens frequently with SwiftUI Previews, where the `View` already bears the type of the store and one can use `.init(`. 

I wonder if the checks should happen in the `ViewStore`'s too. I guess so and this seems doable using the same approach as the store and passing the token to `ViewStore`.

I would be interested in @mbrandonw or @stephencelis feedback at some point, especially regarding #811 (but this is not an urgent matter).

mbrandonw:
Hey @tgrapperon, thanks again for pushing this forward!

Right now we consider #811 an alternative to this PR. We are weary of introducing an API that takes an explicit `DispatchQueue` because that makes it seem as if the `Store` is going to implementing locking logic with that queue, and that is not the case. It's only used to provide runtime warnings.

We're hoping #811 strikes a balance between all of the goals we have: we want some thread checking and we want people to be able to use the `Store` on background threads. It's difficult to support both of those things fully, so we are going to make the compromise that stores created via `Store.init` get checked at runtime and are forced to run on the main queue, and stores created via `Store.unchecked` are not checked at all.

What are you thoughts on that? We'd be curious to hear from @mluisbrown and @lukeredpath too.

mbrandonw:
> I wonder if the checks should happen in the ViewStore's too. I guess so and this seems doable using the same approach as the store and passing the token to ViewStore.

`ViewStore`'s get checks automatically by virtue of the fact that their `.send`s just go straight through to the underlying store, so we don't think they need any extra checks.

tgrapperon:
> ViewStore's get checks automatically by virtue of the fact that their .sends just go straight through to the underlying store, so we don't think they need any extra checks.

I had the getter side in mind, as I thought it could be possible to tap in the store's state from another thread when initializing the `ViewStore`. I wasn't sure that `CurrentValueSubject` is thread safe when you get its value (the publisher are, if I understand [this forums.swift response](https://forums.swift.org/t/thread-safety-for-combine-publishers/29491/13)  correctly, but there's no mention for `value`). Since updating `value` is equivalent to using `send(`, I guess there's no issue on the other way too. So I now agree that `ViewStore` checks are probably unnecessary.

I agree about misleading the user into thinking the store will dispatch the operations automatically. I've added a note in factory documentation for this reason, but I think it can be still problematic.

I have no issue with the middle ground solution, which was my first idea (I was unaware of `DispatchSpecificKey`). 
I was not super enthusiast about surfacing some type that will most likely feel out of place when converted to the new concurrency model, but this seemed the only choice offered by current GCD APIs.

Thanks to you, @lukeredpath, and @mluisbrown for reviewing the PR!

lukeredpath:
FWIW, I think keeping the checks for a main queue store covers the majority use case and having the unchecked store for background queue usage is fine - if you’re using TCA in a background queue it implies that you have a good reason and will be taking care to use the store in a thread safe way anyway. The naming (unchecked) makes it clear you’re on your own too.

mluisbrown:
I agree that #811 is a good compromise. As @lukeredpath mentioned, anyone using Stores off the main thread probably knows what they're doing so shouldn't need the checks.

tgrapperon:
I'm closing the PR in favor of #811.
Thanks to everyone involved!

PR #811, opened 2021-09-17
Perform thread check only when store is created on main queue.

mbrandonw:
This relaxes the thread checking in the store to only those stores created on the main queue. This still allows people to create stores on background threads and send actions on background threads, but such stores will not get thread checking.

See https://github.com/pointfreeco/swift-composable-architecture/discussions/802 for a lot more discussion on why we want to do this.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -294,7 +294,7 @@ public final class Store<State, Action> {
       reducer: .init { localState, localAction, _ in
         isSending = true
         defer { isSending = false }
-        self.send(fromLocalAction(localAction), originatingFrom: nil)
+        self.send(fromLocalAction(localAction))
         localState = toLocalState(self.state.value)
         return .none
       },
@@ -343,7 +343,7 @@ public final class Store<State, Action> {
         let localStore = Store<LocalState, LocalAction>(
           initialState: localState,
           reducer: .init { localState, localAction, _ in
-            self.send(fromLocalAction(localAction), originatingFrom: nil)
+            self.send(fromLocalAction(localAction))
             localState = extractLocalState(self.state.value) ?? localState
             return .none
           },
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -121,7 +121,7 @@ public final class Store<State, Action> {
   private let reducer: (inout State, Action) -> Effect<Action, Never>
   private var bufferedActions: [Action] = []
   #if DEBUG
-    private let initialThread = Thread.current
+  private let isMainThreadStore: Bool
   #endif
 
   /// Initializes a store from an initial state, a reducer, and an environment.
@@ -137,6 +137,11 @@ public final class Store<State, Action> {
   ) {
     self.state = CurrentValueSubject(initialState)
     self.reducer = { state, action in reducer.run(&state, action, environment) }
+
+    #if DEBUG
+    DispatchQueue.main.setSpecific(key: mainQueueKey, value: value)
+    self.isMainThreadStore = DispatchQueue.getSpecific(key: mainQueueKey) == value
+    #endif
   }
 
   /// Scopes the store to one that exposes local state and actions.
@@ -289,7 +294,7 @@ public final class Store<State, Action> {
       reducer: .init { localState, localAction, _ in
         isSending = true
         defer { isSending = false }
-        self.send(fromLocalAction(localAction))
+        self.send(fromLocalAction(localAction), originatingFrom: nil)
         localState = toLocalState(self.state.value)
         return .none
       },
@@ -338,7 +343,7 @@ public final class Store<State, Action> {
         let localStore = Store<LocalState, LocalAction>(
           initialState: localState,
           reducer: .init { localState, localAction, _ in
-            self.send(fromLocalAction(localAction))
+            self.send(fromLocalAction(localAction), originatingFrom: nil)
             localState = extractLocalState(self.state.value) ?? localState
             return .none
           },
@@ -368,8 +373,8 @@ public final class Store<State, Action> {
     self.publisherScope(state: toLocalState, action: { $0 })
   }
 
-  func send(_ action: Action, isFromViewStore: Bool = true) {
-    self.threadCheck(status: .send(action, isFromViewStore: isFromViewStore))
+  func send(_ action: Action, originatingFrom originatingAction: Action?) {
+    self.threadCheck(status: .send(action, originatingAction: originatingAction))
 
     self.bufferedActions.append(action)
     guard !self.isSending else { return }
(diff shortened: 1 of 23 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -368,8 +373,8 @@ public final class Store<State, Action> {
     self.publisherScope(state: toLocalState, action: { $0 })
   }
 
-  func send(_ action: Action, isFromViewStore: Bool = true) {
-    self.threadCheck(status: .send(action, isFromViewStore: isFromViewStore))
+  func send(_ action: Action, originatingFrom originatingAction: Action?) {
+    self.threadCheck(status: .send(action, originatingAction: originatingAction))
```
mbrandonw, line 377: This is unrelated to the purpose of this PR, but by threading the originating action for effect emissions we can provide better messaging when the thread check fails.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -418,62 +423,63 @@ public final class Store<State, Action> {
...
+          called on the main thread.
           """
 
-      case let .send(action, isFromViewStore: true):
+      case let .send(action, originatingAction: nil):
         message = """
-          "ViewStore.send(\(debugCaseOutput(action)))" was called on the wrong thread. Make sure \
-          that "ViewStore.send" is always called on the same thread the store was created on.
+          "ViewStore.send(\(debugCaseOutput(action)))" was called on a non-main thread. Make sure \
+          that "ViewStore.send" is always called on the main thread..
```
stephencelis, line 453:
```suggestion
          that "ViewStore.send" is always called on the main thread.
```

stephencelis (approved):

Sources/ComposableArchitecture/Store.swift
```diff
@@ -368,8 +373,8 @@ public final class Store<State, Action> {
     self.publisherScope(state: toLocalState, action: { $0 })
   }
 
-  func send(_ action: Action, isFromViewStore: Bool = true) {
-    self.threadCheck(status: .send(action, isFromViewStore: isFromViewStore))
+  func send(_ action: Action, originatingFrom originatingAction: Action?) {
```
stephencelis, line 376:
```suggestion
  func send(_ action: Action, originatingFrom originatingAction: Action? = nil) {
```

mluisbrown (approved):

Sources/ComposableArchitecture/Store.swift
```diff
@@ -111,6 +111,21 @@ import Foundation
...
+/// #### Thread safety checks
+///
+/// The store performs some basic thread safety checks in order to help catch mistakes. Stores
+/// constructed via the initializer ``Store/init(initialState:reducer:environment:)`` are assumed
+/// to run only on the main thread, and so a check is executed immediately to make sure that is the
+/// case. Further, all actions sent to the store and all scopes (see ``Store/scope(state:action:)``)
+/// of the store are also checked to make sure that work is performed on the main thread.
+///
+/// If you need a store that runs on a non-main thread, which should be very rare and you should
+/// have a very good reason to do so, then you can consturct a store via the
```
tgrapperon, line 123: There's a typo at "construct"

Tests/ComposableArchitectureTests/StoreTests.swift
```diff
@@ -447,4 +447,26 @@ final class StoreTests: XCTestCase {
...
+        viewStore.send(())
+        DispatchQueue.global().asyncAfter(deadline: .now() + 0.1) {
+          viewStore.send(())
+        }
+      }
+    }
+
+    // Allow asychnronous code to run for a bit to make sure the warning breakpoint is
+    // never tripped.
+    _ = XCTWaiter.wait(for: [.init()], timeout: 1)
```
tgrapperon, line 470: I think we are always waiting 1s and failing silently here. Is it the expected behavior?

tgrapperon: What about:
```swift
func testNonMainQueueStore() {
  var expectations: [XCTestExpectation] = []
  for i in 1 ... 100 {
    let expectation = XCTestExpectation(description: "\(i)th iteration is complete")
    expectations.append(expectation)
    DispatchQueue.global().async {
      let viewStore = ViewStore(
        Store.unchecked(
          initialState: 0,
          reducer: Reducer<Int, Void, Void> {
            state, _, _ in
            state += 1
            if state == 2 { expectation.fulfill() }
            return .none
          },
          environment: ()
        )
      )
      viewStore.send(())
      DispatchQueue.global().asyncAfter(deadline: .now() + 0.1) {
        viewStore.send(())
      }
    }
  }

  wait(for: expectations, timeout: 1)
}
```
On my machine, it runs the test in 0.13s.

mbrandonw: Great suggestion, thanks!

tgrapperon: In a true TCA spirit, I now think that the reducer's environment should be the expectation!

Tests/ComposableArchitectureTests/StoreTests.swift
```diff
@@ -447,4 +447,34 @@ final class StoreTests: XCTestCase {
...
+  func testNonMainQueueStore() {
+    var expectations: [XCTestExpectation] = []
+    for i in 1 ... 100 {
+      let expectation = XCTestExpectation(description: "\(i)th iteration is complete")
+      expectations.append(expectation)
+      DispatchQueue.global().async {
+        let viewStore = ViewStore(
+          Store.unchecked(
+            initialState: 0,
+            reducer: Reducer<Int, Void, Void> {
```
tgrapperon, line 460:
```suggestion
            reducer: Reducer<Int, Void, XCTestExpectation> {
```

Tests/ComposableArchitectureTests/StoreTests.swift
```diff
@@ -447,4 +447,34 @@ final class StoreTests: XCTestCase {
...
+              state, _, _ in
```
tgrapperon, line 461:
```suggestion
              state, _, expectation in
```

Tests/ComposableArchitectureTests/StoreTests.swift
```diff
@@ -447,4 +447,34 @@ final class StoreTests: XCTestCase {
...
+              state += 1
+              if state == 2 {
+                return .fireAndForget { expectation.fulfill() }
+              }
+              return .none
+            },
+            environment: ()
```
tgrapperon, line 468:
```suggestion
            environment: expectation
```

tgrapperon (changes requested):
I think we can make the expectation the environment of the reducer.

Tests/ComposableArchitectureTests/StoreTests.swift
mbrandonw: haha, sure thing!

PR #812, opened 2021-09-17
Clean up speech recognition case study.

mbrandonw:
This caught my eye and thought we could clean it up like we have some of the other long-living clients.


stephencelis (approved):
Much better!

PR #813, opened 2021-09-17
Remove BindingAction and BindableState usage for compiler version < 5.4

srhtshn:
The project was giving error in `Deprecations.swift` file while compiling with Xcode 12.4 due to `BindingAction` and `BindableState` definitions are not found. The reason was that definitions are wrapped into  compiler version >= 5.4 check with the release 0.26.0. 

For the deprecations that uses `BindingAction` and  `BindableState`:
- One is moved into compiler version check.
- Others are removed for older compiler versions. (simply, I removed the deprecations in `else` block)


Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -58,96 +58,31 @@ extension Store {
   }
 }
 
-extension ViewStore {
```
stephencelis, line 61: Ah I think the main issue is this should be wrapped in `#if compiler(>=5.4)` and we should keep the deprecation notices around.

Just curious, but what's preventing you from using Xcode 12.5?

srhtshn: Ah sorry 🤦, thank you. Let me update the PR.

About the question, there are just some other priorities in the project. We'll be upgrading to 12.5.1 soon.

srhtshn: Sorry @stephencelis . I had couple copy of the TCA source code and I messed with git clients. Let me close that PR and open a fresh one.

PR #814, opened 2021-09-17
Remove BindingAction and BindableState usage for compiler version < 5.4

srhtshn:
The project was giving error in `Deprecations.swift` file while compiling with Xcode 12.4 due to `BindingAction` and `BindableState` definitions are not found. The reason was that definitions are wrapped into  compiler version >= 5.4 check with the release 0.26.0. 

For the deprecations that uses `BindingAction` and  `BindableState`:
- One is moved into compiler version check.
- Others are removed for older compiler versions. (simply, I removed the deprecations in `else` block)


Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -58,96 +58,33 @@ extension Store {
...
-  public subscript<Value>(
-    dynamicMember keyPath: WritableKeyPath<State, BindableState<Value>>
-  ) -> Binding<Value>
-  where Action: BindableAction, Action.State == State, Value: Equatable {
-    self.binding(
-      get: { $0[keyPath: keyPath].wrappedValue },
-      send: { .binding(.set(keyPath, $0)) }
-    )
-  }
-}
```
stephencelis, line 76: Deprecations in this file are kept in order, so we can know when a particular interface was deprecated. Can you bring this back and wrap it with `#if compiler(>=5.4)` here?

Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -58,96 +58,33 @@ extension Store {
...
-
-// NB: Deprecated after 0.25.0:
-
 #if compiler(>=5.4)
-  extension BindingAction {
```
stephencelis, line 81: We want to keep these deprecations around for Xcode 12.5+, they are old interfaces.

Sources/ComposableArchitecture/Internal/Deprecations.swift
stephencelis: Just a clarification: I think this PR should be a diff of 2 lines that adds an `#if` above and `#endif` below this extension.

srhtshn: Actually, I just did the same thing. 
- Added `#if compiler(>=5.4)`  and fixed the indentation. 
- Removed the deprecations in else part. 

However github diff visualizer makes it look complicated. On Xcode it's more clear

srhtshn: The deprecations are in correct order at the moment

Sources/ComposableArchitecture/Internal/Deprecations.swift
srhtshn: This is because also github's diff visualizer 😢 . It's in the same line (one below actually) within the file.

Sources/ComposableArchitecture/Internal/Deprecations.swift
stephencelis: Are you sure you're pushed? I pulled down your branch and am seeing the same as reflected on GitHub.

Sources/ComposableArchitecture/Internal/Deprecations.swift
stephencelis: The diff is 16 additions and 79 deletions. Even if the visualizer was tripping up, the additions should be 2 more than whatever deletions are detected. It looks like maybe your changes weren't pushed?

Sources/ComposableArchitecture/Internal/Deprecations.swift
srhtshn: Yes, I am quite confident this time. Here how it looks on my Xcode diff
Screen Shot 2021-09-18 at 02 11 42

 Although diff looks quite weird, line numbers are matching.

Sources/ComposableArchitecture/Internal/Deprecations.swift
srhtshn: The # of deletions are higher, because I removed the deprecations, in `#else` part. Because they were using `BindingAction` and `BindableState` which is not defined for compiler versions < 5.4. However diff calculation got weird because of the similar texts. Although, I did not changed anything in this `line 81`, it seems to be removed and re-added in `line 82` in the new file.

srhtshn:
I've reverted the changes, and pushed in 2 commits. Although diff changes are the same, it's easier to follow now. Sorry if I caused too much confusion. I should have done like this at the first place

stephencelis:
@srhtshn Ah, sorry about that! Those helpers really should be available to 12.4, but we had some code that wasn't properly bracketed. I've opened #815 to address, because we really do want to keep those deprecations around. Thanks for bringing this to our attention!

srhtshn:
Ah now I understood. You're welcome! Thanks for addresing this so fast.

PR #825, opened 2021-09-23
✨ Add support for multiline text alignment in alerts

ferologics:
Fixes #824

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/TextState.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/TextState.swift
@@ -1,5 +1,6 @@
 import CustomDump
 import SwiftUI
+import CasePaths
 
 /// An equatable description of SwiftUI `Text`. Useful for storing rich text in state for the
 /// purpose of rendering in a view hierarchy.
@@ -68,6 +69,7 @@ public struct TextState: Equatable, Hashable {
     case strikethrough(active: Bool, color: Color?)
     case tracking(CGFloat)
     case underline(active: Bool, color: Color?)
+    case multilineTextAlignment(TextAlignment)
   }
 
   fileprivate enum Storage: Equatable, Hashable {
@@ -204,6 +206,12 @@ extension TextState {
     `self`.modifiers.append(.underline(active: active, color: color))
     return `self`
   }
+
+  public func multilineTextAlignment(_ alignment: TextAlignment)  -> Self {
+    var `self` = self
+    `self`.modifiers.append(.multilineTextAlignment(alignment))
+    return `self`
+  }
 }
 
 extension Text {
@@ -239,6 +247,8 @@ extension Text {
         return text.tracking(tracking)
       case let .underline(active, color):
         return text.underline(active, color: color)
+      case .multilineTextAlignment:
+        return text // applied in view body
       }
     }
   }
@@ -246,7 +256,11 @@ extension Text {
 
 extension TextState: View {
   public var body: some View {
-    Text(self)
+    if let alignment = modifiers.compactMap(/Modifier.multilineTextAlignment).first {
+      Text(self).multilineTextAlignment(alignment)
+    } else {
+      Text(self)
+    }
   }
 }
 
@@ -361,7 +375,8 @@ extension TextState: CustomDumpRepresentable {
           .fontWeight(.none),
           .foregroundColor(.none),
           .strikethrough(active: false, color: _),
-          .underline(active: false, color: _):
+          .underline(active: false, color: _),
+          .multilineTextAlignment(_):
           break
         }
       }


stephencelis (commented):
@ferologics Thanks for the discussion and PR!

Interesting! I guess this is just now possible in iOS 15 with the new alert APIs that work with view builders and not simple `Text`? Seems like other view modifiers that work, as well, like `textCase`. I wonder if it's documented anywhere which modifiers are safe to use in alerts and confirmation dialogs?

Regardless, I think we could definitely improve the situation here, so thanks for bringing this up! We'll chat about it next week and try to think through the best way to accommodate an ever growing number of modifiers 😆

Sources/ComposableArchitecture/SwiftUI/TextState.swift
```diff
@@ -239,14 +247,20 @@ extension Text {
...
+        return text // applied in view body
       }
     }
   }
 }
 
 extension TextState: View {
   public var body: some View {
-    Text(self)
+    if let alignment = modifiers.compactMap(/Modifier.multilineTextAlignment).first {
```
tgrapperon, line 259:
```suggestion
    if let alignment = modifiers.compactMap(/Modifier.multilineTextAlignment).last {
```
In order to be homogenous with the way modifiers are applied when comparing states (via `reduce`, where the last one of a kind wins), I think this should be `last` here.

tgrapperon:
> I wonder if it's documented anywhere which modifiers are safe to use in alerts and confirmation dialogs?

@stephencelis: I'm not aware of such a thing. There is a similar issue with Menus, or `navigationTitle` among many others, and what's effectively working also depends on the platforms (and probably the OS versions too).
Maybe one can harness `swift-snapshot-testing` and an exhaustive modifiers test suite to build and maintain such a set. I can give it a try if you find the idea worthy.

ferologics:
@stephencelis Ideally the `.alert` TCA modifier would also take custom `message` ViewBuilder in a form of `(State) -> (some View)`. This way TCA could avoid maintaining an ever growing list of view modifiers, and provide greater flexibility to it's users.

According to the [documentation](https://developer.apple.com/documentation/swiftui/view/alert(_:ispresented:presenting:actions:message:)-8584l) the `message` has to be a `Text` label:
> On iOS, tvOS, and watchOS, alerts only support controls with labels that are Text. Passing any other type of view results in the content being omitted.

but there appears to be some limitation on the styling of it:

> Only unstyled text is supported for the message.

It'd be worthwhile to experiment here to understand better what "unstyled" means here.

stephencelis:
> @stephencelis Ideally the `.alert` TCA modifier would also take custom `message` ViewBuilder in a form of `(State) -> (some View)`. This way TCA could avoid maintaining an ever growing list of view modifiers, and provide greater flexibility to it's users.

One of the reasons we have `TextState` is to allow for these things to be tested, since this state is given full coverage with the test store. Perhaps we need to figure out where to draw the line, though.

> It'd be worthwhile to experiment here to understand better what "unstyled" means here.

For sure! It's kinda unfortunate that the interface isn't strict `Text` anymore, since it means more invalid runtime states, but maybe we can correct for this in TCA.

ferologics:
> One of the reasons we have TextState is to allow for these things to be tested, since this state is given full coverage with the test store. Perhaps we need to figure out where to draw the line, though.

Yes, I see testing is a pivotal concern of any changes to be made in the API. And while testability is one of the prime objectives of the library, maintaining compatibility with SwiftUI should probably take priority here (considering the fact that the list of view modifiers is bound to expand).

> ... It's kinda unfortunate that the interface isn't strict Text anymore, since it means more invalid runtime states ...

Agreed, this has been my experience as well. The domain modeling is too lax and the extensive list ViewModifiers that can be applied seemingly to any view is misleading. Incorrectly applied modifiers lead to unpredictable behavior.

Just to confirm, I took a peek at the [list of available modifiers](https://developer.apple.com/documentation/swiftui/text-view-modifiers) for `Text`. Unfortunately it is the same list for all SwitfUI views, even though they are linked under the appropriate View section (`List`/`Text`/ etc.). Hopefully this is a hint to a future API improvement 🤞.

image

That being said, there are some "sections" that explicitly mention Text modifiers. Perhaps there's some logic to this after all 🤔 Not sure if scraping the documentation is the best idea, but maybe the only "working" solution in this context.

image

> ... maybe we can correct for this in TCA.

It would definitely be nice to have a more restrictive list of modifiers that can be applied. I'm not sure about the maintainability aspect of it, but perhaps some form of code generation could be utilized for this.

stephencelis:
> Yes, I see testing is a pivotal concern of any changes to be made in the API. And while testability is one of the prime objectives of the library, maintaining compatibility with SwiftUI should probably take priority here (considering the fact that the list of view modifiers is bound to expand).
> [...] 
> It would definitely be nice to have a more restrictive list of modifiers that can be applied. I'm not sure about the maintainability aspect of it, but perhaps some form of code generation could be utilized for this.

Another thing to keep in mind is that `TextState` and `AlertState` are completely additive APIs in TCA, and are not required, so one can still drive a vanilla `alert` modifier with a view store.

ferologics:
Indeed! that is a great fallback in this case, where multiline text alignment is not there yet.

Also, it just occurred to me that with the advent of iOS 15 the newfound focus on accessibility in SwiftUI should be kept in mind. Maybe it isn't the most standard/popular part of the API, but still there is a handful of modifiers there that the `TextState` / `AlertState` could/should support.

PR #829, opened 2021-09-24
Don't print to console when dumping state for ViewStore.debug

mbrandonw:
While debugging something I realized that `ViewStore.debug` was printing too much to the console.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/WithViewStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/WithViewStore.swift
@@ -63,13 +63,15 @@ public struct WithViewStore<State, Action, Content> {
   fileprivate var _body: Content {
     #if DEBUG
       if let prefix = self.prefix {
+        var stateDump = ""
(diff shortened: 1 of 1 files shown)


stephencelis (approved):

hadiidbouk:
I would like to add that this will fix the problem that I introduced in version 0.28.0.

When you have an `NSManagedObjectID` in your state, the app will just crash when using `.debug()`, however, it seems like a problem inside the `CustomDump` package, you will have problems when trying to dump an `NSManagedObjectID` value.

mbrandonw:
@hadiidbouk Interesting. We're not sure why it would crash, but also in general it's probably not a good idea to store core data reference types in state. Instead you could hold onto the `uriRepresentation()` of the object id, and then convert it back into an object id using `managedObjectID(forURIRepresentation:)`.

But, if you want to open a [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions) with more details on the crash there may be some people in the community who can help.

hadiidbouk:
Yes you are definitely right

PR #830, opened 2021-09-24
Fix typo in `BindingAction.pullback(_:)` docs

malonehedges:
Noticed a small typo where the action enum case was missing a closing parentheses. :P


stephencelis (approved):
Ah good catch! Thanks!

PR #833, opened 2021-09-28
Alert state button action

kdeda:
I'm developing a Mac application using TCA on my spare time.
I also have a custom SheetView that works in conjunction with AlertState.Button
To use it with TCA I had to made a tinny change to 
```
extension AlertState.Button 
     public func toSwiftUIAction(send: @escaping (Action) -> Void) -> () -> Void
```

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Alert.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Alert.swift
@@ -330,7 +330,7 @@ extension AlertState.Button: Hashable where Action: Hashable {
 }
 
 extension AlertState.Button {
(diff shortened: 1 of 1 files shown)


kdeda:
Commit 1, simply contains the demo app and does not need merged
I'v added a new commit that contains a demo macOS app.
../swift-composable-architecture/Examples/ModalSheet/DemoApp/DemoApp.xcodeproj

The simple app will show the partially completed SheetView which makes use of the new buttonAction() to create a 'fancy' macOS modal sheet.
The SheetView on this demo is tailored to be used by my existing app WhatSize and so it's still in flux.

kdeda:
Commit 2: **We need the public access here to access this from other** 
Is the one that can be merged if you elect to.

PR #847, opened 2021-10-11
feat: Let a store know about it's parent

maximkrouk:
This feature might be needed for recursive navigation in UIKit
I can give you access to a private repo to ensure that it's needed
> TLDR
> - On change of ids in the recursive navigation stack I update recursive controllers in RecursiveNavigationController
> - After controllers stack is updated accordingly to the state I scope parent module to this controllers
> - I want to avoid redundant scoping and currently I set parent store as an associated object to the child to check if the store shouldn't be scoped
> - I have to use a custom scope method to set an associated object (both a custom method and an associatedobject api is a bit confusing)

Implementation
- I created `_ParentTraking` internal protocol to erase store's type for a parent and to specify an ability to get the next parent
- `Store` is now `_ParentTracking`
- I added 2 instance methods to `Store` type `isParent(for:)` and `isDirectParent(for:)` to get the information if the store is a parent to another one

The change is additive and won't affect any users

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -133,6 +133,7 @@ public final class Store<State, Action> {
   var effectCancellables: [UUID: AnyCancellable] = [:]
   private var isSending = false
   var parentCancellable: AnyCancellable?
+  var parentStoreID: ObjectIdentifier?
   private let reducer: (inout State, Action) -> Effect<Action, Never>
   var state: CurrentValueSubject<State, Never>
   #if DEBUG
@@ -341,6 +342,7 @@ public final class Store<State, Action> {
         guard !isSending else { return }
         localStore?.state.value = toLocalState(newValue)
       }
+    localStore.parentStoreID = ObjectIdentifier(self)
     return localStore
   }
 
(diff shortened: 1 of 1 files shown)


stephencelis:
@maximkrouk Thanks for taking the time to work on a PR!

> I can give you access to a private repo to ensure that it's needed

This, or a simplified example, would be helpful, I think, in understanding the problem.

maximkrouk:
I gave you and @mbrandonw access for 2 private repos, `swift-standard-extensions` is a dependency for the `composable-architecture-extensions`, the second one contains an example and recursive navigation implementation for UIKit.
Note that both packages are on the prototyping stage, and no one depends on them, so feel free to tell me what you think or create a new branch and commit anything you want 🙂

maximkrouk:
Ping 🌚

maximkrouk:
Maybe some thoughts about complex UIKit cell states?
Currently, I came to something like this:
```swift
// simplified example
public final class MyViewController: UIViewController {
  private(set) public var store: Store<State, Action>?
  private var cellsBuffer: [MyCompossableCell: IndexPath] = [:]

  func cellForRowAt(...) {
    cellsBuffer[cell] = indexPath
    // link cell store to the current one on reuse
    // probably this step may be skipped if cell.store.state.id == store.state.items[indexPath].id
    cell.setStore(store.scope(...))
  }

  func setStore(...) {
     self.store = store
     // update cell stores to be linked to the current store if the current one was changed
     cellsBuffer.forEach { cell, indexPath in
       cell.setStore(store.scope(...))
     }
  }
}
```

mbrandonw:
Hi @maximkrouk, we are going to close this without merging. Please feel free to open a discussion if you want to discuss this more.

PR #863, opened 2021-10-18
✨ Add support for accessibility modifiers on TextState

ferologics:
The `accessibilityLabel` modifier does not work all that well with the `TextState` as there's a different return type for iOS 14 and iOS 15. The former is defined as an extension on `ModifiedContent` while latter on `Text`. With that in mind, it makes sense to support the more recent modifiers.

I currently managed to build this only for iOS. For macOS the compiler complains about wrong return type `ModifiedContent<...>`. The source code designates  `@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)` so it should in theory work.

TODO:
- [x] Add custom dump of accessibility modifiers
- [x] Compile for other than iOS target platform

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/TextState.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/TextState.swift
@@ -58,6 +58,9 @@ public struct TextState: Equatable, Hashable {
   fileprivate let storage: Storage
 
   fileprivate enum Modifier: Equatable, Hashable {
+    case accessibilityHeading(AccessibilityHeadingLevel)
+    case accessibilityLabel(AccessibilityLabelValue)
+    case accessibilityTextContentType(AccessibilityTextContentType)
     case baselineOffset(CGFloat)
     case bold
     case font(Font?)
@@ -68,9 +71,6 @@ public struct TextState: Equatable, Hashable {
     case strikethrough(active: Bool, color: Color?)
     case tracking(CGFloat)
     case underline(active: Bool, color: Color?)
-    case accessibilityLabel(AccessibilityLabelValue)
-    case accessibilityTextContentType(AccessibilityTextContentType)
-    case accessibilityHeading(AccessibilityHeadingLevel)
   }
 
   fileprivate enum Storage: Equatable, Hashable {
@@ -321,6 +321,12 @@ extension TextState {
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 extension TextState {
+  public func accessibilityHeading(_ headingLevel: AccessibilityHeadingLevel) -> Self {
+    var `self` = self
+    `self`.modifiers.append(.accessibilityHeading(headingLevel))
+    return `self`
+  }
+
   public func accessibilityLabel(_ string: String) -> Self {
     var `self` = self
     `self`.modifiers.append(.accessibilityLabel(.init(string)))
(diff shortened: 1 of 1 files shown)


ferologics:
I just built for macOS using Xcode `Version 13.1 (13A1030d)` (RC) and the error about incompatible return type is gone.

3a4oT:
Thanks for the work! Do you have any updates on this?

ferologics:
@3a4oT it's more or less finished. Still missing custom dump for the new modifiers, but I'm waiting for green light from @stephencelis before adding this (in case we'd want to revert to previous implementation -- see preceding commit).

stephencelis:
@ferologics Have you looked into the CI failures?

ferologics:
@stephencelis only now 🙃 it looks like the `AccessibilityHeadingLevel` and `AccessibilityTextContentType` symbols are missing when the library is compiled with Xcode version lower than `13.1`. I think I'm supposed to wrap the part that code that references such symbols in a `#if compiler(>=5.5.1) { ... }`?

stephencelis:
@ferologics Ideally we don't force Xcode 13.1 after merging, so if you can figure out the right compiler directive, please let us know! Otherwise it may take us a little bit to investigate ourselves, but we definitely are interested in getting this functionality into the library.

ferologics:
@stephencelis unfortunately Xcode 13.1 (Swift `5.5.1`) is the minimum required version. This is due to the API `ModifiedContent` return type mismatch. You can verify this by modifying [compiler directive](https://github.com/pointfreeco/swift-composable-architecture/pull/863/commits/ed69614a88891be85d0f2d33cd91dcfd3682092e#diff-2718aa7a03231b3de0d78e8c9b9f9535ae3da36ecfb9ee8e2ab428b757990ab4R296) to `#if compiler(>=5.5)` and running in Xcode 13.0 like so:

image

3a4oT:
friendly ping! can we merge it?

Sources/ComposableArchitecture/SwiftUI/TextState.swift
```diff
@@ -68,6 +68,9 @@ public struct TextState: Equatable, Hashable {
     case strikethrough(active: Bool, color: Color?)
     case tracking(CGFloat)
     case underline(active: Bool, color: Color?)
+    case accessibilityLabel(AccessibilityLabelValue)
+    case accessibilityTextContentType(AccessibilityTextContentType)
+    case accessibilityHeading(AccessibilityHeadingLevel)
```
stephencelis, line 73: Just a quick change for alphabetical order:
```suggestion
    case accessibilityHeading(AccessibilityHeadingLevel)
    case accessibilityTextContentType(AccessibilityTextContentType)
```
May move these to the beginning of the enum in the future, but should be good for now!

stephencelis:
> friendly ping! can we merge it?

We'll be taking another look at this next week! Sorry for the delay, just want to make sure we make the time to look it over again before merging.

ferologics:
I updated the PR with `customDumpValue`s for the new accessibility modifiers, and ordered the code alphabetically where possible 🙂

stephencelis:
@ferologics Finally got around to reviewing this in its current state. Thanks for your patience! The one thing I was wondering is if `AccessibilityLabelValue` could simply be `TextState`? This type seems to be functionally the same, so I think it would simplify things. Want to give it a shot?

ferologics:
@stephencelis absolutely! I was reluctant to do this initially because `TextState` supports concatenation. As you can see in my last commit, in `case .concatenated(_, _):` I raise an `assertionFailure`. Practically speaking the exposed public API won't allow for such state, so I think it should be fine.

stephencelis (approved):
@ferologics Looks great! Thanks so much for working on this!

PR #864, opened 2021-10-19
[Feature] - BindingAction.append operator

thebarndog:
Sometimes with bindable properties that are arrays, its useful to append a value rather than just set it. To that end, this PR adds an `.append` operator.

Example:

```swift
.binding(.append(\.$identifiers, id))
```

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Binding.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Binding.swift
@@ -329,6 +329,21 @@ public struct BindingAction<Root>: Equatable {
         valueIsEqualTo: { $0 as? Value == value }
       )
     }
+      
+    public static func append<Element>(
+      _ keyPath: WritableKeyPath<Root, BindableState<[Element]>>,
+      _ element: Element
+      ) -> Self {
+        .init(
+        keyPath: keyPath,
+        set: {
+            let value = $0[keyPath: keyPath].wrappedValue
+            $0[keyPath: keyPath].wrappedValue = value + [element]
+        },
+        value: element,
+        valueIsEqualTo: { _ in false }
+          )
+      }
 
     /// Matches a binding action by its key path.
     ///


thebarndog:
> You can definitely do this, but a thing we want to try to do is prevent business logic from "escaping" the reducer. As soon as a view has unfettered write access to a piece of state, you run the risk of introducing business logic that cannot be tested.

I think I failed to clarify, I'm only using `.binding` inside my reducers, I'm not even using `viewStore.binding(_:)`. I use `.binding` inside my reducers to just set simple `Bool` or `String` properties so I can avoid having the boilerplate actions like `setUserName` or `setIsAuthorized` and so on. From the perspective of the view layer I agree, it should only be mutated either by dispatching actions or setting up bindings via the `ViewStore`.


> As an alternative, what do you think of defining an "appended" key path on array?
> 
> ```swift
> extension Array {
>   var append: Element? {
>     get { self.last }
>     set {
>       if let newValue = newValue {
>         self.append(newValue)
>       } else if !self.isEmpty {
>         self.removeLast()
>       }
>     } 
>   }
> }
> ```
> 
> And then use it:
> 
> ```swift
> .binding(\.$identifiers.append, id))
> ```
> 
> This could be defined in your own application or even a library.

That's really interesting, I like that approach. I'll probably end up closing this PR in favor of that extension considering its lightly and non-intrusive.

Quick clarification question:

> You can definitely do this, but a thing we want to try to do is prevent business logic from "escaping" the reducer. As soon as a view has unfettered write access to a piece of state, you run the risk of introducing business logic that cannot be tested.

When you say "escaping", do you mean using `.binding(.set(\.$property, newValue))` outside of a reducer?

iampatbrown:
Hey @thebarndog, cool idea.  Do you have a short example of what the reducer would look like? Is there a benefit to using `.binding(.set(\.$isAuthorized, false))` over `state.isAuthorized = false`?

thebarndog:
Another idea for an extension @stephencelis, reducing boilerplate from

```swift
case let .logIn(provider):
    return .authenicate(provider: provider)
        .flatMap { user in 
            return Publishers.Merge(
                Just(AppAction.binding(.set(\.$currentUser, user)),
                Just(AppAction.binding(.set(\.$connectedAccounts.append, provider))
            )
        }
```

```swift
case let .logIn(provider):
    return .authenicate(provider: provider)
        .on(success: 
            .merge(
                .init(value: .binding(.set(\.$currentUser, user)),
                .init(value: .binding(.set(\.$connectedAccounts.append, provider))
             ),
              failure: .none
         )
```

where `.on(success:failure:)` is an extension on `Effect` that looks like `func on(success: Effect, failure: Effect)`. Also useful would be `Effect.if(_:then:)` for `Effect`s that produce `Bool`s

thebarndog:
I'm going to go ahead and close this since there are some better ways to accomplish the same thing

PR #868, opened 2021-10-23
[Feature] - Async/Await fire and forget method for Effect

thebarndog:
This PR adds a `fireAndForget` method to `Effect` for the async/await concurrency apis. The function is defined as:

https://github.com/pointfreeco/swift-composable-architecture/blob/81c3da674d6f400e55984d6ddb36db71169d257a/Sources/ComposableArchitecture/Beta/Concurrency.swift#L107-L136

It takes a `Sendable` escaping closure and the implementation is very similar to the non-async fire and forget method.

### Points to Address

Two quick things I wanted to make sure were squared away before getting this reviewed and merged:

1. This project uses a fairly non-standard spacing scheme and I had a hard time matching the syntax formatting in the package. What's the right indention for this project, two spaces? And is there a linting file I can run?

2. I was curious about this comment in the code:

```
    /// Note that due to the lack of tools to control the execution of asynchronous work in Swift,
    /// it is not recommended to use this function in reducers directly. Doing so will introduce
    /// thread hops into your effects that will make testing difficult. You will be responsible for
    /// adding explicit expectations to wait for small amounts of time so that effects can deliver
    /// their output.
```

By "thread hopping" are you referring to the fact that the new concurrency model doesn't guarantee that a thread that suspends will be the same one that resumes the work? I was just confused because when I hear "thread hopping" I think of full-blown context switches like what happens in GCD.

It sounds like its less of an issue with concurrency and more an issue with XCTest always forcing things to be on the main thread?

### DIFF
--- a/Sources/ComposableArchitecture/Beta/Concurrency.swift
+++ b/Sources/ComposableArchitecture/Beta/Concurrency.swift
@@ -111,7 +111,7 @@ import SwiftUI
       ///   - priority: Priority of the underlying task.
       ///   - work: The work to execute.
       /// - Returns: An effect wrapping the given asynchronous work.
-      static func fireAndForget(priority: Task.Priority?, _ work: @Sendable @escaping () async -> Void) -> Effect {
+      static func fireAndForget(priority: TaskPriority?, _ work: @Sendable @escaping () async -> Void) -> Effect {
           Deferred { () -> AnyPublisher<Output, Failure> in
               // create a future with void and never types
               Future<Void, Never> { seal in
--- a/Sources/ComposableArchitecture/Beta/Concurrency.swift
+++ b/Sources/ComposableArchitecture/Beta/Concurrency.swift
@@ -111,7 +111,7 @@ import SwiftUI
       ///   - priority: Priority of the underlying task.
       ///   - work: The work to execute.
       /// - Returns: An effect wrapping the given asynchronous work.
-      static func fireAndForget(priority: TaskPriority?, _ work: @Sendable @escaping () async -> Void) -> Effect {
+      static func fireAndForget(priority: TaskPriority? = nil, _ work: @Sendable @escaping () async -> Void) -> Effect {
           Deferred { () -> AnyPublisher<Output, Failure> in
               // create a future with void and never types
               Future<Void, Never> { seal in
--- a/Sources/ComposableArchitecture/Beta/Concurrency.swift
+++ b/Sources/ComposableArchitecture/Beta/Concurrency.swift
@@ -111,7 +111,7 @@ import SwiftUI
       ///   - priority: Priority of the underlying task.
       ///   - work: The work to execute.
       /// - Returns: An effect wrapping the given asynchronous work.
-      static func fireAndForget(priority: Task.Priority?, _ work: @escaping () async -> Void) -> Effect {
+      static func fireAndForget(priority: Task.Priority?, _ work: @Sendable @escaping () async -> Void) -> Effect {
           Deferred { () -> AnyPublisher<Output, Failure> in
               // create a future with void and never types
               Future<Void, Never> { seal in
--- a/Sources/ComposableArchitecture/Beta/Concurrency.swift
+++ b/Sources/ComposableArchitecture/Beta/Concurrency.swift
@@ -103,6 +103,36 @@ import SwiftUI
       }
       .eraseToEffect()
     }
+      
+      /// Wraps an asynchronous unit of work that returns no value in an effect.
+      ///
+      /// This function is useful for executing one-off work in an asynchronous context that can execute and be forgetten about.
+      /// - Parameters:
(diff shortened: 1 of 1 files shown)


thebarndog:
> I think using a test scheduler with TestStore works pretty well in this case.

It does but as it stands today, schedulers and even Combine, is soft-deprecated because of async await and while Combine will absolutely be around for at least a few more years, it's worth to start thinking now about how to migrate and what new testing constructs will exist.

> Do you have an example of some async work you would typically do in the reducer?

Absolutely! One practical example from something I'm working on right now is I have a `MultipeerService` in my environment and in one of my reducers, I kick off starting multipeer connectivity so that looks like


```swift
case .authenticationResponse(result<AppAction, Error>):
    .fireAndForget {
        await environment.multipeerService.startDiscovery() 
        await environment.multipeerService.startAdvertising()
    }
```

I think you are right though in that a better way to approach this is to move stuff into the service and just have the service return an effect.

stephencelis:
Hey @thebarndog! Thanks for opening the PR and starting a discussion. Given that we already have `Effect.fireAndForget`, we think an async version makes sense to introduce, as well.

First, to address your questions:

> This project uses a fairly non-standard spacing scheme and I had a hard time matching the syntax formatting in the package. What's the right indention for this project, two spaces? And is there a linting file I can run?

We auto-format on merge, so no need to worry about this. We use 2-space formatting, though (same as Swift standard library), so you could change your Xcode settings for the project, or you could install the tool we use for auto-formatting, [swift-format](https://github.com/apple/swift-format), and run `make format` from the Terminal from inside the TCA project directory).

> I was curious about this comment in the code:
>
> […]
>
> By "thread hopping" are you referring to the fact that the new concurrency model doesn't guarantee that a thread that suspends will be the same one that resumes the work? I was just confused because when I hear "thread hopping" I think of full-blown context switches like what happens in GCD.
>
> It sounds like its less of an issue with concurrency and more an issue with XCTest always forcing things to be on the main thread?

@iampatbrown explained the crux of the problem here, but it's a matter of being able to test features in a completely synchronous manner.

Per your response:

> The test suggestion is a good one, that's a nice way to hide dependencies but I don't know that it's necessary. XCTest has great support for Swift concurrency and there's no reason to hide that api. Something about messing with queues in tests just makes me nervous.

Unfortunately, XCTest's `async`/`await` support does not help us here. The asynchronous lifecycle of an effect is not exposed by the store in any way, so there's nothing to "`await`" in a TCA test. If you take a look at @iampatbrown's example test again, it's unclear how you would use `await` to let XCTest know to suspend and for how long. All of that scheduling is internal to the store/reducer, so you can't do anything but sprinkle `XCTWaiter` calls throughout your tests instead.

The only mechanism we have for controlling the lifecycle of an effect is via the test store's environment, typically via an `ImmediateScheduler` that will immediately perform any work, or via a `TestScheduler`.

The problem with `async`/`await` is there's no way to control its execution in a synchronous way, which means using `Effect.task` (or your new proposed `Effect.fireAndForget`) is similar to throwing a `DispatchQueue.main.async { ... }` in an effect: you're now back to needing to throw `XCTWaiter` calls in your test. In the future, Swift concurrency will hopefully add support for controlling the execution/scheduling of async work (maybe via the "custom executors"), but it does not appear to be possible to do so at this time.

So this is the "thread hopping" we refer to, and as such, we still don't think it's appropriate to do `async`/`await` work directly in a reducer, and instead, all calls to `Effect.task` (and your new helper) should only be done inside "live" effects that the environment vends out.

But with all that said, we _do_ think that this PR is worth merging, but we want to bring in and document all the caveats we have documented with `Effect.task`. We also think you should be careful in the `MultipeerService` example you have above: because you're doing this `async`/`await` work in your reducer, it means it's not super easy to test that this work is being performed.

thebarndog:
> We auto-format on merge, so no need to worry about this. We use 2-space formatting, though (same as Swift standard library), so you could change your Xcode settings for the project, or you could install the tool we use for auto-formatting, [swift-format](https://github.com/apple/swift-format), and run `make format` from the Terminal from inside the TCA project directory).

That may be what the Swift standard library does but its not Xcode's default setting and frankly it makes code so much harder to read and far less beautiful. Four spaces for an indention is standard across almost every major framework except the standard library so I would defer to the community on that point for the standard. Unfortunately even though Apple is the arbiter of Swift, they often have some backwards ideas about styling and formatting (wrote Swift at Apple, can confirm).

>  We also think you should be careful in the MultipeerService example you have above: because you're doing this async/await work in your reducer, it means it's not super easy to test that this work is being performed.

Totally fair but only other option really is to make an entirely separate composable module like ComposableLocation and I'm personally not a fan of that design. It works well enough but I find that the pattern is really confusing and hard to follow and I spend a lot of time trying to decipher what the module is doing.

Also this PR is really old so I'm just going to close it. Unfortunately I won't have time to re-issue it.

PR #878, opened 2021-11-02
Added effect helper for tasks returning Result

surajbarkale:
I ended up wrapping the URLSession in a layer in my application. Using async for this layer helped to make the code clearer. Since I had introduced typed errors, I did not want to go back to untyped throws. Which means I ended up writing async function returning `Result`.

I added this small extension on Effect to easily bridge this use case.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/FactClient.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/FactClient.swift
@@ -2,6 +2,31 @@ import Combine
 import ComposableArchitecture
 import XCTestDynamicOverlay
 
+#if compiler(>=5.5)
+enum ApiError: Equatable, Error {
+  case invalidResponse
+  case httpStatus(Int)
+  case urlSession(NSError)
+}
+
+extension URLSession {
+  func dataResult(from url: URL) async -> Result<Data, ApiError> {
+    do {
+      let (data, response) = try await data(from: url)
+      guard let response = response as? HTTPURLResponse else {
+        return .failure(.invalidResponse)
+      }
+      guard response.statusCode < 300 else {
+        return .failure(.httpStatus(response.statusCode))
+      }
+      return .success(data)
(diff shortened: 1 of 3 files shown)


stephencelis:
Hi @surajbarkale! Thanks for taking the time to PR and start the discussion! And sorry for the delay. @mbrandonw and I were both pretty busy and unavailable this past month.

Do you have a more fleshed out example of how you're using your proposed helper? Swift's lack of untyped errors is definitely a pain point in the language, but it also seems like it's impossible to work around the issue when interfacing with throwing code 😣 We were hoping that ergonomics would improve whenever Swift gets around to adding support for typed errors in the future.

surajbarkale:
In my app I am wrapping all errors from URLSession in `URLSesssionError` equatable struct. So I have added following overload:
```swift
extension URLSession {
  func dataResult(for request: URLRequest) async -> Result<(Data, URLResponse), URLSessionError> {
    do {
      return .success(try await data(for: request))
    } catch let error {
      return .failure(URLSessionError(error))
    }
  }
}
```

This overload along with https://github.com/JohnSundell/CollectionConcurrencyKit allows me to write async code in the environment provider and wrap it up in an effect when communicating with UI domain.

sadiklar:
Yyu

stephencelis:
> In my app I am wrapping all errors from URLSession in `URLSessionError` equatable struct. So I have added following overload:

Without a full reducer I'm still not quite sure I get the full picture. If you have a `URLSession` wrapper, couldn't this live in your environment instead? Then your reducer could invoke this dependency and call `catchToEffect()` to get a typed `Result` that you can bundle up in an action. I'm still not quite understanding why the reducer would need to be typed.

surajbarkale:
Yes this will not affect the reducer at all. I have updated the `FactClient` in case studies to show how this can be used. This extension helps to transform publisher based code to async code in environment with minimal changes.

PR #886, opened 2021-11-09
Reducer failure type

JacksonUtsch:
### Here is the support for a failure type in the reducer. It introduces a fourth generic to the `Reducer` type. 

Motivation is to allow error handling to become more dynamic. 

Two reducer methods are included, `catch` and `assertNoFailure` which are wrappers around the Combine equivalents. This allows for reducers to be modified/errors to be handled before entering their parent `Store` 

An HTTP API reducer would be expected to result in errors, (JSON Decoding, 404, timeout, auth, etc..) these can be handled via `catch` to alert the user in a more global scope (AppState.error or equivalent), log to console, report to remote error logging, etc..

With the fourth generic added, it would be a breaking change. 
(Existing projects can adapt by changing `Reducer<State, Action, Environment>` to `Reducer<State, Action, Environment, Never>`)

Looking for feedback and suggestions


stephencelis:
Hi @JacksonUtsch! Thanks for taking the time to explore this and share! And sorry for the delayed response. @mbrandonw and I didn't have a ton of free time this past month.

The change you're proposing is pretty big! It introduces another generic to an already generic-heavy type, so it'd be good to very clearly motivate things. The main example you give of API logic is usually handled in the dependency layer of TCA, and not in a reducer, so it's hard for me to picture exactly what you have in mind. Do you think you could introduce a case study / demo app to your branch that gives a concrete example of a failing reducer in use?

It also might be worth opening a discussion for this change instead of a draft PR. You could then link to your branch from that discussion. That way more folks can weigh in, in a threaded capacity. What do you think?

stephencelis:
@JacksonUtsch Thanks for resurfacing https://github.com/pointfreeco/swift-composable-architecture/discussions/880, that's helpful! In the interest of keeping things tidy do you mind if we close this till there's been more discussion/consideration?

JacksonUtsch:
Sure thing. 👍

PR #887, opened 2021-11-11
Add AlertState mapping

skyler-cricut:
Modify `AlertState`, making the ID configurable, so `AlertStates` can be composed.

This change makes the following possible:

```Swift
struct ParentState {
    var child: ChildState!

    var alert: AlertState<ParentAction>?
    var combinedAlert: AlertState<ParentAction>? {
        if let alert = alert {
            return alert
        } else if let alert = child.alert {
            return alert.map { .child($0) }
        } else {
            return nil
        }
    }
}

enum ParentAction {
    case child(ChildAction)
}

struct ChildState {
    var alert: AlertState<ChildAction>?
}

enum ChildAction { }
```

Without this change, the above could technically be done, but the mapped `AlertState` would not have the same identity as the original, which creates UI bugs.

It is useful to compose alerts especially when a parent state has multiple children (i.e. an Array) that could each have their own AlertState.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Alert.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Alert.swift
@@ -100,38 +100,44 @@ import SwiftUI
 /// ```
 ///
 public struct AlertState<Action> {
-  public let id = UUID()
+  public let id: UUID
   public var buttons: [Button]
   public var message: TextState?
   public var title: TextState
 
   @available(iOS 15, macOS 12, tvOS 15, watchOS 8, *)
   public init(
+    id: UUID = UUID(),
     title: TextState,
     message: TextState? = nil,
     buttons: [Button]
   ) {
+    self.id = id
     self.title = title
     self.message = message
     self.buttons = buttons
   }
 
   public init(
+    id: UUID = UUID(),
     title: TextState,
     message: TextState? = nil,
     dismissButton: Button? = nil
   ) {
+    self.id = id
     self.title = title
     self.message = message
     self.buttons = dismissButton.map { [$0] } ?? []
   }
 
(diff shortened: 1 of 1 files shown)


stephencelis:
Hi @skyler-cricut! Thanks for taking the time to put together a PR. It's definitely a fun transformation. A few questions come to mind, though:

Can you describe when you think a parent should be responsible for displaying alerts for all children? Why shouldn't a child domain display its own alerts?

One thought is that if a child can be built, previewed, and run in isolation, we'd want it to be responsible for showing alerts in order to exercise that functionality.

skyler-cricut:
Hi, @stephencelis. TCA has been very useful to my large team—thanks for making it!

> One thought is that if a child can be built, previewed, and run in isolation, we'd want it to be responsible for showing alerts in order to exercise that functionality.

That's a good point, and maybe we're approaching this the wrong way.

In the project we're developing, we have a parent state which contains many children, but the children can each have their own error, and many children could feasibly have errors simultaneously. Since the views associated with the children are items in a list, they can't all show their alerts at once—they block each other. In a preview or example app with just the child this interference doesn't happen; there are no other alerts appearing.

I thought it made sense in a case like this for the parent to manage the logic for which alerts to display in which order, etc.

Our workaround was to make sure that no two children have alert states at the same time as each other, which worked for our case. I don't know that it would always work, though.

stephencelis:
@skyler-cricut Better late than never, but cherry-picked your commit into SwiftUINavigation here: https://github.com/pointfreeco/swiftui-navigation/pull/71

Made a few changes along the way, but if you still need this functionality it shipped today!

importRyan:
Thank you!

PR #888, opened 2021-11-11
Remove duplicate debug

kgrigsby59:
The tic-tac-toe example double prints all the actions. This is because .debug is applied twice to the appReducer. Once in RootView.swift and once in AppCore.swift. This PR removes it from RootView.swift.


mbrandonw (approved):
Good catch, thanks!

sadiklar:
18 Kas 2021 Per 02:46 tarihinde Soner Sadiklar ***@***.***>
şunu yazdı:

>

PR #900, opened 2021-11-17
Removing redundant state duplications

dankinsoid:
`CurrentValueSubject` and `.removeDupilcates()` led to redundant duplications of state and substates in every scope and `ViewStore`


stephencelis:
Hi @dankinsoid! Thanks for taking the time to put together a PR. However, we're having a little trouble understanding the changes and the motivation for them from the diff alone. Can you put together some more information? Ideally, could you describe how the problem shows up in a real-world code sample and how this diff fixes things?

dankinsoid:
Hi! 
I will give a slightly abstract example
```swift
struct LargeAppState {
    var largeSubState: LargeSubState
}

let store = Store<LargeAppState, AppAction>(...) // Original store, it stores LargeAppState

let substore1 = store.scope(\.largeSubState) // This substore stores LargeSubState and syncronizes with original store \.largeSubState
let substore2 = store.scope(\.largeSubState) // This substore stores another LargeSubState and syncronizes too

let viewStore = ViewStore(store) // This view store contains .removeDuplicates publisher, that stores LargeAppState value
let viewSubstore1 = ViewStore(substore1) // This substore contains publisher, that stores LargeSubState value
let viewSubstore2 = ViewStore(substore2) // This substore contains publisher, that stores LargeSubState value
```
Every substore and view store allocates additional memory for states and substates (if it's value types), instead of linking to original store's state, so memory usage grows ≈ O(ln) (?)
I'm not sure if this is a serious problem, since I did not take measurements, and this effect may be negligible due to COW for collections

dankinsoid:
The purpose of my proposal is O(1) memory usage when creating scopes and view stores

mbrandonw:
Hi @dankinsoid, we are going to close this without merging. Please feel free to open a discussion if you want to discuss it further.

PR #901, opened 2021-11-24
Update Xcode version from 13.0 to 13.1

ferologics:
This would be useful to have before merging #863


stephencelis (approved):
Looks good! Thanks!

PR #910, opened 2021-12-07
Add reset button to form binding case

onevcat:
There is a `.resetButtonTapped` action which indicates that a reset method should exist, but it is not there yet.

This pull request adds a reset button to the view and it sends the `.resetButtonTapped` to trigger a reset.


stephencelis (approved):
Ha good catch, thanks! 😅

PR #924, opened 2021-12-16
Fix UIKit Case Study warning

stephencelis:
Noticed a deprecation warning for using an old TCA API. I'm not 100% satisfied with the fix yet...but it's probably better than leaving as is.

### DIFF
--- a/Examples/CaseStudies/tvOSCaseStudies/FocusView.swift
+++ b/Examples/CaseStudies/tvOSCaseStudies/FocusView.swift
@@ -48,7 +48,13 @@ let focusReducer = Reducer<FocusState, FocusAction, FocusEnvironment> {
             .padding()
 
           HStack(spacing: 40) {
-            ForEach(1..<11) { index in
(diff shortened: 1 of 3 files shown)


Examples/CaseStudies/tvOSCaseStudies/FocusView.swift
```diff
@@ -48,7 +48,13 @@ let focusReducer = Reducer<FocusState, FocusAction, FocusEnvironment> {
...
 
           HStack(spacing: 40) {
-            ForEach(1..<11) { index in
+            ForEach(1..<6) { index in
+              Button(numbers[index]) {}
+                .prefersDefaultFocus(viewStore.currentFocus == index, in: self.namespace)
+            }
+          }
+          HStack(spacing: 40) {
+            ForEach(6..<11) { index in
```
stephencelis, line 57: Unrelated fix to render the tvOS focus study more cleanly: 
image

mbrandonw (approved):

PR #925, opened 2021-12-16
Handle `.fireAndForget` scheduling internally 

iampatbrown:
Just an idea following a discussion about `.fireAndForget`. Not too sure of the implications but thought I'd check if it's worth exploring further.


iampatbrown:
I'm aborting mission... for now :P Just realised this wouldn't do what I was thinking it would.

PR #930, opened 2021-12-20
Add AnimatedAction

tgrapperon:
This PR allows to promote any action into an animated action from the call site, including those who have unanimated asynchronous effects. 

It works in a similar fashion as `Binding`s: You conform the `Action` to some protocol with a special `case`, and you create a new reducer using the `.animations()` higher order reducer. When you send an action you want to animate, you wrap it as
```
.animated(.with(.spring(), action: theAction))
```

Under the hood, the higher order reducer re-emits the action on the scheduler's animation or transaction. The scheduler is specified as a `KeyPath` from the `Environment`.

I've encountered a few cases where none of the `ViewStore` animation variants I tried was working in sibling features when simply setting a value in the local state (which bubbled up, then down to the siblings). Re-emitting the action on `.mainQueue.animation()` did solve the issue, so I've implemented this abstraction. I don't know how to properly hierarchize this kind of animation with the other ones (from `binding.animation` and `ViewStore.send` for example). It's already a complex topic by itself, and maybe some generalization is possible. 

What do you think about it?

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
@@ -153,7 +153,7 @@ struct AnimationsView: View {
             .padding([.horizontal, .bottom])
           #if compiler(>=5.4)
           Button("Center") {
-            viewStore.send(.animated(.with(.spring(), action: .tapped(proxy.center))))
+            viewStore.send(.tapped(proxy.center).animation(.spring()))
           }
           .padding([.horizontal, .bottom])
           #endif
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
@@ -13,7 +13,9 @@ private let readMe = """
   To animate changes made to state through a binding, use the `.animation` method on `Binding`.
 
   To animate asynchronous changes made to state via effects, use the `.animation` method provided \
-  by the CombineSchedulers library to receive asynchronous actions in an animated fashion.
+  by the CombineSchedulers library to receive asynchronous actions in an animated fashion. You \
+  can also use the `.animations()` higher reducer with `.animated(AnimatedAction<Action>)` actions \
+  to receive and handle actions in the same animated way.
 
   Try it out by tapping or dragging anywhere on the screen to move the dot, and by flipping the \
   toggle at the bottom of the screen.
@@ -45,7 +47,8 @@ struct AnimationsState: Equatable {
   var isCircleScaled = false
 }
 
-enum AnimationsAction: Equatable {
+enum AnimationsAction: Equatable, AnimatableAction {
+  indirect case animated(AnimatedAction<Self>)
   case circleScaleToggleChanged(Bool)
   case dismissAlert
   case rainbowButtonTapped
@@ -63,6 +66,8 @@ let animationsReducer = Reducer<AnimationsState, AnimationsAction, AnimationsEnv
   state, action, environment in
 
   switch action {
+  case .animated:
+    return .none
   case let .circleScaleToggleChanged(isScaled):
     state.isCircleScaled = isScaled
     return .none
@@ -102,6 +107,7 @@ let animationsReducer = Reducer<AnimationsState, AnimationsAction, AnimationsEnv
     return .none
   }
 }
+.animations(scheduler: \.mainQueue)
 
 struct AnimationsView: View {
   @Environment(\.colorScheme) var colorScheme
@@ -145,6 +151,12 @@ struct AnimationsView: View {
           .padding()
           Button("Rainbow") { viewStore.send(.rainbowButtonTapped, animation: .linear) }
             .padding([.horizontal, .bottom])
+          #if compiler(>=5.4)
+          Button("Center") {
+            viewStore.send(.animated(.with(.spring(), action: .tapped(proxy.center))))
+          }
+          .padding([.horizontal, .bottom])
+          #endif
           Button("Reset") { viewStore.send(.resetButtonTapped) }
             .padding([.horizontal, .bottom])
         }
@@ -152,6 +164,10 @@ struct AnimationsView: View {
       }
     }
   }
+}
+
+fileprivate extension GeometryProxy {
+  var center: CGPoint { .init(x: size.width / 2, y: size.height / 2) }
 }
 
 struct AnimationsView_Previews: PreviewProvider {
@@ -184,3 +200,10 @@ struct AnimationsView_Previews: PreviewProvider {
     }
   }
 }
+
+#if compiler(<5.4)
+  // Make the `.animation()` higher reducer compile on older versions of Swift.
+  public extension Reducer where Action: AnimatableAction {
+    func animations<S>(scheduler: @escaping (Environment) -> S) -> Self where S: Scheduler { self }
+  }
+#endif
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
@@ -202,7 +202,7 @@ struct AnimationsView_Previews: PreviewProvider {
 }
 
 #if compiler(<5.4)
-  // Make the `.animation()` higher reducer compile on older versions of Swift.
+  // Makes the `.animation()` higher reducer compile on older versions of Swift.
   public extension Reducer where Action: AnimatableAction {
     func animations<S>(scheduler: @escaping (Environment) -> S) -> Self where S: Scheduler { self }
   }
--- a/Sources/ComposableArchitecture/SwiftUI/Animation.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Animation.swift
@@ -22,19 +22,19 @@ public struct AnimatedAction<Action> {
   
   /// Creates an animated action from an action and a animation.
   /// - Parameters:
-  ///   - animation: an `Animation`
   ///   - action: an `Action`
+  ///   - animation: an `Animation`
   /// - Returns: An action animating `action` with the animation `animation`.
-  public static func with(_ animation: Animation, action: Action) -> Self {
+  public static func action(_ action: Action, with animation: Animation) -> Self {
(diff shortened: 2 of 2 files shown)


tgrapperon:
Should we keep the order `.with(_: Animation, action: Action)`, like SwiftUI's block syntax for animations, or switch to the more fluent `.action(_ action: Action, with animation: Animation)` ? This last case will render as:
```swift
.animated(.action(.buttonTapped, with: .spring()))
```
The `AnimatedAction` wrapper is mostly used to provide pseudo-equatability for the action without making `Transaction` conforms to `Equatable`. If we drop `Transaction`s, we can forego the wrapper to simply require `indirect case animated(Action, Animation)` instead, as `Animation` is already `Equatable`.

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
```diff
@@ -153,7 +153,7 @@ struct AnimationsView: View {
             .padding([.horizontal, .bottom])
           #if compiler(>=5.4)
           Button("Center") {
-            viewStore.send(.animated(.with(.spring(), action: .tapped(proxy.center))))
+            viewStore.send(.tapped(proxy.center).animation(.spring()))
```
iampatbrown, line 156: NICE!

tgrapperon:
Thanks for the feedback @iampatbrown!

The `Effect` method is a nice one to have. If the scheduler was parametrized like in this PR, you could maybe avoid the test branching, and gain more flexibility for SwiftUI Previews too. You're most likely creating this effect at some place you have access to the `Environment` and to an eventual `mainQueue` dependency. I don't really understand the `delay(0)` when used in conjunction with `onReceive`. From my experience, `delay` makes the rest of the chain receive on its scheduler argument (but I may be wrong). I'm not fully understanding why you need a `delay` there. To make the action being processed in a second pass instead of synchronously?

One difference is that this PR allows to take animations decisions from the `View` itself instead of hardcoding it in the reducer. It can adapt to cases where you don't want to animate the first appearance but you want to animate changes for example. This can be done from the reducer, but it couples it more tightly with its `View`. Animations are not tested if I'm not mistaken, and I have the feeling they belong more to the `View` side than the model (but I may be wrong).

About the issue I had, I can't really extract something yet, but I'm in a super complex setup `SwiftUI` wise, with many optimization routes and custom `View`s. I had animations when changing the parameter programmatically, but not when using a custom control that was updating the parameter when scrolling a `ScrollView`, so it could be related to the `Runloop` or something else. I'll try to replicate the issue in a toy project if I can.

You're right about the api fluency! I've reordered the arguments, and I've also extended the protocol, so you can quickly create `.animated()` versions by calling `.animation(...)` or `.transition(...)` on the action you want to send itself. This is way nicer!

I need to check what happens if someone calls repeatedly these methods (I'm afraid it could deadlock). I guess some safeguards can be installed if required.

tgrapperon:
Am I right assuming that:
```swift
withAnimation(animation) {…}
``` 
is equivalent to:
```swift
withTransaction(Transaction(animation: animation)) {…}
```
?

iampatbrown:
From my understanding they are equivalent.

iampatbrown:
> I'm not fully understanding why you need a `delay` there too

I'm not exactly sure why it works... but it fixed issues I had going from 
```swift
nil -> state1 (without animation)
state1 -> state2 (with animation)
```
I agree about the view being responsible for animations. I'll have to play around a bit more. 

I do normally use the environment mainQueue or the ` .animation(_:value:)` view modifier. This is just an alternative I'm testing out :)

stephencelis:
Hey @tgrapperon! Sorry for the delay. We're just starting to go through our holiday backlog.

This looks like a fun, alternative way of handling animation! I'm having a little bit of trouble understanding the problem it's addressing that the existing solutions don't solve yet. Do you think you could help me out?

For example, in the case study you added this:

```swift
viewStore.send(.tapped(proxy.center).animation(.spring()))
```

This is equivalent to the following, right?

```swift
viewStore.send(.tapped(proxy.center), animation: .spring())
```

Or am I missing something?

You mention this:

> I've encountered a few cases where none of the `ViewStore` animation variants I tried was working in sibling features when simply setting a value in the local state (which bubbled up, then down to the siblings).

Do you think you could provide a demonstration of this?

I really do like that this code is completely additive, so one idea would be to ship this as another library for folks to take it for a spin. We really should start promoting some of these libraries better and give more visibility, like swift-composable-environment. We'll be looking into that soon.

tgrapperon:
Thanks for the feedback @stephencelis!

> This is equivalent to the following, right?

In this case, yes, as the changes to the state are synchronous with the action being received. When the `Store` did finish reducing the `.tapped` action, the state has the modified values, and this is caught up in the `withAnimation` block from `ViewStore(send:animation:)`. But the approach from this PR would also work if the changes to the state were asynchronous. For example, if the `.tapped` action triggers a `.task` that publishes results to the state later, out of `withAnimation` scope. The case study could be updated with such an action, where `ViewStore(send:animation:)` wouldn't animate. 

That's the rationale of the PR: turn any action, even with asynchronous effects, into an animatable action, and from the UI's side. This is an alternative to parametrizing an existing action with an `Animation` or `Transition` associated value, which may force to rewrite many things. 

For example, this can be exploited conveniently when we want to update something without animations during `onAppear`, but with animations when the user is interacting, all of this without clogging the reducer's logic with purely UI code (_and untestable until you update `swift-snapshot-testing` with videos!😉_). It would also provide a unique way to trigger animations from a `ViewStore` (bindings excluded, but maybe one can mimic the `.animation()` modifier there too, and provide only one story for animating changes in TCA).

About the issue I had, it may be related to some programming error or inconsistencies on my side. I'm developing a big project with many animations. More precisely, this is a guitar/music theory app with many interactive views, and each one responds to user input occurring in other parts of the UI, with animations, transitions, etc. Some parts are complex, and the (deterministic) layout is computed in a background task, with the results being published later.
As there are many different views, and as I've updated the way I'm handling animations as new use-cases appeared, some views are still observing the state for animations themselves, and other ones aren't making any decisions on this matter (the new approach I'm using). That's why I'm not sure I can reproduce the issue in a toy app. But I'm on the point of assembling many features into bigger ones, so if I spot something, I'll try to track down the issue and extract something if what's happening is not clear.

Thanks for mentioning `swift-composable-environment`! [^1]. I can extract this PR as a side project indeed. Let me know if you think that's a better approach. In the meantime, I'm converting the PR to a draft.

[^1]: I've recently released a very WIP version of [`swift-derived-state`](https://github.com/tgrapperon/swift-derived-state) which is a protocolized version of the `StateBinding` I've discussed in #451. Faster, but unfortunately still too slow because of generic `KeyPath`'s (benchmarks provided). However, it could be manageable with code-generation of ad-hoc equivalents during builds with the [next version of SPM](https://github.com/apple/swift-evolution/blob/main/proposals/0303-swiftpm-extensible-build-tools.md). 
I didn't write the documentation yet, but tests are showing a few use-cases.
The user could use the slower, but faster to implement, `DerivedState` when iterating the app, and then convert it to ad-hoc code automatically when the API is stabilized. But I digress!

shawnkoh:
> I've been using this helper for animating side effects.
> 
> ```swift
> extension Effect {
>   public func animation(_ animation: Animation? = .default) -> Self {
>     #if DEBUG
>     if ProcessInfo.processInfo.environment["XCTestConfigurationFilePath"] != nil {
>       return self
>     }
>     #endif
>     return self
>       .delay(for: 0, scheduler: DispatchQueue.main)
>       .receive(on: DispatchQueue.main.animation(animation))
>       .eraseToEffect()
>   }
> }
> ```
> 
> It's probably not ideal but I've found it useful. The main thing I needed it for was to change state then animate a `matchedGeometryEffect`. This is how it's used in the reducer
> 
> ```swift
> case let .tapped(ingredient):
>   state.selection = .init(ingredient: ingredient)
>   return Effect(value: .openCard).animation(.openCard)
> 
> case .openCard:
>   state.selection?.displayMode = .cardFront
>   return .none 
> ```
> 
>  > I've encountered a few cases where none of the ViewStore animation variants I tried was working in sibling features when simply setting a value in the local state (which bubbled up, then down to the siblings)
> 
> Did you have a small example of this? I wouldn't mind trying a few things out.
> 
> Regarding the fluency, I don't mind `send(_ action: Action, animation: Animation?)` and would possibly try to keep it similar either by changing the `animation:` label or passing some sort of wrapper eg: `animation: ActionAnimation`
> 
> ```swift
> viewStore.send(.animated(.with(.spring(), action: .tapped(proxy.center))))
> ```
> 
> This feels a bit noisy to me and I think it can be hidden.
> 
> ```swift
> viewStore.send(.animated(.action(.buttonTapped, with: .spring())))
> ```
> 
> This seems better and sometimes my animations can end up pretty long so it works well for me on the right.
> 
> Anyway, some cool ideas, keen to see it take shape.

Sorry this is an aside: but could you explain the use case for delaying by 0 on the main scheduler?

iampatbrown:
@shawnkoh It allows SwiftUI to receive an initial state change `A -> B` without an animation, then perform the animation on the second state change `B -> C`. It's similar to `DispatchQueue.main.async { ... }`. Without the delay it goes `A -> C`. 

I use deferred now though. Here is the updated extension:

```swift
import ComposableArchitecture
import SwiftUI

#if DEBUG
extension ProcessInfo {
  fileprivate static var isRunningXCodeUnitTests: Bool {
    processInfo.environment["XCTestConfigurationFilePath"] != nil
  }

  fileprivate static var isRunningSPMTests: Bool {
    processInfo.arguments.contains(where: { $0.hasSuffix("xctest") })
  }

  fileprivate static var isRunningUnitTests: Bool {
    Self.isRunningXCodeUnitTests || Self.isRunningSPMTests
  }
}
#endif

extension Effect {
  public func animation(_ animation: Animation? = .default) -> Self {
    #if DEBUG
    // NB: Perform effect synchronously during unit tests
    if ProcessInfo.isRunningUnitTests { return self }
    #endif
    return self.deferred(for: 0, scheduler: DispatchQueue.main.animation(animation))
  }
}
```

If you use `deferred` with a scheduler you provide to the environment, you need to use the immediate scheduler during tests. The motivation behind this extension was to avoid passing a queue to a deeply nested feature. 

I hope that makes sense.

tgrapperon:
I will implement this feature as a third-party library.

stephencelis:
Please do share it wish us when you do! We'd like to start aggregating a list somewhere.

PR #935, opened 2021-12-22
Reuse `Store`s & `ViewStore`s

tgrapperon:
Today's discussion about `StateObject`'s (#934) made me try to tackle this related problem again, as `@StateObject` would ultimately not be required if the `ViewStore` instances are preserved.
This is by no mean a complete PR, as there are probably many sharp edges.

To sum up, the `Store` caches weakly its scoped children and viewStores. Each store owns its cache and reuses instances when possible.

### ViewStore
The designated way to create a viewStore is by calling the `.viewStore` property on Store's of `Equatable`, or the `.viewStore(deduplicate:)` method in the general case. When possible, the same viewStore instance will be returned each time. `ViewStore(store)` doesn't return a cached instance (because it's an initializer of a reference type, we can't assign an existing instance to `self`).  `WithViewStore`'s are using the cached value when possible, however.

### ScopeIdentifier
There is a new type called `ScopeIdentifier` that identifies a specific scope configuration. This `scopeIdentifier` is used to key the cached child stores instances. The same type is used to key the `ViewStore`'s cache.

We should provide a `ScopeIdentifier` when scoping a store or creating a `ViewStore`. `ForEachStore`, `SwitchStore`, and `CaseLet` views are accepting such `ScopeIdentifier?` as an argument. If the value is `nil`, caching is disabled. `IfLsetStore` doesn't need any `ScopeIdentifier` as the operation is univocal.

Because providing a `ScopeIdentifier` each time a store is scoped is annoying, a contextual identifier can be automatically derived using the `#fileID` and `#line` of the calling function. Furthermore, a static `SharedStoreConfiguration` type was added. This type has one static property called `isAutomaticReuseOfStoreAndViewStoreInstancesEnabled: Bool` . 

**When this value is set to `true` (the default for now), current clients of TCA are automatically reusing `Store` and `ViewStore` without having to touch the client's code**.

This automatic `scopeIdentifier` derivation should work in most cases, and this is probably the most complex issue with the subject. At the top of my head, it will fail:
- If the same store is scoped twice on the same line. One can make automatic identifiers a little more resilient by augmenting them with `LocalState.self` and `LocalAction.self`, or maybe with the `#column` value.
- If scoping is parametric and can change during the lifetime of the `store`, that is if the functions (not their values) that are transforming `LocalState` and `LocalAction` are changing. This looks quite contrived, but this is theoretically possible.

ViewStore reusing is more simple, as the only variable is the deduplication function which is involved less frequently and very likely static.

I particularly like the `store.viewStore` naming, as I find it more logical. Maybe it should be ultimately renamed `view`.

I didn't write the documentation for the new overloads yet, as I'm awaiting feedback. @mbrandonw, @stephencelis, please let me know if something is not clear.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/00-Core.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/00-Core.swift
@@ -304,7 +304,7 @@ let rootReducer = Reducer<RootState, RootAction, RootEnvironment>.combine(
       environment: { .init(mainQueue: $0.mainQueue, webSocket: $0.webSocket) }
     )
 )
-//.debug()
+.debug()
 .signpost()
 
 private func liveFetchNumber() -> Effect<Int, Never> {
--- a/Examples/CaseStudies/SwiftUICaseStudies/00-Core.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/00-Core.swift
@@ -304,7 +304,7 @@ let rootReducer = Reducer<RootState, RootAction, RootEnvironment>.combine(
       environment: { .init(mainQueue: $0.mainQueue, webSocket: $0.webSocket) }
     )
 )
-.debug()
+//.debug()
 .signpost()
 
 private func liveFetchNumber() -> Effect<Int, Never> {
--- a/Examples/CaseStudies/SwiftUICaseStudies/02-Effects-Timers.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/02-Effects-Timers.swift
@@ -60,7 +60,7 @@ struct TimersView: View {
   @ObservedObject var viewStore: ViewStore<TimersState, TimersAction>
 
   init(store: Store<TimersState, TimersAction>) {
-    self.viewStore = ViewStore(store)
+    self.viewStore = store.viewStore
   }
 
   var body: some View {
--- a/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-ElmLikeSubscriptions.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-ElmLikeSubscriptions.swift
@@ -85,7 +85,7 @@ struct ClockView: View {
   @ObservedObject var viewStore: ViewStore<ClockState, ClockAction>
 
   init(store: Store<ClockState, ClockAction>) {
-    self.viewStore = ViewStore(store)
+    self.viewStore = store.viewStore
   }
 
   var body: some View {
--- a/Examples/CaseStudies/UIKitCaseStudies/CounterViewController.swift
+++ b/Examples/CaseStudies/UIKitCaseStudies/CounterViewController.swift
@@ -31,7 +31,7 @@ final class CounterViewController: UIViewController {
   var cancellables: Set<AnyCancellable> = []
 
   init(store: Store<CounterState, CounterAction>) {
-    self.viewStore = ViewStore(store)
+    self.viewStore = store.viewStore
     super.init(nibName: nil, bundle: nil)
   }
 
--- a/Examples/CaseStudies/UIKitCaseStudies/ListsOfState.swift
+++ b/Examples/CaseStudies/UIKitCaseStudies/ListsOfState.swift
@@ -29,7 +29,7 @@ final class CountersTableViewController: UITableViewController {
 
   init(store: Store<CounterListState, CounterListAction>) {
     self.store = store
-    self.viewStore = ViewStore(store)
+    self.viewStore = store.viewStore
     super.init(nibName: nil, bundle: nil)
   }
 
--- a/Examples/CaseStudies/UIKitCaseStudies/LoadThenNavigate.swift
+++ b/Examples/CaseStudies/UIKitCaseStudies/LoadThenNavigate.swift
@@ -61,7 +61,7 @@ class LazyNavigationViewController: UIViewController {
 
   init(store: Store<LazyNavigationState, LazyNavigationAction>) {
     self.store = store
-    self.viewStore = ViewStore(store)
+    self.viewStore = store.viewStore
     super.init(nibName: nil, bundle: nil)
   }
 
--- a/Examples/CaseStudies/UIKitCaseStudies/NavigateAndLoad.swift
+++ b/Examples/CaseStudies/UIKitCaseStudies/NavigateAndLoad.swift
@@ -60,7 +60,7 @@ class EagerNavigationViewController: UIViewController {
 
   init(store: Store<EagerNavigationState, EagerNavigationAction>) {
     self.store = store
-    self.viewStore = ViewStore(store)
+    self.viewStore = store.viewStore
     super.init(nibName: nil, bundle: nil)
   }
 
--- a/Examples/TicTacToe/tic-tac-toe/Sources/GameUIKit/GameViewController.swift
+++ b/Examples/TicTacToe/tic-tac-toe/Sources/GameUIKit/GameViewController.swift
@@ -29,7 +29,7 @@ public final class GameViewController: UIViewController {
 
   public init(store: Store<GameState, GameAction>) {
     self.store = store
-    self.viewStore = ViewStore(store.scope(state: ViewState.init))
+    self.viewStore = store.scope(state: ViewState.init).viewStore
     super.init(nibName: nil, bundle: nil)
   }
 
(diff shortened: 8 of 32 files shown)


tgrapperon:
I've refactored a few things:
- Automatic `ScopeIdentifier` derivation now also uses the call site column, so this should tackle the possible mismatches with multiple scopes of the same store on the same line.
- `isAutomaticReuseOfStoreAndViewStoreInstancesEnabled` was renamed to `shouldInferScopeIdentifiers`
- The two booleans used to configure reuse logs were split into an `OptionSet` to provide more control. The two principal modes are `.nonReusable` to spot non-reusable `Store` and `ViewStore` instantiations, and `all` to inspect all reuse activity. They can be set as `SharedStoreConfiguration.printOptions = .all` for example.
- The `SharedStoreConfiguration` is most likely transitory and only there to give a better understanding of the reuse process. It should probably be removed at some point.
- I'm not satisfied by presenting a `ScopeIdentifier` as a reuse identifier for `ViewStore`. ScopeIdentifier` should maybe be renamed to a more general `ReuseIdentifier`.
- `ViewStore(store)` initializer is now displaying a runtime warning when `SharedStoreConfiguration.printOptions` contains `.nonReusableViewStore`. This initializer should probably be deprecated in favor of `store.viewStore` at some point (this would be wild!). Fortunately, the fix would be simple: 
```swift
ViewStore(store, removeDuplicates:…) -> store.viewStore(removeDuplicates:…), and
ViewStore(store) -> store.viewStore
```
_Maybe it's too late to discuss the naming of `ViewStore` vs `StoreView` (or maybe it was already discussed). I'll maybe open a separate discussion to see how `StoreView` would be received/perceived if this PR goes farther and `store.viewStore` makes it into the library. `store.view` would be a pretty neat way to provide read/write access to a store, but it would feel a little awkward if it returned a `ViewStore` and not a `StoreView`._

iampatbrown:
Hey @tgrapperon.

Some good work. I haven't had a chance to fully go through it but I wanted to give my initial thoughts. 

- This is a fairly big change with some new concepts and could potentially be broken up into multiple changes.
- I'm guessing there will likely be changes to `Store` as more concurrency support is implemented. Maybe caching stores is something that could be considered at the same time. 
- Is there a way to demonstrate the performance benefits of using cached stores? 
- Is it possible to achieve some of this behaviour using a property wrapper similar to `@StateObject`?
- I think `store.viewStore` or `store.view` might blur the lines between `Store` and `ViewStore`, possibly leading people to misuse `store.viewStore.state` or at that point wonder why it's not just `store.state`.
- The `ScopeIdentifier` id generation is pretty cool.
- I have a feeling the current implementation is causing a retain cycle. I just did some small tests using the `Todos` example and noticed stores were being retained for deleted todo items.

I'll test this out a bit more in the New Year and let you know how it goes :)

tgrapperon:
Thanks for the feedback @iampatbrown. A few comments until you experiment further:

> This is a fairly big change with some new concepts and could potentially be broken up into multiple changes.

Indeed. I'm just starting the discussion, but I guess it's easier to discuss on some code than on abstract ideas.

>  I'm guessing there will likely be changes to Store as more concurrency support is implemented. Maybe caching stores is something that could be considered at the same time.

I agree. This, or the potential switch to protocolized `Reducer`s if it happens someday.

> Is there a way to demonstrate the performance benefits of using cached stores?

Pure model benchmarks are not showing benefits. I don't know how it translates to a realistic app, with a complex model, and SwiftUI receiving new ObservedObject or States. I guess there should be cases where looking up the store in a dictionary and returning the same instance is faster than sending a new instance to SwiftUI to let it check if something has changed. We unfortunately don't have dedicated UI benchmarks yet. I haven't checked the effects on memory usage either, as this is quite difficult to measure.

> Is it possible to achieve some of this behaviour using a property wrapper similar to @StateObject? 

I'm not sure to understand exactly the question: The reusing part, or creating a property wrapper conforming to `DynamicProperty` for SwiftUI? For the former, I need to think about it. For the latter, I guess that `@ObservedObject` is fine as we are owning and reusing the `ViewStore` instances. `@StateObject` is a very special beast and it's almost as difficult to parametrize it as `@State`. It is mainly used to keep the app's object model in the root view, or as some kind of more refined `@State` for inner views. 

>  I think store.viewStore or store.view might blur the lines between Store and ViewStore, possibly leading people to misuse store.viewStore.state or at that point wonder why it's not just store.state.

That's a byproduct of keeping the `ViewStore` instances in the `Store`s. I like the `store.view(Store)` because we understand that the store is the subject, and we derive a special thing to poke into it. The `Store` is an opaque thing, we can't touch it bare hands, and we need to open a window to see the state or send actions. It also bears some resemblance with `String`'s UTF's views which are not instantiated manually. But this is a neverending debate, I guess.  I don't really see however how one could misuse `store.viewStore.state` more than `ViewStore(store).state`. 

> The ScopeIdentifier id generation is pretty cool.

Thanks! I was trying to make `CasePath` equatable, etc., and I had this epiphany under the shower. There are corner cases though. Maybe this trick can be reused elsewhere!

> I have a feeling the current implementation is causing a retain cycle.

I had the same impression, but I also had the same one when disabling the part that is setting the cache. Maybe SwiftUI keeps things around a little bit. If you log something in `Store.deinit` or `ViewStore.deinit`, you can see that sometimes nothing is released when navigating from a feature back to the main list in the case study app for example (even with the current, non-reusing, version of TCA). But it is also possible that I've missed something. Please let me know if you find anything!

Thanks again for the feedback!

iampatbrown:
> it's easier to discuss on some code than on abstract ideas.

Yeah the code helps heaps :)

> or the potential switch to protocolized Reducer

That's essentially what I was thinking about as well. Potentially some big changes to the internals which may impact this PR.

> > Is it possible to achieve some of this behaviour using a property wrapper similar to @StateObject?
>
> I'm not sure to understand exactly the question

I meant more could the caching be achieved without modifying the internals. Either by wrapping a store in something else or using a property wrapper + projectedValue. Was just thinking as an alternative, that would also make it easier to test things out in the short term. 

> I guess. I don't really see however how one could misuse store.viewStore.state more than ViewStore(store).state. Do you have an example?

```swift
TextField("Type here", text: store.viewStore.binding(\.$text))
vs
TextField("Type here", text: ViewStore(store).binding(\.$text))
```
I think the `ViewStore(store).binding(\.$text)` example is easier to recognize as being used incorrectly. Also using `store.viewStore.state` anywhere in a `View` without observing changes might occur more frequently compared to `ViewStore(store).state` 

> Please let me know if you find anything!

Will do :)

tgrapperon:
About the property wrapper, this could be tricky, but potentially doable at first sight. Right now, the `Store` holds weak references to the scoped stores and `ViewStore`. I guess it's possible to store them aside of the store, in some wrapper type. One would need to interact from this type (scoping, `ViewStore`'ing). And one would also need to keep a reference to this wrapper instance, which is something we don't need to do with the `Store`'s solution. I'll experiment with it once I have a few moments.

maximkrouk:
Didn't have time to take a closer look, but looks a bit related to https://github.com/pointfreeco/swift-composable-architecture/pull/847#issuecomment-1005526981

tgrapperon:
There's not a lot of traction around this PR, so I'm closing it for now.

maximkrouk:
@iampatbrown 

> I think `store.viewStore` or `store.view` might blur the lines between `Store` and `ViewStore`, possibly leading people to misuse `store.viewStore.state` or at that point wonder why it's not just `store.state`.

Fair, but actually this part of the design causes some questions by itself, I'd vote to make some stores' API public (like state, pure state publisher and `send` method), it might be helpful in some simple cases.
Also, I'm not sure if StorePublisher works as a performance tool for UIKit, it can minimize redundant SwiftUI.View renders, but it causes more heavy operations by comparing more data (I came to using `[dynamicMember: \.items.count]`, instead of `.items.count` for UIKit, so that collections are not compared for the publisher, only their `count` property).

> Is there a way to demonstrate the performance benefits of using cached stores?

I think UIKit scoping for collection view cells is the case, you don't have to create a new store if the cell already setup, but this scoping should not rely on the file and line.

-----
@tgrapperon 

> Maybe it's too late to discuss the naming of ViewStore vs StoreView (or maybe it was already discussed). I'll maybe open a separate discussion to see how StoreView would be received/perceived if this PR goes farther and store.viewStore makes it into the library. store.view would be a pretty neat way to provide read/write access to a store, but it would feel a little awkward if it returned a ViewStore and not a StoreView.

Probably +1 for that. Maybe it's a bit too late, but `ViewStore` sounds like a store for views, `StoreView` may sound like a UI-component, but from the system perspective it's more like a View-layer for Store (if the Store is our UI-less app/system, so it's a way to interact with a store)

stephencelis:
@tgrapperon We definitely think there's something to your exploration here, but we haven't had time in the last month since holiday to look into it further. Thanks for sharing, though! We hope to give things a closer look soon.

tgrapperon:
No problems. It was more a proof of concept than anything else, to serve as a basis for discussion.

PR #945, opened 2022-01-04
Improved way to get SwiftUI image handle

mcfans:
Hi!
I love this new runtime warning feature, however I think there's a better way to get SwiftUI image handle than hardcode a SwiftUI symbol name. In this PR, I iterate over dyld images and search SwiftUI directly. I think this approach is more straightforward.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift
+++ b/Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift
@@ -8,9 +8,18 @@
   // Feedback filed: https://gist.github.com/stephencelis/a8d06383ed6ccde3e5ef5d1b3ad52bbc
   let rw = (
     dso: { () -> UnsafeMutableRawPointer in
-      var info = Dl_info()
-      dladdr(dlsym(dlopen(nil, RTLD_LAZY), "LocalizedString"), &info)
-      return info.dli_fbase
+      let count = _dyld_image_count()
+      for i in 0..<count {
+        if let name = _dyld_get_image_name(i) {
+          let swiftString = String(cString: name)
+          if swiftString.hasSuffix("SwiftUI") {
+            if let pointer = UnsafeMutableRawPointer(bitPattern: _dyld_get_image_vmaddr_slide(i)) {
+              return pointer
+            }
+          }
+        }
+      }
+      fatalError("Import SwiftUI to use this")
     }(),
     log: OSLog(subsystem: "com.apple.runtime-issues", category: "ComposableArchitecture")
   )
--- a/Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift
+++ b/Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift
@@ -12,9 +12,9 @@
       for i in 0..<count {
         if let name = _dyld_get_image_name(i) {
           let swiftString = String(cString: name)
-          if swiftString.hasSuffix("SwiftUI") {
-            if let pointer = UnsafeMutableRawPointer(bitPattern: _dyld_get_image_vmaddr_slide(i)) {
-              return pointer
+          if swiftString.hasSuffix("/SwiftUI") {
+            if let header = _dyld_get_image_header(i) {
+              return UnsafeMutableRawPointer(mutating: UnsafeRawPointer(header))
             }
           }
         }
--- a/Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift
+++ b/Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift
@@ -19,7 +19,7 @@
           }
         }
       }
-      fatalError("Import SwiftUI to use this")
+      return UnsafeMutableRawPointer(mutating: #dsohandle)
     }(),
     log: OSLog(subsystem: "com.apple.runtime-issues", category: "ComposableArchitecture")
   )


Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift
```diff
@@ -8,9 +8,18 @@
...
   let rw = (
     dso: { () -> UnsafeMutableRawPointer in
-      var info = Dl_info()
-      dladdr(dlsym(dlopen(nil, RTLD_LAZY), "LocalizedString"), &info)
-      return info.dli_fbase
+      let count = _dyld_image_count()
+      for i in 0..<count {
+        if let name = _dyld_get_image_name(i) {
+          let swiftString = String(cString: name)
+          if swiftString.hasSuffix("SwiftUI") {
```
stephencelis, line 15: Would there be a potential problem if another library has "SwiftUI" as a suffix?

Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift
```diff
@@ -8,9 +8,18 @@
...
+            if let pointer = UnsafeMutableRawPointer(bitPattern: _dyld_get_image_vmaddr_slide(i)) {
+              return pointer
+            }
+          }
+        }
+      }
+      fatalError("Import SwiftUI to use this")
```
stephencelis, line 22: Is there a way to avoid introducing a fatal error here? Our hope would be this would be a no-op, though maybe it already is a fatal error hiding in plain sight in the existing implementation.

stephencelis (commented):
Hey @mcfans! Thanks for taking the time to PR. I was hoping you could provide a little more info as to what the changes do to make it "better" or "more straightforward"? Is it somehow more performant, or are the `_dyld_*` functions more widely known? We don't work with dynamic library code very often, but we're more familiar with the functions we reached for 😅

mcfans:
> Hey @mcfans! Thanks for taking the time to PR. I was hoping you could provide a little more info as to what the changes do to make it "better" or "more straightforward"? Is it somehow more performant, or are the `_dyld_*` functions more widely known? We don't work with dynamic library code very often, but we're more familiar with the functions we reached for 😅

Oh I mean the logic is more straightforward. The new logic is that when you need the SwiftUI image handle here and we just search in

mcfans:
> Hey @mcfans! Thanks for taking the time to PR. I was hoping you could provide a little more info as to what the changes do to make it "better" or "more straightforward"? Is it somehow more performant, or are the `_dyld_*` functions more widely known? We don't work with dynamic library code very often, but we're more familiar with the functions we reached for 😅

Oh I mean this logic is more straightforward. Old behavior is get handle of main program -> find some SwiftUI symbol by this handle -> find this symbol’s image ->  get it’s address. New behavior is get all images -> find one named SwiftUI -> return it’s address.
New behavior doesn’t require a hard code symbol name. Although it used more code than previous solution, it just a for loop using dyld API.

Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift
mcfans: yeah I changed this to /SwiftUI so some library named XXSwiftUI should be filtered

stephencelis:
> Oh I mean this logic is more straightforward. Old behavior is get handle of main program -> find some SwiftUI symbol by this handle -> find this symbol’s image -> get it’s address. New behavior is get all images -> find one named SwiftUI -> return it’s address.
> New behavior doesn’t require a hard code symbol name. Although it used more code than previous solution, it just a for loop using dyld API.

Ah yeah. I definitely agree that the loop/predicate against the module name is less direct than a lookup that comes from a symbol inside the module. I think ultimately we're a little worried about introducing this change right now, though, mainly because we're comfortable with the code that's currently running and don't foresee it breaking anytime soon, and because there are some subtle changes in the PR that we haven't had time to fully grok.

Are you OK with us closing this for now? If for some reason there is a functional problem with our current solution, we'd be happy to address it, but we aren't aware of one right now.

stephencelis (commented):
This fixes #958, though runtime warnings don't appear to be `os_log`-privileged for iOS 13 (maybe those warnings were added in 14?)

We could rely on a different symbol that was available back in iOS 13. From the following command depending on iOS Simulator install location:

```shell
nm -gU '/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 13.7.simruntime/Contents/Resources/RuntimeRoot/System/Library/Frameworks/SwiftUI.framework/SwiftUI' | perl -e 'print sort { length($a) <=> length($b) } <>'
```

But this PR seems preferable to me now. Thanks @mcfans!

stephencelis (approved):

JosephDuffy:
I've been implementing this myself over the last few days and was put off using the `_dyld_` functions because they are not thread safe. I also found it to be possible to use `dladdr` across platforms by using `$s7SwiftUI4AxisON` for the symbol, which works on all platforms. It's possible to use `_getMangledTypeName` but that's not included on all platforms that support SwiftUI since it was added in Swift 5.3.

One thing to note is that the man page says when using `dlopen` you should call `dlclose` when finished.

I'm still testing which platforms this works on. iOS 14.5 is working but iOS 14.3 it doesn't show the runtime warning.

stephencelis:
> One thing to note is that the man page says when using `dlopen` you should call `dlclose` when finished.

Since this is debug code and a handle to a dynamic library that I don't think would ever be deallocated on `dlclose`, do you think this is necessary? Is failing to call `dlclose` an issue in some regard?

> I'm still testing which platforms this works on. iOS 14.5 is working but iOS 14.3 it doesn't show the runtime warning.

Please let us know what you find!

PR #949, opened 2022-01-06
Update `.cancellable` implementation to fix #946

### LINKED ISSUE #946: Only first value published when using Effect.merge() and .cancellable()
**Describe the bug**
I've found that using `Effect.merge()` and `.cancellable()` together can in some situations cause an effect to end after its first value is published, rather than waiting for all values to be published. Here's a reduced test case, with a couple of variations that do not fail as well:

**To Reproduce**

```swift
import Combine 
import ComposableArchitecture 
import XCTest 

class MergeCancelTest: XCTestCase { 
  var cancellables: Set<AnyCancellable> = []

  func testNestedMergeCancellation() {
    // The test passes if you change this to Effect.concatenate
    let effect = Effect<Int, Never>.merge(
      (1...2).publisher
        // The test passes if you add .receive(on: testScheduler) and advance the scheduler before XCTAssertEqual
        .eraseToEffect()
        // The test passes if you remove this .cancellable() call
        .cancellable(id: 1)
    )
    // The test passes if you remove this .cancellable() call
    .cancellable(id: 2)
     
    var output: [Int] = []
    effect
      .sink { output.append($0) }
      .store(in: &cancellables)
     
    XCTAssertEqual(output, [1, 2])
  }
} 
```

I've wrapped this up in a test project which shows the failing case plus the multiple variations mentioned in the comments: [MergeCancel.zip](https://github.com/pointfreeco/swift-composable-architecture/files/7809799/MergeCancel.zip)

**Expected behavior**
Both values (1 and 2) should be passed to the `.sink` closure and the test should pass. Instead it fails with this error:

``` 
MergeCancelTests/MergeCancelTests.swift:32: error: -[MergeCancelTests.MergeCancelTest testNestedMergeCancellation] : XCTAssertEqual failed: ("[1]") is not equal to ("[1, 2]")
``` 

As the comments indicate, I know of three ways to make this test pass: 

1. Change `Effect.merge` to `Effect.concatenate`
2. Change the inner publisher to use a non-immediate scheduler
3. Remove either (or both) of the `.cancellable(id:)` calls

(In my current project I've chosen option 2 to get my tests passing, but it makes the tests more complicated than necessary.)

I'm pretty sure the bug is related to `Effect.cancellable` using a `PassthroughSubject` internally, and subscribing that subject to the publisher before the subject has been connected to any downstream subscribers. But I don't understand it fully, because clearly that works just fine in most cases.

**Environment**
  - swift-composable-architecture version 0.32.0
  - Xcode 13.0
  - Swift 5.5
  - OS: macOS 12.0

iampatbrown:
Possible fix for #946 
Main issue with `PassthroughSubject` is that it could only forward 1 value from an upstream `PassthroughSubject`, ie. nested cancellable effects. This should fix that by subscribing directly to the upstream publisher.

### DIFF
--- a/Examples/VoiceMemos/VoiceMemosTests/VoiceMemosTests.swift
+++ b/Examples/VoiceMemos/VoiceMemosTests/VoiceMemosTests.swift
@@ -113,7 +113,7 @@ class VoiceMemosTests: XCTestCase {
     environment.audioRecorder.startRecording = { _ in
       audioRecorderSubject.eraseToEffect()
     }
-    environment.mainRunLoop = .immediate
+    environment.mainRunLoop = self.mainRunLoop.eraseToAnyScheduler()
     environment.temporaryDirectory = { .init(fileURLWithPath: "/tmp") }
     environment.uuid = { UUID(uuidString: "DEADBEEF-DEAD-BEEF-DEAD-BEEFDEADBEEF")! }
 
@@ -124,6 +124,7 @@ class VoiceMemosTests: XCTestCase {
     )
 
     store.send(.recordButtonTapped)
+    self.mainRunLoop.advance(by: 0.5)
     store.receive(.recordPermissionResponse(true)) {
       $0.audioRecorderPermission = .allowed
       $0.currentRecording = .init(
@@ -133,6 +134,7 @@ class VoiceMemosTests: XCTestCase {
       )
     }
     audioRecorderSubject.send(completion: .failure(.couldntActivateAudioSession))
+    self.mainRunLoop.advance(by: 0.5)
     store.receive(.audioRecorder(.failure(.couldntActivateAudioSession))) {
       $0.alert = .init(title: .init("Voice memo recording failed."))
       $0.currentRecording = nil
@@ -189,7 +191,7 @@ class VoiceMemosTests: XCTestCase {
   func testPlayMemoFailure() {
     var environment = VoiceMemosEnvironment.failing
     environment.audioPlayer.play = { _ in Effect(error: .decodeErrorDidOccur) }
-    environment.mainRunLoop = .immediate
+    environment.mainRunLoop = self.mainRunLoop.eraseToAnyScheduler()
 
     let url = URL(string: "https://www.pointfree.co/functions")!
     let store = TestStore(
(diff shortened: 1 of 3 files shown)


Examples/VoiceMemos/VoiceMemosTests/VoiceMemosTests.swift
```diff
@@ -124,6 +124,7 @@ class VoiceMemosTests: XCTestCase {
     )
 
     store.send(.recordButtonTapped)
```
iampatbrown, line 126: `.recordButtonTapped` ends up returning https://github.com/pointfreeco/swift-composable-architecture/blob/313dd217dcd1d0478118ec5d15225fd473c1564a/Examples/VoiceMemos/VoiceMemos/VoiceMemos.swift#L70-L76
using `RunLoop.immediate` results in a failed test because `.currentRecordingTimerUpdated` now also gets received once before `.audioRecorder(.failure(.couldntActivateAudioSession))` is received. I'm unsure if this is an undesired result of the changed.

Adding following test  to `TimerTests.swift` demonstrates the difference in behaviour.

```swift
 func testTimerImmediate() {
    let scheduler = DispatchQueue.immediate

    var count = 0

    Effect.timer(id: 1, every: .seconds(1), on: scheduler)
      .sink { _ in count += 1 }

    XCTAssertNoDifference(count, 1) // This PR

    // XCTAssertNoDifference(count, 0) // Current implementation
  }
```

iampatbrown: For the current implementation, the timer is triggered as soon as the `PassThroughSubject` subscribes to it, essentially dropping the value before the effect is subscribed to downstream. 

https://github.com/pointfreeco/swift-composable-architecture/blob/313dd217dcd1d0478118ec5d15225fd473c1564a/Sources/ComposableArchitecture/Effects/Cancellation.swift#L36-L37

So the timer auto connects and outputs value at line 37 which is lost before returning the effect.

In this PR the timer auto connects at `.sink { _ in count += 1 }`

stephencelis: Interesting find! I think overall this PR is an improvement in the expectation that at least 1 thing emits from a timer running on an immediate scheduler, even if it is a breaking change.

stephencelis (approved):
Great work @iampatbrown! Thanks for digging into the problem.

PR #970, opened 2022-01-15
Fix filename

rono23:
Probably the file name is wrong, so I fixed it.

| Before | After |
|---|---|
| image | image |


stephencelis (approved):
Good catch, thanks!

PR #976, opened 2022-01-21
Cancel in flight effects inline instead of returning concatenated effect

iampatbrown:
While investigating #968 I realised the in flight cancellation could probably be done inline. Avoids an extra call to `cancellablesLock` which resulted in ~70% performance bump

```swift
benchmark("cancelInFlight concatenate") {
  let _ = Effect(value: 1)
    ._cancellable(id: 1, cancelInFlight: true)
    .sink { _ in }
}

benchmark("cancelInFlight inline") {
  let _ = Effect(value: 1)
    .cancellable(id: 1, cancelInFlight: true)
    .sink { _ in }
}

// cancelInFlight concatenate 16829.000 ns ±  29.10 %      78900
// cancelInFlight inline       9898.000 ns ±  28.35 %     136377
```

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -29,40 +29,40 @@ extension Effect {
   ///     canceled before starting this new one.
   /// - Returns: A new effect that is capable of being canceled by an identifier.
   public func cancellable(id: AnyHashable, cancelInFlight: Bool = false) -> Effect {
-    let effect = Deferred {
-      ()
-        -> Publishers.HandleEvents<
-          Publishers.PrefixUntilOutput<Self, PassthroughSubject<Void, Never>>
-        > in
-      cancellablesLock.lock()
-      defer { cancellablesLock.unlock() }
+    Deferred { () -> Publishers.HandleEvents<
+      Publishers.PrefixUntilOutput<Self, PassthroughSubject<Void, Never>>
+    > in
+    cancellablesLock.lock()
+    defer { cancellablesLock.unlock() }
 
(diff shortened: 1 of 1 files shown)


tgrapperon:
Great @iampatbrown ! Quite a dramatic improvement. Out of curiosity, is most of the time saved by avoiding one additional lock?

tgrapperon:
Great find @iampatbrown. This is a dramatic improvement. If you still have the benchmarks at hand, what would it give without any locking at all, out of curiosity?

stephencelis (approved):
Good call, thanks for PRing! Hopefully keeps things as performant or more with #977 😄

iampatbrown:
Yeah, I think the overhead of `.concatenate` was the main performance hit. cc. @tgrapperon

tgrapperon:
Thanks for the reply @iampatbrown. I deleted my question about it because I thought it wasn't very relevant after spotting that you mentioned this fact in your first comment.

This made me wonder if cancellationCancellables, and by transitivity `Effect`s shouldn't be bound the the `Store` instead of sharing a top-level dictionary. This should allow avoiding the locking if I'm not mistaken, and also protect against setups with multiple TCA stacks, where one would be able to cancel an effect from another store by mistake. This could maybe be a requirement if `Store`s are converted to actors at some point (I don't know for sure about this).

mbrandonw:
> and also protect against setups with multiple TCA stacks, where one would be able to cancel an effect from another store by mistake

As long as you are using private or reducer-scoped ids it is impossible to cancel an effect from another store (or any reducer really).

mbrandonw:
@tgrapperon Yeah that's a great point, thanks for sharing!

PR #977, opened 2022-01-21
Don't consider an effect in-flight till it's kicked off

stephencelis:
A potential fix for #968.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -50,12 +50,15 @@ extension Effect {
         }
       }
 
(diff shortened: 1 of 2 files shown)


mbrandonw (approved):
We chatted about this this morning and looks good to me, but curious about other people's thoughts so we should keep it open for a bit longer.

stephencelis:
@iampatbrown What do you think of this being the semantics around "in-flight"? It seems like chaining multiple `.cancellable`s onto something that hasn't been kicked off shouldn't allow a pending effect to be cancelled.

iampatbrown:
@stephencelis perfect. Semantically it makes more sense as well. I'd probably still consider inlining the cancellation though, but happy with this PR as is :)

PR #978, opened 2022-01-22
Add `EffectID` property wrapper

tgrapperon:
_Edited to reflect the final API_

Following the discussion about store-bound cancellationIDs in #976, taking inspiration from the `scopeIdentifier` defined in #935, and using Swift 5.4 local property wrappers, I'm proposing this `EffectID` property wrapper to streamline cancellation identifiers declarations, and make them automatically inherit some root-store context identifier.

These two things can be separated into two PRs if needed. If the property wrapper can be defined and used in a third-party library, store-binding of effects requires some internal changes as well as the property wrapper.

Defining:
```swift
@EffectID var timerId
```
inside a reducer is all that is needed. The user can directly use `timerId` as a cancellable and root store-specific identifier. The user can also provide some `Hashable` value to decorate the identifier with even more contextual data:
```swift
@EffectID var timerId = state.timerID
```

A few comments:
- Accessing an `@EffectID` variable returns an `EffectID.Value`, which is an opaque hashable value. This is the only way to construct this type for the user.

- In order to be automatically bound to a store, the `@EffectID` wrapper needs to be declared in some reducer. It will work as a regular hashable identifier if declared elsewhere. This can be spotted during debug runs, and the user can be warned. _EDIT: This is more relaxed now. The property wrapper can be declared anywhere one can use a property wrapper. But it still needs to be accessed from a reducer. Runtime warnings are raised if the user fails to do so._

- I've experimented with a generic `wrappedValue`, but the API is less convenient and more confusing for the user. Furthermore, using `Void` as a generic when the user doesn't provide a value raises a warning, and using another fallback type like `String` or `Int` seems awkward for the user.
_EDIT: The canonical way to define the identifier is without a value. Providing a default value adds more information to the returned identifier (on top of the bare identifier itself)._

- Right now, there should be a slight performance impact caused by locking the `currentContextID` on each action received. There are maybe ways to improve this.

- I've updated two cases studies only. I'll update all occurrences if this PR goes further. I'll also add additional tests for the property wrapper itself.


tgrapperon:
I've refactored a few things:

- `@CancellationID` is renamed `@EffectID`
- Locking was removed. The current thread's `threadDictionary` is now used to store the `currentContextID` and the value is then accessed without locking. I suspect that this is faster than locking, but this should be benchmarked. 
- Some API is exposed so the user can assign an arbitrary context to a store and to the `EffectID` property wrapper. By default, the `ObjectIdentifier` of the store is used as a context.

Some unrelated test is failing on Xcode 12.4. All tests are passing on my machine under 13.2.1. I'll try to download Xcode 12.4 to check what's going on.

I'll write some documentation and add a few tests later.

tgrapperon:
I've removed the API to manually define a context for `Store`. It was only a source of bugs we're trying to avoid.

I've also refactored the identifier generated by `@EffectID`. The property wrapper can be used everywhere, but it can only be accessed when some `Reducer` is reducing an action. Otherwise, a runtime warning is raised (only once per `id`) when this identifier is compared. It should even be possible to use it in some function argument, but I didn't experiment with this yet.

After building successfully, Xcode 12.4 from CI is now segfaulting during build. I'm downloading this version right now to check what's going on.

Tests/ComposableArchitectureTests/EffectIdentifierTests.swift
```diff
@@ -27,7 +27,7 @@
       XCTAssertNotEqual(id1, id2_2)
       XCTAssertNotEqual(id2_1, id2_2)
     }
-
+    #if compiler(>=5.5)
```
tgrapperon, line 31: Local property wrappers are available in Swift 5.4, contrary to the commit's message, but Xcode 12.5.1 had an issue with them (see https://forums.swift.org/t/segmentation-fault-11-with-local-property-wrappers/47905).
I bumped the requirement to Swift 5.5 to avoid the problem.

tgrapperon:
A little bike-shedding: should the property wrapper be named:
- `@EffectID` (Apple way, cf. CoreData, CloudKit, etc.; Swift too? (`Identifiable.ID`));
- `@EffectId` (implicit current TCA style, when creating `Hashable` id's);
- `@EffectIdentifier` (Apple uses it too, but maybe only for properties. I can't think of a type ending with `Identifier` right now)?

I personally like the "blocking" effect created by the uppercase "D" at the end. The "tI" is a little difficult to parse and the uppercase "D" seems to help there too.

If `@EffectID` stays, should the examples be updated with uppercase "D" or lowercase "d" for the variable names?
`@EffectID var timerID` or `@EffectID var timerId`?

Should all occurrences of a `Hashable` id struct be replaced by this wrapper if the PR is accepted, or should it be presented as an alternative?

tgrapperon:
Ok, I've renamed `EffectIdentifier` as `EffectID.Value` and cleaned up a few remnants of the exposed context API.

I'm not touching the PR anymore until someone reviews it!

tgrapperon:
Now, this is ready. I'm not touching the PR anymore.
What are your thoughts?

Sources/ComposableArchitecture/Store.swift
```diff
@@ -134,6 +134,7 @@ import Foundation
 /// sends user actions.
 public final class Store<State, Action> {
   private var bufferedActions: [Action] = []
+  private lazy var contextID = ObjectIdentifier(self)
```
tgrapperon, line 137: Is this creating a retain circle? 
Some `UUID` can be used instead if needed.

iampatbrown:
Hey @tgrapperon 

I haven't worked with document-based apps much so take my comments with a grain of salt. Also, I apologise in advance for being brief/blunt. I'm tight on time but wanted to give some feedback and admire the effort you've put into this PR.

- `EffectID` doesn't quite capture what this is. `EffectID` sounds like something to identify effects.
- I think your use of magic literals is pretty cool, but hesitant on them being used like this in the library. I thought they were more for debugging. Would that mean every `EffectID` adds a string literal to the release build? I'm unsure what implications this has but it might be something people want to avoid.
- Could the lazy contextID be a computed property?
- Would scoped stores contextID change and prevent effects from cancelling? 
- It might be possible to incorporate this into cancellable instead of a wrapper.
- I experienced around a 9% hit running the standard benchmarks.

I'm going to revisit this in the future. Keeping the [proto](https://github.com/pointfreeco/swift-composable-architecture/tree/proto) branch in mind, I can imagine some sort of DocumentGroupReducer / Dependecy for handling this.

Even though I'm not really on board with this PR, I don't really have enough experience with the problem it solves. 

Anyway, thanks again @tgrapperon. I always enjoy seeing your work :)

tgrapperon:
Thanks for reviewing the PR @iampatbrown!
Before going into a point per point answer, I would like to mention again that this PR is made of two things:
- A convenient way to define stable identifiers for effect cancellation with the property wrapper, which feels very natural and a little less contrived than defining a whole new `Hashable` type;
- A way to avoid collisions when canceling effects. This can happen most likely with document-based apps (where one may have multiple similar root stores in the same process) if we don't make sure to augment cancellation identifiers with something specific to each root store. This is the part that is responsible for the performance hit. If necessary, this can be dropped in a first time. Without this feature, the property wrapper can have no impact at all on performance.

> EffectID doesn't quite capture what this is. EffectID sounds like something to identify effects.

The first draft used `@CancellationID` instead. I've switched to `EffectID` because it looked more idiomatic when called from reducers. The fact that ongoing effects sharing the same ID are canceled then re-emitted is more an implementation detail for the end-user. But we can go back to `@CancellationID` if needed.

> I think your use of magic literals is pretty cool, but hesitant on them being used like this in the library. I thought they were more for debugging. Would that mean every EffectID adds a string literal to the release build? I'm unsure what implications this has but it might be something people want to avoid.

I've seen (maybe in Swift forums) people saying that it was leaking information into the binary. This can be problematic for the `#file` keyword which is outputting the full file path. As I'm understanding it [from the documentation](https://docs.swift.org/swift-book/ReferenceManual/Expressions.html#ID390), the Swift team seems to see `#fileID` as something acceptable to ship and privacy-wise.

> Could the lazy contextID be a computed property?

Not really. It is inherited from the parent's store and needs to be stored as we don't hold a reference to the parent. The lazy value is only used to seed the root store. It could be assigned in the initializer too, but it would have meant adding a new private initializer to `Store`.

> Would scoped stores contextID change and prevent effects from cancelling?

Changing the `contextID` would indeed lose the effects, but I can't think about a way this can happen. The `contextID` is private and changed to match the parent's one iff the store is scoped (that is when the scoped store is instantiated). Do you see somewhere else it can be changed? 

> It might be possible to incorporate this into cancellable instead of a wrapper.

I think this is indeed possible with some precautions. We can append internally the `currentContextID` to the id provided by the user when the effect is formed. There are probably corner cases that need to be checked. But the property wrapper is the good part!

> I experienced around a 9% hit running the standard benchmarks.

This is huge, but bear in mind that benchmarks are extremely biased. They execute so fast that the smallest difference may double the execution time. I've benchmarked a 40ns penalty when locking the action being received. I didn't check recursive locks that happen when the property wrapper is defined and accessed, but at worse, let's say it takes the same time. We're around an 80ns penalty for each action received per reducer using a property wrapper (if the id is used 80, 40 otherwise). We're still in the ns realm. Of course, within an app with 100 reducers using each one 100 ids, we would top around 1ms if locking perf is linear. This is not at all negligible, but I guess such an app would have other issues!
As I said earlier, the locking/safety part can be postponed in a future PR if better solutions are required.

The proto branch would probably allow a more direct approach, as the `contextID` could be stored in the `Reducer` itself, as an internal dependency that is passed along pullbacks. With such reducers, I even think that cancellation identifiers could be also be defined as stored properties. I haven't checked the branch in quite some time, but it seems there's a lot of room to find a solution that avoids locking there.

iampatbrown:
> It is inherited from the parent's store and needs to be stored as we don't hold a reference to the parent. 

I overlooked this. I thought there was a different context per scope as well. Could you bypass the call to `withContext` on the scoped stores then?

> bear in mind that benchmarks are extremely biased. 

Agreed. I was more just adding it as additional info.

Thanks for the response. I'll get back to you on the other points a bit later :)

tgrapperon:
> Could you bypass the call to withContext on the scoped stores then?

The `contextID` is something that identifies the whole TCA tree originating from a root store. The objective is to avoid two distinct stores (in the sense belong in trees originating from different root store instances) to cross-cancel effects. There is not really a need for something more precise than identifying the trees, hence, their root store.

With one context per store, the *Store* instance that created the effect will be responsible for canceling it, which is not something already required by the architecture (nor desirable I guess).

The `withContext` sets the global `currentContextID` for the time the reducer executes, and with a lock, in case a different store of the same type (more precisely using the same reducers), but in another thread, would reduce some action at the same time. 

While thinking more about it, it seems that the lock is not even required with pure document-based apps, as everyone runs on the main thread. In other words, the lock only starts to be required if some unchecked root store is instantiated (if I recall correctly, this is the designated way to instantiate non-main thread stores, am I right?). It can potentially offer an optimization route in "all main-thread" cases. I don't know if "using the lock or not" can be triggered dynamically, or if a new API should be used to opt-in (implicitly of course)

The context switches are still required for document-based apps, but without locking, `withContext` should be more or less transparent.

tgrapperon:
I've slightly modified `contextID` access. When on the main thread, a value specific to this thread is touched without precaution. Background stores are using locking on a value they all share.

To sum up, in the current state:
- A negligible performance impact when running on the main thread, including document-based apps.
- A slight performance hit. caused by locking for background stores. This shouldn't be required if only one of these exists. so maybe an API similar to the one used for `unchecked` Stores could be implemented.

This makes me think that, because `EffectID.Value` is capable of storing if it was issued on the main thread or not, one could save a few ns using the same technique of having separate pools to bypass `cancellationCancellables` locking, and requiring the `cancelID` arguments to be an `EffectID.Value` s obtained with the property wrapper (in this case, the `@CancellationID` naming would indeed be more appropriate). Methods with a free cancel identifier could be either deprecated[^1] or fall back to the background cancellable pool. In this last case, using the property wrapper would be safer and technically more performant than the actual implementation.

[^1]: Interestingly, a `callAsFunction()` method that returns `self` can be implemented on `EffectID.Value` temporarily, so the conversion from private `Hashable` identifiers would simply require to replace `struct TimerID: Hashable` occurrences by `@EffectID var TimerID` and nothing more! This is not a serious idea, but it can be fun to see what's possible with modern versions of Swift!

What do you think about it?

tgrapperon:
After playing a little with it, it's not absolutely obvious. As I'm getting it, the `Defer` closure from `Effect.cancellable(id:cancelInFlight:)` can execute on any thread, even when originating from a main-thread store. Am I right? If the user writes
```swift
return someEffect
  .cancellable(id: 1)
  .subscribe(on: DispatchQueue.global())
  .receive(on: DispatchQueue.main)
  .eraseToEffect()
```
To my understanding, the `Defer` block from `.cancellable` will execute on `DispatchQueue.global()`. Is it the case?

tgrapperon:
It occurred to me that the `currentContextID` can be stored using the new [`TaskLocal`](https://developer.apple.com/documentation/swift/tasklocal) property wrapper.
As I'm getting it, its access is thread-safe by design. 
When this property wrapper is not available, the previous scheme (with locking on background threads) is used.
~~I haven't checked the performance impact yet, but I guess some thread-agnostic solution is desirable in the long run.~~

EDIT: `TaskLocal` is taking a little more time than accessing the `Thread.threadDictionary` twice (~245ns vs 230ns), and almost 3-4x more than locking (~245ns vs 70ns), so I reverted the commit.

I've also slightly improved privacy concerns with `#fileID`. In some cases, the filename could have leaked in logs, if the user dumped something containing an `EffectID.Value`. With Release builds, the `EffectID` is using the filename's hash instead of the filename. This is slightly bothering because tested code is now different from released code, and I'm not sure this is not solving non-existent problems.

I'm not sure about the validity of these last two commits, so please tell me if you think they should be reversed.

tgrapperon:
The `contextID` can probably be set in a top-level reducer, meaning that this feature could be implemented as a third-party library instead. 
Thanks for the review @iampatbrown.

mbrandonw:
Hi @tgrapperon, sorry for not getting back sooner, but I did think that was a very interesting exploration you did.

In general I was a little worried about incurring the cost of locking or thread/task storage changes for each `send`, though your benchmarks do show that at the very least the lock isn't too expensive. I personally would have felt more comfortable with the task storage option because it at least plays nicely with async/await (which we have big plans for soon) and is just a more modern tool for this kind of thing, even though it was the slowest.

Also, as you say in your last comment, this technically can be achieved as some kind of reducer operator that is applied for each store created. So you don't get the functionality "for free", but at least it is possible.

Thanks again for experimenting with this. It's very interesting stuff!

tgrapperon:
Thanks for the feedback @mbrandonw. Please note that in its last version, the PR wasn't locking at all on main thread stores, even with document-based apps! (main thread stores had a specific `contextID` value they directly accessed without locking). So for the vast majority of TCA apps, the additional cost per `send` was equivalent to checking `Thread.isMainThread` and setting a value.
The `TaskLocal` experiment was interesting. I guess it's backed by a dictionary under the hood (that's what the last line of the documentation implies), so it explains why the performance is comparable to `Thread.threadDictionary`. I guess it's good to know for future explorations.

iampatbrown:
@tgrapperon No worries. You did some great ground work here.

tgrapperon:
Hey @mbrandonw & @stephencelis!
As I was working on extracting `@EffectID` as a third-party library, it occurred to me that the story about document-based apps could also apply to any collection of features: Let say you have a ListRow feature, and each row host a timer (or any cancellable effect). In order for each row to work independently, you have to "leak" some identifier at the row level to define distinct effect identifiers. In other words, each row is aware of a higher structure it must acknowledge when forming effect identifiers. This is not optimal, as the user may also have to store an identifier in the `State/Environment`.

Extracting this library as a third party made me refactor it into some higher-order reducer that I named `.namespace<ID: Hashable>(_ id:ID)`. I also created overloads to derive the `id` from a function/keyPath from `State` or `Environment`. This higher-order reducer can be created from any reducer, whatever its position, and applies downstream. If this feature would be brought into TCA, `forEach` reducers could automatically "namespace" the pulled-back reducer with the `IdentifiedCollection.ID` or  `Dictionary.Key`. As a result, `Element`'s features could use nonparameterized `@EffectID`to to define their effect identifier and truly work in isolation.
This use case is much more common than document-based apps. If you've already added an additional `id` property to the `Hashable` structs you're using as identifiers, you may have encountered situations where namespacing/`@EffectID` would have automatically handled this instead. I personally experienced this configuration many times.

You can check a preliminary implementation at [tgrapperon/swift-composable-architecture_effect-identifier](https://github.com/tgrapperon/swift-composable-architecture_effect-identifier). There is an example document-based app that shows how to "namespace" the root reducer (without it, the app doesn't work correctly, as foreseen), but it would work the same if each document would live in the row of a list (and even without changing a line the feature's code).
(in the example, the `State` carries an identifier, but only for presentation purposes).

As the performance impact for main-thread stores is almost null, and since this new use-case could benefit many users,  is there still life in this PR with this new reducer-based approach, and automatic application to `forEach` pullbacks? 
I can reopen and refactor the PR accordingly if needed.

tgrapperon:
A final message for this PR. I opted for the third-party route. I added a few pullback overloads that I'm almost sure couldn't make it to TCA. The library is now open-source at https://github.com/tgrapperon/composable-effect-identifier, with some basic documentation and a very cool example where I'm building 4 apps (1 shoebox, 1 document-based, macOS&iOS) that are handling many timers simultaneously, around 1 timer feature that is absolutely unaware that other timers may exist.

Hopefully, opportunities to tackle the problem in a more elegant way in Core-TCA may present again on the `proto` branch or when converting TCA to the new concurrency paradigm.

tgrapperon:
Thanks to everyone for the feedback!

tgrapperon:
As a real final note, and since there was some interest in the performance of various thread-safe strategies, I've benchmarked a few scenarios, and the results are quite interesting:
- Locking is still the best all-purpose solution, and the impact is very small.
- `@TaskLocal` performs very well on reads, but not on writes.
- `Actors` are not good enough for this purpose, at least with the implementation I've written.

```
== Thread Safety Experiments ==================================================================
                                    Value  Difference Variation  Performance  Error  Iterations
-- Thread Safe Value Access - Get -------------------------------------------------------------
Direct Access                   (*)  51 ns                                   ±0.24 %    1000000
Main Thread Check                    55 ns      +5 ns    +9.00 %       x0.92 ±1.07 %    1000000
Thread.current.threadDictionary     442 ns    +391 ns  +773.77 %       x0.11 ±0.09 %    1000000
NSRecursiveLock                      81 ns     +30 ns   +59.83 %       x0.63 ±0.17 %    1000000
@TaskLocal                           62 ns     +12 ns   +23.45 %       x0.81 ±0.18 %    1000000
Actor                               583 ns    +532 ns +1052.44 %       x0.09 ±0.43 %    1000000
-- Thread Safe Value Access - Set -------------------------------------------------------------
Direct Access                   (*)  55 ns                                   ±0.19 %    1000000
Main Thread Check                    62 ns      +7 ns   +11.85 %       x0.89 ±0.22 %    1000000
Thread.current.threadDictionary     456 ns    +401 ns  +726.31 %       x0.12 ±0.08 %    1000000
NSRecursiveLock                      88 ns     +33 ns   +59.75 %       x0.63 ±0.15 %    1000000
@TaskLocal                          263 ns    +208 ns  +376.72 %       x0.21 ±0.11 %    1000000
Actor                               554 ns    +499 ns  +904.69 %       x0.10 ±0.19 %    1000000
===============================================================================================
```

You can find the code at https://github.com/tgrapperon/thread-safety-experiments. I guess it may help in the future.
Benchmarks were realized with my own `swift-measure` library, but there's also a `swift-benchmark` branch that uses Google's `swift-benchmarks` instead. Results are different in absolute, but their interpretation is similar:
```
name                                                           time       std        iterations
-----------------------------------------------------------------------------------------------
Thread Safe Value Access - Get.Direct Access                    25.000 ns ± 319.11 %    1000000
Thread Safe Value Access - Get.Main Thread Check                32.000 ns ± 374.82 %    1000000
Thread Safe Value Access - Get.Thread.current.threadDictionary 383.000 ns ±  99.03 %    1000000
Thread Safe Value Access - Get.NSRecursiveLock                  61.000 ns ± 186.22 %    1000000
Thread Safe Value Access - Get.@TaskLocal                       42.000 ns ± 283.91 %    1000000
Thread Safe Value Access - Get.Actor                           217.000 ns ± 691.59 %    1000000

Thread Safe Value Access - Set.Direct Access                    26.000 ns ± 436.62 %    1000000
Thread Safe Value Access - Set.Main Thread Check                31.000 ns ± 794.86 %    1000000
Thread Safe Value Access - Set.Thread.current.threadDictionary 412.000 ns ± 233.52 %    1000000
Thread Safe Value Access - Set.NSRecursiveLock                  64.000 ns ± 395.22 %    1000000
Thread Safe Value Access - Set.@TaskLocal                      214.000 ns ±  83.85 %    1000000
Thread Safe Value Access - Set.Actor                           195.000 ns ± 410.73 %    1000000
```

I'll double-check if `swift-measure` is correct. `swift-benchmark` standard deviation also often goes in the 500%+, which is quite suspicious. I'll investigate.

Feel free to PR if you have other ideas to check!

PR #984, opened 2022-01-25
Fix UIKitCaseStudies' List example

bjford:
Hi there!

Noticed that the UIKitCaseStudies' List example had a small problem with `countLabel` of `CounterViewController`.
It wasn't updating accordingly when incrementing/decrementing.
This PR fixes it.

### DIFF
--- a/Examples/CaseStudies/UIKitCaseStudies/CounterViewController.swift
+++ b/Examples/CaseStudies/UIKitCaseStudies/CounterViewController.swift
@@ -28,7 +28,7 @@ let counterReducer = Reducer<CounterState, CounterAction, CounterEnvironment> { 
 
 final class CounterViewController: UIViewController {
   let viewStore: ViewStore<CounterState, CounterAction>
-  var cancellables: Set<AnyCancellable> = []
+  private var cancellables: Set<AnyCancellable> = []
 
   init(store: Store<CounterState, CounterAction>) {
     self.viewStore = ViewStore(store)
--- a/Examples/CaseStudies/UIKitCaseStudies/ListsOfState.swift
+++ b/Examples/CaseStudies/UIKitCaseStudies/ListsOfState.swift
@@ -63,11 +63,12 @@ final class CountersTableViewController: UITableViewController {
   }
 
   override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
-    let counter = self.viewStore.counters[indexPath.row]
+    let indexPathRow = indexPath.row
+    let counter = self.viewStore.counters[indexPathRow]
     self.navigationController?.pushViewController(
       CounterViewController(
         store: self.store.scope(
-          state: { _ in counter },
+          state: \.counters[indexPathRow],
           action: { .counter(id: counter.id, action: $0) }
         )
       ),


stephencelis (approved):
Ah thanks! That case study is a bit out-of-date 😅

bjford:
> Ah thanks! That case study is a bit out-of-date 😅

Np 👍 :)
While fixing it, I was considering to add some unit-tests for the UIKitCaseStudies. 
What do you think? Would it be useful?

stephencelis:
> While fixing it, I was considering to add some unit-tests for the UIKitCaseStudies. What do you think? Would it be useful?

@bjford One issue with the current studies is they rely on a deprecated version of `forEach`, and we haven't investigated what it would mean to upgrade to the latest one.

We don't want to send you on any wild goose chases, though! 😅

bjford:
> > While fixing it, I was considering to add some unit-tests for the UIKitCaseStudies. What do you think? Would it be useful?
> 
> @bjford One issue with the current studies is they rely on a deprecated version of `forEach`, and we haven't investigated what it would mean to upgrade to the latest one.
> 
> We don't want to send you on any wild goose chases, though! 😅

Hi again @stephencelis!
I've been looking into what you said, but I couldn't figure out whether the `UIKitCaseStudies` uses a deprecated version of the `.forEach` operator. 

I can see its deprecations:
image

However when I jump to the definition of `.forEach` then I'm redirected here https://github.com/pointfreeco/swift-composable-architecture/blob/dd084dc147dadede0ad25f829a43ceafed2b9abc/Sources/ComposableArchitecture/Reducer.swift#L754

stephencelis:
@bjford That might just be Xcode getting confused. That's the non-deprecated version that uses `IdentifiedArray` instead of `Array`, so it should be good to use!

bjford:
> @bjford That might just be Xcode getting confused. That's the non-deprecated version that uses `IdentifiedArray` instead of `Array`, so it should be good to use!

Ah okay, so just to clarify `UIKitCaseStudies` is using the proper version of `.forEach`.

stephencelis:
@bjford Oh ha! I completely forgot that I already fixed that 😅 https://github.com/pointfreeco/swift-composable-architecture/pull/924

bjford:
> @bjford Oh ha! I completely forgot that I already fixed that 😅 #924

@stephencelis awesome! 💪

tylerjames:
Hi,  still getting to know TCA and actually just encountered this problem (because I hadn't updated the library).

I'm looking at the solution in your commit and I can't quite figure out *why* it didn't work before.

Previously it was accessing the counter for the corresponding row like this:
`let counter = self.viewStore.counters[indexPath.row]`

and in the fixed version it's accessing it like this:
```swift
self.navigationController?.pushViewController(
      CounterViewController(
        store: self.store.scope(
          state: \.counters[indexPathRow],
          action: { .counter(id: counter.id, action: $0) }
        )
```

Why does accessing the counter state via keypath vs accessing it using `self.viewStore.counters` make a difference?

bjford:
> Hi, still getting to know TCA and actually just encountered this problem (because I hadn't updated the library).
> 
> I'm looking at the solution in your commit and I can't quite figure out _why_ it didn't work before.
> 
> Previously it was accessing the counter for the corresponding row like this: `let counter = self.viewStore.counters[indexPath.row]`
> 
> and in the fixed version it's accessing it like this:
> 
> ```swift
> self.navigationController?.pushViewController(
>       CounterViewController(
>         store: self.store.scope(
>           state: \.counters[indexPathRow],
>           action: { .counter(id: counter.id, action: $0) }
>         )
> ```
> 
> Why does accessing the counter state via keypath vs accessing it using `self.viewStore.counters` make a difference?

Hi @tylerjames! 

As far as I understand it makes a difference because using key path you'll get a func in the form of `(State) -> LocalState`: that is what the first argument of `.scope` expects. [check here](https://github.com/pointfreeco/swift-composable-architecture/blob/12ae834de95e9b78ad74634456c70e9b199e8f3e/Sources/ComposableArchitecture/Store.swift#L326)
The following will also work `state: { $0.counters[indexPathRow] }`: no key path here.
These two solution are consuming the `State` that is fed in the closure.

The problem is that using `{ _ in counter }` won't make it because it is ignoring the `State` that is fed in the closure. You must use that `State`.

Hope that this makes sense and answers your question.

tylerjames:
Yes, that makes perfect sense. Thank you.

PR #987, opened 2022-01-29
Remove `Sendable` warning in `ViewStore.suspend(while:)`

tgrapperon:
Swift 5.6 (from Xcode 13.3 beta 1) complains that the `Box` value touched in the `Sendable` handler of `.withTaskCancellationHandler` is not `Sendable`. This kind of warnings will show by default with Swift 5.6 and [will become errors with Swift 6](https://forums.swift.org/t/concurrency-in-swift-5-and-6/49337#concurrency-in-swift-5-and-6-2).

This PR marks `Box` as `@unchecked Sendable`. The `@unchecked` is required by `Box.Value==AnyCancellable` which is not `Sendable` by itself. This removes the warning at build time, but it fixes nothing on the concurrency aspect. Conditional conformance of `Box` to `Sendable` would require us to conform `AnyCancellable` to `Sendable` ourselves, which is most likely a bad idea.

I've also marked `Box` as final, [as recommended](https://developer.apple.com/swift/blog/?id=27). (I've furthermore wrongfully mentioned it as `ViewStore.Box` in the commit message. This is a top-level class)

I don't know if there are better/safer alternatives without migrating more things to "async/await".

EDIT: CI failed in some unrelated test.

### DIFF
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -477,7 +477,7 @@ private struct HashableWrapper<Value>: Hashable {
     }
   }
 
-  private class Box<Value> {
+  private class Box<Value>: @unchecked Sendable {
     var wrappedValue: Value
 
     init(wrappedValue: Value) {
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -477,7 +477,7 @@ private struct HashableWrapper<Value>: Hashable {
     }
   }
 
-  private class Box<Value>: @unchecked Sendable {
+  private final class Box<Value>: @unchecked Sendable {
     var wrappedValue: Value
 
     init(wrappedValue: Value) {


tgrapperon:
https://github.com/pointfreeco/swift-composable-architecture/blob/67ce7e5839ab7331f3a8dcb5d31783dafd9e8107/Sources/ComposableArchitecture/ViewStore.swift#L470

Is there a configuration where the publisher can cancel without emitting once, and thus where `resume` is never called?

stephencelis:
@tgrapperon Confused a bit about the failure. It's saying a test failed but not printing any info about it 😕

Going to kick off one more CI build but not sure I'll have a chance to investigate further right now.

tgrapperon:
CI shows: [L1257](https://github.com/pointfreeco/swift-composable-architecture/runs/4993018253?check_suite_focus=true#step:4:1257) 
```
Restarting after unexpected exit, crash, or test timeout in EffectCancellationTests.testCancellationAfterDelay_WithTestScheduler(); summary will include totals from previous launches.
```
Really odd. This test expects nothing, so it looks like some external event.

tgrapperon:
Starting from Xcode 13.3b3, we can also use the [`@preconcurrency`](https://github.com/apple/swift-evolution/blob/main/proposals/0337-support-incremental-migration-to-concurrency-checking.md#preconcurrency-attribute-on-nominal-declarations) attribute to fix the issue. Is it a better solution than making `Box` `@unchecked Sendable`?

jshier:
No, that's not the intent of the attribute. If you've made a type conform to `Sendable`'s informal requirements but can't make it automatically conform, `@unchecked Sendable` is the expected solution. `@preconcurrency` is useful for types you can't evaluate at all, like older Obj-C code. I wouldn't expect it to be that useful for Swift code except as a mitigation as you update a codebase to support concurrency incrementally.

tgrapperon:
@jshier Ok, thanks for the clarification!

stephencelis:
@tgrapperon Now that 13.3 RC's out, I no longer see a warning here. Do you think this PR is no longer necessary (for the moment)?

tgrapperon:
I can confirm, the warning is gone.
This PR serves no purpose then.

PR #997, opened 2022-02-06
Fix memory leak caused by `Effect.cancellable`

### LINKED ISSUE #993: Memory leak in long-living effects example
**Describe the bug**
There're memory leaks detected when I have TCA store subscribes to long-running effects which have `cancallable(id:)` attached to. 

**To Reproduce**
Take "Long-living effects" as an example. Go to the screen then tap on "Navigate to another screen", then tap back. The expectation is the long-running effect gets canceled and no memory leak.
Screen Shot 2022-02-05 at 13 21 59

Below is a screenshot of the leak debugger.
Screen Shot 2022-02-05 at 13 21 51

Zip up a project that reproduces the behavior and attach it by dragging it here.
Check out the example project from https://github.com/pointfreeco/swift-composable-architecture. 
```swift
let longLivingEffectsReducer = Reducer<
  LongLivingEffectsState, LongLivingEffectsAction, LongLivingEffectsEnvironment
> { state, action, environment in

  struct UserDidTakeScreenshotNotificationId: Hashable {}

  switch action {

  ...

  case .onAppear:
    // When the view appears, start the effect that emits when screenshots are taken.
    return environment.userDidTakeScreenshot
      .map { LongLivingEffectsAction.userDidTakeScreenshotNotification }
      .cancellable(id: UserDidTakeScreenshotNotificationId()) // Without this line, the leak goes away. but cancellation is required to clean subscription up as we know.

  case .onDisappear:
    // When view disappears, stop the effect.
    return .cancel(id: UserDidTakeScreenshotNotificationId())
  }
}

```

**Expected behavior**
`cancellable(id:) should not cause memory leaks.

**Environment**
  - swift-composable-architecture version 0.33.1 (Latest)
  - Xcode 13.2.1
  - Swift 5.5.2
  - OS: iOS 15.2

**Additional context**
Add any more context about the problem here.

iampatbrown:
This PR should fix #993 which was introduced by #949.

It looks like `PrefixUntilOutput` might keep the output publisher alive until receiving a completion. This is something I overlooked.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -47,6 +47,7 @@ extension Effect {
       cancellationCancellable = AnyCancellable {
         cancellablesLock.sync {
           cancellationSubject.send(())
+          cancellationSubject.send(completion: .finished)
           cancellationCancellables[id]?.remove(cancellationCancellable)
           if cancellationCancellables[id]?.isEmpty == .some(true) {
             cancellationCancellables[id] = nil


iampatbrown:
@BeerPoldet and @tgrapperon do you mind confirming this fixes the leak on your end?

tgrapperon:
@iampatbrown I can confirm, no leaks on my side. 
I'm not the one who has dealt with them the most though. @BeerPoldet is more knowledgeable about the issue and their response should prevail.

iampatbrown:
@tgrapperon Thanks for checking mate. Will wait to see how @BeerPoldet goes.

BeerPoldet:
@iampatbrown  @tgrapperon Thank you both of you for your fast responses and actions.

I can confirm that the leaks have gone after several tries. I hope that this fix can go soon. 
Screen Shot 2022-02-07 at 08 33 26

iampatbrown:
Great news @BeerPoldet :)

> I hope that this fix can go soon.

Once @mbrandonw or @stephencelis have had a chance to take a look it shouldn't be too long. 

Thanks again for picking this up. Really awesome effort by you and @tgrapperon debugging this!

tgrapperon:
Congrats to you for finding such a nice solution, @iampatbrown!

iampatbrown:
@tgrapperon I initially thought we might have to create a custom `Publisher` :P glad it was a straight forward fix. More of an oversight in my initial implementation :) 

Actually, @BeerPoldet I'd love to know how you originally picked up on this? I'm not too familiar with debugging the memory graph.

stephencelis (approved):
Thanks for tracking this down!

PR #999, opened 2022-02-07
Fix edit while filtered bug in todos example app

omaralbeik:
## Changes
Fixed a bug in Todos example app where moving items returns the wrong todos indices if the completed filter is enabled

> The issue is happening because the `source` and `destination` are captured from the `filteredTodos` array while the move operation is happening in the original `todos` array

[before]: https://user-images.githubusercontent.com/8127757/152877651-fbfecfc6-0c20-4cbd-b31c-341b96f07be9.gif
[after]: https://user-images.githubusercontent.com/8127757/152877638-78e43d4a-4e8b-4bc6-950d-d67e30337554.gif

### Before/After
| Before | After |
|-|-|
| ![Before][before] | ![After][after] |

### DIFF
--- a/Examples/Todos/Todos/Todos.swift
+++ b/Examples/Todos/Todos/Todos.swift
@@ -66,7 +66,13 @@ let appReducer = Reducer<AppState, AppAction, AppEnvironment>.combine(
       return .none
 
     case let .move(source, destination):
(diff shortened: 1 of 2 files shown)


stephencelis (approved):
Good find! 😅

Thanks for the fix _and_ test!

Issue #1000, opened 2022-02-11
How to handle task cancellation with ViewStore.send(_:while:)


lukeredpath:
I'm experimenting with integrating the `ViewStore` async support with the SwiftUI `.task` modifier and I can't seem to figure out how cancellation is supposed to work.

Looking at the source of `ViewStore.suspend` it looks like the iOS 14 implementation does handle cancellation but its unclear whether the iOS 15 version, that uses the `AsyncPublisher` API in Combine does handle cancellation.

The `.task` modifier in SwiftUI starts a task on appear and is supposed to cancel the task on disappear, or if the provided `id` parameter changes. In my scenario, I have a `List` and each view in the list has its own row store and view store. I'm trying to start a task when the row appears and I've got this working. As a simplified example:

```swift
struct RowState: Equatable, Identifiable {
  let id: UUID
  var isRunningTask: Bool = false
}

enum RowAction: Equatable {
  case runTask
  case taskFinished
}

let rowReducer = Reducer<...> { state, action, environment in
  switch action {
  case .runTask:
    // simulate some async work
    state.isRunningTask = true
    return Effect(value: .taskFinished)
      .delay(for: 2.0, on: DispatchQueue.main)
      .eraseToEffect()

  case .taskFinished:
    state.isRunningTask = false
    return .none
  }
}
```

In the view:

```swift
struct RowView: View {
  let store: Store<RowState, RowAction>

  var body: some View {
    Text("Is running task? \(viewStore.isRunningTask ? "Yes" : "No")")
        .task(id: viewStore.id) {
            await viewStore.send(.runTask, while: \.isRunningTask)
        }
  }
}
```

In the case where task is an expensive operation, like loading some asset from the network, how would we handle cancellation of whatever effect is triggered by `.runTask` when SwiftUI tries to cancel the task? 

Its hard to tell exactly what is happening but I'm guessing that even if the task is being cancelled, all it is doing is cancelling the subscription to the change in the state of the underlying `ViewStore` - it does not actually cancel the effect returned from the reducer in the `runTask` action.

lukeredpath:
I've opened the PR above after experimenting with solutions to this and this seems to work nicely, at the expense of losing the simplified iOS 15 implementation that uses `AsyncPublisher` under the hood - I'm open to suggestions for making it work with that too.

iampatbrown:
[...]

PR #1001, opened 2022-02-11
Add support for task cancellation when using ViewStore.send(_:while).

lukeredpath:
This adds the ability for sending a cancellation action to the view store when calling `ViewStore.send(_:while)` inside a SwiftUI `.task` modifier, in turn making it possible to start and cancel long-running effects..

To implement this, I had to remove the iOS 15 specific implementation that uses the `AsyncPublisher` API as I couldn't see any obvious way of hooking into the cancellation (or if it even supports cancellation).

I originally considered just exposing the `cancellationHandler` parameter on the `send` methods but I believe that the most common use case when calling `.send` inside a `.task` modifier will be to send a cancellation action to the store in order to cancel a long-running effect.

If users need to perform some custom cancellation handling, they have the option of calling `ViewStore.send()` followed by `ViewStore.suspend()` inside the `.task` modifier and passing in a custom cancellation handler that way, but I've optimised the high-level APIs for what I think most people will need.

Solves #1000

### DIFF
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -524,7 +524,7 @@ private struct HashableWrapper<Value>: Hashable {
     ///   - cancellationHandler: A closure that will be called if the suspend task is cancelled.
     public func suspend(
       while predicate: @escaping (State) -> Bool,
-      onCancel cancellationHandler: (() -> Void)?
+      onCancel cancellationHandler: (() -> Void)? = nil
     ) async {
       let cancellable = Box<AnyCancellable?>(wrappedValue: nil)
       try? await withTaskCancellationHandler(
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -406,6 +406,68 @@ private struct HashableWrapper<Value>: Hashable {
     /// to `.refreshable` that the work has finished which will cause the loading indicator to
     /// disappear.
     ///
+    /// This method also takes an optional cancellation action that will be sent to the store when the
+    /// suspend task is cancelled - this can be useful when calling this method from inside a SwiftUI
+    /// `.task` modifier, as the task will be cancelled automatically when the view moves off screen.
+    /// If you're using this to start a long running effect when the view appears and waiting for it to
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/ViewStore.swift
```diff
@@ -433,47 +501,55 @@ private struct HashableWrapper<Value>: Hashable {
...
+    ///   - cancellationHandler: A closure that will be called if the suspend task is cancelled.
+    public func suspend(
+      while predicate: @escaping (State) -> Bool,
+      onCancel cancellationHandler: (() -> Void)? = nil
+    ) async {
+      let cancellable = Box<AnyCancellable?>(wrappedValue: nil)
+      try? await withTaskCancellationHandler(
+        handler: {
+          cancellable.wrappedValue?.cancel()
+          cancellationHandler?()
```
lukeredpath, line 533: Swift is warning that `cancellationHandler` isn't `@Sendable` - the same warning is reported for `cancellable` which is fixed by #987 - one possible solution here would be - following the merging of #987, to move the cancellation handler into `Box` so only that is being sent across and calling it that way.

lukeredpath: With the above fix in mind, I've refactored this so the Box holds on to the cancellation handler: https://github.com/pointfreeco/swift-composable-architecture/pull/1001/commits/8b16883df8b90350ce28d1b49e53938cfc75cc95

stephencelis:
Hey @lukeredpath! Thanks for submitting this PR, and sorry for not responding sooner. This is definitely a problem worth solving, and we've been aware of the issue and have been experimenting with some alternative solutions that we'd love to share soon after we iron out more of the details.

stephencelis:
Hey @lukeredpath! We believe this functionality is more-or-less covered in our [Concurrency Beta](https://github.com/pointfreeco/swift-composable-architecture/discussions/1186) branch, and were curious if that suits your needs. The main difference in API is we do not expose an `onCancel` handler directly. Instead we believe you can lean on `withTaskCancellationHandler` to be notified of such events.

lukeredpath:
@stephencelis I've only given it an initial glance but yes, I think the concurrency beta handles this nicely.

PR #1009, opened 2022-02-24
`Effect.cancel(ids:)` prefer sequence of ids over variadic list.

iampatbrown:
This PR disfavors the variadic cancellation to prevent a sequence of ids from being treated as a single cancellation id eg: 

```swift
case let .delete(ids):
  for id in ids {
    state.rows.remove(id: id)
  }
  return .cancel(ids: ids) // currently treated as AnyHashable(ids) instead of [AnyHashable]
```

It also changes the array parameter to a sequence to prevent the same issue for occurring with sub sequences.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -96,7 +96,7 @@ extension Effect {
 
   /// An effect that will cancel multiple currently in-flight effects with the given identifiers.
   ///
-  /// - Parameter ids: An sequence of effect identifiers.
+  /// - Parameter ids: A sequence of effect identifiers.
   /// - Returns: A new effect that will cancel any currently in-flight effects with the given
   ///   identifiers.
   public static func cancel<S: Sequence>(ids: S) -> Effect where S.Element == AnyHashable {
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -89,6 +89,7 @@ extension Effect {
   /// - Parameter ids: A variadic list of effect identifiers.
   /// - Returns: A new effect that will cancel any currently in-flight effects with the given
   ///   identifiers.
+  @_disfavoredOverload
   public static func cancel(ids: AnyHashable...) -> Effect {
     .cancel(ids: ids)
   }
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -96,10 +96,10 @@ extension Effect {
 
   /// An effect that will cancel multiple currently in-flight effects with the given identifiers.
   ///
-  /// - Parameter ids: An array of effect identifiers.
+  /// - Parameter ids: An sequence of effect identifiers.
   /// - Returns: A new effect that will cancel any currently in-flight effects with the given
   ///   identifiers.
-  public static func cancel(ids: [AnyHashable]) -> Effect {
+  public static func cancel<S: Sequence>(ids: S) -> Effect where S.Element == AnyHashable {
     .merge(ids.map(Effect.cancel(id:)))
   }
 }
--- a/Tests/ComposableArchitectureTests/EffectCancellationTests.swift
+++ b/Tests/ComposableArchitectureTests/EffectCancellationTests.swift
@@ -324,7 +324,7 @@ final class EffectCancellationTests: XCTestCase {
 
     Effect<Int, Never>.merge(
       .cancel(ids: CancellationId(id: 1), CancellationId(id: 2)),
-      .cancel(ids: [CancellationId(id: 4), CancellationId(id: 5)])
+      .cancel(ids: [CancellationId(id: 4), CancellationId(id: 5)][...])
     )
       .sink { _ in }
       .store(in: &self.cancellables)
--- a/Tests/ComposableArchitectureTests/EffectCancellationTests.swift
+++ b/Tests/ComposableArchitectureTests/EffectCancellationTests.swift
@@ -302,4 +302,34 @@ final class EffectCancellationTests: XCTestCase {
 
     XCTAssertEqual(output, [1, 2])
   }
+  
+  func testMultipleCancellations() {
+    let scheduler = DispatchQueue.test
+    var values: [Int] = []
+
+    struct CancellationId: Hashable {
+      var id: Int
+    }
+
+    let effects = [1, 2, 3, 4, 5].map { id in
+      Just(id)
+        .delay(for: 1, scheduler: scheduler)
+        .eraseToEffect()
+        .cancellable(id: CancellationId(id: id))
+    }
+
+    Effect<Int, Never>.merge(effects)
+      .sink { values.append($0) }
+      .store(in: &self.cancellables)
+
+    Effect<Int, Never>.merge(
+      .cancel(ids: CancellationId(id: 1), CancellationId(id: 2)),
+      .cancel(ids: [CancellationId(id: 4), CancellationId(id: 5)])
+    )
+      .sink { _ in }
+      .store(in: &self.cancellables)
+
+    scheduler.advance(by: 1)
+    XCTAssertEqual(values, [3])
+  }
 }


tgrapperon:
Good catch! This was a sneaky one!
And the `Sequence` conversion seems to be a win/win: more flexible and safer.
The `S.Element == AnyHashable` requires elements to be casted explicitely to `AnyHashable`, isn't it? That is, `String`s are still taking the variadic overload, and are not seen as a `Sequence` of `Characters`, themselves promoted as `AnyHashable`?

iampatbrown:
> The S.Element == AnyHashable requires elements to be casted explicitely to AnyHashable, isn't it? That is, Strings are still taking the variadic overload, and are not seen as a Sequence of Characters, themselves promoted as AnyHashable?

Ohhh good thinking... I'd need to confirm. Maybe that's why it was originally an array too.

tgrapperon:
That's my original concern: If it infers `["abc", 123]` as an array of `AnyHashable` without having to cast elements, I don't see immediately why it won't see "abc" as a `Sequence` of `AnyHashable` too. Maybe `ExpressibleByLiterals` could interfere too, so it could be worth checking if this is working as expected with variables and with literals.

iampatbrown:
Good call. I'm going to convert to draft for now and tests this out a bit later :)

Thanks @tgrapperon!

iampatbrown:
Whipped up a quick playground to compare the current implementation with this PR

**before:**
```swift
func anyHashableCount(_ hashables: [AnyHashable]) -> Int {
  hashables.count
}

func anyHashableCount(_ hashables: AnyHashable...) -> Int {
  hashables.count
}
```

**after:**
```swift
@_disfavoredOverload
func anyHashableCount(_ hashables: AnyHashable...) -> Int {
  hashables.count
}

func anyHashableCount<S: Sequence>(_ hashables: S) -> Int where S.Element == AnyHashable {
  Array(hashables).count
}
```

I tested `Int`, `String`, `UUID` and a `Hashable` struct using a variety of arguments. Here is an example for `Int`:

```swift
anyHashableCount(123) // One
anyHashableCount(1, 2, 3) // Three
anyHashableCount([1, 2, 3]) // Array
anyHashableCount([1, 2, 3] as AnyHashable) // Array as AnyHashable
anyHashableCount([1, 2, 3] as [AnyHashable]) // Array as [AnyHashable]
anyHashableCount([1, 2, 3][...]) // ArraySlice
anyHashableCount(Set([1, 2, 3])) // Set
```

There results can be summarised as follows: 

**before:**
| arguments              | expected count | `Int` | `String` | `UUID` | `Hashable` |
|------------------------|:--------------:|:-----:|:--------:|:------:|:----------:|
| One                    | 1              | ✅     | ✅        | ✅      | ✅          |
| Three                  | 3              | ✅     | ✅        | ✅      | ✅          |
| Array                  | 3              | ✅     | ✅        | ❌      | ❌          |
| Array as AnyHashable   | 1              | ✅     | ✅        | ✅      | ✅          |
| Array as [AnyHashable] | 3              | ✅     | ✅        | ✅      | ✅          |
| ArraySlice             | 3              | ❌     | ❌        | ❌      | ❌          |
| Set                    | 3              | ❌     | ❌        | ❌      | ❌          |

**after:**
| arguments              | expected count | `Int` | `String` | `UUID` | `Hashable` |
|------------------------|:--------------:|:-----:|:--------:|:------:|:----------:|
| One                    | 1              | ✅     | ✅        | ✅      | ✅          |
| Three                  | 3              | ✅     | ✅        | ✅      | ✅          |
| Array                  | 3              | ✅     | ✅        | ✅      | ✅          |
| Array as AnyHashable   | 1              | ✅     | ✅        | ✅      | ✅          |
| Array as [AnyHashable] | 3              | ✅     | ✅        | ✅      | ✅          |
| ArraySlice             | 3              | ✅     | ✅        | ✅      | ✅          |
| Set                    | 3              | ✅     | ✅        | ✅      | ✅         |

@tgrapperon can you think of any other scenarios we should test out?

Here is the full playground if you want to test it out as well. 

<details>
  <summary>Click to expand!</summary>
 
 ```swift
 import Foundation

// func anyHashableCount(_ hashables: [AnyHashable]) -> Int {
//  hashables.count
// }

@_disfavoredOverload
func anyHashableCount(_ hashables: AnyHashable...) -> Int {
  hashables.count
}

func anyHashableCount<S: Sequence>(_ hashables: S) -> Int where S.Element == AnyHashable {
  Array(hashables).count
}

anyHashableCount(123) == 1 // before: ✅ after: ✅
anyHashableCount(1, 2, 3) == 3 // before: ✅ after: ✅
anyHashableCount([1, 2, 3]) == 3 // before: ✅ after: ✅
anyHashableCount([1, 2, 3] as AnyHashable) == 1 // before: ✅ after: ✅
anyHashableCount([1, 2, 3] as [AnyHashable]) == 3 // before: ✅ after: ✅
anyHashableCount([1, 2, 3][...]) == 3 // before: ❌ after: ✅
anyHashableCount(Set([1, 2, 3])) == 3 // before: ❌ after: ✅

anyHashableCount("123") == 1 // before: ✅ after: ✅
anyHashableCount("1", "2", "3") == 3 // before: ✅ after: ✅
anyHashableCount(["1", "2", "3"]) == 3 // before: ✅ after: ✅
anyHashableCount(["1", "2", "3"] as AnyHashable) == 1 // before: ✅ after: ✅
anyHashableCount(["1", "2", "3"] as [AnyHashable]) == 3 // before: ✅ after: ✅
anyHashableCount(["1", "2", "3"][...]) == 3 // before: ❌ after: ✅
anyHashableCount(Set(["1", "2", "3"])) == 3 // before: ❌ after: ✅

let uuid1 = UUID(uuidString: "00000000-0000-0000-0000-000000000001")!
let uuid2 = UUID(uuidString: "00000000-0000-0000-0000-000000000002")!
let uuid3 = UUID(uuidString: "00000000-0000-0000-0000-000000000003")!

anyHashableCount(uuid1) == 1 // before: ✅ after: ✅
anyHashableCount(uuid1, uuid2, uuid3) == 3 // before: ✅ after: ✅
anyHashableCount([uuid1, uuid2, uuid3]) == 3 // before: ❌ after: ✅
anyHashableCount([uuid1, uuid2, uuid3] as AnyHashable) == 1 // before: ✅ after: ✅
anyHashableCount([uuid1, uuid2, uuid3] as [AnyHashable]) == 3 // before: ✅ after: ✅
anyHashableCount([uuid1, uuid2, uuid3][...]) == 3 // before: ❌ after: ✅
anyHashableCount(Set([uuid1, uuid2, uuid3])) == 3 // before: ❌ after: ✅

struct ID: Hashable {
  var id: Int
}

let id1 = ID(id: 1)
let id2 = ID(id: 2)
let id3 = ID(id: 3)

anyHashableCount(id1) == 1 // before: ✅ after: ✅
anyHashableCount(id1, id2, id3) == 3 // before: ✅ after: ✅
anyHashableCount([id1, id2, id3]) == 3 // before: ❌ after: ✅
anyHashableCount([id1, id2, id3] as AnyHashable) == 1 // before: ✅ after: ✅
anyHashableCount([id1, id2, id3] as [AnyHashable]) == 3 // before: ✅ after: ✅
anyHashableCount([id1, id2, id3][...]) == 3 // before: ❌ after: ✅
anyHashableCount(Set([id1, id2, id3])) == 3 // before: ❌ after: ✅
 ```
</details>

iampatbrown:
I forgot to check using mixed types. I'll do that tomorrow :P

mbrandonw:
Nice find @iampatbrown! Very subtle. I'll wait to merge until you get that last test of mixed types in. Thanks for uncovering this.

tgrapperon:
I'll let @iampatbrown make the final conclusions, but it looks fine on my side. I've only managed to make a `String` go through the sequence overload by wrapping it: `Array("abc")`, that is a `[String.Element]==[Character]` that is coerced to a `Sequence` of `AnyHashable`. If this is the expected behavior, I don't see why `String` which is itself a `Sequence` of `Character` doesn't get the same treatment. 
If I remove the variadic overload, a simple `string` argument doesn't go through the remaining function as it:
```swift
error: global function 'anyHashableCount' requires 
the types 'String.Element' (aka 'Character') 
and 'AnyHashable' be equivalent
```
So, I don't really understand yet why a `[String.Element]` is an acceptable `Sequence` of `AnyHashable` when `String` isn't.

But the PR works fine on my side, and a simple `String` argument will go through the variadic overload as expected. 
I've tested with mixed types, and it also branches as expected, but I'll let @iampatbrown conclude.

iampatbrown:
@tgrapperon @mbrandonw looks good on my end too. Tested a bit more and looked at where `_HasCustomAnyHashableRepresentation` from [swift/AnyHashable](https://github.com/apple/swift/blob/3d039251b4b3fb0f744d090cd964a81d85458130/stdlib/public/core/AnyHashable.swift) is being used and nothing jumped out. Happy to merge this PR as is.

tgrapperon:
> `_HasCustomAnyHashableRepresentation` from [swift/AnyHashable](https://github.com/apple/swift/blob/3d039251b4b3fb0f744d090cd964a81d85458130/stdlib/public/core/AnyHashable.swift)…

Thanks for the pointer @iampatbrown, this is very interesting! I was suspecting some kind of special treatment, but for `String` only, but it turns out it's more complex than that. `AnyHashable` is much more subtle than simply erasing equality and hashing.

lukeredpath:
I'm trying to update our app to 0.34 and I'm getting stuck with this. We have a merged effect that includes a `.cancel` effect. This code was previously compiling:

```swift
return .merge(
  ...,

  .cancel(ids: [One(), Two(), Three()])
)
```

Where the cancel IDs are just `struct One: Hashable {}` for example. If I change this to a variadic list it works, but the array version should work too shouldn't it?

iampatbrown:
Hey @lukeredpath looks like this one slipped through the cracks... in the short term this can be fixed by explicitly casting to `.cancel(ids: [One(), Two(), Three()] as [AnyHashable])`

Reintroducing the `func cancel(ids: [AnyHashable])` overload will fix this at a library level though. @mbrandonw @stephencelis are you fine for me to PR this change? It would require 3 overloads in total. I have verified it works against all the previous tests, however, you may have some additional thoughts.

stephencelis:
@iampatbrown We chatted about it a bit today and think that maybe instead of adding another overload we could just simplify things to a single `[AnyHashable]` interface. The variadic overload is maybe not worth it, ergonomically, and the sequence-based overload doesn't play nicely with Swift's type system (might be worth filing a bug with Apple).

Want to PR that change? If you don't have the time we can try getting around to it soon.

iampatbrown:
> We chatted about it a bit today and think that maybe instead of adding another overload we could just simplify things to a single [AnyHashable] interface. 

@stephencelis I agree. Keeping the variadic overload would allow things like `Set` and `OrderedSet` (ie. `IdentifiedArray.ids` to behave incorrectly without the user knowing. That just doesn't sit right with me, even if it's a rare edge case.

I'll aim to create a PR later today. If you two have anymore thoughts, let me know.

PR #1015, opened 2022-03-09
Fix `WithViewStore` issues with `View`s using escaping closures

tgrapperon:
This PR follows the discussion #1012 and implements the solution proposed by @iampatbrown.

I've reordered the "Animations" case from `CaseStudies`, and it indeed fixes the issue with `GeometryReader`. 
It also seems to solve the `Stepper` issues! I'll check with `ScrollViewReader` later.
I'll hunt other uses and mentions in the documentation in a second step if the implementation is greenlighted.

I've used `newInstance()` instead of `copy()` because the latter could suggest that both instances would behave separately once generated. This is of course open to discussion.

### DIFF
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -282,7 +282,7 @@ public final class ViewStore<State, Action>: ObservableObject {
 
 extension ViewStore {
   func newInstance() -> ViewStore<State, Action> {
-    ViewStore(_send: _send, _state: _state)
+    ViewStore(_send: _send, _state: _state, viewCancellable: viewCancellable)
   }
 }
 
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -76,21 +76,18 @@ public final class ViewStore<State, Action>: ObservableObject {
 
     self.viewCancellable = store.state
       .removeDuplicates(by: isDuplicate)
-      .sink { [weak self] in
-        guard let self = self else { return }
-        self.objectWillChange.send()
-        self._state.value = $0
+      .sink { [weak objectWillChange = self.objectWillChange, weak _state = self._state] in
+        guard let objectWillChange = objectWillChange, let _state = _state else { return }
+        objectWillChange.send()
+        _state.value = $0
       }
   }
 
-  fileprivate init(
-    _send: @escaping (Action) -> Void,
-    _state: CurrentValueRelay<State>,
-    viewCancellable: AnyCancellable? = nil
-  ) {
-    self._send = _send
-    self._state = _state
-    self.viewCancellable = viewCancellable
+  internal init(_ viewStore: ViewStore<State, Action>) {
+    self._send = viewStore._send
+    self._state = viewStore._state
+    self.objectWillChange = viewStore.objectWillChange
+    self.viewCancellable = viewStore.viewCancellable
   }
 
   /// A publisher that emits when state changes.
@@ -277,12 +274,6 @@ public final class ViewStore<State, Action>: ObservableObject {
   ) -> LocalState {
     get { state.rawValue(self.state) }
     set { self.send(action.rawValue(newValue)) }
-  }
-}
-
-extension ViewStore {
-  func newInstance() -> ViewStore<State, Action> {
-    ViewStore(_send: _send, _state: _state, viewCancellable: viewCancellable)
   }
 }
 
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -83,6 +83,16 @@ public final class ViewStore<State, Action>: ObservableObject {
       }
   }
 
+  fileprivate init(
+    _send: @escaping (Action) -> Void,
+    _state: CurrentValueRelay<State>,
+    viewCancellable: AnyCancellable? = nil
+  ) {
+    self._send = _send
+    self._state = _state
+    self.viewCancellable = viewCancellable
+  }
+
   /// A publisher that emits when state changes.
   ///
   /// This publisher supports dynamic member lookup so that you can pluck out a specific field in
@@ -267,6 +277,12 @@ public final class ViewStore<State, Action>: ObservableObject {
   ) -> LocalState {
     get { state.rawValue(self.state) }
     set { self.send(action.rawValue(newValue)) }
+  }
+}
+
+extension ViewStore {
+  func newInstance() -> ViewStore<State, Action> {
+    ViewStore(_send: _send, _state: _state, viewCancellable: viewCancellable)
   }
 }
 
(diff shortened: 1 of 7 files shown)


Sources/ComposableArchitecture/ViewStore.swift
```diff
@@ -83,6 +83,16 @@ public final class ViewStore<State, Action>: ObservableObject {
       }
   }
 
+  fileprivate init(
+    _send: @escaping (Action) -> Void,
+    _state: CurrentValueRelay<State>,
+    viewCancellable: AnyCancellable? = nil
```
iampatbrown, line 89: `viewCancellable` is only really relevant to the root `ViewStore` so I'm not sure it needs to be included. I originally included `objectWillChange` in case the copied view store was observed in some way. This actually makes me wonder if creating a new `ViewStore` may break `.binding()` animations.

tgrapperon: Dropping `viewCancellable` was indeed something I considered. I quickly tried to find cases where it would be needed and I found none yet, but I opted for making a _bona fide_ `ViewStore` in any case. As I'm getting it, keeping `viewCancellable` protects against the user sneaking it outside `SwiftUI` in `onAppear` or `task` for example (I may have done things like that in the past!). Without the cancellable, the escaped instance's state will unsync. Do you see issues in duplicating the `viewCancellable`?

I'll check with animations later, but they're still working in the "Animation" case study where an animated binding is used to toggle "Big mode".

tgrapperon: Ah, the `CurrentValueRelay` is a class, so it will be updated by the parent. I'll need to think more about it then!

tgrapperon: You're probably right. The `AnyCancellable` is only a reference to the existing subscription from the parent, and the `CurrentValueRelay` is the same instance, so there is no need to keep one additional reference to the subscription. I'll update the PR in this sense if it doesn't break anything.

iampatbrown: > the CurrentValueRelay is a class

Yeah. The `viewCancellable` will only cancel the sink subscription to the `Store` which is linked to the original `ViewStore`. Hypothetically this could be contained in a single class which would then allow `ViewStore` to be independently copied by passing that reference around. Essentially making `ViewStore` more like a wrapper. I just didn't want to introduce too much of a change.

> they're still working in the "Animation" case study where an animated binding is used to toggle "Big mode"

I confirmed animations were working too.

tgrapperon: I've updated and reverted the change! As long as the `@ObservedObject` `ViewStore` is alive, there is no issue, but if it goes away and someone still holds the `viewStore` somewhere outside of `SwiftUI`, updating will stop.
I think it should be kept for this reason. Don't you think?

iampatbrown: > updating will stop.

It will stop regardless. If the `@ObservedObject` `ViewStore` doesn't exist. The subscription becomes a no-op:

```swift
self.viewCancellable = store.state
  .removeDuplicates(by: isDuplicate)
  .sink { [weak self] in
    guard let self = self else { return }
    self.objectWillChange.send()
    self._state.value = $0
  }
```

This can be fixed by capturing `ObservableObjectPublisher` and `CurrentValueRelay` separately (or wrapping them up together). But like I said, I didn't suggested it to avoid introducing additional changes.

iampatbrown: essentially something like this: 

```swift
  public init(
    _ store: Store<State, Action>,
    removeDuplicates isDuplicate: @escaping (State, State) -> Bool
  ) {
    self._send = { store.send($0) }
    self._state = CurrentValueRelay(store.state.value)

    self.viewCancellable = store.state
      .removeDuplicates(by: isDuplicate)
      .sink { [weak objectWillChange = self.objectWillChange, weak _state = self._state] in
        guard let objectWillChange = objectWillChange, let _state = _state else { return }
        objectWillChange.send()
        _state.value = $0
      }
  }

  private init(viewStore: ViewStore<State, Action>) {
    self._send = viewStore._send
    self._state = viewStore._state
    self.objectWillChange = viewStore.objectWillChange
    self.viewCancellable = viewStore.viewCancellable
  }
```

That would then work the way you're describing, and holding on to `viewCancellable` would be necessary.

tgrapperon: > It will stop regardless.

You're right, I missed that.

Your capturing implementation seems fine to me, and it doesn't look much more complex than the current one. You're just capturing more narrowly. The `objectWillChange` translocation shouldn't be problematic as this publisher doesn't reference `self` and we don't let the framework synthetize-it to begin with.
I'll refactor in this spirit!
You should get all the credit for this PR!

tgrapperon:
The PR was updated according to @iampatbrown comments.
The `viewCancellable` now captures the `_state` and `objectWillChange` instances instead of `self`.
The `ViewStore` passed to the closure could then escape it and continue to receive updates outside `SwiftUI`, even if the original `ViewStore` instance disappears.

Sources/ComposableArchitecture/ViewStore.swift
iampatbrown: Cool. The updated init looks good too. 

> You're just capturing more narrowly

Which I thought could introduce additional overhead, but it's likely nothing. 

> You should get all the credit for this PR!

Team effort :)

tgrapperon: I've updated the PR. One novel situation is having two different `@ObservedObject` observing different instances, but sharing the same `objectWillChange` publisher. It can happen if the `viewStore` is passed as an argument to an inner view that installs it as `@ObservedObject`. It should most likely work, but I'll check if SwiftUI hasn't specific optimizations around `@ObservedObject`.

iampatbrown: > It can happen if the viewStore is passed as an argument to an inner view that installs it as @ObservedObject

That's the main reason for sharing the objectWillChange publisher. Even though it's probably not applicable because if the parent view changes the child view will be passed a new view store to use with `@ObservedObject`.

tgrapperon: > Which I thought could introduce additional overhead, but it's likely nothing.

Benchmarking shows no difference, in any case.

tgrapperon:
We now seem to have a better understanding of what was causing the issue, why the workaround worked, and why they would probably not fix this.

Right now, the PR fixes all known issues with `WithViewStore`:
- Contained `GeometryReader`s
- Contained `ScrollViewReader`s
- `Stepper`s initialized with `onIncrement/onDecrement` closures.
- Most likely issues from #1012.

I've removed mentions to these known issues and refactored the sample code to use top-level `WithViewStore`s where the `ViewStore`s were installed manually as `@ObservedObject`s.

There are probably other cases that were missed, but I feel that this PR would fix them too.
We haven't found adverse side effects to the change yet.

stephencelis (approved):
Great investigative work! Just tested this branch and brought it into isowords and it's working great.

PR #1017, opened 2022-03-12
Add icons variants for required resolutions

tgrapperon:
Example apps are showing warnings about missing icon sizes.
I've upscaled the existing iPhone icon and added the required variants. All example apps are now building without warnings.
The Asset file could probably be extracted in a `Resources` folder and shared among projects to simplify maintenance if needed.


tgrapperon:
Ok, it seems as good as it can be without centralizing resources or using lossy compression.
The PR adds 200kB in total to the repository to remove build warnings for example apps.

PR #1018, opened 2022-03-12
Add specialized overloads to `Effect.merge` and `Reducer.combine`

iampatbrown:
This PR adds additional overloads to `Effect.merge` and `Reducer.combine` to take advantage of the specialized merge publishers rather than defaulting to `Publishers.MergeMany`. Additional benchmarks have also been added.

```
name                                      time         std        iterations
----------------------------------------------------------------------------
Effects.merge([a, b])                      7275.000 ns ±  36.93 %     184188 <- Before 
Effects.merge(a, b)                        6001.000 ns ±  42.96 %     226572 <- After
Effects.merge([a, b, c, d, e, f, g, h])   15660.000 ns ±  29.94 %      82214 <- Before
Effects.merge(a, b, c, d, e, f, g, h)     12124.000 ns ±  38.70 %     105266 <- After
```

```
name                                      time         std        iterations
----------------------------------------------------------------------------
Reducer.combine([a, b])                   12251.000 ns ±  53.21 %     102758 <- Before
Reducer.combine(a, b)                      9703.000 ns ±  36.59 %     129784 <- After
Reducer.combine([a, b, c, d, e, f, g, h]) 21670.000 ns ±  40.42 %      60340 <- Before
Reducer.combine(a, b, c, d, e, f, g, h)   15953.000 ns ±  37.13 %      74157 <- After
```

Notes: 
- Could revert benchmark changes. They are mainly to give an indication of changes in performance. 
- Alternative parameter names could be used `combine(_ a: Reducer, _ b: Reducer)` vs `combine(_ r0: Reducer, _ r1: Reducer)`
- Unsure about related docs and formatting.

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -221,11 +221,133 @@ public struct Effect<Output, Failure: Error>: Publisher {
       }
   }
 
+  /// Merges two effects together into a single effect, which runs the effects at the same time.
+  ///
+  /// - Parameters:
+  ///   - a: An effect.
+  ///   - b: A second effect.
+  /// - Returns: A new effect
+  public static func merge(_ a: Effect, _ b: Effect) -> Effect {
+    Publishers.Merge(a, b).eraseToEffect()
+  }
+
+  /// Merges three effects together into a single effect, which runs the effects at the same time.
+  ///
+  /// - Parameters:
+  ///   - a: An effect.
+  ///   - b: A second effect.
+  ///   - c: A third effect.
+  /// - Returns: A new effect
+  public static func merge(_ a: Effect, _ b: Effect, _ c: Effect) -> Effect {
+    Publishers.Merge3(a, b, c).eraseToEffect()
+  }
+
+  /// Merges four effects together into a single effect, which runs the effects at the same time.
+  ///
+  /// - Parameters:
+  ///   - a: An effect.
+  ///   - b: A second effect.
+  ///   - c: A third effect.
+  ///   - d: A fourth effect.
+  /// - Returns: A new effect
+  public static func merge(_ a: Effect, _ b: Effect, _ c: Effect, _ d: Effect) -> Effect {
+    Publishers.Merge4(a, b, c, d).eraseToEffect()
+  }
+
+  /// Merges five effects together into a single effect, which runs the effects at the same time.
+  ///
+  /// - Parameters:
+  ///   - a: An effect.
+  ///   - b: A second effect.
+  ///   - c: A third effect.
+  ///   - d: A fourth effect.
+  ///   - e: A fifth effect.
+  /// - Returns: A new effect
+  public static func merge(
+    _ a: Effect,
+    _ b: Effect,
+    _ c: Effect,
+    _ d: Effect,
+    _ e: Effect
+  ) -> Effect {
+    Publishers.Merge5(a, b, c, d, e).eraseToEffect()
+  }
+
+  /// Merges six effects together into a single effect, which runs the effects at the same time.
+  ///
+  /// - Parameters:
+  ///   - a: An effect.
+  ///   - b: A second effect.
+  ///   - c: A third effect.
+  ///   - d: A fourth effect.
+  ///   - e: A fifth effect.
+  ///   - f: A sixth effect.
+  /// - Returns: A new effect
+  public static func merge(
+    _ a: Effect,
+    _ b: Effect,
+    _ c: Effect,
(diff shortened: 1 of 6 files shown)


iampatbrown:
@mbrandonw @stephencelis CI failed at `ReducerTests.testDebug` which I don't think is impacted by this change.

> Asynchronous wait failed: Exceeded timeout of 2 seconds, with unfulfilled expectations: "logs".

https://github.com/pointfreeco/swift-composable-architecture/blob/d924b9ad27d2a2ccb0ada2639f5255084ff63382/Tests/ComposableArchitectureTests/ReducerTests.swift#L113-L120

Think this is similar to https://github.com/pointfreeco/swift-composable-architecture/pull/181

tgrapperon:
Great find @iampatbrown, the gain in performance seems very substantial, even for two effects only.
I know that @mbrandonw and @stephencelis are generally wary about adding additional overloads/API surface. Each case should be discussed of course, and the compiler has made great progress in this area, so it's maybe not an issue as of today. However, it seems there's a case for an `EffectBuilder` resultBuilder. I wonder what form it could take in this case. That wouldn't probably solve the number of overloads, but it can maybe simplify what the user will see for code completion. Maybe there are also features from `resultBuilder` that can be exploited here to provide new functionalities to the user, like conditional blocks for example. I have a few reducers where I know I could use such a builder to add optional effects inline instead of using an array that I need to name. 
Another positive side effect of `resultBuilder`s is that they provide a level of API indirection that gives more freedom when reorganizing the internal without touching the public API. This can prove useful if the library migrates away from `Combine` at some point.
But it's just an idea like that, and we can discuss it elsewhere if you think it's too off-topic for this PR.

iampatbrown:
> I know that @mbrandonw and @stephencelis are generally wary about adding additional overloads/API surface.

Understandable. That's one of the reasons I've made this a draft. I do think there's an opportunity to use result builder for sure. If you feel like starting up a separate discussion go for it :) My hope was this PR would have minimal impact on the API and allow existing code to benefit from the performance improvements.  

I recognise this may become redundant as alternatives to `Combine` are introduced, however, I thought there was a decent enough performance gain to justify the change. I also think `.combine` and `.merge` are used pretty heavily at the moment.

iampatbrown:
@tgrapperon As an alternative it's possible to use `switch reducers.count` and `switch effects.count` and specialize that way. I just found the overloads benchmarked slightly better.

tgrapperon:
>  I thought there was a decent enough performance gain to justify the change

Indeed, I was just pitching the `EffectBuilder` idea. I'll probably open another PR later to explore further.

> As an alternative it's possible to use switch reducers.count and switch effects.count

As I'm getting it, users using the variadic overload will automatically switch to the specific overload without having to touch their code, so the current PR approach doesn't effectively increase the API surface as much as it can seem at first sight. The additional burden on the compiler is probably acceptable too.

tgrapperon:
@iampatbrown, you said you were unsure about documentation, so it was probably about this: There is a 25-30% penalty in using the array variants, but there is nothing yet to inform the user about the difference. Do you think some `/// - Note:` in the doc would be enough, or should the array variant be made more difficult to reach in order to favorize the faster specialized version when possible (using a labeled argument for example[^1])?

I personally never used the array variant for `Reducer`s, but I'm often using the array variant for `Effect`s when I need to return effects conditionally. With this PR in place, I would wrongly assume I'm optimizing SLOC's without consequences.

[^1]: If a label is added to the array variant, the unlabelled variant could probably be deprecated with a fix-it that would also invite to switch to the specialized variant. I'm unsure if this deprecation works with respect to the previous fix you did with `@_disfavoredOverload`, or if array variants would fall back again by mistake into the variadic overload? (I don't know how Swift ranks `@_disfavoredOverload` vs deprecated methods. It probably doesn't change anything, but I need to verify)

After thinking more about it, the `EffectBuilder` idea is indeed completely off-topic for the scope of this PR: It adds no new API/features, and many users will assume they're using the variadic variant where they're instead using the specialized one. It's indeed a completly different question.

stephencelis:
@iampatbrown Cool exploration! This PR and @tgrapperon's give us a lot to think about. We definitely want to revisit some of these ideas when we get around to exploring `Reducer` as a protocol, and what builders bring to the table then (especially with constrained opaque types on the horizon).

Because of this we may hold off on merging these things for now since the improvements, while nice, are also pretty small in the grand scheme of things, and the added surface area can be intimidating (we've seen firsthand how `swift-parsing` can be a lil steeper to dive into than a TCA project). Have you heard of `Reducer.combine` being a bottleneck anywhere?

PR #1019, opened 2022-03-13
[Experimental] Add an `EffectBuilder` result builder

tgrapperon:
Hello,
This is an experiment following the idea pitched in #1018. An `EffectBuilder` result builder type is introduced. It leads to an interesting (but quite dangerous, see below) syntax. There are two entry points to the reducer:
- An `Effect` initializer with a builder closure (two variants in reality, see below)
- The main `Reducer` initializer!

A few notes/observations:
- The builder is mostly an array builder of merged effects.
- Each expression in the reducer generates an `Effect`! It generates `.none` most of the time for `state` changes or variable declarations. This is quite inefficient, but more permissive than SwiftUI's `ViewBuilder`.
- I've added a `isNone` property to `Effect` in order to filter them out easily when merging them. This can maybe be used to optimize the current version of the library too.
- The builder syntax doesn't pair well with static methods, as any static`.xxx` is often understood as a method call on the previous line.
- Because of this, and to avoid redefining new types, the `Effect` initializer accepts an argument-less closure or a closure with an argument that is the type of the effect and upon which we can call static methods:
```swift
Effect {
  $0.fireAndForget { … } // $0 == Effect<Output, Failure>
}
```
- I made `Output/Action` build `Effect(value:)` and `Failure` build `Effect(error:)`, mostly to see how it would look.
- I've added a new Counter-Builder CaseStudy that uses the `EffectBuilder` (opt-in by not using `return`s). The result is quite interesting, but we can generate effects by mistake too easily!
- The builder was implemented as a subtype of `Effect` to share its generics, mostly by convenience. It probably interferes with type inference in some configurations.

In its current shape, the builder is maybe too permissive. We generate merged `Effects` very easily. Cases are also automatically promoted to `.none` by default. Static methods can't be called using inference most of the time. Introducting new top-level types instead is possible but would quickly clash with existing names like `Future` for example, and would furthermore prevent to use code completion for discovery.

It is very nice that both `EffectBuilder` and raw closure variants of the `Reducer` initializer can be used by coercion. Adding `@Effect<>.Builder` to the `Reducer initializer didn't break a single line of existing code in tests and `CaseStudies`.

This PR is only for experimental purposes. Feel free to provide any feedback!
@stephencelis, I'm doing a draft PR as it is easier to pinpoint lines of codes to discuss and evolve things. Please let me know if you prefer that I create discussions for this kind of thing instead.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Counter-EffectBuilder.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Counter-EffectBuilder.swift
@@ -1,0 +1,88 @@
+import ComposableArchitecture
+import SwiftUI
+
+private let readMe = """
+  This screen demonstrates the basics of the Composable Architecture in an archetypal counter \
+  application.
+
+  The domain of the application is modeled using simple data types that correspond to the mutable \
+  state of the application and any actions that can affect that state or the outside world.
+  """
+
+struct CounterBuilderState: Equatable {
+  var count = 0
+}
+
+enum CounterBuilderAction: Equatable {
+  case decrementButtonTapped
+  case incrementButtonTapped
+  
+  case doubleIncrementButtonTapped
+  case doubleDecrementButtonTapped
+}
+
+struct CounterBuilderEnvironment {}
+
+let counterBuilderReducer = Reducer<CounterBuilderState, CounterBuilderAction, CounterBuilderEnvironment> { state, action, _ in
+  switch action {
+  case .decrementButtonTapped:
+    state.count -= 1
+  case .incrementButtonTapped:
+    state.count += 1
+  case .doubleDecrementButtonTapped:
+    if state.count >= 2 {
+      CounterBuilderAction.decrementButtonTapped
+      CounterBuilderAction.decrementButtonTapped
+    }
+  case .doubleIncrementButtonTapped:
+    CounterBuilderAction.incrementButtonTapped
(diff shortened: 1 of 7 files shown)


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Counter-EffectBuilder.swift
```diff
@@ -0,0 +1,88 @@
...
+  case doubleIncrementButtonTapped
+  case doubleDecrementButtonTapped
+}
+
+struct CounterBuilderEnvironment {}
+
+let counterBuilderReducer = Reducer<CounterBuilderState, CounterBuilderAction, CounterBuilderEnvironment> { state, action, _ in
+  switch action {
+  case .decrementButtonTapped:
+    state.count -= 1
```
tgrapperon, line 29: `state.count -= 1` is built as `.none` by the builder's `Void` expression.

tgrapperon:
Another observation is that using the result builder in `Reducer`s blocks won't currently work with specializations like the ones being currently used in `swift-parsing` or proposed in #1018. Most statements will build `Void` as `Effect.none`. This will clutter the combinatoric, making the generation of specialized overloads effectively impossible. To make it work in `Reducer`s, the new `reducing` methods of result builders would be required to coalesce/compactify all theses `.none` prefixes.

That being said, and since it was mostly an experiment, I'm closing this PR for now. Thanks for your attention!

PR #1023, opened 2022-03-15
Runtime Warning Finesse

stephencelis:
Makes it so runtime warnings invoke `XCTFail` during tests.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/Deprecations.swift
+++ b/Sources/ComposableArchitecture/Internal/Deprecations.swift
@@ -3,10 +3,6 @@ import Combine
 import SwiftUI
 import XCTestDynamicOverlay
 
-#if DEBUG
-  import os
-#endif
-
 // NB: Deprecated after 0.31.0:
 
 extension Reducer {
@@ -548,8 +544,7 @@ extension Reducer {
       }
       if index >= globalState[keyPath: toLocalState].endIndex {
         #if DEBUG
-          os_log(
-            .fault, dso: rw.dso, log: rw.log,
+          runtimeWarning(
             """
             A "forEach" reducer at "%@:%d" received an action when state contained no element at \
             that index. …
@@ -563,21 +558,21 @@ extension Reducer {
             reasons:
 
             • This "forEach" reducer was combined with or run from another reducer that removed \
-            the element at this index when it handled this action. To fix this make sure that \
-            this "forEach" reducer is run before any other reducers that can move or remove \
-            elements from state. This ensures that "forEach" reducers can handle their actions \
-            for the element at the intended index.
+            the element at this index when it handled this action. To fix this make sure that this \
+            "forEach" reducer is run before any other reducers that can move or remove elements \
+            from state. This ensures that "forEach" reducers can handle their actions for the \
+            element at the intended index.
 
             • An in-flight effect emitted this action while state contained no element at this \
             index. While it may be perfectly reasonable to ignore this action, you may want to \
             cancel the associated effect when moving or removing an element. If your "forEach" \
-            reducer returns any long-living effects, you should use the identifier-based \
-            "forEach" instead.
+            reducer returns any long-living effects, you should use the identifier-based "forEach" \
+            instead.
 
-            • This action was sent to the store while its state contained no element at this \
-            index. To fix this make sure that actions for this reducer can only be sent to a \
-            view store when its state contains an element at this index. In SwiftUI \
-            applications, use "ForEachStore".
+            • This action was sent to the store while its state contained no element at this index \
+            To fix this make sure that actions for this reducer can only be sent to a view store \
+            when its state contains an element at this index. In SwiftUI applications, use \
+            "ForEachStore".
             """,
             "\(file)",
             line,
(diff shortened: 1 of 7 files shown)


Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -548,8 +544,7 @@ extension Reducer {
       }
       if index >= globalState[keyPath: toLocalState].endIndex {
         #if DEBUG
-          os_log(
-            .fault, dso: rw.dso, log: rw.log,
+          runtimeWarning(
```
stephencelis, line 547: Moved things to a helper function, which means the purple warning unfortunately happens there now. It's not super helpful to highlight even here, though.

mbrandonw (approved):

Sources/ComposableArchitecture/Internal/Deprecations.swift
iampatbrown: You could use [@_transparent](https://github.com/apple/swift/blob/main/docs/TransparentAttr.md) if you wanted to get the warning back to the `runtimeWarning` call site

stephencelis: Oh great! TIL :smile:

You up to PR that?

iampatbrown: Yeah no worries

lukeredpath:
@stephencelis having this automatically trigger an XCTFail without any way to disable this feels like quite a strongly opinionated approach and breaking change to include - I can think of a number of tests in our test suite that this would cause to fail because we are taking a shortcut and sending child actions just to test the parent domain's handling of that action, to simplify tests. Would you consider some way to selectively disable this?

stephencelis:
@lukeredpath Are you seeing new test failures due to this change? Runtime warnings (and XCTest failures) should only happen if you send a child action to _nil_ state, which would be an invalid thing to do and probably should cause a test failure, no?

Remember that previously to the introduction of runtime warnings we were setting a breakpoint instead, which is even more disruptive.

lukeredpath:
We have a couple of tests in our suite that do produce runtime warnings that would fail with this change - previously with the breakpoint implementation you could disable this by passing `breakpointOnNil: false` which we had to do in a few places due to weird SwiftUI bugs.

To give an example, we might have some kind of Parent > Optional child domain relationship. We might send `.actionA` to the parent domain which initialises the child domain state and then triggers some action in the child domain which starts of some complicated series of effects and callbacks and then a final action to indicate something was completed. We handle this final completion action in the parent reducer to perform some additional logic.

For the sake of brevity we don't always want to have to set up all of the mocks and handle all of the child domain actions for this flow just to test the behaviour of the parent domain, so we might _simulate_ the completion by sending the final child action to the store. We don't care about the child domain's behaviour and so it's state is `nil`. This approach is used in a few of our tests and will fail with this change.

Another test in our suite that would fail is for some actual TCA utility code we have for dealing with pulling back reducers over optional paths using a `CasePath` - this the `OptionalPath` stuff on one of the TCA branches that we are using in our app. We have some tests for this functionality and have a test specifically around this behaviour that triggers a runtime warning, but that's because we are trying to test this actual behaviour:

```swift
func testReducerOverOptionalStateUsingOptionalPath() {
        let sayHelloReducer = Reducer<String, Action, Void> { state, action, _ in
            switch action {
            case .sayHello:
                state = "Hello there!"
            default:
                break
            }
            return .none
        }

        let optionalReducer: Reducer<SomeState, Action, Void> = sayHelloReducer
            .pullback(
                state: OptionalPath(\SomeState.optionalString),
                action: /.self,
                environment: { _ in () }
            )

        var state = SomeState()
        _ = optionalReducer(&state, .sayHello, ())
        XCTAssertEqual(nil, state.optionalString, "Expected reducer to not run when optional state is nil")

        state = SomeState(optionalString: "")
        _ = optionalReducer(&state, .sayHello, ())
        XCTAssertEqual("Hello there!", state.optionalString, "Expected string to be set")
    }
```

We just ignore the runtime warnings in this case but a failure would be something we can't ignore, we'd probably have to remove this test.

stephencelis:
@lukeredpath We wanted to share some of our thoughts with you to see what you think. For each of the two points you bring up, are you saying that prior to runtime warnings, when you'd run your test suite, you'd hit a few breakpoints, and then need to manually hit continue for each one?

We actually think we made a mistake in releasing runtime warnings _without_ the corresponding `XCTFail`. We rushed to soften the experience of running an app, especially when it comes to SwiftUI bugs, but in doing so we also muted potential programming errors in tests.

We also think that providing an escape hatch like `breakpointOnNil` was ultimately a mistake, but a necessary temporary measure given the lack of better tooling at the time.

We think it's totally fine, though, to develop workarounds that make the testing experience nicer for complex flows (as has been discussed recently), with the understanding that you are opting out of some safety built into the library. We want to make sure the default library experience catches these potential programming errors.

We discussed a couple paths forward you could take:

1. You could use `XCTExpectFailure` to wrap calls that produce these warnings. There's a matcher block that you can use to expect a specific failure message, so that actual failures aren't missed. We don't change runtime warning messages often, so hopefully you wouldn't see too much churn with this approach across releases.

2. All of our `pullback` operations are quite simple/fundamental in the sense that they don't contain much logic other than gluing everything together. Because of this, you can define your own `{quiet,unsafe}Pullback`s in your projects. It sounds like you already maintain the optional pullback? If you choose to maintain these versions that are a little less noisy, you are free to tune the noise dial to whatever works for you.

Are either of these good options?

PR #1026, opened 2022-03-17
Update .spi.yml

finestructure:
Fixes https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1626

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,5 +1,13 @@
 version: 1
 builder:
   configs:
+  - platform: ios
+    scheme: ComposableArchitecture
+  - platform: macos-xcodebuild
+    scheme: ComposableArchitecture
+  - platform: macos-xcodebuild-arm
+    scheme: ComposableArchitecture
+  - platform: tvos
+    scheme: ComposableArchitecture
   - platform: watchos
     scheme: ComposableArchitecture_watchOS


finestructure:
Unfortunately this will be required whenever the particular circumstances of the underlying issue https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1532 are met.

In a nutshell, the issue there was that under certain particular circumstances, `xcodebuild -list` will pick up schemes from xcode projects in _subdirectories_ (like your examples and showcases) 😦.

We're making several attempts to discover the correct scheme to build but in TCA's case we end up picking `swift-composable-architecture-Package`. _Normally_, that's the right scheme but not in your case.

You're always going to be on the safe side if you specify the correct schemes in `.spi.yml`. But in order to be able to determine that ahead of time, here's what should work: Whenever `{package name}-Package` (which is what we should end up discovering) is _not_ the correct scheme, provide an override in `.spi.yml`.

Our `xcodebuild` commands are the following, in case you want to preflight the builds:

```
env DEVELOPER_DIR="/Applications/Xcode_13.2.1.app" xcrun xcodebuild -IDEClonedSourcePackagesDirPathOverride="$PWD/.dependencies" -derivedDataPath "$PWD/.derivedData" build -scheme "\(scheme)" -destination "generic/platform=ios"
```

`ios` can be replaced with `tvos` and `watchos`.

macOS builds use `-destination 'platform=macos,arch=x86_64'` and `-destination 'platform=macos,arch=arm64'` for Swift 5.5 and above, `-arch x86_64`, `-arch arm64` for earlier versions.

I know this is quite annoying and fiddly - sorry about that! I really hope that in the early 2030s or so we can drop the `xcodebuild` code path 🤣

finestructure:
> Unfortunately this will be required whenever the particular circumstances of the underlying issue [SwiftPackageIndex/SwiftPackageIndex-Server#1532](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1532) are met.

Actually, in this case that particular issue simply exposed that we were building the wrong scheme all along. The issue of TCA requiring a custom scheme would have need fixing either way, we just failed to notice it due to the example projects polluting the "scheme space".

PR #1037, opened 2022-03-23
Upgrade swift-case-paths to 0.8.0

nsillik:
The latest release of swift-case-paths fixes an [issue](https://github.com/pointfreeco/swift-case-paths/issues/63)
that can lead to crashes when using TCA. This commit upgrades the version to include the fix.


stephencelis (approved):
Reasonable to us! Thanks!

PR #1041, opened 2022-03-29
Effect.cancel(ids:) prefer array of ids and deprecate variadic overload

iampatbrown:
As discussed in #1009

Making this a draft so I can add more information tomorrow. Ran out of time. 

I had a quick look on GitHub and only found a couple of projects using the variadic overload.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/Deprecations.swift
+++ b/Sources/ComposableArchitecture/Internal/Deprecations.swift
@@ -3,6 +3,20 @@ import Combine
 import SwiftUI
 import XCTestDynamicOverlay
 
+// NB: Deprecated after 0.34.0:
+
+extension Effect {
+  @available(
+    *,
+    deprecated,
+    message: "For improved safety, using a variadic list is no longer supported. Use an array of identifiers instead. For more on this change, see: https://github.com/pointfreeco/swift-composable-architecture/pull/1041"
+  )
+  @_disfavoredOverload
+  public static func cancel(ids: AnyHashable...) -> Effect {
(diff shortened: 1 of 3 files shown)


Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -3,6 +3,20 @@ import Combine
 import SwiftUI
 import XCTestDynamicOverlay
 
+// NB: Deprecated after 0.34.0:
+
+extension Effect {
+  @available(
+    *,
+    deprecated,
+    message: "For improved safety, using a variadic list is no longer supported. Use an array of identifiers instead. For more on this change, see: https://github.com/pointfreeco/swift-composable-architecture/pull/1041"
```
iampatbrown, line 12: @stephencelis are you fine with this? The plan is to add a some context to the PR outlining the issue related to supporting both `AnyHashable...` and `[AnyHashable]`. Not sure if safety is the right word though.

stephencelis, line 12:
```suggestion
    message: "Using a variadic list is no longer supported. Use an array of identifiers instead. For more on this change, see: https://github.com/pointfreeco/swift-composable-architecture/pull/1041"
```

stephencelis (approved):
Thanks!

Sources/ComposableArchitecture/Internal/Deprecations.swift
iampatbrown: Good change. I'll update the PR description tonight.

PR #1050, opened 2022-04-06
Add a map-style overload to eraseToEffect

klundberg:
I'm wondering if there's appetite for this sort of change? I find myself needing to map after erasing to an effect (similar to how catching to an effect also usually needs a map), since many of my dependencies' publishers don't return effects of ready-to-use actions. A convenience operator that combines `eraseToEffect` and `map` as I've implemented would be useful to have, and I bet other users of TCA would benefit as well. If it's not a big enough benefit to include in the core library, no big deal though!

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -315,6 +315,28 @@ extension Publisher {
     Effect(self)
   }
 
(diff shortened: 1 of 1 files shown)


stephencelis (approved):
Makes sense to us! Thanks for PRing!

PR #1055, opened 2022-04-08
Let consumers decide whether to print out messages for other build configurations than DEBUG

bjford:
Currently the `.debug()` reducer defined in `ReducerDebugging.swift` prints only in `DEBUG` builds.
As stated [here](https://github.com/pointfreeco/swift-composable-architecture/blob/1a2dd2d505ba968243cf5a67bf0151abef6099f5/Sources/ComposableArchitecture/Debugging/ReducerDebugging.swift#L34)

These changes let consumers decide whether to print messages for other build configurations than `DEBUG`.
The motivation and use-case for these changes is the following: 
If your app relies heavily on a third-party library which runs its production code/implementation only while in RELEASE mode, then the `.debug()` reducer won't print out any messages, making it difficult to debug.

What do you think about it?

### DIFF
--- a/Sources/ComposableArchitecture/Debugging/ReducerDebugging.swift
+++ b/Sources/ComposableArchitecture/Debugging/ReducerDebugging.swift
@@ -31,11 +31,10 @@ public enum ActionFormat {
 extension Reducer {
   /// Prints debug messages describing all received actions and state mutations.
   ///
+  /// Printing is only done in debug (`#if DEBUG`) builds.
+  ///
   /// - Parameters:
   ///   - prefix: A string with which to prefix all debug messages.
-  ///   - shouldPrintInOtherBuildConfiguration: A boolean value that indicates
-  ///     whether to print out messages for other build configurations (e.g. `RELEASE`, `ALPHA`, etc..)
-  ///     than `DEBUG`.
   ///   - toDebugEnvironment: A function that transforms an environment into a debug environment by
   ///     describing a print function and a queue to print from. Defaults to a function that ignores
   ///     the environment and returns a default ``DebugEnvironment`` that uses Swift's `print`
@@ -43,7 +42,6 @@ extension Reducer {
   /// - Returns: A reducer that prints debug messages for all received actions.
   public func debug(
     _ prefix: String = "",
-    shouldPrintInOtherBuildConfigurations: Bool = false,
     actionFormat: ActionFormat = .prettyPrint,
     environment toDebugEnvironment: @escaping (Environment) -> DebugEnvironment = { _ in
       DebugEnvironment()
@@ -51,7 +49,6 @@ extension Reducer {
   ) -> Reducer {
     self.debug(
       prefix,
-      shouldPrintInOtherBuildConfigurations: shouldPrintInOtherBuildConfigurations,
       state: { $0 },
       action: .self,
       actionFormat: actionFormat,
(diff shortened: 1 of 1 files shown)


AlexisQapa:
IMHO the API would be cleaner by taking a closure to delegate if the printing should be done. You could then default the closure to be a check if DEBUG thought I'm not sure if the method name should then stay as is. Maybe it's better to have a print() operator with the configurability and keep debug as is.

tgrapperon:
Shouldn't a build flag be more appropriate for this purpose? It feels a little odd to configure build-specific behaviors from a standard function argument, and it seems to be something you want to opt-in globally anyway, and not on a per reducer basis. Some launch argument or environment variable could maybe be used instead. This would unclutter the API for a very specific (or power-user-oriented) use case, and this would protect against shipping a binary that prints by mistake.

Furthermore, a side effect of this PR is that the `debug()` implementation now ships with any RELEASE build. I didn't explore what are the ramifications, but it looks like it will bring bits `swift-custom-dump` with it. Unfortunately, we can't use alternative build configurations when using SPM, so it seems to be unavoidable. Does this PR increase the size of an optimized binary by a lot?

bjford:
> Shouldn't a build flag be more appropriate for this purpose? It feels a little odd to configure build-specific behaviors from a standard function argument, and it seems to be something you want to opt-in globally anyway, and not on a per reducer basis. Some launch argument or environment variable could maybe be used instead. This would unclutter the API for a very specific (or power-user-oriented) use case, and this would protect against shipping a binary that prints by mistake.
> 
> Furthermore, a side effect of this PR is that the `debug()` implementation now ships with any RELEASE build. I didn't explore what are the ramifications, but it looks like it will bring bits of `swift-custom-dump` with it. Unfortunately, we can't use alternative build configurations when using SPM, so it seems to be unavoidable. Does this PR increase the size of an optimized binary by a lot?

@tgrapperon @AlexisQapa thank you for your feedback!
I've tried to use an environment variable but then I got the following error `Expressions are not allowed at the top level` while using it in the sample projects.
For now, I've added a `.print()` operator that prints for any build configurations, so at least the `.debug()` operator isn't not cluttered.
I'll investigate the idea of using a build flag as described [here](https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#additional-flags)

@tgrapperon how can I check the size of an optimized binary? I tried to do it, but I'm not sure I'm doing it properly. Is it by building the package in release mode and then checking the size of the `release` folder under the build folder, as described [here](https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#release)?

tgrapperon:
@bjford Maybe you can wrap you environment checks inside the `debug()` function. You can't do it in top level, but you should be able to `guard` against it and early return in `debug()`. But I'm afraid it's not very efficient, and I'm not sure there's a way for us to control if code is stripped or not when using `Release` mode. It seems we can only use `DEBUG` when working with SPM.

About binary size, yes, I meant checking the generated binary size when built using either `Release` mode or the `-O` argument from command line, and comparing the result with or without your change.

I wonder if this shouldn't be better extracted as a third party library: this is a rather narrow use-case, and it looks overkill to make every app ship with debug code. You would be able to quickly import this hypothetical 3rd party library when debugging it in `Release` mode and comment the dependency when shipping the app. Would it be doable with the current public API? It would probably respond to every concern I had about performance, binary size and API.

bjford:
@tgrapperon You have a very good point there! I've tried and with the current public API it's not doable because `debugCaseOutput` func is internal ([here](https://github.com/pointfreeco/swift-composable-architecture/blob/2828dc44f6e3f81d84bcaba72c1ab1c0121d66f6/Sources/ComposableArchitecture/Debugging/ReducerInstrumentation.swift#L90)). 
Do you think it could be made public? otherwise I can just remove it in the hypothetical 3rd party library, and have only `.prettyPrint` as `actionFormat` ([here](https://github.com/pointfreeco/swift-composable-architecture/blob/06b38d780e464a299d1f9c2030a2154c9c69ad4c/Sources/ComposableArchitecture/Debugging/ReducerDebugging.swift#L180)

bjford:
@tgrapperon Yes, that would be fine by me. 👍

bjford:
@tgrapperon 
here it is: https://github.com/bjford/swift-composable-debug-extension
I guess I can close the PR now :)

tgrapperon:
> here it is: https://github.com/bjford/swift-composable-debug-extension

It looks like it's still private (or the address is wrong)

bjford:
> > here it is: https://github.com/bjford/swift-composable-debug-extension
> 
> It looks like it's still private (or the address is wrong)

ooops, try now :)

PR #1064, opened 2022-04-27
Fix invalid web socket server URL

tatsuz0u:
> echo.websocket.org was shut down sometime in 2021 when the company was acquired and changed their name. Try going to [www.websocket.org](http://www.websocket.org/) and you'll see the service is not longer available.
>
> https://stackoverflow.com/a/70438302

I replaced `wss://echo.websocket.org` with `wss://echo.websocket.events` (which is provided by the answerer of the link above) since the `.org` one is no longer available.


stephencelis (approved):
Good find! Thanks for taking the time to PR a fix :smile:

PR #1077, opened 2022-05-03
Improve performance of type-safe effect cancel token lookup

stephencelis:
Our current recommendation of using a dedicated type for effect cancel tokens unfortunately leads to the equivalent of a linear search due to Swift's default hashing behavior:

```swift
struct A: Hashable {}
struct B: Hashable {}

A().hashValue == B().hashValue // true
```

This PR introduces an internal wrapper type that includes the type's unique identifier in the hashing logic, which should improve the hashing of these identifiers:

```swift
struct CancelToken: Hashable {
  let id: AnyHashable
  let discriminator: ObjectIdentifier

  init(id: AnyHashable) {
    self.id = id
    self.discriminator = ObjectIdentifier(type(of: id.base))
  }
}

CancelToken(id: A()).hashValue == CancelToken(id: A()).hashValue // true
CancelToken(id: B()).hashValue == CancelToken(id: B()).hashValue // true
CancelToken(id: A()).hashValue == CancelToken(id: B()).hashValue // false
```

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -37,6 +37,7 @@ extension Effect {
       cancellablesLock.lock()
       defer { cancellablesLock.unlock() }
 
+      let id = CancelToken(id: id)
       if cancelInFlight {
         cancellationCancellables[id]?.forEach { $0.cancel() }
       }
@@ -94,5 +95,15 @@ extension Effect {
   }
 }
 
-var cancellationCancellables: [AnyHashable: Set<AnyCancellable>] = [:]
+struct CancelToken: Hashable {
+  let id: AnyHashable
+  let discriminator: ObjectIdentifier
+
+  init(id: AnyHashable) {
+    self.id = id
+    self.discriminator = ObjectIdentifier(type(of: id.base))
+  }
+}
+
+var cancellationCancellables: [CancelToken: Set<AnyCancellable>] = [:]
 let cancellablesLock = NSRecursiveLock()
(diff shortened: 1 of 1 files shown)


mbrandonw (approved):

tgrapperon:
We even have `A().hashValue == Hasher().finalize()`!
It may seem surprising at first, but it corresponds to the way you would implement `Hashable` manually for the empty struct `A`.

stephencelis:
Some naive benchmarks. Even with just 10 poorly-hashed cancellable effects in-flight, the wrapper with `type(of:)` overhead performs better:

```swift
name                         time           std        iterations
-----------------------------------------------------------------
CancelToken                      292.000 ns ±  57.27 %    1000000
Empty AnyHashable               1208.000 ns ±  21.68 %    1000000
```

With 10,000 long-living cancellable effects (probably unlikely but anything is possible!):

```swift
name                         time           std        iterations
-----------------------------------------------------------------
CancelToken                      292.000 ns ±  57.27 %    1000000
Empty AnyHashable            1088917.000 ns ±   4.24 %       1286
```

tgrapperon:
@stephencelis Nice! I was thinking about benchmarking them, as I was worrying a little about `ObjectIdentifier`. Glory to the hash table then!

About the initial issue and default `Hashable` implementation, it is also interesting to confirm that the ordering of your struct's fields matters:
```swift
struct A: Hashable {
  let int: Int = 0
  let string: String = ""
}

struct B: Hashable {
  let int: Int = 0
  let string: String = ""
}

struct C: Hashable {
  let string: String = ""
  let int: Int = 0
}

A().hashValue == B().hashValue // true
A().hashValue == C().hashValue // false
```

PR #1078, opened 2022-05-03
Add cancel overloads that take types

stephencelis:
While it's easy enough to create instances of hashable structs for cancel tokens, it's effectively using static info under the hood (more so as of #1077). So let's consider adding overloads that take `Any.Type` instead. In use:

```diff
-struct CancelId: Hashable {}
-return .cancel(id: CancelId())
+enum CancelId {}
+return .cancel(id: CancelId.self)
```

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -5,9 +5,10 @@ extension Effect {
   /// Turns an effect into one that is capable of being canceled.
   ///
   /// To turn an effect into a cancellable one you must provide an identifier, which is used in
-  /// ``Effect/cancel(id:)`` to identify which in-flight effect should be canceled. Any hashable
-  /// value can be used for the identifier, such as a string, but you can add a bit of protection
-  /// against typos by defining a new type that conforms to `Hashable`, such as an empty struct:
+  /// ``Effect/cancel(id:)-iun1`` to identify which in-flight effect should be canceled. Any
+  /// hashable value can be used for the identifier, such as a string, but you can add a bit of
+  /// protection against typos by defining a new type for the identifier, or by defining a custom
+  /// hashable type:
   ///
   /// ```swift
   /// struct LoadUserId: Hashable {}
@@ -72,17 +73,43 @@ extension Effect {
     .eraseToEffect()
   }
 
+  /// Turns an effect into one that is capable of being canceled.
+  ///
+  /// A convenience for calling ``Effect/cancellable(id:cancelInFlight:)-17skv`` with a static type
+  /// as the effect's unique identifier.
+  ///
+  /// - Parameters:
+  ///   - id: A unique type identifying the effect.
+  ///   - cancelInFlight: Determines if any in-flight effect with the same identifier should be
+  ///     canceled before starting this new one.
+  /// - Returns: A new effect that is capable of being canceled by an identifier.
+  public func cancellable(id: Any.Type, cancelInFlight: Bool = false) -> Effect {
+    self.cancellable(id: ObjectIdentifier(id), cancelInFlight: cancelInFlight)
+  }
+
   /// An effect that will cancel any currently in-flight effect with the given identifier.
   ///
   /// - Parameter id: An effect identifier.
   /// - Returns: A new effect that will cancel any currently in-flight effect with the given
   ///   identifier.
   public static func cancel(id: AnyHashable) -> Effect {
-    return .fireAndForget {
+    .fireAndForget {
       cancellablesLock.sync {
         cancellationCancellables[.init(id: id)]?.forEach { $0.cancel() }
       }
     }
+  }
+
+  /// An effect that will cancel any currently in-flight effect with the given identifier.
+  ///
+  /// A convenience for calling ``Effect/cancel(id:)-iun1`` with a static type as the effect's
+  /// unique identifier.
+  ///
+  /// - Parameter id: A unique type identifying the effect.
+  /// - Returns: A new effect that will cancel any currently in-flight effect with the given
+  ///   identifier.
+  public static func cancel(id: Any.Type) -> Effect {
+    .cancel(id: ObjectIdentifier(id))
   }
 
   /// An effect that will cancel multiple currently in-flight effects with the given identifiers.
@@ -91,6 +118,18 @@ extension Effect {
   /// - Returns: A new effect that will cancel any currently in-flight effects with the given
   ///   identifiers.
   public static func cancel(ids: [AnyHashable]) -> Effect {
+    .merge(ids.map(Effect.cancel(id:)))
+  }
+
+  /// An effect that will cancel multiple currently in-flight effects with the given identifiers.
+  ///
+  /// A convenience for calling ``Effect/cancel(ids:)-dmwy`` with a static type as the effect's
+  /// unique identifier.
+  ///
+  /// - Parameter ids: An array of unique types identifying the effects.
+  /// - Returns: A new effect that will cancel any currently in-flight effects with the given
+  ///   identifiers.
+  public static func cancel(ids: [Any.Type]) -> Effect {
     .merge(ids.map(Effect.cancel(id:)))
   }
 }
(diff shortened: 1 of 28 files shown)


Sources/ComposableArchitecture/Effects/Cancellation.swift
```diff
@@ -93,6 +120,18 @@ extension Effect {
   public static func cancel(ids: [AnyHashable]) -> Effect {
```
stephencelis, line 120: The 2 array overloads that take `ids: [AnyHashable]` or `ids: [Any.Type]` are the main strange thing about this PR, but in practice I imagine people adopt one style or the other wherever `cancel(ids:)` is used...

tgrapperon: The new `enum` approach is appealing, but you also sometimes need parametric identifiers:
```swift
struct MyEffectIdentifier: Hashable { 
  var id: String
 }
```
I suspect that many users will switch to the leaner enum way in general and fall back to the "instance" way when needed. I didn't check thoroughly the PR yet: is using both styles in the same project problematic (aside from overload resolution at build time)?

stephencelis: @tgrapperon You should be able to reach for whichever suits you for each effect! This is mainly a convenience around `.cancel(id: ObjectIdentifier(Token.self))` that flattens things.

tgrapperon: @stephencelis Ok, so the overload is merely an oddity. I don't think this will cause additive cognitive issues for the user, as the current `.cancel` method is already accepting almost everything you throw at it, and accepting any type is merely a prolongation of this.
And for the compiler, their arguments can never be ambiguous, so it should not really cause any additional strain.

This is a great addition to the library! It doesn't solve the potential collisions in documents-based apps (I think this would require protocolized reducers to provide a convenient API), but it already streamlines identifiers declaration a lot!

mbrandonw (approved):

Sources/ComposableArchitecture/Effects/Cancellation.swift
iampatbrown: @tgrapperon This change actually lends itself to accommodating document-based apps. I've used something similar to `CancelToken` that included a stack of reducer ids `var reducerIds: [AnyHashable]`. I have a feeling when I did some benchmarks it worked out more performant to use an array instead of a struct. ie: `let id = [id] + reducerIds`. It's been a while since I've used it though.

Here's an implementation I found that used an array but It works the same with a `CancelToken`

```swift
extension Reducer {
  public func id<ID: Hashable>(_ id: ID) -> Self {
    Self { state, action, environment in
      withReducerId(id) { self.run(&state, action, environment) }
    }
  }
  
  public func id<ID: Hashable>(_ toId: @escaping (State) -> ID) -> Self {
    Self { state, action, environment in
      withReducerId(toId(state)) { self.run(&state, action, environment) }
    }
  }
}

extension Hashable where Self == AnyHashable {
  public static func withReducerId<ID: Hashable>(_ id: ID) -> AnyHashable {
    reducerIdsLock.sync { reducerIds + [id] }
  }
}

private func withReducerId<ID: Hashable, Result>(_ id: ID, _ body: () -> Result) -> Result {
  reducerIdsLock.sync {
    reducerIds.append(id)
    defer { reducerIds.removeLast() }
    return body()
  }
}

private var reducerIds: [AnyHashable] = ["co.pointfree.ComposableArchitecture.ReducerId"]
private let reducerIdsLock = NSRecursiveLock()
```

Looking at it now this one was a standalone extension. I also had another one that was more like this PR.

tgrapperon: @iampatbrown Interesting. If I'm getting it right, it looks similar to what I'm doing in [`composable-effect-identifier`](https://github.com/tgrapperon/composable-effect-identifier)(CEI) where [`EffectNamespace`](https://github.com/tgrapperon/composable-effect-identifier/blob/d58c6b8082151fca4cf9a744e9f308ff0ec42ceb/Sources/ComposableEffectIdentifier/EffectID.swift#L275)  is playing the role of `reducerIds`. In CEI, the `EffectID` is decorated when it is accessed and sent through `.cancel(id:)`, whereas I imagine that in your case, it is decorated "on the other side", when the `CancelToken` is formed. My approach requires the user to use some `EffectID` values (they have other benefits), whereas yours seems indeed to be able to work with any type.

With the protocolized version I'm foreseeing, the chain would be injected and stored as field values when reducers are pulled back or created. This should allow foregoing locking, and this is what I had in mind in my comment.

In CEI, I'm using a specific, lock-free route for main thread reducers: Thanks to `Thread.isMainThread` being very fast to check, if it's `true`, I'm writing directly to a dedicated main-thread container without any precaution. If it returns `false`, I'm using locking on a separate non-main-tread container. Maybe pairing this with your approach would allow solving (or at least improving) the collision issue in an efficient and convenient (if not transparent) way for the user[^1].

[^1]: I have not experimented a lot with `async-await` with regard to this issue (I only know that `@TaskLocal` and `Actors` are less efficient than locking). As we are invited to migrate away from threads, I don't know if `Thread.isMainThread` is something that we should use, or if any modern counterpart exists with `MainActor` for example.

This does not only concern document-based apps by the way. Having segregated identifiers greatly improves reusability, composition, and separation of concerns (as presented in the Example of CEI, where rows of a list are isolated features unaware of the existence of their siblings, and implemented without using any `rowID` for their effects).

iampatbrown: > This does not only concern document-based apps by the way

My use case was for reducer reusability. I actually used your examples in CEI as test cases to make sure it worked :) 
For rows I would use `rowReducer.id(\.id).forEach(...)`

tgrapperon: The main issue with this design is that it makes it difficult to cancel from the parent some effect a child reducer has produced. If the child reducer is still active, you can probably delegate this to it, but otherwise, you need to reconstruct the identifiers chain manually, which is something quite inconvenient for the user, and probably very difficult in some cases.

One way to achieve this would be to have some kind of query API for CancelTokens to retrieve the ones you want from the cancellation pool. 

As some `proto` branch is moving, another solution would be to have something like SwiftUI `Preference`, where child reducers can inform parents of their existence or other parameters. In the parent, you can then work on a subset of effects produced by their children. There are many performance concerns and implementation subtleties[^1], but it can open a world of opt-in delegate-style callbacks for features, which is something not yet formalized in TCA.

[^1]: For example, you need to account for reducers that are gone, but with effects still in flight. In this case, you should probably have tombstones, or even more evolved data types like CRDT or history trackers. This suggests that "effective" reducers types (the ones that are hosting dependencies and reducing the actions to produce effects, like `UI/NSView[Controller]` in SwiftUI) should probably be different from "declarative" reducers (which are more akin to SwiftUI `View`s). Such a separation would also probably allow optimization routes in their "effective" setup: You have a declarative` Reducer` setup, and some internal builder/factory is constructing an equivalent "effective" `Reducer` that performs the work.
But I'll stop there, as this is mostly speculative, and maybe not possible at all.

iampatbrown: > it makes it difficult to cancel from the parent 

I'd normally use a `.presents` extension on reducer for this.  

```swift
extension Reducer {
  public func presents<LocalState, LocalAction, LocalEnvironment, ID>(
    forEach localReducer: Reducer<LocalState, LocalAction, LocalEnvironment>,
    state toLocalState: WritableKeyPath<State, IdentifiedArray<ID, LocalState>>,
    action toLocalAction: CasePath<Action, (ID, LocalAction)>,
    hasCancellableEffects: @escaping (LocalAction) -> Bool,
    environment toLocalEnvironment: @escaping (Environment) -> LocalEnvironment
  ) -> Self { 
     ...
  }
}  
```

With a `CasePath` overload as well, which ends up looking something like this:

```swift
let listReducer = Reducer<ListState, ListAction, Void> { state, action, _ in
  ...
}.presents(
  forEach: rowReducer, // or rowReducer.id(\.id) if needed
  state: \ListState.rows,
  action: /ListAction.row,
  hasCancellableEffects: /RowAction.onAppear,
  environment: {}
)
```

This is probably a little off topic from the PR now though :P Maybe we could revisit this in a discussion at a later time. Looks like there might be some changes coming up that would impact all of this.

PR #1082, opened 2022-05-05
add `timeout` that can be cancel

ytyubox:
I see some Effects with time, really think they are great! Just want to contribute timeout.

Big thanks for share this project!

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Timeouting.swift
+++ b/Sources/ComposableArchitecture/Effects/Timeouting.swift
@@ -1,0 +1,39 @@
+import Combine
+
+extension Effect {
+    /// Turns an effect into one that can be timeouted.
+    ///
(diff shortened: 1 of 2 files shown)


mbrandonw:
Hi @ytyubox, thanks for taking the time to share this with us.

We don't think this operator necessarily needs to be added to the library since it is easily expressed using existing operators, in particular `.timeout` from Combine and `.cancellable` from TCA.

The only time-based operator that ships with TCA is `debounce`, but it isn't simply a re-definition of Combine's debounce. It debounces with respect to actions sent rather than debouncing the emissions of the effect. Implementing this kind of effect cannot be done simply with existing operators, and so that is why we ship the operator.

So, I'm going to close this for now.

ytyubox:
Make sense. Thanks!
 
https://github.com/pointfreeco/swift-composable-architecture/discussions/952#discussioncomment-1925713

PR #1083, opened 2022-05-06
Async fire-and-forget

mbrandonw:
This has come up before (#868), and we think it will be useful, especially with some new features coming soon to TCA that make it play nicer with Swift's concurrency tools.


stephencelis (approved):

iampatbrown:
@mbrandonw What are your thoughts on including the priority parameter as well?

mbrandonw:
@iampatbrown  Oh good point, will add that too!

PR #1088, opened 2022-05-10
Ignore errors thrown in fireAndForget.

mbrandonw:


Sources/ComposableArchitecture/Effects/Concurrency.swift
```diff
@@ -104,7 +104,7 @@ import SwiftUI
     }
 
     /// Creates an effect that executes some work in the real world that doesn't need to feed data
-    /// back into the store.
+    /// back into the store and ignores any errors.
```
stephencelis, line 107:
```suggestion
    /// back into the store. Like a throwing `Task`, it will terminate with the first error.
```

stephencelis (approved):
Nice looks good! One small suggest, but take it or leave it.

Sources/ComposableArchitecture/Effects/Concurrency.swift
```diff
@@ -104,7 +104,7 @@ import SwiftUI
     }
 
     /// Creates an effect that executes some work in the real world that doesn't need to feed data
-    /// back into the store.
+    /// back into the store. Like a throwing `Task`, it will terminate with the first error.
```
mbrandonw, line 107:
```suggestion
    /// back into the store. If an error is thrown, the effect will complete and the error will be ignored.
```

PR #1089, opened 2022-05-12
Swift 5.7 fixes

stephencelis:


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-FocusState.swift
```diff
@@ -75,7 +75,7 @@
   }
 
   extension View {
-    func synchronize<Value: Equatable>(
```
stephencelis, line 78: These conformances are redundant and Swift 5.7 is louder about letting us know!

Sources/ComposableArchitecture/ViewStore.swift
```diff
@@ -436,6 +436,7 @@ private struct HashableWrapper<Value>: Hashable {
     ///   - action: An action.
     ///   - predicate: A predicate on `State` that determines for how long this method should
     ///     suspend.
+    @MainActor
```
stephencelis, line 439: Swift 5.7 no longer guarantees that this method runs on the main actor, but this is required by SwiftUI.

mbrandonw (approved):

PR #1094, opened 2022-05-17
Add SPI docc config

finestructure:
This adds settings to opt into DocC generation and hosting on SPI

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -5,9 +5,8 @@ builder:
     scheme: ComposableArchitecture
   - platform: macos-xcodebuild
     scheme: ComposableArchitecture
-  - platform: macos-xcodebuild-arm
-    scheme: ComposableArchitecture
   - platform: tvos
     scheme: ComposableArchitecture
   - platform: watchos
     scheme: ComposableArchitecture_watchOS
+  - documentation_targets: [ComposableArchitecture]
--- a/Package.swift
+++ b/Package.swift
@@ -50,3 +50,11 @@ let package = Package(
     ),
   ]
 )
+
+
+#if swift(>=5.6)
+// Add the documentation compiler plugin if possible
+package.dependencies.append(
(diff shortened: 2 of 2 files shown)


.spi.yml
```diff
@@ -5,9 +5,8 @@ builder:
     scheme: ComposableArchitecture
   - platform: macos-xcodebuild
     scheme: ComposableArchitecture
-  - platform: macos-xcodebuild-arm
-    scheme: ComposableArchitecture
```
finestructure, line 9: Unrelated but while I was at it I removed this - we're running all builds on ARM now and the `macos-xcodebuild-arm` platform variant doesn't exist anymore.

.spi.yml
```diff
@@ -5,9 +5,8 @@ builder:
...
   - platform: tvos
     scheme: ComposableArchitecture
   - platform: watchos
     scheme: ComposableArchitecture_watchOS
+  - documentation_targets: [ComposableArchitecture]
```
finestructure, line 12: This sets the list of documentation targets for all configs (but we're only actually building docs for `macos-spm` / `5.6` unless an explicit platform and/or Swift version is configured.)

finestructure, line 12: This sets the list of documentation targets for all configs (but we're only actually building docs for `macos-spm` / `5.6` unless an explicit platform and/or Swift version is configured.)

stephencelis:
@finestructure Sounds good! Just ping us when it's ready.

finestructure:
We should be good to go now!

It'll take 24h after merging into `main` for it to be picked up by our build system (which also generates the docs), because we're coalescing default branch revisions.

NB: This will be exposed on staging, I'll let you know once is up!

stephencelis (approved):
Sounds good! Thanks @finestructure!

finestructure:
> It'll take 24h after merging into `main` for it to be picked up by our build system (which also generates the docs), because we're coalescing default branch revisions.

FYI I nudged this along ahead of the build cycle

PR #1101, opened 2022-05-19
Add throttle overloads that take types

elkraneo:
Given that Effect cancellation now has endpoints that can take types as identifiers and those were extended to debounce, it seems natural to expose the same functionality for throttle.


stephencelis (approved):
Must've missed this one! Thanks!

PR #1107, opened 2022-05-24
Cleanup

stephencelis:
A few small things.

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -317,8 +317,7 @@ extension Publisher {
 
   /// Turns any publisher into an ``Effect``.
   ///
-  /// This is a convenience operator for writing ``Effect/eraseToEffect()`` followed by a
-  /// ``Effect/map(_:)``.
+  /// This is a convenience operator for writing ``Effect/eraseToEffect()`` followed by `map`.
   ///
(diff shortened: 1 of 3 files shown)


Sources/ComposableArchitecture/Effect.swift
```diff
@@ -317,8 +317,7 @@ extension Publisher {
 
   /// Turns any publisher into an ``Effect``.
   ///
-  /// This is a convenience operator for writing ``Effect/eraseToEffect()`` followed by a
-  /// ``Effect/map(_:)``.
+  /// This is a convenience operator for writing ``Effect/eraseToEffect()`` followed by `map`.
```
stephencelis, line 320: This `map` doesn't link correctly and causes doc warnings.

Sources/ComposableArchitecture/TestSupport/TestStore.swift
```diff
@@ -177,7 +176,7 @@
     private var inFlightEffects: Set<LongLivingEffect> = []
     var receivedActions: [(action: Action, state: State)] = []
     private let reducer: Reducer<State, Action, Environment>
-    private var snapshotState: State
```
stephencelis, line 180: You mentioned `state` reads better and I agree!

mbrandonw (approved):

PR #1110, opened 2022-05-26
Update Search Demo to use Open-Meteo

stephencelis:
[MetaWeather](https://www.metaweather.com) has been offline for awhile now, with no sign of return or status page/account that I could find. [Open-Meteo](https://open-meteo.com/en) appears to have a more active presence, so let's rely on it for now, instead.


stephencelis:
I think this demo could be polished up a bit more before merging. Might give things another look tomorrow.

mbrandonw (approved):

PR #1113, opened 2022-05-30
Conform `WithViewStore` to `Commands`

tgrapperon:
Following #1112, this makes `WithViewStore` conform to `Commands` when appropriate. This should allow to bundle `Commands` in their own type like Apple does in their [`Landmarks`](https://developer.apple.com/tutorials/swiftui) tutorials, and still have access to `WithViewStore`. This brings up two additional questions:
- Should we review SwiftUI for similar situations, like `ToolbarContent` for example, and conform `WithViewStore` for them too?
- Should we reimplement Apple SwiftUI tutorial apps using TCA (it's maybe already done somewhere else)?

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/WithViewStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/WithViewStore.swift
@@ -153,7 +153,7 @@ extension WithViewStore: DynamicViewContent where State: Collection, Content: Dy
 @available(iOS 14, macOS 11, tvOS 14, watchOS 7, *)
 extension WithViewStore: Scene where Content: Scene {
   /// Initializes a structure that transforms a store into an observable view store in order to
-  /// compute views from store state.
+  /// compute scenes from store state.
   ///
   /// - Parameters:
   ///   - store: A store.
--- a/Sources/ComposableArchitecture/SwiftUI/WithViewStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/WithViewStore.swift
@@ -216,3 +216,75 @@ extension WithViewStore where State == Void, Content: Scene {
     self.init(store, removeDuplicates: ==, file: file, line: line, content: content)
   }
 }
+
+@available(iOS 14.0, macOS 11.0, *)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+extension WithViewStore: Commands where Content: Commands {
+  /// Initializes a structure that transforms a store into an observable view store in order to
+  /// compute commands from store state.
+  ///
+  /// - Parameters:
+  ///   - store: A store.
+  ///   - isDuplicate: A function to determine when two `State` values are equal. When values are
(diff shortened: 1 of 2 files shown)


stephencelis (approved):

stephencelis:
Thanks @tgrapperon! To answer your questions:

> Should we review SwiftUI for similar situations, like `ToolbarContent` for example, and conform `WithViewStore` for them too?

We think so. As long as we embrace `WithViewStore` as a construct it should ideally work wherever folks expect it to work.

The proliferation of overloads is unfortunate but I can't see a way around it given the `Content` conditional conformance, and some SwiftUI APIs must see the same problem.

Want to update this PR with others that you find? Or would you like me to merge as-is?

> Should we reimplement Apple SwiftUI tutorial apps using TCA (it's maybe already done somewhere else)?

That'd be fun! We've discussed it in the past but haven't gotten around to it. The Scrumdinger app in particular would be a good example.

tgrapperon:
> Want to update this PR with others that you find? Or would you like me to merge as-is?

I'll try to update the PR later today or tomorrow. I'll ping you when it's ready.

iampatbrown:
@tgrapperon Nice work. Initially I was worried about the overloads. But it doesn't seem too bad. `TableColumnContent` and `TableRowContent` might be better suited for `ForEachStore`

> Should we reimplement Apple SwiftUI tutorial apps using TCA

Here's a version of [Scrumdinger](https://github.com/iampatbrown/ScrumdingerTCA) I put together a while a go. Feel free to take anything that's useful.

tgrapperon:
@iampatbrown, I agree about `WithViewStore` & `Table`. We furthermore already have built-in witnesses for `TableRowContent` and `TableColumnContent`, so we will very unlikely encounter a situation where we will need our own conforming type.

I was trying to make them work simply because they exist, without assuming any use case.  Assuming `WithViewStore` conformance is operational, I can imagine using `WithViewStore`'s to conditionally define `TableColumns` or `TableRows`. But in both case, we can achieve the same result with an outer `WithViewStore` exposing only the relevant properties.

Congrats about the Scrumdinger port! I'll look at it out of curiosity. It should probably be either linked or embedded into TCA. Many folks are considering Apple's SwiftUI samples as canon, and it can be very interesting for newcomers (or more experienced users) to compare the differences with TCA side by side. I wonder if there is a way to ease the comparison. I'll think about it!

iampatbrown:
> I wonder if there is a way to ease the comparison.

I think levering DocC Tutorials is probably a good start. Being able to see how a project evolves can be pretty useful. Then the Apple and TCA versions can also be compared in increments rather than as a whole.

tgrapperon:
@mbrandonw Thanks, I thought autocomplete was using alphabetical ordering.

stephencelis (approved):
Thanks @tgrapperon!

tgrapperon:
There are three more potential candidates from Swift Charts. I'll add them in a new PR once they're out of beta.

PR #1122, opened 2022-06-02
Make TestStore's state visible

tgrapperon:
This should allow people to perform assertions outside of the send/check sequence.
This follows https://twitter.com/alexito4/status/1532380971567308800

### DIFF
--- a/Sources/ComposableArchitecture/TestSupport/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestSupport/TestStore.swift
@@ -176,7 +176,7 @@
     private var inFlightEffects: Set<LongLivingEffect> = []
     var receivedActions: [(action: Action, state: State)] = []
     private let reducer: Reducer<State, Action, Environment>
-    private var state: State
+    public private(set) var state: State
     private var store: Store<State, TestAction>!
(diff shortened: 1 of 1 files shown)


jshier:
Personally, I dislike any test structure that puts assertions out of line with the main test execution. I consider TCA's mutation checkers to themselves be assertions, so adding additional assertions conflates requirements, sort of like nesting assertions directly. To me this:

```swift
testStore.receive(...) {
  $0.value = newValue
}
XCTAssertEqual(testStore.state.otherValue, someOtherValue)
```

is cleaner and clearer than

```swift
testStore.receive(...) {
  $0.value = newValue
  XCTAssertEqual($0.otherValue, someOtherValue)
}
```

as it properly separates my assertions about mutations and other assertions. Technically, yes, they do the same thing, but clarity is paramount in tests. Plus it just seems wrong to assert against the locally mutated value instead of the actual underlying state.

mbrandonw:
I've started a discussion here https://github.com/pointfreeco/swift-composable-architecture/discussions/1124! Let's move over there so that we don't have a long, interesting discussion trapped in a closed PR.

PR #1126, opened 2022-06-03
Replace relative links with absolute links in README

tockrock:
I've noticed that most (but not all) of the links in the README to different sections of the repo were relative ones.

I wanted to suggest changing all the links to absolute links.
This would help future translation efforts by allowing the links to work in external gists without modification.

Hope this helps!

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,11 +51,11 @@ You can watch all of the episodes [here](https://www.pointfree.co/collections/co
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


stephencelis:
@tockrock The reason we prefer relative URLs in the repo is so that they work across branches and historic commits. If we used absolute URLs, non-`main`-branch links will always point to `main`, to something that may no longer exist (or exist yet!).

Sorry that this complicates translation efforts, but hope you understand!

tockrock:
If you have other priorities, I think it's totally reasonable! Thanks for the explanation.
I just wanted to send a PR incase there wasn't a strong one.

Closing this PR.

PR #1129, opened 2022-06-03
Test Store documentation updates

stephencelis:
Just a few small things I noticed.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/Deprecations.swift
+++ b/Sources/ComposableArchitecture/Internal/Deprecations.swift
@@ -130,10 +130,10 @@ extension Reducer {
       func assert(step: Step) {
         switch step.type {
         case let .send(action, update):
-          self.send(action, file: step.file, line: step.line, update)
+          self.send(action, update, file: step.file, line: step.line)
 
         case let .receive(expectedAction, update):
-          self.receive(expectedAction, file: step.file, line: step.line, update)
+          self.receive(expectedAction, update, file: step.file, line: step.line)
 
         case let .environment(work):
           if !self.receivedActions.isEmpty {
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -130,10 +130,10 @@ extension Reducer {
       func assert(step: Step) {
         switch step.type {
         case let .send(action, update):
-          self.send(action, file: step.file, line: step.line, update)
+          self.send(action, update, file: step.file, line: step.line)
```
stephencelis, line 133: Trailing closures used to not work when default arguments came after them, but that was fixed a long time ago, so I think it should be mostly safe to reorder them. I doubt many folks depend on `file:` and `line:` externally, and those that do should get a handy fix-it whenever they upgrade.

mbrandonw (approved):

Sources/ComposableArchitecture/TestSupport/TestStore.swift
```diff
@@ -335,11 +344,19 @@
...
+    ///     store. The mutable state sent to this closure must be modified to match the state of the
+    ///     store after processing the given action. Do not provide a closure if no change is
+    ///     expected.
     public func send(
       _ action: LocalAction,
+      _ updateExpectingResult: ((inout LocalState) throws -> Void)? = nil,
       file: StaticString = #file,
-      line: UInt = #line,
-      _ update: ((inout LocalState) throws -> Void)? = nil
+      line: UInt = #line
```
jeffersonsetiawan, line 359: Just wondering about this change, is there any specific reason you flip the order of this parameter @stephencelis ?

In my project, there is a need to provide the file and line.
Previous code looks like this:
```swift
store.send(.x, file: file, line: line) { ... }
```

but after this change, the trailing closure won't work and need to do this instead.
```swift
store.send(.x, { .... }, file: file, line: line)
```

stephencelis: @jeffersonsetiawan Sorry about that! We didn't envision folks leveraging the `file:line:` defaults. The original order is an artifact of Swift not supporting trailing closure syntax in this case, but that was fixed a few versions back, so we decided to update things.

PR #1132, opened 2022-06-04
Fix logic for new 'expected state to change' failure.

mbrandonw:
While messing with some stuff today I realized that the new "expected state to change" failures from #1091 can lead to some confusing failures. For example, this test:

```swift
func testBasics() {
  let store = TestStore(
    initialState: 0,
    reducer: Reducer<Int, Void, Void> { state, _, _ in state += 1; return .none },
    environment: ()
  )

  store.send(()) { _ = $0 }
}
```

gives two failures:

> 🛑 A state change does not match expectation: …
>
>    − 0
>    \+ 1
>
> (Expected: −, Actual: +)

> 🛑 Expected state to change, but no change occurred.
>
> The trailing closure made no observable modifications to state. If no change to state is expected, omit the trailing closure.

Both are technically true since my closure did not mutate the state to the correct value _and_ my closure didn't actually make any changes.

However, I feel if the state change assertion fails then we should probably skip the "expected state to change".

What do you think @rcarver and @stephencelis?

Further, this stuff was confusing me enough that I realized it would be really nice to have dedicated tests for our error messaging. This helped me figure out things more easily, and I think Stephen is working on some updates to the copy too so it would be nice to have test coverage on that.


Tests/ComposableArchitectureTests/TestStoreFailureTests.swift
```diff
@@ -0,0 +1,212 @@
...
+    } issueMatcher: {
+      $0.compactDescription == """
+        A state change does not match expectation: …
+
+            − TestStoreFailureTests.State(count: 0)
+            + TestStoreFailureTests.State(count: 1)
+
+        (Expected: −, Actual: +)
+        """
+    }
```
mbrandonw, line 60: This test would not pass without the changes I made to `TestStore` because it would generate 2 failures, and so the `issueMatch` would not be correct.

rcarver:
Ah interesting! While both error messages are technically accurate, agreed that it can be reduced to more important error which will resolve both issues. Sorry I missed covering this case!

stephencelis (approved):

PR #1134, opened 2022-06-06
Add animation operator to Effect.

mbrandonw:
This was brought up a long time ago by @mayoff in the Swift [forums](https://forums.swift.org/t/an-alternative-take-on-animating-asynchronous-effects/45000), but at the time we didn't think it was necessary since scheduler animations seemed to cover all its use cases. However, with some of the concurrency updates we will be bringing to the library soon this kind of operator is going to be handy when used with `Effect.task`.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Animation.swift
+++ b/Sources/ComposableArchitecture/Effects/Animation.swift
@@ -1,7 +1,7 @@
 import Combine
 import SwiftUI
 
-extension Publisher {
+extension Effect {
   /// Wraps the emission of each element with SwiftUI's `withAnimation`.
   ///
   /// This publisher is most useful when using with ``Effect/task(priority:operation:)-2czg0``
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/TestSupport/TestStore.swift
```diff
@@ -176,8 +176,8 @@
 
     /// The current state.
     ///
-    /// When read from a trailing closure assertion in ``send`` or ``receive``, it will equal the
-    /// `inout` state passed to the closure.
+    /// When read from a trailing closure assertion in ``send(_:_:file:line:)`` or
+    /// ``receive(_:_:file:line:)``, it will equal the `inout` state passed to the closure.
```
mbrandonw, line 180: unrelated docc warning fixes.

Sources/ComposableArchitecture/Effects/Animation.swift
```diff
@@ -0,0 +1,64 @@
+import Combine
+import SwiftUI
+
+extension Publisher {
```
stephencelis, line 4: Given long-term thoughts on the `Effect` type, should we define this directly on `Effect` instead, or should we pull it out as a Combine helper and define it in `CombineSchedulers`, instead (though it admittedly isn't a scheduler-specific operation)?

stephencelis (approved):
Could merge as is, but if you agree it should be defined on `Effect`, could take that change.

Issue #1138, opened 2022-06-10
[Behavior] Combine Publisher Testable concurrency when observed in different places


fila95:
**Describe the bug**
Not sure it properly a bug, just a behaviour that might need to be fixed. 
Setup:
- There is a publisher stored in the main environment which is passed to 3 features (for reference let's call them 1,2,3).
- All these features start observing its values when an initial action is sent
- The order of which such actions are called is random

This seems to be something relatable to Combine which does not send values to subscribers in-order. When it comes to testing though (in this case writing an integration test), it becomes really hard to use the `TestStore` to capture exactly what happened, leading to many fails each different from one execution to another.

**To Reproduce**
I went ahead and prepared an Xcode project to demonstrate the issue. Just look at the test suite:
- There is one file called `AppStack` which contains the logic
- The other contains the test

Just run the test 3-4 times and see the randomness ;) 

[SampleIssue.zip](https://github.com/pointfreeco/swift-composable-architecture/files/8877636/SampleIssue.zip)

**Expected behavior**
The expectation is that the actions are called in-order of subscription especially since they are bound to the same scheduler. This I fear is not controllable so I was thinking about this possible solution:
Updating the `TestStore` providing an additional `receive` action (or editing the existing one) which does not check only the first received action but compared the one expected with the whole in-flight-actions bucket

**Screenshots**
N/A

**Environment**
  - swift-composable-architecture version 0.34.0-0.36.0 for sure
  - Xcode All 13 versions
  - Swift 5.6
  - OS: iOS 13, 14, 15

**Additional context**
I'd love to hear your thoughts and I'll be more than happy to help with a fix for this ;)

fila95:
Yeah definitely, thanks for taking some time to assess the issue ;)

PR #1139, opened 2022-06-10
[Feature] Let `TestStore` support unordered actions

fila95:
Tentative resolution for: https://github.com/pointfreeco/swift-composable-architecture/issues/1138
The issue with this approach is that if an reducer modifies the same slices of state we can't unsure correctness..
Do you have any other suggestion? 
This is a proof of concept, tests will be added later ;)

### DIFF
--- a/Sources/ComposableArchitecture/TestSupport/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestSupport/TestStore.swift
@@ -503,6 +503,64 @@
         self.line = line
       }
     }
+    
+    /// Asserts an action was received from an effect and asserts when state changes.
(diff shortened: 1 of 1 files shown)


stephencelis:
Hey @fila95! Thanks for the very detailed issue, repro, and PR! I used a mirror to examine the insides of a `PassthroughSubject`, and it looks like they store their subscribers in a `Set<ConduitBase>`, so emissions are probably non-deterministic as a result.

We've been actively working on introducing more `async`-friendly APIs to TCA (which introduces plenty of additional non-determinism), and so your PR actually looks a little bit like one of our explorations. We'll be diving into this soon, but your PR looks like a good solution. The main thing I think we'll want to figure out is API name/design bikeshedding. Hope to have time to dive into that soon!

PR #1152, opened 2022-06-18
Show runtime warnings when the stack allocation reaches 95%

tgrapperon:
Following #752, this PR makes `Store` show runtime warnings when the stack depth reaches 95%. There are many things to discuss, so obviously a draft for now.

Right now, each store can present such warnings, but it should maybe be restricted to the root store only.

As runtime warnings are inconvenient to write with their `StaticString`, I'm waiting for feedback to decide if we should produce different messages depending on the nature of the event (when receiving an action, the action itself could be the type responsible for a stack overflow).

I've unified Thread checks and Stack checks under the same banner and renamed `ThreadStatus` into `StoreEvent` accordingly.

Right now, the threshold is at 95%. This is tricky because we don't want false positives, but we also want at least one warning to be emitted before reaching the overflow. In the discussions, I've seen folks reporting 20kB states. When the stack size is 512kB (iOS, background), that's around 5%, so I guess 95% it's a fair value. We can probably go down to 90/92% to have more headroom. We could also adjust automatically according to the stack size if we assume that a 20kB state is already problematic, and emit warnings when we reach around 50kB from the end.

When the warning is emitted, the user can see the current stack status and the store's state size. I've added a few comments about workarounds. Nothing definitive of course, but I think you'll get the idea.

What do you think about this?

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -527,15 +527,15 @@ public final class Store<State, Action> {
       guard self.stackChecksEnabled else { return }
       // Disable in scoped stores
       guard self.parentCancellable == nil else { return }
-      let threshold: Double = 0.95
+      let threshold: UInt = 50_000
       let status = StackStatus()
-      guard status.usedFraction > threshold else { return }
+      guard status.available < threshold else { return }
       
       // We warn only once, but scoped stores could pass through on `init` as their
       // parent cancellable is nil at this point. So we use the `threadDictionary`.
       // This path is only hit once per store lifetime, and only when the stack depth is
       // beyond the threshold.
-      let threadDictionaryKey = "co.pointfree.ComposableArchitecture.stack-checked-warned"
+      let threadDictionaryKey = "co.pointfree.ComposableArchitecture.stack-check-warned"
       guard Thread.current.threadDictionary[threadDictionaryKey] == nil else { return }
       defer {
         self.stackChecksEnabled = false
@@ -568,7 +568,8 @@ public final class Store<State, Action> {
         #advice-use-copy-on-write-semantics-for-large-values" for example). If the property is an \
         "enum", you can simply mark it as "indirect", producing the same effect.
         
-        "Array", "Set", "Dictionary" or "IdentifiedArray" are already storing their content on the heap.
+        "Array", "Set", "Dictionary" or "IdentifiedArray" are already storing their content on the \
+        heap.
         """,
         [
           String(format: "%.0f%%", status.usedFraction * 100),
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -426,105 +426,104 @@ public final class Store<State, Action> {
     #endif
   }
 
-  @inline(__always)
-  private func threadCheck(event: Event) {
-    #if DEBUG
-      guard self.mainThreadChecksEnabled && !Thread.isMainThread
-      else { return }
-
-      switch event {
-      case let .effectCompletion(action):
-        runtimeWarning(
-          """
-          An effect completed on a non-main thread. …
-
-            Effect returned from:
-              %@
-
-          Make sure to use ".receive(on:)" on any effects that execute on background threads to \
-          receive their output on the main thread, or create your store via "Store.unchecked" to \
-          opt out of the main thread checker.
-
-          The "Store" class is not thread-safe, and so all interactions with an instance of \
-          "Store" (including all of its scopes and derived view stores) must be done on the same \
-          thread.
-          """,
-          [debugCaseOutput(action)]
-        )
-
-      case .`init`:
-        runtimeWarning(
-          """
-          A store initialized on a non-main thread. …
-
-          If a store is intended to be used on a background thread, create it via \
-          "Store.unchecked" to opt out of the main thread checker.
-
-          The "Store" class is not thread-safe, and so all interactions with an instance of \
-          "Store" (including all of its scopes and derived view stores) must be done on the same \
-          thread.
-          """
-        )
-
-      case .scope:
-        runtimeWarning(
-          """
-          "Store.scope" was called on a non-main thread. …
-
-          Make sure to use "Store.scope" on the main thread, or create your store via \
-          "Store.unchecked" to opt out of the main thread checker.
-
-          The "Store" class is not thread-safe, and so all interactions with an instance of \
-          "Store" (including all of its scopes and derived view stores) must be done on the same \
-          thread.
-          """
-        )
-
-      case let .send(action, originatingAction: nil):
-        runtimeWarning(
-          """
-          "ViewStore.send" was called on a non-main thread with: %@ …
-
-          Make sure that "ViewStore.send" is always called on the main thread, or create your \
-          store via "Store.unchecked" to opt out of the main thread checker.
-
-          The "Store" class is not thread-safe, and so all interactions with an instance of \
-          "Store" (including all of its scopes and derived view stores) must be done on the same \
-          thread.
-          """,
-          [debugCaseOutput(action)]
-        )
-
-      case let .send(action, originatingAction: .some(originatingAction)):
-        runtimeWarning(
-          """
-          An effect published an action on a non-main thread. …
-
-            Effect published:
-              %@
-
-            Effect returned from:
-              %@
-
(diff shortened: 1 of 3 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -408,20 +411,28 @@ public final class Store<State, Action> {
...
+  
+  @inline(__always)
+  private func performChecks(event: Event) {
+    #if DEBUG
+    threadCheck(event: event)
+    stackCheck(event: event)
+    #endif
+  }
 
   @inline(__always)
```
tgrapperon, line 429: This could probably be removed if we make everything transit through `performChecks`

Sources/ComposableArchitecture/Store.swift
```diff
@@ -179,7 +181,8 @@ public final class Store<State, Action> {
       initialState: initialState,
       reducer: reducer,
       environment: environment,
-      mainThreadChecksEnabled: false
+      mainThreadChecksEnabled: false,
+      stackChecksEnabled: true
```
tgrapperon, line 185: Because this initializer is mostly used for background stores for which the stack size is smaller, we probably want stack checks to be enabled.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -327,7 +330,7 @@ public final class Store<State, Action> {
     state toLocalState: @escaping (State) -> LocalState,
     action fromLocalAction: @escaping (LocalAction) -> Action
   ) -> Store<LocalState, LocalAction> {
-    self.threadCheck(status: .scope)
+    self.performChecks(event: .scope)
     var isSending = false
     let localStore = Store<LocalState, LocalAction>(
```
tgrapperon, line 335: If we only want the root store to emit, we can flip `stackChecksEnabled` to `false` for these instances.

tgrapperon:
The CI failure was weird. I've only lowered the trigger to 85% and it failed on 13.2.1. Could runtime warnings make tests fail? In this case, it could probably mean that we reached 85% in the failing tests on 13.2.1, which would be a surprise! Or it's some unrelated flakiness.

tgrapperon:
After thinking a little more about it, I now think it should only warn once per thread's lifetime. I don't see the point in warning the user more than this, as all warnings are likely sharing the same causes.
As a consequence, I also think that we should not warn in scoped store: we may report a totally acceptable leaf state size, and potentially miss a larger one at the root. The root store will always participate, so it will likely spot the incoming overflow anyway.
I'll push another commit in this spirit.

Sources/ComposableArchitecture/Store.swift
tgrapperon: I finally switched it back to `false` in order to be coherent with "unchecked".

Sources/ComposableArchitecture/TestSupport/TestStore.swift
```diff
@@ -235,6 +235,9 @@
         },
         environment: ()
       )
+      #if DEBUG
+      store.stackChecksEnabled = false
```
tgrapperon, line 239: Technically, `TestStore` can still emit a warning when it initializes. The workaround would be to make the `Store.init` that expose checks flags internal instead of private. This would also allow to remove this line.

Sources/ComposableArchitecture/TestSupport/TestStore.swift
```diff
@@ -233,7 +233,9 @@
             .eraseToEffect { .init(origin: .receive($0), file: action.file, line: action.line) }
 
         },
-        environment: ()
+        environment: (),
+        mainThreadChecksEnabled: true,
```
tgrapperon, line 237: I've kept `true` as it was the current value, but it possible that we would want `false`. In this case, we could make `Store.init` private again and use `Store.unchecked` for `TestStore`.
I'm disabling RW for `TestStore` because of the failure I had in CI a few commits above. I didn't try to reproduce it, and it is also probable that disabling them is no needed after all.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -509,20 +519,82 @@ public final class Store<State, Action> {
...
+          "\(status.available)",
+          "\(MemoryLayout<State>.size)",
+          "\(MemoryLayout<State>.size)"
+        ]
+      )
+    }
+  #endif
 
-  private init<Environment>(
+  init<Environment>(
```
tgrapperon, line 585: Made internal in order to be used by `TestStore` and disable Stack checks in order to pass CI on Xcode 13.2.1.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -509,20 +519,82 @@ public final class Store<State, Action> {
...
     initialState: State,
     reducer: Reducer<State, Action, Environment>,
     environment: Environment,
-    mainThreadChecksEnabled: Bool
+    mainThreadChecksEnabled: Bool,
```
tgrapperon, line 589: I'm wondering if this should be abstracted into a type that would store all store configuration or checks flags. Like a `StoreConfiguration` type.

Sources/ComposableArchitecture/TestSupport/TestStore.swift
tgrapperon: Maybe runtime warnings should be promoted into test failures, but I don't know if this is possible without refactoring too much. This is not the purpose of this PR anyway.

tgrapperon:
I can't explain why it fails on CI: Failures are not involving StackChecks, they seem to disappear if I raise the threshold from 85 to 95%, but on my machine, executing the same command on the same Xcode version, they all pass, and they're still passing with a threshold at 15%.

iampatbrown:
@tgrapperon The failure you're getting is:
```
[ComposableArchitectureTests.ReducerTests testDebug] :
Asynchronous wait failed: Exceeded timeout of 2 seconds, with unfulfilled expectations: "logs".
```
Pretty sure `testDebug` is just a bit flaky. I've had it fail a few times before. CI doesn't have as much power as your local machine which is why you're seeing the discrepancy. 
https://github.com/pointfreeco/swift-composable-architecture/blob/9c47990ee5d2e2e979237404e936ee062c9a36e3/Tests/ComposableArchitectureTests/ReducerTests.swift#L130

iampatbrown:
From https://github.com/pointfreeco/swift-composable-architecture/pull/1018#issuecomment-1065983632
> @mbrandonw @stephencelis CI failed at ReducerTests.testDebug which I don't think is impacted by this change.
>
>Asynchronous wait failed: Exceeded timeout of 2 seconds, with unfulfilled expectations: "logs".
[swift-composable-architecture/Tests/ComposableArchitectureTests/ReducerTests.swift](https://github.com/pointfreeco/swift-composable-architecture/blob/d924b9ad27d2a2ccb0ada2639f5255084ff63382/Tests/ComposableArchitectureTests/ReducerTests.swift#L130)
Line 130 in [d924b9a](https://github.com/pointfreeco/swift-composable-architecture/commit/d924b9ad27d2a2ccb0ada2639f5255084ff63382)
 self.wait(for: [logsExpectation], timeout: 2) 
>Think this is similar to https://github.com/pointfreeco/swift-composable-architecture/pull/181

tgrapperon:
@iampatbrown Thanks for the feedback. What's puzzling me is I can probably make this go away by raising the threshold for stack-related warnings from 85% to 95%, as this is the only change I made the first time it happened. This seems completely unrelated, and the threshold is not met in both cases. And the low-level stuff happens before checking it anyway. The only things that follow are the thread dictionary thing (that wasn't implemented when the failure first appeared in this PR), and the RW, which I suspected to be the cause. I'll raise it back to 95% in the meantime.

tgrapperon:
@iampatbrown Also thanks for pointing out the correct test, I was reading its assertion as the error message from the test. It makes a little more sense!

mbrandonw:
Hey @tgrapperon, thanks for exploring this. It's definitely interesting, but we want to confirm that the purple runtime warnings show when running on a device, and not just the simulator. Have you tested that?

tgrapperon:
I've refined the warning message to show the type of the `State` and added "bytes" units.
I'm sure a native speaker could formulate this more idiomatically.
Runtime Warning message

mackoj:
Thanks for taking the time to work on this this look really nice !

tgrapperon:
With the release of the reducer protocol, I feel that there will be less traction for this PR. I'll close it for now and possibly revisit if folks are reporting stack overflows even with protocols.

PR #1153, opened 2022-06-19
Fix deprecated "TestStore.assert" APIs

stephencelis:
The new exhaustive check around state changes was not backported, so some tests could start to fail when upgrading TCA.

@vermont42 discovered this when running the tests from our intro series of episodes: https://github.com/pointfreeco/episode-code-samples/issues/118


mbrandonw (approved):
Dang, good find!

PR #1155, opened 2022-06-22
Move cancellation ids to end of scopes.

mbrandonw:
You used to get warnings for moving ID types to the end of scopes after `return` statements (see https://github.com/apple/swift/issues/56805), but that's not the case anymore. By moving the IDs to the end of the scope we can remove unnecessary details from the top of our reducers.


mbrandonw:
Gonna close this without merging. While interesting, we have some changes coming that somewhat alleviate the need for this.

PR #1156, opened 2022-06-22
Clean up sample code

stephencelis:
We employ a few tricks that might be confusing folks that are unfamiliar with them, so this dials some things down in the sample code and docs. We could maybe take it further, but this is a start.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/IfLetStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/IfLetStore.swift
@@ -10,25 +10,27 @@ import SwiftUI
 /// ```swift
 /// IfLetStore(
 ///   store.scope(state: \SearchState.results, action: SearchAction.results),
-///   then: SearchResultsView.init(store:),
-///   else: { Text("Loading search results...") }
-/// )
+/// ) {
+///   SearchResultsView(store: $0)
+/// } else {
+///   Text("Loading search results...")
+/// }
 /// ```
 ///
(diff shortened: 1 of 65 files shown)


mbrandonw (approved):

Sources/ComposableArchitecture/SwiftUI/IfLetStore.swift
```diff
@@ -10,25 +10,27 @@ import SwiftUI
 /// ```swift
 /// IfLetStore(
 ///   store.scope(state: \SearchState.results, action: SearchAction.results),
-///   then: SearchResultsView.init(store:),
-///   else: { Text("Loading search results...") }
-/// )
+/// ) {
+///   SearchResultsView(store: $0)
+/// } else {
```
gohanlon, line 15: Typo, should be `else:`.

mbrandonw: @gohanlon good eye! fixed here b66fe3d7ea7a4e16f0a0942f15c246fcf11be094

gohanlon:
👍 I usually prefer explicit types rather than implicit `.init`s, unless doing so creates too much verbosity. Having explicit types helps the compiler produce more useful errors when a mistake is made elsewhere, and it also feels like less strain on the compiler.

Aside: I printed out several of the case studies when learning. First thing I did was manually annotate the type of every `.init`. Really made me aware of how much I lean on Xcode's tools to reveal type information that's "buried" when reading unfamiliar code.

PR #1158, opened 2022-06-23
Add automatic binding support

ZevEisenberg:
This is a proof-of-concept that it is possible to make `Reducer.init(_:)` automatically add `.binding()` when the action is a `BindableAction<State>`. I made it after forgetting to add `.binding()` one too many times, and also hearing from friends that they frequently make this mistake, and even go so far as to avoid `BindableAction` entirely because of how easy it is to forget.

It's a proof of concept because it has a major problem: if you use this in existing code with explicit `.binding()`, all your `.binding()` calls will be duplicated. Would love to talk about options to avoid this. Here's one option:

If we decide that automatic binding is good, and we never want `.binding()` to be explicit, we could change `.binding()` to be a deprecated no-op with a suggestion to delete it.

Another option would be to add a private property to `Reducer` called `hasHadBindingApplied` or something like that, and `.binding()` would look for that property and become a no-op if it's true. That option sounds less composable, however, so I don't love it.

Also, I haven't looked at the upcoming protocol-based Reducer work, so I don't know how any of this would translate over.

### DIFF
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -898,3 +898,21 @@ public struct Reducer<State, Action, Environment> {
     self.reducer(&state, action, environment)
   }
 }
+
+// NB: `BindableAction` can produce crashes in Xcode 12.4 (Swift 5.3) and earlier due to an enum
+//     protocol witness bug: https://bugs.swift.org/browse/SR-14041
+#if compiler(>=5.4)
+public extension Reducer where Action: BindableAction, Action.State == State {
+  init(_ theReducer: @escaping (inout State, Action, Environment) -> Effect<Action, Never>) {
+    self.reducer = { state, action, environment in
+      guard let bindingAction = (/Action.binding).extract(from: action)
+      else {
+        return theReducer(&state, action, environment)
+      }
+
+      bindingAction.set(&state)
+      return theReducer(&state, action, environment)
+    }
+  }
+}
+#endif
--- a/Tests/ComposableArchitectureTests/BindingTests.swift
+++ b/Tests/ComposableArchitectureTests/BindingTests.swift
@@ -35,5 +35,38 @@
 
       XCTAssertNoDifference(viewStore.state, .init(nested: .init(field: "Hello!")))
     }
+
+    func testAutomaticBinding() {
+      struct State: Equatable {
+        @BindableState var nested = Nested()
+
+        struct Nested: Equatable {
+          var field = ""
+        }
+      }
+
+      enum Action: BindableAction, Equatable {
+        case binding(BindingAction<State>)
+      }
+
+      let reducer = Reducer<State, Action, ()> { state, action, _ in
+        switch action {
+        case .binding(\.$nested.field):
+          state.nested.field += "!"
+          return .none
+        default:
+          return .none
+        }
+      }
+      // look, ma, no .binding()!
+
+      let store = Store(initialState: .init(), reducer: reducer, environment: ())
+
+      let viewStore = ViewStore(store)
+
+      viewStore.binding(\.$nested.field).wrappedValue = "Hello"
+
+      XCTAssertNoDifference(viewStore.state, .init(nested: .init(field: "Hello!")))
+    }
   }
 #endif


Sources/ComposableArchitecture/Reducer.swift
```diff
@@ -898,3 +898,21 @@ public struct Reducer<State, Action, Environment> {
...
+public extension Reducer where Action: BindableAction, Action.State == State {
+  init(_ theReducer: @escaping (inout State, Action, Environment) -> Effect<Action, Never>) {
+    self.reducer = { state, action, environment in
+      guard let bindingAction = (/Action.binding).extract(from: action)
+      else {
+        return theReducer(&state, action, environment)
+      }
+
+      bindingAction.set(&state)
+      return theReducer(&state, action, environment)
```
ZevEisenberg, line 914: Borrowed from `.binding()` method from Binding.swift because I couldn't figure out how to use `.binding()` inside this initializer. Open to suggestions.

stephencelis:
@ZevEisenberg Thanks for exploring! This definitely give us some things to consider, but I also have to throw a wrench in things 🫣

We've been planning on making a switch to `Reducer` being a protocol for some time, and with Swift 5.7 on the horizon the language finally has the tools to make it an ergonomic reality. Basically:

```swift
protocol ReducerProtocol<State, Action> {
  associatedtype State
  associatedtype Action
  func reduce(into state: inout State, action: Action) {}
}
```

What's not clear to me, however, is how one could insert this kind of logic automatically into a `ReducerProtocol` conformance...any ideas?

The only idea that comes to mind is maybe a `BindableReducerProtocol` conformance that inherits `ReducerProtocol`...but I haven't worked out the details and am not sure it would work yet.

tgrapperon:
Could this be handled by the builder? I feel that it could be possible to automatically insert `Binding` reducers if the action is `BindableAction`, either when accumulating, or as a final result step if no `Binding` reducer is present in the chain[^1]. I haven't experienced a lot with the their new 5.7 affordances, so it could also be impossible.

[^1]: For example, if the initial result is a `BindingReducer` when Action is `BindableAction`, we could bypass subsequent `BindingReducers` when accumulating. This would also protect against binding twice, but it method would only work with one protocol and not extend to `Navigation` for example

Furthermore, if the builder is a protocol, third parties would also be able to extend this approach to add their own behaviors. Navigation will likely have the same issue, isn't it?

stephencelis:
Added an alternate idea that simply warns instead of trying to automatically insert this logic: #1159

Thoughts?

> Could this be handled by the builder? I feel that it could be possible to automatically insert `Binding` reducers if the action is `BindableAction`, either when accumulating, or as a final result step if no `Binding` reducer is present in the chain[1](#user-content-fn-1-e50b44ddf5b4cea688d5da46cb0e7d46). I haven't experienced a lot with the their new 5.7 affordances, so it could also be impossible.
> 
> Furthermore, if the builder is a protocol, third parties would also be able to extend this approach to add their own behaviors. Navigation will likely have the same issue, isn't it?

@tgrapperon This is definitely interesting! And something to keep in mind generally. The main issues I foresee is that there's currently no requirement that a reducer is created in a builder context, so you could still have opportunities to lose out on this behavior. Also, if you're combining multiple reducers together that all work on bindable state/actions, they would be performing this work multiple times.

It seems that to avoid pitfalls, we should generally rely on explicitness over magic (I came to similar conclusions with parser builders magically moving between substring and UTF-8 inputs).

tgrapperon:
@stephencelis I agree and I prefer explicitness over magic. The current issue with non-protocolized reducers is that you have pullbacks at the top and fluent "factorizable"[^1] reducers at the end, which make it easy to miss them. With the reducer builder, we will most likely write pullbacks and bindings at the same place:
```swift
var body: some Reducer {
  Scope(…)
  Scope(…)
  Binding()
  Reduce {
    …
  }
}
```
It will be much easier to read, and with all components on the same side, it would also easier to spot what's missing. We can maybe also imagine a construct like `SwitchStore` where you would need to handle each case as a build block:
```swift
var body: some SwitchReducer {
  Case(/Action.scoped) {
    Scope(…)
    
  }
  Case(/Action.binding) {
    Binding()
    Reduce {
      … // additional binding effects
    }
  }
}
```
Or even directly switching using the builder's capabilities and passing an action parameter: 
```swift
var body: some Reducer { action in
  switch action {
  case .scoped:
    Scope(…)
  case .binding:
    Binding(…)
    Reduce {
      … 
    }
  }
}
```
I haven't checked how much any of this is realizable, but it could be interesting, at least as an alternative approach (maybe using another, dedicated, builder).
[^1]: "factorizable" in the sense that we can rewrite the resulting reducer as the combination of two other non-emtpy reducers (and I think I just found a new thing to explore…). This includes `.binding()` but not `filter` for example.

stephencelis:
@ZevEisenberg Cool to close this out for now in favor of #1159? Definitely an interesting exploration, though, and look forward to anything else you find on the protocol branches.

ZevEisenberg:
Feel free to close. I still have some ideas related to protocols, which ~this margin is too narrow to contain~ I will try to play with if I get time.

stephencelis:
Closing for now then, thanks! If you explore anything else, please do share in a GitHub discussion when you have the chance 😄

PR #1159, opened 2022-06-23
Add runtime warnings for unused binding actions

stephencelis:
An alternative approach to #1158.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Binding.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Binding.swift
@@ -277,11 +277,13 @@ import SwiftUI
     /// - Parameter keyPath: A key path to a specific bindable state.
     /// - Returns: A new binding.
     public func binding<Value: Equatable>(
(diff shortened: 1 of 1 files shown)


stephencelis:
Thanks @ZevEisenberg for inspiring me to look into other approaches! :smile:

Sources/ComposableArchitecture/SwiftUI/Binding.swift
```diff
@@ -346,6 +364,50 @@ public struct BindingAction<Root>: Equatable {
...
+              """
+            runtimeWarning(
+              """
+              A binding action created at "%@:%d" never invoked its setter:
+
+                Action:
+                  %@
+
+              Enhance reducers that receive binding actions with the "Reducer.binding()" modifier \
+              to ensure the setter will mutate associated state.
```
mbrandonw, line 398: How do you feel about something like this?

```suggestion
              A binding action created at "%@:%d" was not properly handled:

                Action:
                  %@

              To fix this invoke the "binding()" method on your feature's reducer.
```

Issue #1160, opened 2022-06-24
Tests pass with unexpected state changes


umbertovolta:
**Describe the bug**
Since the latest release `0.37.0`, when the `updateExpectingResult` closure is `nil`, tests pass even if the action sent/received changes the state.

**To Reproduce**
You can verify the bug by simply removing the closure from the final `receive` of the test case `testCompleteTodo` inside `TodoTests`.

```swift
  func testCompleteTodo() {

    // ...

    store.receive(.sortCompletedTodos)
//    {
//      $0.todos = [
//        $0.todos[1],
//        $0.todos[0],
//      ]
//    }
  }

```

**Expected behavior**
The test case should fail and complain about an unexpected change.

**Environment**
  - swift-composable-architecture version `0.37.0`
  - Xcode `13.4.1`
  - Swift `5.7`
  - OS: `iOS 15.5`

**Additional context**
I'm going to open a PR with a proposal for the fix. :)

PR #1161, opened 2022-06-24
[Fix] Check the state does not actually change when no changes are expected

umbertovolta:
Proposal of fix for #1160.

I've also added 2 new tests to avoid regression.

Let me know what you think. Thank you! :)


stephencelis (approved):
Thanks for the catch! And tests! We'll merge and cut a release right away.

umbertovolta:
Kudos to @mikchelbaev for the catch. 🙌🏻

Thank you!

Issue #1162, opened 2022-06-24
Unexpected Unhandled BindingAction with TestStore


roanutil:
**Describe the bug**
If a TestStore receives a BindingAction the runtime warning and test failure are triggered. If a BindingAction is sent to a TestStore it behaves as expected and recognizes that the action is handled by the Reducer's binding() modifier.

**To Reproduce**
A test case reproducing the bug has been added to my fork of TCA.
https://github.com/roanutil/swift-composable-architecture/blob/reproduce-binding-invocation-bug-with-test-store/Tests/ComposableArchitectureTests/TestStoreReceivesBindingActionTests.swift

```swift
import ComposableArchitecture
import Foundation
import XCTest

final class TestStoreReceiveBindingActionTests: XCTestCase {

    func testBindingInvocation() throws {
        let store = TestStore(initialState: FeatureState(), reducer: featureReducer, environment: ())
        store.send(FeatureAction.sayHello)
        store.receive(.set(\.$text, "Hello World!")) { state in
            state.text = "Hello World!"
        }
    }
}

struct FeatureState: Equatable {
    @BindableState var text: String

    init(text: String = "") {
        self.text = text
    }
}

enum FeatureAction: Equatable, BindableAction {
    /// An action that in turn sends a binding action
    case sayHello
    case binding(BindingAction<FeatureState>)
}

let featureReducer = Reducer<FeatureState, FeatureAction, Void> { _, action, _ in
    switch action {
    case .sayHello:
        return Effect(value: .set(\.$text, "Hello World!"))
    case .binding:
        return .none
    }
}.binding()
```

**Expected behavior**
I expected the TestStore to recognize that the BindingAction is handled.

**Screenshots**
N/A

**Environment**
  - swift-composable-architecture version 0.38.0
  - Xcode 13.4.1
  - Swift 5.6
  - OS: iOS 15.5

**Additional context**
Add any more context about the problem here.

stephencelis:
@roanutil Want to see if pinning to the branch at #1163 fixes things? If it does I'll cut a patch release right away.

roanutil:
@stephencelis Yes that fixes it. Thanks for the quick response!

stephencelis:
[0.38.1](https://github.com/pointfreeco/swift-composable-architecture/releases/tag/0.38.1) has the fix! Thanks again for reporting :smile:

PR #1163, opened 2022-06-24
Only warn for unused binding actions sent from view stores

stephencelis:
#1162 points out that the warning can appear at inopportune times, like in tests. We should push the warning to the view layer, instead.


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Forms.swift
```diff
@@ -46,7 +46,7 @@
       return .none
     }
   }
-  .binding()
+//  .binding()
```
stephencelis, line 49:
```suggestion
  .binding()
```

mbrandonw (approved):

PR #1164, opened 2022-06-25
Modernize bindings in Todos sample app

filblue:


stephencelis:
@filblue Thanks for taking the time to explore and share this. We've actually gotten a PR for this before, but opted not to take this change:

https://github.com/pointfreeco/swift-composable-architecture/pull/784#discussion_r708677611

Because `ViewState` and `ViewAction` makes things more complex, we think it's better to keep things simple as a demo. I think we'll probably want to develop an article on best practices for bindable state and actions in the future, though, for guidance here.

wtruppel:
@stephencelis That's great! I look forward to seeing the new tools. But please keep in mind that not everyone will be on iOS 16. We, in particular, are supporting iOS 14 and above. It would be great if the new navigation tools don't depend on the new `NavigationStack` APIs.

PR #1168, opened 2022-06-28
"Failing" -> "Unimplemented"

stephencelis:
We've found that the "failing" naming of dependencies can be confusing. While it's accurate in isolation, when you're writing tests, "failure" is already a loaded term. We'd like to move in the direction of the "unimplemented" terminology (which is funnily enough where we started, when we had some preexisting dependencies that would `fatalError`).

This PR:

- Renames `Effect.failing` to `Effect.unimplemented`, soft-deprecating `Effect.failing` in the process. We can warn in a future version.
- Uses `XCTUnimplemented` and `UnimplementedScheduler` in examples


mbrandonw (approved):
much better

PR #1171, opened 2022-06-29
Collapse case study readmes behind disclosure groups

stephencelis:
These READMEs can be quite a lot all at once, and on smaller devices can even completely hide the content.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndConfirmationDialogs.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndConfirmationDialogs.swift
@@ -89,14 +89,18 @@ struct AlertAndConfirmationDialogView: View {
   var body: some View {
     WithViewStore(self.store) { viewStore in
       Form {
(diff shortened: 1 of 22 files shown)


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndConfirmationDialogs.swift
```diff
@@ -89,14 +89,18 @@ struct AlertAndConfirmationDialogView: View {
...
     WithViewStore(self.store) { viewStore in
       Form {
-        Section(header: Text(template: readMe, .caption)) {
-          Text("Count: \(viewStore.count)")
-          Button("Alert") { viewStore.send(.alertButtonTapped) }
-          Button("Confirmation Dialog") { viewStore.send(.confirmationDialogButtonTapped) }
+        Section {
+          DisclosureGroup("About this case study") {
+            Text(template: readMe)
+          }
```
mbrandonw, line 95: very tiny suggestion, but should we just have a `about(String) -> some View` function that does this, or an `AboutView`? just to considerate the "About this case study" and whatnot?

PR #1189, opened 2022-07-07
Concurrency Beta

mbrandonw:
This PR is our ongoing work on adding concurrency features to the Composable Architecture. We laid out the plans for this beta in details in this discussion: #1186.

Feel free to ask questions or make comments on specific lines of code in this PR, and ask broader questions in the discussion thread. We will also be opening smaller PR's against this branch for better visibility into how the beta is evolving over the next few weeks.


Sources/ComposableArchitecture/Effect.swift
```diff
@@ -264,141 +85,380 @@ public struct Effect<Output, Failure: Error>: Publisher {
...
+    catch handler: (@Sendable (Error) async -> Output)? = nil,
+    file: StaticString = #file,
+    fileID: StaticString = #fileID,
+    line: UInt = #line
+  ) -> Self {
+    Deferred<Publishers.HandleEvents<PassthroughSubject<Output, Failure>>> {
+      let subject = PassthroughSubject<Output, Failure>()
+      let task = Task(priority: priority) { @MainActor in
+        await withTaskCancellationHandler {
+          if Thread.isMainThread {
```
larryonoff, line 240: I'm not sure, but Thread.isMainThread may return invalid results (when working with Dispatch), this's why ReactiveSwift framework introduced UIScheduler, see https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Scheduler.swift#L92

mbrandonw: @larryonoff Thanks for going through the PR with a fine toothed comb and uncovering this. 🙂

We are actually going to remove the cancellation handler from `Effect.task` and `Effect.run` for right now. We only added it to support doing:

```swift
withUnsafeCurrentTask { $0?.cancel() }
```

...but that actually kind of behaves weirdly and we're not sure if it's a Swift concurrency bug or what.

Also, we're not sure handling cancellation in this way is correct, otherwise _every_ unstructured task ever created should technically start with a `withTaskCancellationHandler`, which would be weird.

We will keep thinking about it and see if we should bring it back later.

larryonoff:
@mbrandonw

I observed that e.g. class Box is not marked as final. I assume that it isn't expected to be inherited in future. So it would be great to mark it final to increase performance by reducing dynamic dispatch (small but wise optimisation), see [article](https://developer.apple.com/swift/blog/?id=27).

It would also be great to revise other classes for the same optimisation.

Thanks!

mbrandonw:
Hi @larryonoff, good catch! We plan to make that class internal too since the library probably shouldn't be vending such a generic type. Thanks for the reminder :)

stephencelis:
@larryonoff Our latest updates to the beta actually just made `Box` internal (we'd forgotten to do so when we moved it earlier). If you were relying on it, can you let us know how? We believe that `ActorIsolated` (which we used to call `SendableState`) and `UncheckedSendable` are tools more tuned to the problem.

larryonoff:
> @larryonoff Our latest updates to the beta actually just made `Box` internal (we'd forgotten to do so when we moved it earlier). If you were relying on it, can you let us know how? We believe that `ActorIsolated` (which we used to call `SendableState`) and `UncheckedSendable` are tools more tuned to the problem.

Thanks for asking. I use this branch a little. Just sometimes checking changes and looking forward to release :)

Sources/ComposableArchitecture/Effect.swift
```diff
@@ -1,185 +1,331 @@
...
+                file: file,
+                line: line
+              )
+            #endif
+            return
+          }
+          await subject.send(handler(error))
+        }
+      }
+      return subject.handleEvents(receiveCancel: task.cancel)
```
larryonoff, line 129: I have an issue with cancelling task in reactiveswift-composable-architecture in the same place, e.g. changing to `subject.handleEvents(receiveCancel: { task.cancel() })` fixed the issue.

It would be great to fix for current production version, If this's the issue.

What do you think about this?

stephencelis: @larryonoff Can you explain more? I'm not sure I understand.

Are you saying that you had a problem with:

```swift
subject.handleEvents(receiveCancel: task.cancel)
```

That was fixed by opening up the closure?

```swift
subject.handleEvents(receiveCancel: { task.cancel($0) }
```

And are you saying that this applied to reactiveswift-composable-architecture, or are you saying the issue also affects this TCA branch? If you're seeing a problem with our branch, can you tell us how to reproduce? We'd be happy to incorporate the change, but we want to better understand what it's fixing.

larryonoff: @stephencelis sorry for such a blurry description. I wanted to describe the issue as soon as I caught it.

Let me add some details :)

I caught the issue in `reactiveswift-composable-architecture` version of TCA. I had a problem with `subject.handleEvents(receiveCancel: task.cancel)`, which was fixed by opening the closure `subject.handleEvents(receiveCancel: { task.cancel($0) }`.

I'm aware that ReactiveSwift Effect lifetime differs from Combine Effects.

I just wanted to point on the issue, because I thought that It may be easy for you to catch the issue if it's.

I checked now TCA tests and I've found a test `testCancellingTask_Infallible`. So if it passes, everything should be good.

Sources/ComposableArchitecture/Effects/ConcurrencySupport.swift
```diff
@@ -0,0 +1,369 @@
...
+  /// let stream = AsyncStream<Int> { continuation = $0 }
+  /// let continuation = _continuation!
+  ///
+  /// // vs.
+  ///
+  /// let (stream, continuation) = AsyncStream<Int>.streamWithContinuation()
+  /// ```
+  ///
+  /// This tool is usually used for tests where we need to supply an async sequence to a dependency
+  /// endpoint and get access to its continuation so that we can emulating the dependency
```
tgrapperon, line 91:
```suggestion
  /// endpoint and get access to its continuation so that we can emulate the dependency
```

Sources/ComposableArchitecture/Effects/ConcurrencySupport.swift
```diff
@@ -0,0 +1,369 @@
...
+  /// emitting data. For example, suppose you have a dependency exposes an async sequence for
```
tgrapperon, line 92:
```suggestion
  /// emitting data. For example, suppose you have a dependency exposing an async sequence for
```

Sources/ComposableArchitecture/Effects/ConcurrencySupport.swift
```diff
@@ -0,0 +1,369 @@
...
+  /// let stream = AsyncThrowingStream<Int, Error> { continuation = $0 }
+  /// let continuation = _continuation!
+  ///
+  /// // vs.
+  ///
+  /// let (stream, continuation) = AsyncThrowingStream<Int, Error>.streamWithContinuation()
+  /// ```
+  ///
+  /// This tool is usually used for tests where we need to supply an async sequence to a dependency
+  /// endpoint and get access to its continuation so that we can emulating the dependency
```
tgrapperon, line 188:
```suggestion
  /// endpoint and get access to its continuation so that we can emulate the dependency
```

Sources/ComposableArchitecture/Effects/ConcurrencySupport.swift
```diff
@@ -0,0 +1,369 @@
...
+  /// emitting data. For example, suppose you have a dependency exposes an async sequence for
```
tgrapperon, line 189:
```suggestion
  /// emitting data. For example, suppose you have a dependency exposing an async sequence for
```

tgrapperon (commented):
Mostly typos. There are changes I'm not sure about.

PR #1201, opened 2022-07-11
Small updates to animations case study.

mbrandonw:
Backports the cancellation logic from #1189 and writes a test.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
@@ -40,7 +40,7 @@ extension Effect where Failure == Never {
 
 struct AnimationsState: Equatable {
   var alert: AlertState<AnimationsAction>?
-  var circleCenter = CGPoint(x: UIScreen.main.bounds.width / 2, y: UIScreen.main.bounds.height / 2)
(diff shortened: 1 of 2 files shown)


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
```diff
@@ -40,7 +40,7 @@ extension Effect where Failure == Never {
 
 struct AnimationsState: Equatable {
   var alert: AlertState<AnimationsAction>?
-  var circleCenter = CGPoint(x: UIScreen.main.bounds.width / 2, y: UIScreen.main.bounds.height / 2)
+  var circleCenter = CGPoint(x: 175, y: 300)
```
mbrandonw, line 43: We have to hard code this due to isolation errors in Xcode 14.

stephencelis (approved):

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
tgrapperon: Since we have a GeometryReader already in place, we can pass the initial center at `onAppear`, isn't it? This is a frequent pattern (initializing values from SwiftUI), so it may be worth presenting an example if this isn't happening in another study.

PR #1203, opened 2022-07-11
Set the circle in the center of the view in the animation's study

tgrapperon:
The reuse of the `.tapped` action is maybe not optimal though.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
@@ -114,7 +114,30 @@ struct AnimationsView: View {
       VStack(alignment: .leading) {
         Text(template: readMe, .body)
           .padding()
-          .allowsHitTesting(false)
+          .gesture(
+            DragGesture(minimumDistance: 0).onChanged { gesture in
+              viewStore.send(
+                .tapped(gesture.location),
+                animation: .interactiveSpring(response: 0.25, dampingFraction: 0.1)
+              )
+            }
+          )
+          .overlay {
+            GeometryReader { proxy in
+              Circle()
+                .fill(viewStore.circleColor)
+                .colorInvert()
+                .blendMode(.difference)
+                .frame(width: 50, height: 50)
+                .scaleEffect(viewStore.isCircleScaled ? 2 : 1)
+                .position(
+                  x: viewStore.circleCenter?.x ?? proxy.size.width / 2,
+                  y: viewStore.circleCenter?.y ?? proxy.size.height / 2
+                )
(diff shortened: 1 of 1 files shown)


tgrapperon:
Ah sorry, this doesn't pair well with `.reset`.
I'm marking as draft to check if I can make it work without adding too much.

iampatbrown:
One option could be to store a translation, a `CGSize` that resets to `.zero`,  and work from that.

tgrapperon:
Indeed. I'm using a `CGPoint?` instead but it's the same idea, but I'm now reorganizing the whole hierarchy with a `GeometryReader` in an overlay which is more up to todays standards I guess. The vertical offset that shifted the dot from your finger wasn't something done on purpose, but an happy little accident of dividing the height off the `ZStack/Text` by the height of the view.

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
```diff
@@ -111,53 +111,53 @@ struct AnimationsView: View {
...
+            .fill(viewStore.circleColor)
+            .colorInvert()
+            .blendMode(.difference)
+            .frame(width: 50, height: 50)
+            .scaleEffect(viewStore.isCircleScaled ? 2 : 1)
+            .position(
+              x: viewStore.circleCenter?.x ?? proxy.size.width / 2,
+              y: viewStore.circleCenter?.y ?? proxy.size.height / 2
+            )
+            .offset(y: viewStore.circleCenter == nil ? 0 : -44)
```
tgrapperon, line 143: Should we keep this effect, or center the dot on the gesture's location?

tgrapperon:
@iampatbrown I've reorganized a little the hierarchy. I'm not sure if the "active" view should be the text view like it was, or the whole view. On an iPad for example, the principal text view could be much shorter in height. If we use the whole screen instead, we can't have the `DragGesture` with `minimumDistance=0` because it captures all touch events, including those dedicated to the buttons. So I put the gesture in a background, and the circle in an overlay. What do you think?
With iOS 16's `SpatialTapGesture`, we'll probably be able to put everything in an overlay, along with a `DragGesture` using its standard minimum distance.

tgrapperon:
I did put the gesture back on the `Text`'s only. This is simpler to understand than trying to make the whole view active.
I converted the `ZStack` to `Text+Overlay` because it also makes more sense I think. 
I've also moved the gesture closer to `Text`, so we understand better where it acts. Alternatively, I could put the gesture in the overlay with a `.contentShape(Rectangle())`, but I think it would probably be a little more complex to parse.
I think the studies could be modernized in several places, but it is maybe better to wait for iOS 16 before making a pass.

stephencelis (approved):
Fun solution to the problem! Thanks @tgrapperon!

PR #1206, opened 2022-07-12
Deprecate unchecked store

mbrandonw:
In the near future we want to move `Store` and `ViewStore` to `@MainActor`, and that unfortunately doesn't play nicely with [`Store.unchecked`](https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/store/unchecked(initialstate:reducer:environment:)). So, we'd like to deprecate it.

In the more distant future, when we can completely disentangle Combine from the library, we actually want to make `Store` just an actor, not a `@MainActor`. This means you would get the benefits of "unchecked" stores without having to think about it. We don't know when this will be possible, but it is how we think things will eventually evolve.

If you do use `Store.unchecked` today, there are a few options:

* Re-evaluate why you moved to unchecked stores and see if there are other options. Do you use unchecked stores because you were performing intense CPU work in the reducer? If so, perhaps you can move that work to effects that run on background threads. Or, do you use unchecked stores because you send actions at high frequency and the main thread couldn't keep up? If so, perhaps actions can be batched, throttled and/or accumulated and sent at a lower frequency.
* Copy-and-paste `Store.swift` in order to remove all the thread checks, and make `state` and `send` public. This new type of store won't be usable with all the other tools in the library, but nearly all those tools are UI-based and so bound to the main thread anyway.


Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -5,6 +5,36 @@ import XCTestDynamicOverlay
...
+/// - Parameters:
+///   - initialState: The state to start the application in.
+///   - reducer: The reducer that powers the business logic of the application.
+///   - environment: The environment of dependencies for the application.
+@available(
+  *, deprecated,
+   message:
+    """
+    If you use this initializer, please open a discussion on GitHub and let us know how: \
+    https://github.com/pointfreeco/swift-composable-architecture/discussions/new
```
stephencelis, line 20: Nice, multiline strings work here?

Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -5,6 +5,36 @@ import XCTestDynamicOverlay
 
 // NB: Deprecated after 0.38.0:
```
stephencelis, line 6: I think you want to move this deprecation above and add a "Deprecated after 0.38.2" section.

Tests/ComposableArchitectureTests/TestStoreFailureTests.swift
```diff
@@ -1,3 +1,4 @@
+import Combine
```
stephencelis, line 1: Snuck in?

mbrandonw: Yup, removed!

PR #1207, opened 2022-07-12
Clean up deprecation formatting

stephencelis:
Didn't realize multiline worked here, but should make it easier to read without word wrap

### DIFF
--- a/Sources/ComposableArchitecture/Internal/Deprecations.swift
+++ b/Sources/ComposableArchitecture/Internal/Deprecations.swift
@@ -31,7 +31,10 @@ extension Effect {
     *,
     deprecated,
     message:
-      "Using a variadic list is no longer supported. Use an array of identifiers instead. For more on this change, see: https://github.com/pointfreeco/swift-composable-architecture/pull/1041"
+      """
+      Using a variadic list is no longer supported. Use an array of identifiers instead. For more \
+      on this change, see: https://github.com/pointfreeco/swift-composable-architecture/pull/1041
+      """
   )
   @_disfavoredOverload
   public static func cancel(ids: AnyHashable...) -> Self {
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -427,7 +447,10 @@ extension Reducer {
   @available(
     *, deprecated,
     message:
-      "'Reducer.binding()' no longer takes an explicit extract function and instead the reducer's 'Action' type must conform to 'BindableAction'"
+      """
+      "'Reducer.binding()' no longer takes an explicit extract function and instead the reducer's \
```
jshier, line 451: Extraneous leading quotes in this message and the next.

mbrandonw (approved):
looks good once @jshier's find is fixed!

Sources/ComposableArchitecture/Internal/Deprecations.swift
stephencelis, line 451:
```suggestion
      'Reducer.binding()' no longer takes an explicit extract function and instead the reducer's \
```

Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -441,7 +464,12 @@ extension ViewStore {
   @available(
     *, deprecated,
     message:
-      "For improved safety, bindable properties must now be wrapped explicitly in 'BindableState'. Bindings are now derived via 'ViewStore.binding' with a key path to that 'BindableState' (for example, 'viewStore.binding(\\.$value)'). For dynamic member lookup to be available, the view store's 'Action' type must also conform to 'BindableAction'."
+      """
+      "For improved safety, bindable properties must now be wrapped explicitly in 'BindableState'. \
```
stephencelis, line 468:
```suggestion
      For improved safety, bindable properties must now be wrapped explicitly in 'BindableState'. \
```

PR #1208, opened 2022-07-13
Modernize SwiftUI Case Studies

tgrapperon:
Hello!
I took the liberty to modernize SwiftUI Case Studies using iOS 15 affordances. To summarize:
- Replaced soft-deprecations I'm aware of (mostly around navigation bars);
- Homogenized the `String/Text` overloads used, using `Text` ones only when necessary;
- Factor common modifiers in parents;
- Switch `Color` modifiers to their `ShapeStyle` counterparts;
- Polished a few things here and there, using new APIs when possible.

Sorry for the lengthy PR!


Examples/CaseStudies/SwiftUICaseStudies/02-Effects-SystemEnvironment.swift
```diff
@@ -203,23 +208,23 @@ struct SystemEnvironment<Environment> {
 }
 
 #if DEBUG
-  import XCTestDynamicOverlay
```
tgrapperon, line 206: Sorry! Nothing changed here, I'm using different rules for `#if Debug` indentation in my `swift-format` Xcode extension.

PR #1210, opened 2022-07-15
Add AlwaysEquatable

alwaysblu:
Added a property wrapper that lets you specify properties so that the view doesn't update when the state's properties change.

### DIFF
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -1,6 +1,48 @@
 import Combine
 import SwiftUI
 
+/// A property wrapper type that always returns true meaning the same value
+/// when performing a comparison operation (==),
+/// even if the values ​​are different.
+///
+/// This type can solve the problem of popping at a specific depth when using NavigationLink,
+/// which occurs when the local state in the global state changes.
+///
+/// For example, if you write the code as below,
+/// ``objectWillChange.send()`` of ViewStore will not be called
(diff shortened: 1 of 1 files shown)


stephencelis:
@alwaysblu Thanks for sharing! Rather than PR this now, would you be open to starting a [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions/new) instead? In general we're pretty careful about adding surface are to the library, and usually try to figure out what surface area we can remove in the future to make the library more approachable and less overwhelming to newcomers. You could then describe the feature a bit more, how you use it in practice, and get feedback from the community.

This property wrapper is definitely interesting, though!

alwaysblu:
> @alwaysblu Thanks for sharing! Rather than PR this now, would you be open to starting a [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions/new) instead? In general we're pretty careful about adding surface area to the library, and usually try to figure out what surface area we can remove in the future to make the library more approachable and less overwhelming to newcomers. You could then describe the feature a bit more, how you use it in practice, and get feedback from the community.
> 
> This property wrapper is definitely interesting, though!

@stephencelis 

I created a [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions/1216).

thanks for the reply :)

mbrandonw:
Hi @alwaysblu, thanks for creating a discussion. We're going to close this for now since we're not sure it belongs in the library, and luckily can be easily added from the outside.

PR #1211, opened 2022-07-15
Use a 'lorem' speech client in previews.

mbrandonw:
The Speech framework doesn't work in previews, so this introduces a little "lorem" client that just spits out some lorem ipsum at a rate of 3 words / second.


stephencelis (approved):
Fun!

PR #1215, opened 2022-07-15
Relax AlertState.init(title: TextState, message: TextState? = nil, buttons: [Button])

kdeda:
We are working on a macOS app supporting BigSur.

``` 
 public init(
    title: TextState,
    message: TextState? = nil,
    buttons: [Button]
  ) {

  // is very similar to

  public init(
    title: TextState,
    message: TextState? = nil,
    dismissButton: Button? = nil
  ) {
```

I want to relax the macOS requirements to 11.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Alert.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Alert.swift
@@ -105,7 +105,7 @@ public struct AlertState<Action> {
   public var message: TextState?
   public var title: TextState
 
-  @available(iOS 15, macOS 12, tvOS 15, watchOS 8, *)
+  @available(iOS 15, macOS 11, tvOS 15, watchOS 8, *)
   public init(
     title: TextState,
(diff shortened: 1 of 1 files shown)


kdeda:
It works well with 2 or 3 buttons.
But it seems we can't display more than 3 buttons on a macOS alert.

stephencelis:
@kdeda Sorry for the delay. We specifically tie this availability to the functionality exposed from SwiftUI, and SwiftUI does not have an array-based interface for buttons pre-macOS 12. That's exactly why you noted this:

> Also a small observation on macOS 12, the underlying apple code will not display more than 3 buttons.

We try out best to avoid these runtime gotchas, so I think we want to keep these availability checks in sync with Apple's frameworks.

stephencelis:
I'm going to close this, but would like to note that if you're OK with the runtime gotchas, you should be able to extend our library with your own initializer that doesn't require the availability checks.

PR #1217, opened 2022-07-16
More clean up in speech demo

mbrandonw:
Found a few small things.

### DIFF
--- a/Examples/SpeechRecognition/SpeechRecognition/SpeechClient/Live.swift
+++ b/Examples/SpeechRecognition/SpeechRecognition/SpeechClient/Live.swift
@@ -24,14 +24,6 @@ extension SpeechClient {
       },
       startTask: { request in
         Effect.run { subscriber in
-          let speechRecognizer = SFSpeechRecognizer(locale: Locale(identifier: "en-US"))!
-          let cancellable = AnyCancellable {
-            audioEngine?.stop()
-            audioEngine?.inputNode.removeTap(onBus: 0)
-            recognitionTask?.cancel()
-            _ = speechRecognizer
-          }
-
           audioEngine = AVAudioEngine()
           let audioSession = AVAudioSession.sharedInstance()
           do {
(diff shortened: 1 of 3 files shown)


Examples/SpeechRecognition/SpeechRecognition/SpeechClient/Live.swift
```diff
@@ -24,24 +24,17 @@ extension SpeechClient {
       },
       startTask: { request in
         Effect.run { subscriber in
-          let speechRecognizer = SFSpeechRecognizer(locale: Locale(identifier: "en-US"))!
-          let cancellable = AnyCancellable {
```
mbrandonw, line 28: this cancellable was way to high up, need to move down below after we create audio engine and recognition task.

Examples/SpeechRecognition/SpeechRecognition/SpeechRecognition.swift
```diff
@@ -12,7 +12,6 @@ private let readMe = """
 struct AppState: Equatable {
   var alert: AlertState<AppAction>?
   var isRecording = false
-  var speechRecognizerAuthorizationStatus = SFSpeechRecognizerAuthorizationStatus.notDetermined
```
mbrandonw, line 15: we don't actually use this anywhere, so removing.

Examples/SpeechRecognition/SpeechRecognitionTests/SpeechRecognitionTests.swift
```diff
@@ -8,18 +8,15 @@ class SpeechRecognitionTests: XCTestCase {
...
-    speechClient.requestAuthorization = { Effect(value: .denied) }
-
     let store = TestStore(
       initialState: AppState(),
       reducer: appReducer,
-      environment: AppEnvironment(
-        mainQueue: .immediate,
-        speechClient: speechClient
-      )
+      environment: .unimplemented
```
mbrandonw, line 14: moving towards the pattern of always defaulting to unimplemented and then overriding environment down below.

stephencelis (approved):

PR #1226, opened 2022-07-26
Fixed broken other versions in README

yimajo:
"Other versions" in README is not marked up.
↓
sccreenshot 2022-07-27 0 47 27

Can be fixed by inserting a blank line after  `</summary>`.


mbrandonw (approved):
Good catch, thanks!

PR #1250, opened 2022-08-08
Move debouncing logic to view.

mbrandonw:
An experiment in what it would look like to leverage debouncing in the view (via `task(id:)` and new async tools in the library), rather than using `Effect.debounce`.

### DIFF
--- a/Examples/Search/Search/SearchApp.swift
+++ b/Examples/Search/Search/SearchApp.swift
@@ -10,8 +10,7 @@ struct SearchApp: App {
           initialState: SearchState(),
           reducer: searchReducer.debug(),
           environment: SearchEnvironment(
-            weatherClient: WeatherClient.live,
-            mainQueue: .main
+            weatherClient: WeatherClient.live
           )
         )
       )
--- a/Examples/Search/Search/SearchView.swift
+++ b/Examples/Search/Search/SearchView.swift
@@ -32,19 +32,21 @@ struct SearchState: Equatable {
 enum SearchAction: Equatable {
   case forecastResponse(Search.Result.ID, TaskResult<Forecast>)
   case searchQueryChanged(String)
+  case searchQueryChangeDebounced
   case searchResponse(TaskResult<Search>)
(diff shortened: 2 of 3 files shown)


stephencelis (approved):
Nice.

tgrapperon:
This is something I'm also doing in my project and it works well so far.
I thought it could maybe be possible to provide a `ViewStore.send` with an `after:` argument. I don't know if the "true" `MainActor` could be assumed for waiting, or if the `ViewStore` should be aware of the `mainQueue` dependency (which is not ambiguous in proto branches).

stephencelis:
@tgrapperon You could probably leverage the new `withTaskCancellation(id:)` helper to get a debounce in-place without the `task(id:)` mechanism.

stephencelis:
> This delay is purely on the UI's side, so it could be nice to be able to take decisions from the view's side.

I think `withTaskCancellation(id:)` will work in the view just fine, FWIW. It's a generic function for task cancellation with a given identifier and doesn't require you to be in an `Effect`. If you set `cancelInFlight: true` it should provide debouncing behavior to any async context.

tgrapperon:
In my case, I was doing, in a `.contextualMenu`:
```swift
Button {
  Task {
    try await Task.sleep(450ms)
    viewStore.send(.action)
  }
} label: {…}
```
450ms is the time it takes for the menu to dismiss and animate changes without glitches (a problem specific to this action and this context, not something endemic to contextual menus).
Because this `Task` is not in a `.task` modifier, it will not be cancelled when the `View` goes away. I don't think `withTaskCancellation(id:)` could fix this in this case. This is not absolutely problematic here because I know that the user can't dismiss the `View` in this interval, and the `viewStore and its state will still be here to receive the action.

This is not related to the PR functionality, only the bit with a `sleep/viewStore.send`, so it is off-topic. I've tried to send with a delayed animation, but it doesn't totally fixes glitches in my case. The safest route is to add an optional `delay: UInt` argument to my action and wait in the reducer accordingly. This is probably what I'll end up doing.

PR #1255, opened 2022-08-09
Clean up Todos sort debouncing

stephencelis:
Slightly more code, but uses simple Swift concurrency instead of relying on an operator.

### DIFF
--- a/Examples/Todos/Todos/Todos.swift
+++ b/Examples/Todos/Todos/Todos.swift
@@ -90,15 +90,18 @@ let appReducer = Reducer<AppState, AppAction, AppEnvironment>.combine(
 
     case .todo(id: _, action: .checkBoxToggled):
       enum TodoCompletionID {}
-      return .task { .sortCompletedTodos }
-        .debounce(id: TodoCompletionID.self, for: 1, scheduler: environment.mainQueue.animation())
+      return .task {
+        try await environment.mainQueue.sleep(for: .seconds(1))
+        return .sortCompletedTodos
+      }
+      .animation()
+      .cancellable(id: TodoCompletionID.self, cancelInFlight: true)
 
     case .todo:
       return .none
     }
   }
 )
-.debug()
 
 struct AppView: View {
   let store: Store<AppState, AppAction>
--- a/Examples/Todos/Todos/TodosApp.swift
+++ b/Examples/Todos/Todos/TodosApp.swift
@@ -8,7 +8,7 @@ struct TodosApp: App {
       AppView(
         store: Store(
           initialState: AppState(),
-          reducer: appReducer,
+          reducer: appReducer.debug(),
           environment: AppEnvironment(
             mainQueue: .main,
             uuid: { UUID() }


Examples/Todos/Todos/Todos.swift
```diff
@@ -90,15 +90,18 @@ let appReducer = Reducer<AppState, AppAction, AppEnvironment>.combine(
...
+      }
+      .animation()
+      .cancellable(id: TodoCompletionID.self, cancelInFlight: true)
 
     case .todo:
       return .none
     }
   }
 )
-.debug()
```
stephencelis, line 101: We prob don't want output during tests.

Examples/Todos/Todos/Todos.swift
```diff
@@ -90,15 +90,18 @@ let appReducer = Reducer<AppState, AppAction, AppEnvironment>.combine(
...
     case .todo(id: _, action: .checkBoxToggled):
       enum TodoCompletionID {}
-      return .task { .sortCompletedTodos }
-        .debounce(id: TodoCompletionID.self, for: 1, scheduler: environment.mainQueue.animation())
+      return .task {
+        try await environment.mainQueue.sleep(for: .seconds(1))
+        return .sortCompletedTodos
+      }
+      .animation()
+      .cancellable(id: TodoCompletionID.self, cancelInFlight: true)
```
stephencelis, line 98: Alternately, if we want it to read more linearly:

```suggestion
      return .run { send in
        await withTaskCancellation(id: TodoCompletionID.self, cancelInFlight: true) {
          try await environment.mainQueue.sleep(for: .seconds(1))
          await send(.sortCompletedTodos, animation: .default)
        }
      }
```

mbrandonw (approved):
noice

Examples/Todos/Todos/Todos.swift
damirstuhec: Curious to learn: was `.run` chosen in this suggestion instead of the `.task` because it's better suited for effects that might not emit, eg. when being potentially canceled?

mbrandonw: Hey @damirstuhec, I believe the reason we went with this style is so that we can send the action with an animation. Otherwise you have to do something like this:

```swift
return .task {
  ...
}
.animation()
```

They are basically equivalent, so which you choose is up to you!

Also, as of 0.40.0, both `Effect.task` and `Effect.run` are powered by the same underlying [machinery](https://github.com/pointfreeco/swift-composable-architecture/blob/main/Sources/ComposableArchitecture/Effect.swift#L40), and so there is no performance penalty to using one or the other. It's purely an aesthetics and functionality choice. Prior to 0.40.0, `Effect.run` was a little more heavy weight because it used [`Publishers.Create`](https://github.com/pointfreeco/swift-composable-architecture/blob/main/Sources/ComposableArchitecture/Internal/Create.swift).

damirstuhec: Great. 👍 Thanks a lot for taking the time to reply!

PR #1259, opened 2022-08-11
Voice memos dependency previews

mbrandonw:
The audio player and recorder functionality for the voice memos demo doesn't actually work in previews, but also the mocks we are using don't show off the true functionality of the feature. This puts in more interesting mocks so that you can actually play around with the feature as if it were running real, live dependencies.

### DIFF
--- a/Examples/VoiceMemos/VoiceMemos/VoiceMemos.swift
+++ b/Examples/VoiceMemos/VoiceMemos/VoiceMemos.swift
@@ -53,15 +53,6 @@ let voiceMemosReducer = Reducer<VoiceMemosState, VoiceMemosAction, VoiceMemosEnv
       }
     ),
   Reducer { state, action, environment in
-//    var newRecordingMemo: RecordingMemoState {
-//      RecordingMemoState(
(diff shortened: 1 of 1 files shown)


Examples/VoiceMemos/VoiceMemos/VoiceMemos.swift
```diff
@@ -53,15 +53,6 @@ let voiceMemosReducer = Reducer<VoiceMemosState, VoiceMemosAction, VoiceMemosEnv
...
     ),
   Reducer { state, action, environment in
-    var newRecordingMemo: RecordingMemoState {
-      RecordingMemoState(
-        date: environment.mainRunLoop.now.date,
-        url: environment.temporaryDirectory()
-          .appendingPathComponent(environment.uuid().uuidString)
-          .appendingPathExtension("m4a")
-      )
-    }
```
mbrandonw, line 63: This breaks previews somehow, so removing for now and inlining the work below.

stephencelis (approved):

PR #1267, opened 2022-08-15
Add `final` keyword to every test class by default

Jager-yoo:
Hi everyone!
In my opinion, the test class (which inherits from the [XCTestCase](https://developer.apple.com/documentation/xctest/xctestcase) class) doesn't need to be overridden.
And I found that some test classes already have `final`, but some don't.

So I want to add `final` keyword to every test class by default.
Please comment if there are cases where it should not be `final class` intentionally, or if you have other opinions.
Thanks!

(ref.) Airbnb Swift Style Guide - [Default classes to `final`.](https://github.com/airbnb/swift#final-classes-by-default)


mbrandonw (approved):
Hi @Jager-yoo, thanks for taking the time to do this. It's definitely OK to mark all tests as `final`. We don't tend to do this by default for tests, so it may get out of sync in the future. Feel free to keep us up-to-date. 😄

Jager-yoo:
Yeah, Sure. I understand that this doesn't have to be mandatory.
I will check periodically, checking for updates in the future. 😉

PR #1276, opened 2022-08-19
Add multiple functions to merge/concatenate effects and/or its outputs

marslin1220:
Enhance the ways (array/closure) to merge/concatenate effects and/or its outputs and add instance functions for chaining.

with array/closure, we can merge different effects/outputs by the conditions:

```Swift
let effect: Effect<Int, Never> = Effect.merge {
  if condition {
    return [Effect(value: 1)]
  } else {
    return [Effect(value: 1), Effect(value: 2)]
  }
}
```

merge the effects with its outputs:

```Swift
enum ExampleAction {
    case action1
    case action2
    case action3
  }

// Instead of:
let mergedEffect: Effect<ExampleAction, Never> = Effect.merge(
  Effect(value: .action1),
  Effect(value: .action2),
  Effect(value: .action3),
)

// We can:
let mergedEffectWithActions: Effect<ExampleAction, Never> = Effect.mergeOutputs(.action1, .action2, .action3)
```

and we can chain the effects/outputs in different ways:

```Swift
let mergedEffect: Effect<ExampleAction, Never> = Effect.mergeOutputs(.action1, .action2).concatenateOutputs(.action3)
```

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -340,6 +340,105 @@ extension Effect {
     .merge(effects)
   }
 
+  /// Merges a variadic list of effects together into a single effect, which runs the effects at the
+  /// same time.
+  ///
+  /// - Parameter effects: A list of effects.
+  /// - Returns: A new effect
+  public func merge(_ effects: Self...) -> Self {
+    self.merge(effects)
+  }
+
+  /// Merges another effect together into a single effect, which runs the effects at the
+  /// same time.
+  ///
+  /// - Parameter effect: An effect.
+  /// - Returns: A new effect
+  public func merge(_ effect: Effect) -> Effect {
+    .merge(self, effect)
+  }
+
+  /// Merges an effect array together into a single effect, which runs the effects at the
+  /// same time.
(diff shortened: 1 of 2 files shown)


stephencelis:
Hi @marslin1220! We appreciate the time you took to PR these changes and share.

We may not be doing enough to communicate this to contributors, but we've been working to minimize surface area in the core library, and also to minimize the Combine publisher-forward interfaces of the `Effect` type, so while these helpers could definitely be handy, they unfortunately add to both of these surface areas. With our recent concurrency release, we think folks should mostly only use `Effect.task`, `Effect.run`, `Effect.fireAndForget`, and `Effect.none`.

The Combine integration isn't going away right away, though, so I think it's fine to share these helpers as a companion package/library if you choose to open source them as one, or if you'd like to share them in the GitHub discussion forums, I'm sure some folks may be interested in using them.

Thanks again! And sorry that the helpers aren't right for the core library at this time!

marslin1220:
Hi @stephencelis , thank you for your response and I agree with you to separate the core library and the helpers.
I will create another repo to share those helpers and share the repo link in this PR. 🤝

PR #1283, opened 2022-08-22
ReducerProtocol

mbrandonw:
See the discussion #1282 for more information on this PR.

If you find any typos or parts that might have bugs or problems, please feel free to make a comment on the line and let us know.


Examples/CaseStudies/SwiftUICaseStudies/03-Navigation-Lists-LoadThenNavigate.swift
```diff
@@ -83,16 +65,22 @@ let loadThenNavigateListReducer =
...
         )
         return .none
       }
     }
-  )
+    .ifLet(\.selection, action: .self) {
+      Scope(state: \Identified<State.Row.ID, Counter.State>.value, action: /Action.counter) {
+        Counter()
+      }
+    }
```
kamcma, line 78: While migrating some code to the new reducer API, I was getting the runtime warning about receiving a child action when child state is nil in a situation where I didn't expect it. Can't guarantee it's not user error, but I was stumped. Went looking through the case studies and found this domain that matches mine: an optional `Identified` value driving list selection navigation. Ran the case study, and found that this composition here too emits the child action received when child state is nil runtime warning. Unsure if the warning is in error, or the sample code, or maybe this is a known thing addressed by the upcoming navigation changes. Feel free to close or disregard if this is known issue for the time being.

stephencelis: @kamcma I think this might just be an order-of-operations thing with the nesting that we hope is solved by some of our navigation explorations. I haven't had time to run the case study, but do you have some steps to reproduce for when I dive in?

kamcma: I'm at the current head of the protocol-beta branch, e00022c. Using Xcode 14b6 on Apple silicon. Launch the SwiftUI case studies app in a simulator, navigate to the lists-load-then-navigate case study, tap the first counter in the list to initiate navigation to the optional counter starting at 1, note runtime warning.

stephencelis: @kamcma That was actually a bug in my translation. I'd swapped the action case paths and used `.self` in the wrong spot! @mbrandonw fixed it the other day here: https://github.com/pointfreeco/swift-composable-architecture/pull/1283/commits/426d5c5932ba831b51cbc79701bbb9ea7891d0a2

Our navigation APIs should help avoid this problem since they will flatten things.

Sources/ComposableArchitecture/Reducer/Reducers/CombineReducers.swift
```diff
@@ -0,0 +1,50 @@
...
+  let reducers: Reducers
+
+  /// Initializes a reducer that combines all of the reducers in the given build block.
+  ///
+  /// - Parameter build: A reducer builder.
+  @inlinable
+  public init(
+    @ReducerBuilderOf<Reducers> _ build: () -> Reducers
+  ) {
+    self.reducers = build()
```
flockoffiles, line 29: I am getting the following compiler error on this line:
CombineReducers.swift:29:10: error: 'let' property 'reducers' may not be initialized directly; use "self.init(...)" or "self = ..." instead

mbrandonw: Can you share which Xcode and Swift version you are using?

flockoffiles: This error only happens if library evolution is enabled (BUILD_LIBRARY_FOR_DISTRIBUTION=YES)

flockoffiles: Basically, we are building ComposableArchitecture into an xcframework (and thus enable the library evolution option). Although I understand that it may not be your intention to support this mode at all, it's still a really nice mode for corporate environments, where direct access to GitHub is disabled.

flockoffiles: I am getting similar errors for other reducers as well (where "self.reducers = ... " line is present).
Not sure if it's a known compiler limitation or a bug though.

flockoffiles: The compiler option we are adding is **-enable-library-evolution**

mbrandonw: Let's start a new discussion for this because I have some questions. Got time?

flockoffiles: Yes. Definitely!

flockoffiles: There is also an older discussion (where I contributed a bit previously) here: https://github.com/pointfreeco/swift-composable-architecture/discussions/890

flockoffiles: As far as I understood, that option had to be enabled to simply produce an xcframework with the proper .swiftinterface file.
As such we don't require binary compatibility between different versions of Composable Architecture. But we still like to be able to compile an xcframework once (on the CI) (for a certain version of Composable Architecture) and then use it for different internal builds.

mbrandonw: As far as I can tell you will just get a warning if you build an xcframework without library evolution letting you know the resulting binary has no compatibility guarantees. But now this discussion is getting a bit long, so I do think we should move to a separate thread. 🙂

flockoffiles: Sure. We can continue on https://github.com/pointfreeco/swift-composable-architecture/discussions/890 just create another discussion if you think it's better...

mbrandonw: I think a separate thread because this is specifically about library evolution. And can you try building without evolution turned on and let me know what happens?

flockoffiles: https://github.com/pointfreeco/swift-composable-architecture/discussions/1442

Sources/ComposableArchitecture/Reducer/Reducers/Scope.swift
```diff
@@ -0,0 +1,278 @@
...
+    case casePath(
+      CasePath<ParentState, Child.State>,
+      file: StaticString,
+      fileID: StaticString,
+      line: UInt
+    )
+    case keyPath(WritableKeyPath<ParentState, Child.State>)
+  }
+
+  public let toChildState: StatePath
```
tgrapperon, line 109: Does this kind of property need to be `public`?
As an aside, I've just discovered the [`@_documentation(visibility: internal)`](https://github.com/apple/swift/pull/60242) attribute which doesn't seem to work yet in Xcode 14.1b3.

Sources/ComposableArchitecture/Reducer/Reducers/Scope.swift
```diff
@@ -0,0 +1,278 @@
...
+  public let toChildAction: CasePath<ParentAction, Child.Action>
+  public let child: Child
```
tgrapperon, line 111:
```suggestion
  @usableFromInline
  let toChildState: StatePath
  @usableFromInline
  let toChildAction: CasePath<ParentAction, Child.Action>
  @usableFromInline
  let child: Child
```
After a quick check, it seems that it's the only occurrence. Other reducer are already proceeding with internal properties.

mbrandonw: I think there's an argument to be made to make the properties of a reducer public. It allows people outside the library to write algorithms against the shape of these types. For example, someone could make their own fusing initializer on `Scope` for when someone does `Scope { Scope { ... } }`.

If there are reducer types that do not have public properties, then maybe we should fix those instead. That is, except for any underscored reducers, which those should be hidden behind `some ReducerProtocol` but we can't do that if we want to support Swift 5.6.

Also, this is how Combine and SwiftUI are designed. `ForEach` exposes its `content` and data, and `Publishers.Map` exposes its upstream and transform.

tgrapperon: No problem if this is indeed deliberate.
A few ones have their properties non-public like `Combine` or `Reduce`, but they both have matching initializers for these properties, so it likely doesn't apply.

mbrandonw: Ah good find, I went ahead and made those public.

Sources/ComposableArchitecture/Documentation.docc/Articles/DependencyManagement.md
```diff
@@ -253,6 +253,25 @@ func testFetchUser() async {
 }
 ```
 
+Often times it is not necessary to create a whole new type to conform to `DependencyKey`. If the
+dependency you are registering is a type that you own, then you can conform it directly to the 
+protocol:
+
+```swift
+private enum APIClient: DependencyKey {
```
tgrapperon, line 261: Maybe a `struct` with a few endpoints would be more explicit than an empty `enum` here?

mbrandonw: Oops yes. Made this change as I was running out the door. Will flesh it out more tomorrow!

PR #1284, opened 2022-08-22
Add previewValue to DependencyKey, and fixed build errors on Xcode 13.

mbrandonw:
Based on a [conversation](https://github.com/pointfreeco/swift-composable-architecture/discussions/1282#discussioncomment-3449849) in #1282, we now allow people to also supply a `previewValue` when conforming to `DependencyKey` which will be used in Xcode previews only. If no `previewValue` is provided it will default to the `testValue`.

I also fixed some errors that prevented TCA from compiling in Xcode 13. This is for the core library only. The case studies and some demo apps will only compile on Xcode 14 since we want to only show the most modern way of interacting with the library in those projects.

### DIFF
--- a/Examples/SpeechRecognition/SpeechRecognition/SpeechClient/Client.swift
+++ b/Examples/SpeechRecognition/SpeechRecognition/SpeechClient/Client.swift
@@ -1,4 +1,5 @@
 import Combine
+import Dependencies
 import Speech
 
 struct SpeechClient {
@@ -14,4 +15,16 @@ struct SpeechClient {
     case couldntStartAudioEngine
     case couldntConfigureAudioSession
   }
+}
+
+extension DependencyValues {
+  var speechClient: SpeechClient {
+    get { self[SpeechClientKey.self] }
+    set { self[SpeechClientKey.self] = newValue }
+  }
+}
+
+enum SpeechClientKey: LiveDependencyKey {
+  static let previewValue = SpeechClient.lorem
+  static let testValue = SpeechClient.unimplemented
 }
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -1,3 +1,5 @@
+import Foundation
+
 // TODO: should we have `@Dependency(\.runtimeWarningsEnabled)` and/or `@Dependency(\.treatWarningsAsErrors)`?
 
 /// A collection of dependencies propagated through a reducer hierarchy.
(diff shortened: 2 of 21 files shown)


Examples/SpeechRecognition/SpeechRecognition/SpeechClient/Client.swift
```diff
@@ -15,3 +16,15 @@ struct SpeechClient {
...
+
+extension DependencyValues {
+  var speechClient: SpeechClient {
+    get { self[SpeechClientKey.self] }
+    set { self[SpeechClientKey.self] = newValue }
+  }
+}
+
+enum SpeechClientKey: LiveDependencyKey {
+  static let previewValue = SpeechClient.lorem
```
mbrandonw, line 28: the speech recognition demo will now use the "lorem" speech client in previews.

Sources/Dependencies/DependencyValues.swift
```diff
@@ -13,23 +15,40 @@
...
       else {
-        let isTesting = self.storage[ObjectIdentifier(IsTestingKey.self)]?.base as? Bool ?? false
-        guard !isTesting else { return Key.testValue }
-        return _liveValue(Key.self) as? Key.Value ?? Key.testValue
+
+        let mode =
+          self.storage[ObjectIdentifier(EnvironmentKey.self)]?.base as? Environment
+          ?? {
+            #if DEBUG
+              if ProcessInfo.processInfo.environment["XCODE_RUNNING_FOR_PREVIEWS"] == "1" {
```
tgrapperon, line 37: I didn't check with this exact setup, but I had warnings while polling this dictionary in async contexts. If so, the value could maybe be stored in a top-level constant property as a workaround.

jshier:
Is there any way (or is it desirable) to automatically fallback from `preview` to `test` when `preview` values aren't provided?

tgrapperon:
@jshier As I'm getting it, this is already what's happening:
```swift
extension DependencyKey {
   public static var previewValue: Value { Self.testValue }
 }
```

jshier:
@tgrapperon A yes, I see. I was just seeing the initialization point.

Sources/Dependencies/DependencyValues.swift
mbrandonw: @tgrapperon Ah interesting, yeah would be a bummer to spew warnings into logs unnecessarily. Just pushed a commit to compute this value a single time.

tgrapperon: It was build-time warnings in my case, because I was touching a mutable shared state. But I was also using it inside by CoreData abstraction to force some system entities to be defined synchronously when previewing. It is possible that in this case, it could be OK. As I'm getting it, this subscript can be hit from any async context, so I guess should have complained too with warnings in place, but I didn't check.

mbrandonw: ah ok, well moving it out to a helper makes the code easier to read anyway, so win-win!

Sources/Dependencies/DependencyValues.swift
```diff
@@ -47,14 +65,38 @@ extension DependencyValues {
...
+private var isPreview = { () -> Bool in
+  #if DEBUG
+    if ProcessInfo.processInfo.environment["XCODE_RUNNING_FOR_PREVIEWS"] == "1" {
+      return true
+    }
+    return false
+  #else
+    return false
+  #endif
+}()
```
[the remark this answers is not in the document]
mbrandonw: yep thanks!

davdroman:
@mbrandonw @stephencelis the shoutout on Twitter made my day, thanks guys :)

stephencelis:
@davdroman We would have tagged you but couldn't find an account!

davdroman:
@stephencelis oh that's quite alright, I don't have one 😄

I used to have an account but I deleted it a few years back. Now I just browse the accounts I like anonymously and sporadically.

PR #1307, opened 2022-08-30
Use `memcmp` to improve `ForEachStore`'s performance

tgrapperon:
This PR's proposes to use `memcmp` to return early if the identifiers of a `ForEachStore` can't be different. This should improve the rendering performance when the `IdentifiedArray`'s identifiers are unchanged.

With the current approach, `ForEachStore`'s internal `ViewStore` compares `OrderedSet<ID>`'s using the canonical `==` to decide if successive values it receives are duplicate or not. `OrderedSet` stack memory consists of two pointers to some internal reference type storage. Because `OrderedSet` is Copy-on-Write, a copy of this private storage is performed when an `OrderedSet` is mutated, and this mutated `OrderedSet` gets new pointers to its own storage. For this reason, **if two `OrderedSet`s are sharing the same pointers, their content can't be different**. Comparing these references is fast and allows us to respond very quickly by the affirmative in this occurrence.
In any case pointers differ, the comparison falls back to `==` where elements are checked one by one, as it is currently performed.

As an example, here are the results when comparing:
- Two different `OrderedSet` values of 10000 `Int`s with the same elements.
- A given `OrderedSet` of 10000 `Int`'s with itself (or a non-mutated "copy").
```
name                         time          std        iterations
----------------------------------------------------------------
Current, same elements       149583.000 ns ±   4.16 %       9211
PR, same elements            149583.000 ns ±   4.31 %       9227

Current, same OrderedSet     149542.000 ns ±   4.63 %       9220
PR, same OrderedSet              42.000 ns ± 247.31 %    1000000
```
This shouldn't affect noticeably the comparison performance of `OrderedSet` with different elements.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
@@ -94,7 +94,10 @@ public struct ForEachStore<
   {
     self.data = store.state.value
     self.content = {
-      WithViewStore(store.scope(state: { $0.ids })) { viewStore in
+      WithViewStore(
+        store.scope(state: { $0.ids }),
+        removeDuplicates: areOrderedSetsDuplicates
+      ) { viewStore in
         ForEach(viewStore.state, id: \.self) { id -> EachContent in
           // NB: We cache elements here to avoid a potential crash where SwiftUI may re-evaluate
           //     views for elements no longer in the collection.
@@ -118,4 +121,13 @@ public struct ForEachStore<
   public var body: some View {
     self.content()
   }
+}
+
+func areOrderedSetsDuplicates<ID: Hashable>(lhs: OrderedSet<ID>, rhs: OrderedSet<ID>) -> Bool {
+  var lhs = lhs
+  var rhs = rhs
+  if memcmp(&lhs, &rhs, MemoryLayout<OrderedSet<ID>>.size) == 0 {
+    return true
+  }
+  return lhs == rhs
 }


iampatbrown:
@tgrapperon Awesome work buddy. This is a great improvement!

jshier:
Would this be something for `swift-collections` as well?

tgrapperon:
This is possible, but it maybe has some very intricate side-effects in the general picture that I'm not prepared to delve into, especially if I would need to discuss it in Swift forums[^1]. I've asked Karoy Lorentey [on Twitter](https://twitter.com/lorentey/status/1564339015238553602) about the change, but the tweet format was maybe too short to detail the use case and implementation[^2]. According to his response, this check should already be performed. However, the two last benchmarks and [the code](https://github.com/apple/swift-collections/blob/main/Sources/OrderedCollections/OrderedSet/OrderedSet%2BEquatable.swift) suggest that this isn't the case.

Because we're falling back to `==` in case of discrepancy, the only case where it would be problematic for us would be if it would return `0` (hence `true`) with the two values containing different elements. Because `OrderedSet` has a value semantic, this is not possible, as it would mean that the same private storage would be containing different elements at the same time.

[^1]: For example, if implemented from the library, one has access to the storage, and one can rely on direct pointer comparison instead of `memcmp`. It's also possible that this can be applied to other types in `swift-collection`, etc.

[^2]: I understand that you can't start to randomly recommend low-level tricks on Twitter when you're in his position. It was already very kind from him to respond.

Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
```diff
@@ -119,3 +122,12 @@ public struct ForEachStore<
     self.content()
   }
 }
+
+func areOrderedSetsDuplicates<ID: Hashable>(lhs: OrderedSet<ID>, rhs: OrderedSet<ID>) -> Bool {
```
stephencelis, line 126:
```suggestion
private func areOrderedSetsDuplicates<ID: Hashable>(lhs: OrderedSet<ID>, rhs: OrderedSet<ID>) -> Bool {
```

stephencelis (approved):
Great find, @tgrapperon! Hopefully it'll result in some noticeable gains for features with large lists of state!

mbrandonw:
Awesome find @tgrapperon!

PR #1310, opened 2022-08-30
Add benchmark for effect merging

stephencelis:
Let's get a simple baseline benchmark in place for merging effects, which is a common operation baked into composing reducers.

### DIFF
--- a/Sources/swift-composable-architecture-benchmark/Common.swift
+++ b/Sources/swift-composable-architecture-benchmark/Common.swift
@@ -1,0 +1,46 @@
+import Benchmark
+
+extension BenchmarkSuite {
+  func benchmark(
+    _ name: String,
+    run: @escaping () throws -> Void,
+    setUp: @escaping () -> Void = {},
+    tearDown: @escaping () -> Void
+  ) {
+    self.register(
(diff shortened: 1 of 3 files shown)


stephencelis:
On my local machine, for reference:

```
name                                        time         std        iterations
------------------------------------------------------------------------------
Effects.Merged Effect.none (create, flat)   11083.000 ns ±  15.38 %     122200
Effects.Merged Effect.none (create, nested) 41042.000 ns ±   6.71 %      33038
Effects.Merged Effect.none (sink)           63667.000 ns ±   6.95 %      21767
```

mbrandonw:
The reason we are adding these benchmarks is because we have found that in many/most of all actions sent into a highly composed application do not produce any real effects (i.e. every layer just returns a `.none`). 

For example, think of a text field in a child feature that is 4 layers deep. Every key stroke sends an action that goes through multiple layers, building up a massively merged `.none` effect where every layer has been additionally erased since reducers must return the `Effect` type.

Turns out that the construction of such an effect can be expensive, and even hitting `.sink` on it can be expensive. So, we are exploring ways to discard `.none` effects without incurring a cost, and even fusing multiple merges to avoid nesting. These benchmarks will give us a basis to understand how things improve with those changes.

mbrandonw (approved):

PR #1317, opened 2022-08-31
Replace UUID.incrementing to newUUIDForTest and remove captured value  from concurrent context in Todos Demo

pitt500:
Hi there, 

Reviewing Todos demo,  I found out that `UUID.incrementing()` is always returning `"00000000-0000-0000-0000-000000000000"`, in other words, the captured value `uuid` is not working correctly. 

This is because we added Sendable to the `uuid` parameter in `AppEnvironment`.

To fix this, I added a few things:
- Added a new action to test two Todos instead of just one. This ensures that we are generating at least two different UUIDs in testAddTodo.
- Added static `newUUIDForTest` method that uses static `uuIdTestCounter` to generate the `UUID` values. uuIdTestCounter starts at 0.
- Added a `setUp` method to initialize `uuIdTestCounter` to zero every time we run a new test.

Take a look and see if it is worth adding it.

Regards,
Pitt

### DIFF
--- a/Examples/Todos/TodosTests/TodosTests.swift
+++ b/Examples/Todos/TodosTests/TodosTests.swift
@@ -6,6 +6,10 @@ import XCTest
 @MainActor
 final class TodosTests: XCTestCase {
   let mainQueue = DispatchQueue.test
+    
+  override class func setUp() {
+    UUID.uuIdTestCounter = 0
+  }
 
   func testAddTodo() async {
     let store = TestStore(
@@ -13,7 +17,7 @@ final class TodosTests: XCTestCase {
       reducer: appReducer,
       environment: AppEnvironment(
         mainQueue: self.mainQueue.eraseToAnyScheduler(),
-        uuid: { UUID.incrementing() }
+        uuid: { UUID.newUUIDForTest }
       )
     )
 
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hey @pitt500, good catch! I think we got a little overzealous in converting some point-free code to use more explicit syntax in order to make the code base friendlier to everyone. Prior to the concurrency merge (#1189) we were specifying that dependency like so:

```swift
uuid: UUID.incrementing
```

That gives the environment a single UUID generating function. Whereas the way it is now:

```swift
uuid: { UUID.incrementing() }
```

...causes us to create a whole new generator every time a UUID is asked for, hence why it never increments.

Would you be interested in updating your PR to minimize the proposed changes so that we just revert those few lines?

pitt500:
> Hey @pitt500, good catch! I think we got a little overzealous in converting some point-free code to use more explicit syntax in order to make the code base friendlier to everyone. Prior to the concurrency merge (#1189) we were specifying that dependency like so:
> 
> ```swift
> uuid: UUID.incrementing
> ```
> 
> That gives the environment a single UUID generating function. Whereas the way it is now:
> 
> ```swift
> uuid: { UUID.incrementing() }
> ```
> 
> ...causes us to create a whole new generator every time a UUID is asked for, hence why it never increments.
> 
> Would you be interested in updating your PR to minimize the proposed changes so that we just revert those few lines?

Hi @mbrandonw , I can revert it to just `uuid: UUID.incrementing`, however, that will generate a warning saying:

`Converting non-sendable function value to '@Sendable () -> UUID' may introduce data races`. Is that a good idea?

mbrandonw:
Hi @pitt500, technically that warning is correct because we haven't proven that closure is safe to use from concurrent contexts, but in the situation of this test everything is run on the `@MainActor`, so it is OK.

Also, once we fully roll out the reducer protocol stuff we will be shipping a thread-safe version of the [UUID generator](https://github.com/pointfreeco/swift-composable-architecture/blob/2c0728ee3e53266708cfef72b12fe9ac86f456e5/Sources/Dependencies/Dependencies/UUID.swift#L127-L149). So these warnings will only be around for a little while.

pitt500:
@mbrandonw Sounds good then, I will close this PR and open a new one with the minimal fix. Thanks!

PR #1323, opened 2022-08-31
Add deprecations to non-view `WithViewStore` initializers

stephencelis:
We've found that `WithViewStore`s overloads can really tax the compiler, so it's time to get prepare to remove them. The vast majority of uses are in SwiftUI views, so we can privilege the helper for that use case, and require a bit of extra work of manually observing a view store via `@ObservedObject` in the other cases.

In the future it would be nice to deprecate `WithViewStore` entirely, but we would want to offer a simpler tool, _e.g._ a property wrapper called `@ObservedStore` that smashes the concepts of `Store` and `ViewStore` together without a lot of boilerplate. This would lessen compiler strain by avoiding escaping closures and generic transforms in the view builder layer. Unfortunately, we still haven't figured out how to get property wrappers to work in a way that we find simple and intuitive.

In the meantime we can hopefully improve performance of the tool we provide by narrowing its focus to the most common builder.

### DIFF
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -20,7 +20,18 @@ import SwiftUI
 /// }
 /// ```
 ///
-/// In UIKit applications a ``ViewStore`` can be created from a ``Store`` and then subscribed to for
+/// It can also be observed directly by views, scenes, commands, and other contexts that support the
+/// `@ObservedObject` property wrapper:
+///
+/// ```swift
+/// @ObservedObject var viewStore: ViewStore<State, Action>
+/// ```
+///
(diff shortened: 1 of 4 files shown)


Sources/ComposableArchitecture/ViewStore.swift
```diff
@@ -45,7 +56,7 @@ import SwiftUI
 /// }
 /// ```
 ///
-/// > Important: The ``ViewStore`` class is not thread-safe, and all interactions with it (and the
```
jshier, line 48: Did you mean to remove the double ticks here? Wasn't it intentional for DocC links? You may want to make these consistent across all doc comments either way, as I see a mix.

mbrandonw (approved):

PR #1325, opened 2022-08-31
Use @StateObject for WithViewStore in iOS 14

mbrandonw:
We currently use `@ObservedObject` in `WithViewStore`, which means anytime the parent view is recreated, a whole new `ViewStore` is created. This causes a lot of view stores to be needlessly created, which also causes `objectWillChange` to be pinged many times since the `removeDuplicates` logic can't be used to its full potential when fresh view stores are being created all over the place.

By swapping out `@ObservedObject` for `@StateObject` we can ensure that the view store is only created a single time for the lifetime of the view, and by keeping view stores alive for longer we can better leverage their `removeDuplicates` work.

For reference, here are the counts of the number of view stores created and `objectWillChange` emissions in our Todos demo app going through the script of:

* Add 5 todos
* Complete the first
* Change the filter to "Active" and then "Completed

…we get the following counts on main:

```
ViewStore.init 40
ViewStore.objectWillChange 56
```

After the changes in this PR, the counts are reduced to:

```
ViewStore.init 7
ViewStore.objectWillChange 23
```

And in our Voice Memos demo application, if we:

* Record a new 4 second memo
* Play the memo back
* Record another 4 second memo
* Play the memo back

…we get the following counts on main:

```
ViewStore.init 171
ViewStore.objectWillChange 255
```

And now those counts are:

```
ViewStore.init 44
ViewStore.objectWillChange 118
```

We think this reduction in objects created should help improve performance and reduce the amount views are recomputed.

### Backwards compatibility 

We want these changes to be backwards compatible with past versions of TCA, which means supporting iOS 13. Since `@StateObject` is iOS 14+, and because `WithViewStore` is highly generic over many types of contexts (see #1323), we had to employ some tricks to make this work. Things will get a lot simpler once we remove the deprecations in #1323.

We hope this is 100% backwards compatible with everyone's code, but because we had to tweak some generic constrains in `ForEachStore` and `SwitchStore` there is a slight chance you get a compiler error if you are depending on the exact shape of those generics. We think this should be _extremely_ rare, but if you run into a problem, please open a discussion.

### DIFF
--- a/Tests/ComposableArchitectureTests/EffectOperationTests.swift
+++ b/Tests/ComposableArchitectureTests/EffectOperationTests.swift
@@ -101,8 +101,16 @@ class EffectOperationTests: XCTestCase {
   func testMergeFuses() async {
     var values = [Int]()
 
-    let effect = Effect<Int, Never>.task { 42 }
-      .merge(with: .task { 1729 })
+    let effect = Effect<Int, Never>.task {
+      try await Task.sleep(nanoseconds: NSEC_PER_SEC / 10)
+      return 42
+    }
+    .merge(
+      with: .task {
+        try await Task.sleep(nanoseconds: NSEC_PER_SEC / 5)
+        return 1729
+      }
+    )
     switch effect.operation {
     case let .run(_, send):
       await send(.init(send: { values.append($0) }))
--- a/.swiftpm/xcode/xcshareddata/xcschemes/ComposableArchitecture.xcscheme
+++ b/.swiftpm/xcode/xcshareddata/xcschemes/ComposableArchitecture.xcscheme
@@ -42,7 +42,7 @@
       </Testables>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Debug"
+      buildConfiguration = "Release"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
       launchStyle = "0"
--- a/.swiftpm/xcode/xcshareddata/xcschemes/ComposableArchitecture.xcscheme
+++ b/.swiftpm/xcode/xcshareddata/xcschemes/ComposableArchitecture.xcscheme
@@ -42,7 +42,7 @@
       </Testables>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Release"
+      buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
       launchStyle = "0"
--- a/Examples/Todos/Todos/Todo.swift
+++ b/Examples/Todos/Todos/Todo.swift
@@ -45,6 +45,6 @@ struct TodoView: View {
       }
       .foregroundColor(viewStore.isComplete ? .gray : nil)
     }
-    .debug()
+    .debug("HELLO")
   }
 }
--- a/Examples/Todos/Todos/Todo.swift
+++ b/Examples/Todos/Todos/Todo.swift
@@ -45,5 +45,6 @@ struct TodoView: View {
       }
       .foregroundColor(viewStore.isComplete ? .gray : nil)
     }
+    .debug()
   }
 }
--- a/Examples/Todos/Todos/Todo.swift
+++ b/Examples/Todos/Todos/Todo.swift
@@ -45,6 +45,5 @@ struct TodoView: View {
       }
       .foregroundColor(viewStore.isComplete ? .gray : nil)
     }
-    .debug("HELLO")
   }
 }
--- a/Examples/Todos/Todos/TodosApp.swift
+++ b/Examples/Todos/Todos/TodosApp.swift
@@ -10,7 +10,8 @@ struct TodosApp: App {
           initialState: AppState(),
           reducer:
             appReducer
-            .debug(),
+//            .debug()
+          ,
           environment: AppEnvironment(
             mainQueue: .main,
             uuid: { UUID() }
--- a/Examples/Todos/Todos/TodosApp.swift
+++ b/Examples/Todos/Todos/TodosApp.swift
@@ -10,8 +10,7 @@ struct TodosApp: App {
           initialState: AppState(),
           reducer:
             appReducer
-//            .debug()
-          ,
+            .debug(),
           environment: AppEnvironment(
             mainQueue: .main,
             uuid: { UUID() }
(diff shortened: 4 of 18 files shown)


Tests/ComposableArchitectureTests/EffectOperationTests.swift
```diff
@@ -101,8 +101,16 @@ class EffectOperationTests: XCTestCase {
...
+    let effect = Effect<Int, Never>.task {
+      try await Task.sleep(nanoseconds: NSEC_PER_SEC / 10)
+      return 42
+    }
+    .merge(
+      with: .task {
+        try await Task.sleep(nanoseconds: NSEC_PER_SEC / 5)
+        return 1729
+      }
+    )
```
mbrandonw, line 113: unrelated to this PR but this should fix the flakey test of these two effects racing.

tgrapperon:
The `@StateObject` approach is very nice. The `AnyView` is unfortunate but I think unavoidable to maintain iOS 13 compatibility with built-in wrappers. I would just like to mention that using `@StateObject` ties the `ViewStore` to the view's identity instead of to the store's instance. This works most of the time, because we (almost?) always use a single closure to scope. Even if the `ViewStore` is not following the same `Store` instance as the one from the view, it usually produces the expected results.
However, if we manage to create an example where the store's behavior changes without changing the view's identity, we can see this phenomenon at play.

For example, if you remplace `CounterView` by the example below, you can check that the first view is not spotting the difference when you toggle the `shouldDouble` flag. It however does with the current version that is in `main`.
```swift
struct CounterView: View {
  let store: Store<CounterState, CounterAction>
  @State var shouldDouble: Bool = false
  var body: some View {
    // Added ---
    WithViewStore(
      store.scope(
        state: shouldDouble ? { 2 * $0.count } : { $0.count }
      )
    ) { Text("\($0.state)") }
    
    WithViewStore(
      store.scope(
        state: { (shouldDouble ? 2 : 1) * $0.count }
      )
    ) { Text("\($0.state)") }
    
    Button(shouldDouble ? "Double" : "Simple") {
      shouldDouble.toggle()
    }
    
    // Current ---
      HStack {
        Button {
          viewStore.send(.decrementButtonTapped)
        } label: {
          Image(systemName: "minus")
        }

        Text("\(viewStore.count)")
          .monospacedDigit()

        Button {
          viewStore.send(.incrementButtonTapped)
        } label: {
          Image(systemName: "plus")
        }
      }
    }
  }
}
```
The function switching is clearly something exotic, and solving it at the root would end up identifying scopes, and thus open the way Store/ViewStores reuses, or even to the creation of a graph. In both cases, way beyond the purpose of this PR.

I don't think there is something inherently wrong with the PR. I just wanted to bring this change of behavior to the attention.

mbrandonw:
@tgrapperon Very interesting, thanks for highlighting that difference! So, if non-TCA state influences the scoping of a `WithViewStore`, we may see a divergence of behavior.

It's a bummer to have that change in behavior, even though it does seem like an edge case, so we should think about it a bit more. Looking more towards the future, we are hoping that we can make stores/scoping a bit lazier so that as soon as the state object is hydrated we never have to do another scoping on that store again.

I played around with this in a way that introduced breaking API changes and it resulted in an even more massive drop off in the counts. But of course, that all hinges on this change.

stephencelis (approved):

mbrandonw:
@tgrapperon Can you think of any times you've needed the behavior you described? Used outside state to influence the scope of a store?

I search for `.scope(` in isowords and all uses of scope that open up a custom closure seem to be pure data transformations based on the input. There is [one case](https://github.com/pointfreeco/isowords/blob/92bc4f8d02562be915fed0fd55d8eac6451ff8aa/Sources/ChangelogFeature/ChangelogView.swift#L170) where we do reference outside state, but it just so happens that outside state is static and so would not be affected by this.

I'm also realizing that any logic you do in these scopes is basically untestable / very hard to test since it is outside the purview of the reducer. So, maybe this is something we should be discouraging people from anyway. I'm wondering if we should proceed with this style and document its behavior.

mbrandonw:
> There is [one case](https://github.com/pointfreeco/isowords/blob/92bc4f8d02562be915fed0fd55d8eac6451ff8aa/Sources/ChangelogFeature/ChangelogView.swift#L170) where we do reference outside state, but it just so happens that outside state is static and so would not be affected by this.

Also that scope closure did _not_ need to be written that way at all. Not sure why we did that! It can be written using only the input of the closure. 🙃

tgrapperon:
@mbrandonw No, I don't think I actually ever needed it. I only constructed it for the purpose of being wrong, as it occurred to me that this was now possible.
Even If I would need something environment dependent from a parent (like a size class for example), I would perform the logic inside the scope closure instead of vending different ones, and it would work as expected.
The only thing that is different is that the `viewStore` in the closure is not the `ViewStore` of the store, but possibly a `ViewStore` of a "ghost" of this store, which is normally not problematic given the way TCA works. Other issues are maybe possible, but I'm not seeing one right now.
The issue is that the weird scoping can happen outside of the `View`, from a parent for example: 
```swift
ChildView(state: horizontalSizeClass == .compact ? { $0.first } : { $0.firstAndSecond })
```
(and even without involving foreign data (but foreign logic)):
```swift
ChildView(state: viewStore.isActive ? { $0.first } : { $0.firstAndSecond })
```
The problem is that this is non local, there is no way to know from the `ChildView`'s standpoint. The closure switch is equivalent to changing the "structural identity" of the scoped store, and this is something that is not exploited (yet) in TCA. If this would be the case, the `ViewStore` instance could be installed into the `ViewState` scoped `Store` that would be itself preserved when scoping from a parent, alleviating the need for a `StateObject`. But this is a very complex topic with deep ramifications. 
I think that, in the mean time, the best is to recommend to avoid the pattern above (which fortunately requires more work than one unique closure that preserves the scoping identity).

tgrapperon:
As long as there is only one closure in the scope, this is fine I think, as everything that needs to be captured will be captured and this won't change during the whole lifetime of the scope (even if the state value changes, the first instance captured by `StateObject` will behave the same as all the next ones, as it has the same function with the same inputs (or similar ones themselves).
Or it is? I now have doubts about what happens when you capture the `viewStore` when scoping from a `WithViewStore` itself. I think it effectively works, as it ultimately flows back to the root `Store`, but I will need to think more about it.

mbrandonw:
We're gonna go ahead and merge this, but everyone please do let us know if you see anything _interesting_ happen while running this code. We're also going to improve the documentation to let people know what they should and shouldn't be doing in `scope` transformations.

iampatbrown:
> We're also going to improve the documentation to let people know what they should and shouldn't be doing in scope transformations.

@mbrandonw Good call. Your implementation looks solid. 

> I searched for .scope( ... realizing that any logic you do in these scopes is basically untestable / very hard to test since it is outside the purview of the reducer. So, maybe this is something we should be discouraging people from anyway. 

I came to the same conclusion after looking through old projects that might have been using a closure like this. In all cases I would have be able to restructure things to avoid running into issues with `@StateObject`

> Looking more towards the future, we are hoping that we can make stores/scoping a bit lazier so that as soon as the state object is hydrated we never have to do another scoping on that store again.

I like the sound of this :)

PR #1328, opened 2022-09-01
Defer action removal until performing last action

iampatbrown:
This PR suggests another possible performance improvement.

Right now we use `.removeFirst()` to access the next action. We could instead use an incrementing index and clear `bufferedActions` at the end of the `send` block. 

There have been [discussions](https://github.com/pointfreeco/swift-composable-architecture/discussions/618#discussioncomment-935699) about using `Deque` instead of `Array`, I think this could be a possible alternative. 

I've held off on this one in the past because I was unclear on the memory implications. With all the performance improvements lately, I thought it would be a good time to revisit it and see what people thought. 

Here's the `Store scoping` benchmarks for `main` vs this pr:

```
name                                   time        std        iterations
------------------------------------------------------------------------
Store scoping.Nested store (main)      5000.000 ns ±  27.54 %     262884
Store scoping.Nested store (this pr)   4167.000 ns ±  38.67 %     300487
```

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -334,8 +334,11 @@ public final class Store<State, Action> {
 
     let tasks = Box<[Task<Void, Never>]>(wrappedValue: [])
 
-    while !self.bufferedActions.isEmpty {
-      let action = self.bufferedActions.removeFirst()
+    var index = self.bufferedActions.startIndex
+    defer { self.bufferedActions = [] }
+    while index < self.bufferedActions.endIndex {
+      defer { index += 1 }
+      let action = self.bufferedActions[index]
       let effect = self.reducer(&currentState, action)
 
       switch effect.operation {


tgrapperon:
Hey @iampatbrown! Great minds think alike, I was experimenting on the same bottleneck that popped-out from profiling. I've arrived to a different approach that uses [a simplified linked list instead](https://github.com/pointfreeco/swift-composable-architecture/compare/main...tgrapperon:swift-composable-architecture:buffered-actions?expand=1). 
Out of curiosity, I've compared `main`, this PR and the linked list:
```
name                                    time           std     iterations
-------------------------------------------------------------------------
Store scoping.Nested store - Baseline   3583.000 ns ±  46.99 %     337603
Store scoping.Nested store - Deferred   3125.000 ns ±  59.69 %     382726
Store scoping.Nested store - LinkedList 2750.000 ns ±  55.60 %     431956
```
So the list is indeed faster, but it increases the surface of the library, so I'll PR independently in a second step, as there are probably things to discuss.
Your PR is a clear and direct win and can probably be merged without reservations.

mbrandonw (approved):
Nice find!

iampatbrown:
> I've implemented a different approach that uses [a simplified linked list](https://github.com/pointfreeco/swift-composable-architecture/compare/main...tgrapperon:swift-composable-architecture:buffered-actions?expand=1) instead. 

@tgrapperon That's quite nice. I've noticed SwiftUI uses linked lists a bit too.

> So the list is indeed faster, but it increases the surface of the library,

I had similar results when I originally compared the two methods. It was over 6 months ago and but don't remember the difference being as significant. I'm sure the recent performance improvements have contributed to this :) I ended up sticking with the `Array` to reduce surface area of the change. 

> I'll PR independently in a second step, as there are probably things to discuss.

Let me know when you do :)

tgrapperon:
I haven't checked, but I wonder if an `ArraySlice<Action>` would have performed similarly to your change, as removing the first element is merely moving the `startIndex`. I'll check out of curiosity!

tgrapperon:
I've just checked with `ArraySlice`: it was performing better than `main`, but very slightly slower than your PR. Out of curiosity, I've checked again with your PR's changes in place, but it doesn't like it at all and we're back in the pre-PR perf (I guess indexed access in the slice goes through a few levels of indirection, which would explain the loss of perf).

>I've noticed SwiftUI uses linked lists a bit too.

Please let me know where at the occasion!

PR #1334, opened 2022-09-04
change AVAudioSession category

entaku0818:
Thank you for many TCA sample projects!
I have a problem in Voicememos that is cant't play and low volume recoding voice.
I changed AVAudioSession setting in my project.
And, I think should change to sample project.


mbrandonw (approved):
Great, thanks for this!

PR #1336, opened 2022-09-05
Add `_StateObject`, an iOS 13 compatible substitute of `StateObject`

tgrapperon:
The latest version of `WithViewStore` now uses a `@StateObject` that holds the `ViewStore` value it initially accessed (#1325). This allows to avoid many `ViewStore` instantiations and state comparisons in some cases.

However, the `@StateObject` property wrapper is only available for iOS 14+ (and other Apple OSes of this generation). Because TCA is currently supporting iOS 13, an availability check is performed to switch over the `StateObject` variant when possible, and fall back to `ObservedObject` like it was the case before. Because of the way result builders are working, this availability check wraps the whole `StateObject` branch in an `AnyView`.

`AnyView` have usually bad press. Their performance in extremal setups (root or leaf view) is usually [on par with fully typed views](https://nalexn.github.io/anyview-vs-group/). However, there are cases where it can seriously degrade performance.

Apparently, a `ScrollView` or a `List` having a nested `ForEach`[^1] can communicate its current visible bounds to it. This seemingly allows `ForEach` to only create and render a subset of its views. This communication seems direct, without involving the `Environment`. However, when this `ForEach` is wrapped in an `AnyView`, the `List/ScrollView` can't see it, and therefore can't communicate its visible bounds to it. As a result, the unconstrained `ForEach` will load all of its hierarchy, which can be consequent (sometimes more than a thousand views). As it does this at each rendered frame, this degrades the performance drastically.

[^1]: Or maybe any `DynamicViewContent`, as `ForEach` is the only built-in type conforming to this protocol.

Because the very generic `WithViewStore` view can be used in almost any position, there were cases where it could have unwillingly uncoupled `List/ScrollView` from their inner `ForEach` by wrapping its `Content` view in some `AnyView` on recent OS.

A `StateObject` is an `ObservedObject` that holds a reference to the `ObservableObject` you pass to it. It captures this object in the form of an escaping autoclosure and creates an instance only on the first effective access in a body.

This PR intends to reimplement `StateObject` using API's compatible with iOS 13. This should not only improve the situation on this OS, but most of all, will allow to avoid the availability check and thus, the `AnyView` wrapping.

This internal property wrapper called `_StateObject` is relatively lean and performs similarly to `StateObject` while exhibiting the same `ViewStore` initialization/deduplication behavior. Unfortunately, conditionally upgrading to `StateObject` on compatible OS is not possible with the way Swift works.

This PR rolls back `WithViewStore` to its recent implementation, and replaces the inner `@ObservedObject` with `@_StateObject`.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/Deprecations.swift
+++ b/Sources/ComposableArchitecture/Internal/Deprecations.swift
@@ -698,10 +698,7 @@ extension ForEachStore {
   where
     Data == [EachState],
     Content == WithViewStore<
-      [ID], (Data.Index, EachAction),
-      _StateObjectViewStore<
-        [ID], (Int, EachAction), ForEach<[(offset: Int, element: ID)], ID, EachContent>
-      >
+      [ID], (Data.Index, EachAction), ForEach<[(offset: Int, element: ID)], ID, EachContent>
     >
   {
     let data = store.state.value
@@ -728,10 +725,7 @@ extension ForEachStore {
   where
     Data == [EachState],
     Content == WithViewStore<
-      [ID], (Data.Index, EachAction),
-      _StateObjectViewStore<
-        [ID], (Int, EachAction), ForEach<[(offset: Int, element: ID)], ID, EachContent>
-      >
+      [ID], (Data.Index, EachAction), ForEach<[(offset: Int, element: ID)], ID, EachContent>
     >,
     EachState: Identifiable,
     EachState.ID == ID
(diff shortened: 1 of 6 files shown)


tgrapperon:
CI failed in the unrelated `TicTacToe/LoginCore/testFlow_DismissEarly_TwoFactor_Integration`.
It passes on my machine.

stephencelis:
@tgrapperon Thanks for thoroughly investigating this and providing a succinct solution that will swap out nicely if/when we decide to drop support for iOS 13. Will merge when it goes green! We've noticed some flakiness in CI lately, and are not sure if it's related to GitHub or changes to TCA. I re-ran your job and hopefully we get a handle on some of these CI failures soon!

stephencelis (approved):

PR #1344, opened 2022-09-06
Fix a few flakey tests

mbrandonw:
Ever since #1312 we've been getting some intermittent test failures. This is happening because we have a few tests that work directly with effects instead of using a `TestStore`, and those tests are leaving behind some artifacts that are being picked up by other tests.

This will not affect anyone using the library. This is only a problem in the library itself. We should probably audit all of those tests to make sure that things are cleaned up, or even better switch them all to use `TestStore` so that everything remains in a closed system.

### DIFF
--- a/Examples/TicTacToe/tic-tac-toe/Tests/TwoFactorSwiftUITests/TwoFactorSwiftUITests.swift
+++ b/Examples/TicTacToe/tic-tac-toe/Tests/TwoFactorSwiftUITests/TwoFactorSwiftUITests.swift
@@ -86,5 +86,7 @@ final class TwoFactorSwiftUITests: XCTestCase {
     await store.send(.alertDismissed) {
       $0.alert = nil
     }
+
+    await store.finish()
   }
 }
--- a/Examples/TicTacToe/tic-tac-toe/Tests/AppCoreTests/AppCoreTests.swift
+++ b/Examples/TicTacToe/tic-tac-toe/Tests/AppCoreTests/AppCoreTests.swift
@@ -124,7 +124,5 @@ final class AppCoreTests: XCTestCase {
     ) {
       $0 = .newGame(NewGameState())
     }
-
-    await store.finish()
   }
 }
--- a/Examples/TicTacToe/tic-tac-toe/Tests/AppCoreTests/AppCoreTests.swift
+++ b/Examples/TicTacToe/tic-tac-toe/Tests/AppCoreTests/AppCoreTests.swift
@@ -54,6 +54,8 @@ final class AppCoreTests: XCTestCase {
     await store.send(.newGame(.logoutButtonTapped)) {
       $0 = .login(LoginState())
     }
+
+    await store.finish()
   }
 
   func testIntegration_TwoFactor() async {
@@ -124,5 +126,7 @@ final class AppCoreTests: XCTestCase {
     ) {
       $0 = .newGame(NewGameState())
     }
+
+    await store.finish()
   }
 }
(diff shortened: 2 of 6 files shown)


Examples/TicTacToe/tic-tac-toe/Tests/TwoFactorSwiftUITests/TwoFactorSwiftUITests.swift
```diff
@@ -86,5 +86,7 @@ final class TwoFactorSwiftUITests: XCTestCase {
     await store.send(.alertDismissed) {
       $0.alert = nil
     }
+
+    await store.finish()
```
rcarver, line 90: I've been seeing flakey tests on Xcode Cloud, and adding this seems to fix it. Should calling `finish()` be required?

mbrandonw: Hey @rcarver, thanks for sharing that example. Again, a bummer :/

Two things though. First, if `.onDrag` doesn't start up any long-living effects, you could do this to ensure your effect finishes:

```
await store.send(.interaction(.onDrag(event.canvasDrag))) {
    $0.root = event.modifiedRoot
    $0.interaction.reorder = .idle
}
.finish()
XCTAssertEqual(renderRequestCount, 2)
```

We also just pushed a commit (0874b77) that hopefully helps a bit. It causes `store.send` to await until effects officially start.

mbrandonw: @rcarver yeah that seems like the same deal as the other wait. More `await Task.yields()` would probably work, or you can await the `finish()` of the task returned by `store.send`.

stephencelis: @rcarver I think the general rule of thumb is if you're testing fire-and-forget effects, you should be `await`-`finish`ing the test store task return from `send` before making those assertions. We should include documentation on this in our testing article.

rcarver: Updates:

* Flakey tests fail at d75d8f7d0930d4f027ea5bc1fad61621e0f3f60a *without* `await finish()`
* Flakey tests pass at d75d8f7d0930d4f027ea5bc1fad61621e0f3f60a *with* `await finish()`

@stephencelis ah, easy enough. The non-determinism is the real issue though. I ran a flakey test 10k times locally and it passed without `await finish()`. On Xcode Cloud it passes maybe 50% of the time?

Would it be more consistent if `await finish()` was required for a fire-and-forget to execute? If so would that cause issue if you're mixing fire-and-forget with long-running effects?

stephencelis: @rcarver The non-determinism is unfortunate, but Swift just doesn't provide the tools to make things deterministic. See this thread:

https://forums.swift.org/t/reliably-testing-code-that-adopts-swift-concurrency/57304

I also don't think we can force the caller to `await task.finish()` because effects are black boxes that can be merged...if you have an idea that can be sketched out, though, please do share!

rcarver: @stephencelis amazing thread, I really appreciate how deep you go on these issues. It's way above my head at this point, but will keep an eye on the patterns now that I know the preferred tools to use. Thanks!

Issue #1349, opened 2022-09-08
"Cannot find type 'Duration' in scope" error in Xcode 14 Release Candidate when targeting macOS


Jeehut:
### Description

I just upgraded to the Xcode 14 Release Candidate and my project no longer builds. I've even setup a new project an only included TCA as a dependency, pointing to the latest stable release, with the same result:
![Screen Shot 2022-09-08 at 10 29 35](https://user-images.githubusercontent.com/6942160/189074513-c48eb274-20ec-4d70-8f75-06113c242a58.png)

I already tried the latest `main` branch, same issue there.

### Checklist

- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues).

### Expected behavior

My project builds without errors.

### Actual behavior

I'm getting an error stating `Sources/ComposableArchitecture/TestStore.swift:277:27: error build: Cannot find type 'Duration' in scope`.

### Steps to reproduce

1. Create a new multiplatform project in Xcode 14 Release Candidate
2. Include TCA as a dependency (using SwiftPM)
3. Set your build target to "Your Mac"
4. Build the project

### The Composable Architecture version information

0.39.1

### Destination operating system

macOS

### Xcode version information

14.0 Release Candidate

### Swift Compiler version information

```shell
swift-driver version: 1.62.8 Apple Swift version 5.7 (swiftlang-5.7.0.127.4 clang-1400.0.29.50)
Target: arm64-apple-macosx12.0
```

Jeehut:
Interestingly, I can build without problems when I set an iOS target, e.g. "iPhone 14". My best guess is that `Duration` is a new type introduced with Swift 5.7, but when I target "My Mac" which is currently running the latest release `macOS 12.5.1`, my macOS system has an older version of Swift and somehow Xcode gets confused and still goes inside the `#if swift(>=5.7)` checks which don't work on a macOS 12 target though. Maybe we have to add a macOS availability check additionally.

Jeehut:
I just opened the TCA project and tried to build itself, which also resulted in the same error. It's confusing to see that there is already a macOS version availability check in this code, but it seems to still fail.

![Screen Shot 2022-09-08 at 10 48 59](https://user-images.githubusercontent.com/6942160/189078639-0bc25977-455c-4758-8a50-6d9600ed1c1d.png)

[...]

PR #1350, opened 2022-09-08
Fix a build error in Xcode 14 when targeting macOS 12

Jeehut:
There seems to be an issue with the new Swift `Duration` type and availability checks on macOS. This is a temporary fix until a proper workaround is found or the bug is fixed by Apple.

Fixes https://github.com/pointfreeco/swift-composable-architecture/issues/1349.


mbrandonw:
Hey @Jeehut, thanks for looking into this so soon after the RC release!

I think a better fix will be #1351 because we do really want to use `Duration` here, and we wouldn't want to have to make a breaking change later once `Duration` is available to macOS.

PR #1355, opened 2022-09-08
Fix release builds for case studies.

mbrandonw:
Somehow the introduction of our own back port of `@StateObject` is causing compile errors in release model when using `WithViewStore` in generic views. For example, when building the case studies target in release mode, the following compiler error happens:

```
Undefined symbols for architecture arm64:
  "_$s22ComposableArchitecture12_StateObjectV7Storage33_BCA98CFE651CB6AE3A3A7598ACD2B887LLCMa", referenced from:
      _$s18SwiftUICaseStudies17DownloadComponentV4bodyQrvg in DownloadComponent.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

It happens in both the reusable download component and favoriting component, both of which use `WithViewStore` in a generic context. We should minimize this a bit and file a bug on the swift repo, but for now we can fix by using an explicit `@ObservedObject`. 🫤

### DIFF
--- a/Tests/ComposableArchitectureTests/EffectCancellationTests.swift
+++ b/Tests/ComposableArchitectureTests/EffectCancellationTests.swift
@@ -197,10 +197,11 @@ final class EffectCancellationTests: XCTestCase {
       DispatchQueue.global(qos: .userInteractive),
       DispatchQueue.global(qos: .utility),
     ]
+    let ids = (1...10).map { _ in UUID() }
 
     let effect = Effect.merge(
       (1...1_000).map { idx -> Effect<Int, Never> in
-        let id = idx % 10
+        let id = ids[idx % 10]
 
         return Effect.merge(
           Just(idx)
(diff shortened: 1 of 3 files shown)


Tests/ComposableArchitectureTests/EffectCancellationTests.swift
```diff
@@ -197,10 +197,11 @@ final class EffectCancellationTests: XCTestCase {
       DispatchQueue.global(qos: .userInteractive),
       DispatchQueue.global(qos: .utility),
     ]
+    let ids = (1...10).map { _ in UUID() }
```
mbrandonw, line 200: not related to this PR, but trying to fix more test leakage. again this doesn't affect users of the library, just the library tests themselves. :/

stephencelis (approved):

PR #1358, opened 2022-09-08
Remove `@StateObject` shim for now

stephencelis:
With the introduction we've noticed a regression when building certain views in release (#1355). So let's revert this change for now till we can come up with an alternative, or till Apple fixes the issue in a future version of Swift.


mbrandonw (approved):
🤬

PR #1360, opened 2022-09-09
allow processing of sync actions resulting from state update

diederich:
Sending an action into the system synchronously from a ViewStore's publisher can result in state updates publishing the wrong, old, value.

This PR fixes it by keeping the `Store` in `sending` state while we publish state updates. Once that's done, we check if there were any new actions coming in, in which case we feed them back and start processing again.


stephencelis (approved):

stephencelis:
Thanks! Very subtle :smile:

PR #1368, opened 2022-09-10
More docs for WithViewStore and new observe argument

mbrandonw:
Spurred by the discussions in #1340 I've added more documentation to both `WithViewStore` and the new `observe` argument in its initializer.

### DIFF
--- a/Sources/ComposableArchitecture/Documentation.docc/Articles/Performance.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Articles/Performance.md
@@ -157,15 +157,15 @@ WithViewStore(
 }
 ```
 
-Alternatively, and recommended, you can introduce a lightweight `ViewState` struct nested inside
-your view whose purpose is to transform the `Store`'s full state into the bare essentials of what
-the view needs:
+Alternatively, and recommended, you can introduce a lightweight, equatable `ViewState` struct
+nested inside your view whose purpose is to transform the `Store`'s full state into the bare
+essentials of what the view needs:
 
 ```swift
 struct AppView: View {
   let store: Store<AppState, AppAction>
   
-  struct ViewState {
+  struct ViewState: Equatable {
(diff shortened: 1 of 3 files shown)


Sources/ComposableArchitecture/Documentation.docc/Articles/Performance.md
```diff
@@ -74,10 +74,92 @@ This means `AppView` does not actually need to observe any state changes. This v
 created a single time, whereas if we observed the store then it would re-compute every time a single
 thing changed in either the activity, search or profile child features.
 
-If sometime in the future we do actually need some state from the store, we can create a localized
-"view state" struct that holds only the bare essentials of state that the view needs to do its
-job. For example, suppose the activity state holds an integer that represents the number of 
-unread activities. Then we could observe changes to only that piece of state like so:
+If sometime in the future we do actually need some state from the store, we can start to observe
```
mbrandonw, line 77: this is an update to the performance article. I realized that first talked about a `selectedTab` piece of state and then later an `unreadActivityCount` state was more natural than the reverse.

Sources/ComposableArchitecture/SwiftUI/WithViewStore.swift
```diff
@@ -186,12 +186,71 @@ public struct WithViewStore<ViewState, ViewAction, Content> {
 // MARK: - View
 
 extension WithViewStore: View where Content: View {
-  /// Initializes a structure that transforms a store into an observable view store in order to
-  /// compute views from state.
+  /// Initializes a structure that transforms a ``Store`` into an observable ``ViewStore`` in order
```
mbrandonw, line 189: I wanted a short explanation of the performance pitfalls directly in the docs for `WithViewStore.init(_:observe:)` so that it's as close to the API as possible. Bummer that it has to be repeated a few times, but I think worth it.

stephencelis (approved):

PR #1369, opened 2022-09-10
Do more to avoid concurrency flakiness in tests

stephencelis:
Unfortunately for us, Swift does not offer a [reliable way to test integrated concurrency](https://forums.swift.org/t/reliably-testing-code-that-adopts-swift-concurrency/57304). Because of this, the Composable Architecture does what it can to insert instructions to the runtime to yield for some time and allow the work of an effect to get kicked off. We do this to try to simplify things for end-TCA users so that they don't need to manually instrument their own dependencies with async streams and channels for simple async work.

It's not perfect, though, especially when CI machines are typically underpowered virtual boxes that perform differently than what we expect locally. There are a couple tests in TCA that are flakey on GitHub CI that we could instrument to prevent flakiness, but instead let's use them to continuously tweak our yield logic.

In the future, hopefully with custom executors and some other tools provided by Swift, we can avoid these workarounds, but for now... 🤞

### DIFF
--- a/Sources/ComposableArchitecture/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestStore.swift
@@ -573,7 +573,9 @@
       if "\(self.file)" == "\(file)" {
         self.line = line
       }
-      await Task.megaYield()
+      // NB: Give concurrency runtime more time to kick off effects so users don't need to manually
+      //     instrument their effects.
+      await Task.megaYield(count: 20)
       return .init(rawValue: task, timeout: self.timeout)
     }
 
@@ -1041,7 +1043,7 @@
   }
 
   extension Task where Success == Never, Failure == Never {
-    static func megaYield(count: Int = 6) async {
+    static func megaYield(count: Int = 10) async {
       for _ in 1...count {
         await Task<Void, Never>.detached(priority: .low) { await Task.yield() }.value
       }


mbrandonw (approved):

diederich:
Thanks for looking into this and making this easier for clients! 🙏 

We were also seeing some flakiness, and while *most often* a single `megaYield` was enough, in 10k runs I saw up to 60 megaYields required (with `count` increased to ~700 ) for a test case to finish work. :-/
To be fair there were multiple layers of `async` involved, but still…

Basically I built a “super-megayield“ 😆 with a separate conditional check and if the condition wasn’t fulfilled, kept calling `megaYield` with increasing `count`s - and log the required iteration count to pass:
```
while !condition {
  iterationCount += 1
  await Task.megaYield(iterationCount * 10)
}
```
You’d think if there’s a condition, you don’t need the `megaYield` at all, though using it gave results more reliable than just waiting for an `XCTExpectation`. I haven’t dug much deeper, but was also wondering if priorities / system load / async overload are contributors.

Have you played with reducing the `Task` `priority` in `megaYield` to `.background` instead of `low`?

stephencelis:
@diederich We're definitely down to tweak this further! If upping the `megaYield` in the library makes things more stable, we'll take that on, especially if it doesn't have an unreasonable impact on test performance. I wonder if you can cook up an example test that requires a lot of yielding?

Unfortunately for us, a lot of our yields are without condition: we just want to wait for an effect to spin up before yielding back to the test, but there's no way for us to really know that the effect has _really_ begun or not.

> Have you played with reducing the `Task` `priority` in `megaYield` to `.background` instead of `low`?

We avoided `background` because depending on the device and other async work being processed, `background` priority tasks might not get picked up for a long time.

diederich:
> I wonder if you can cook up an example test that requires a lot of yielding?

I'll give it a try! Mind that even the case where I saw dozens of `Task.megaYield()` required happened rarely. Maybe once in a thousand tests or less.

> Unfortunately for us, a lot of our yields are without condition: we just want to wait for an effect to spin up before yielding back to the test, but there's no way for us to really know that the effect has really begun or not.

Yeah, makes sense. In that concrete case I added the condition in an overwritten client endpoint in the test.
Which seems OK - and IMHO better that adding them to the public client interfaces (as was mentioned in the swift forums thread).

> We avoided background because depending on the device and other async work being processed, background priority tasks might not get picked up for a long time.

Hehe, right - funny enough, that's basically the argument why I thought it might be good to use 🤣 As in, once *that* returns, any other outstanding Tasks have likely been handled. But I guess you're saying it makes tests run much longer by default?

mbrandonw:
> Hehe, right - funny enough, that's basically the argument why I thought it might be good to use 🤣 As in, once _that_ returns, any other outstanding Tasks have likely been handled. But I guess you're saying it makes tests run much longer by default?

I believe that background tasks may be suspended _indefinitely_ while a device is in lower power mode. But now I can't find any docs that say this, and I'm not sure how often that would even factor into running tests. It may be worth trying out again.

diederich:
> I believe that background tasks may be suspended indefinitely while a device is in lower power mode. 

Ohh - I see. 🙏 That would be unfortunate. 😅 

> But now I can't find any docs that say this, and I'm not sure how often that would even factor into running tests. It may be worth trying out again.

Will give it a try!

PR #1378, opened 2022-09-13
Git-ignore .swiftpm/ directory (and schemes)

stephencelis:


acosmicflamingo:
@stephencelis after updating to 0.44.0, I noticed that "swift-clocks" is appearing as a scheme. Is this intentional or does it need to be part of the changes you had made in this PR?

mbrandonw:
@acosmicflamingo I think this is due to swift-clocks checking in its scheme. Want to open an issue or PR on that repo so that we can track?

mbrandonw:
Oh actually we do ignore the scheme in swift-clocks, but it happened in a 0.1.1 release. Are you currently on 0.1.0? If so, try updating and see if it fixes.

acosmicflamingo:
@mbrandonw whoops! You know, I should really get in the habit of clearing caches and getting the absolute latest. I am using 0.1.1 and it is gone. Yay! Sorry for the false alarm.

PR #1391, opened 2022-09-16
Update example to set badge to the unread count

hyperspacemark:
Just a minor thing I noticed while reading the performance article.


mbrandonw (approved):
Good catch, thanks @hyperspacemark!

PR #1392, opened 2022-09-17
Fix the CaseStudies (UIKit)

Czajnikowski:
It didn't work, at least on my local (`Call to main actor-isolated initializer 'init(store:)' in a synchronous nonisolated context`)😛

Fixed by moving `dataSource` into the `RootViewController`

### DIFF
--- a/Examples/CaseStudies/UIKitCaseStudies/RootViewController.swift
+++ b/Examples/CaseStudies/UIKitCaseStudies/RootViewController.swift
@@ -12,6 +12,7 @@ struct CaseStudy {
   }
 }
 
+@MainActor
 let dataSource: [CaseStudy] = [
   CaseStudy(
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hey @Czajnikowski, good catch!

Instead of nesting everything in `RootViewController`, what do you say about just marking the `let dataSource` as `@MainActor`?

```swift
@MainActor
let dataSource: [CaseStudy] = [
  ...
]
```

That should fix it too, and keep things a bit flatter.

If you make that update to this PR I'll merge.

Czajnikowski:
Tbh I wasn't aware that you can mark a constant with `@MainActor` 🙂 Makes a lot of sense 👌

mbrandonw (approved):
Thanks!

PR #1395, opened 2022-09-19
added Self as default associatedtype value for TestDependencyKey.value

pteasima:
Hey pointfreeguys,

This is a 1 line PR to slightly improve ergonomics of conforming to `DependencyKey` protocol. By choosing a default type for the `Value` associated type, we can omit the typealias when defining a key with only a `liveValue`.

The choice of default type is kinda arbitrary, but imho its quite common (at least in my codebases, hope Im not alone 😅) to want to use `Self`. E.g. `struct API: Dependency { static var liveValue: Self { .init() } }`. For that reason, this is an actual improvement to the everyday ergonomics.

Afaik the change is purely additive, I can't think of any existing valid Swift code that breaks as a result of the change. Can't speak for future Swift code 🤷‍♂️😅 but I think we should be fine.

I've started a [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions/1396) alongside this draft PR. Before merging, I would personally remove the newly added test, unless you feel like we should keep it.

### DIFF
--- a/Tests/ComposableArchitectureTests/DependencyKeyTests.swift
+++ b/Tests/ComposableArchitectureTests/DependencyKeyTests.swift
@@ -30,4 +30,17 @@ final class DependencyKeyTests: XCTestCase {
 
     XCTAssertEqual(42, Key.previewValue)
   }
+  
+  //TODO: intentionally long naming. I think we don't need this test. Its just there to prove that the improved TestDependencyKey protocol now compiles without a typealias when `Value == Self`
+  func testDependencyKeyThatsItsOwnValueDefaultValues() {
+    struct Key: DependencyKey {
+//      typealias Value = Int
+      static let liveValue = Self()
+      var intValue: Int = 42
+    }
+
+    XCTAssertEqual(42, Key.previewValue.intValue)
+    XCTAssertEqual(42, Key.testValue.intValue)
+  }
+
 }
--- a/Tests/ComposableArchitectureTests/DependencyKeyTests.swift
+++ b/Tests/ComposableArchitectureTests/DependencyKeyTests.swift
@@ -30,17 +30,4 @@ final class DependencyKeyTests: XCTestCase {
 
     XCTAssertEqual(42, Key.previewValue)
   }
-  
-  //TODO: intentionally long naming. I think we don't need this test. Its just there to prove that the improved TestDependencyKey protocol now compiles without a typealias when `Value == Self`
-  func testDependencyKeyThatsItsOwnValueDefaultValues() {
-    struct Key: DependencyKey {
-//      typealias Value = Int
-      static let liveValue = Self()
-      var intValue: Int = 42
-    }
-
-    XCTAssertEqual(42, Key.previewValue.intValue)
-    XCTAssertEqual(42, Key.testValue.intValue)
-  }
-
 }
--- a/Sources/Dependencies/DependencyKey.swift
+++ b/Sources/Dependencies/DependencyKey.swift
@@ -40,7 +40,7 @@ extension DependencyKey {
 /// See ``DependencyKey`` to define a static, default value for the live application.
 public protocol TestDependencyKey {
   /// The associated type representing the type of the dependency key's value.
-  associatedtype Value
+  associatedtype Value = Self
   // NB: This associated type should be constrained to `Sendable` when this bug is fixed:
   //     https://github.com/apple/swift/issues/60649
 


Tests/ComposableArchitectureTests/DependencyKeyTests.swift
```diff
@@ -30,4 +30,17 @@ final class DependencyKeyTests: XCTestCase {
...
+    struct Key: DependencyKey {
+//      typealias Value = Int
+      static let liveValue = Self()
+      var intValue: Int = 42
+    }
+
+    XCTAssertEqual(42, Key.previewValue.intValue)
+    XCTAssertEqual(42, Key.testValue.intValue)
+  }
+
```
mbrandonw, line 45:
```suggestion
```

mbrandonw (approved):
Hey @pteasima, chatted with Stephen and we think this is a good change to bring in.

It is worth calling out a gotcha that is technically possible even without your change, but becomes maybe a little easier to run into with this new style.

The problem is if you naively conform the `liveValue` using a computed property rather than a `static let`:

```swift
extension MyClient: DependencyKey {
  static var liveValue: Self {
    Self(...)
  }
}
```

This will cause a new `MyClient` value to be constructed every time you reference it in your reducer, and if your client manages any internal, mutable state that will be a bad thing.

So what you really need to do is this:

```swift
extension MyClient: DependencyKey {
  static let liveValue: Self = { 
    Self(...)
  }()
}
```

It's hopefully not a huge deal, and we plan on writing up docs to make it more clear to people that `liveValue`/`testValue`/etc should almost always be shared, static values and not computed properties.

Thanks for opening up this PR!

pteasima:
Interesting gotcha with the getter being called repeatedly. I think SwiftUI.Environment only loads it once into some internal cache (But I only really do stateless "structs with closures", so maybe I just never noticed).

Thanks for merging the PR!

mbrandonw:
> I think SwiftUI.Environment only loads it once into some internal cache

I'd be interested in knowing if that's true. If it is, maybe it's no big deal if they keep everything just to the main thread.

The reason we don't want to do an internal cache is because that will start to bloat `DependencyValues`, and that value gets copied every time a `Task` is created (for the task locals).

> (But I only really do stateless "structs with closures", so maybe I just never noticed).

Just to make sure we're on the same page, [this](https://github.com/pointfreeco/swift-composable-architecture/blob/bcdcb3d7091d57a3e9cf5370b345a9309d205d39/Examples/SpeechRecognition/SpeechRecognition/SpeechClient/Live.swift#L6-L7) is an example of what I'm talking about. If you need any kind of internal data inside your dependency you will need to make sure it's only created a single time. I think that's a pretty common situation.

pteasima:
Thanks for the clarification. I think we're good.

I'll let you know if I get a chance to confirm SwiftUI behavior.

tgrapperon:
About default `Environment` values, it seems that they're at least cached by `@Environment`. For example:
```swift
struct Key: EnvironmentKey {
  static var defaultValue: Int {
    print("defaultValue")
    return 42
  }
}

extension EnvironmentValues {
  var int: Int {
    get { self[Key.self] }
    set { self[Key.self] = newValue }
  }
}

struct ContentView: View {
  @State var tag = 0
  var body: some View {
    VStack {
      KeyView(tag: tag)
      Button("Tap") {
        tag += 1
      }
    }
    .padding()
  }
}

struct KeyView: View {
  var tag: Int
  @Environment(\.int) var int
  var body: some View {
    Text("\(int + tag)")
  }
}
```
Displaying the view prints 3 "defaultValue" when the view appears, but none after that, even while interacting with the button. If one adds `.id(tag)` to `KeyView(tag.tag)`, one resets the state of this view each time, and an additional "defaultValue" gets printed at each tap.

So it seems that `@Environment` caches the value it gets, which makes sense as it's diffing it to avoid `body`'s runs when it doesn't change (and that's also why they wrap functions-like dependencies in structs with `callAsFunction`).
I don't know what happens at the `EnvironmentValue`'s storage level though, but it doesn't seem they do anything special to keep the default values around, and doing so would probably make `@Environment` diffing less efficient, as it will have to compare the concrete default values instead of `nil`. Maybe they abstract this storage too, but I didn't spot anything like that the last time I inspected `EnvironmentValues` using reflection.

stephencelis:
We're open to providing a cache to avoid this gotcha, but we don't really see a simple way of doing so without rethinking our entire design.

The current design leverages `@TaskLocal` and we get thread-safety and scoping for free. If we were instead to have a mutable collection of state that caches values as they are accessed, we'd need to do some/all of the following:

1. We'd need to box the dependency value storage in some way for these mutations to be preferred
2. We'd need to add some copy-on-write logic to allow for dependency values to be treated as a value type
3. We'd need to lock access to this storage for it to continue to be thread-safe
4. We'd need to change storage to be a stack of dictionaries (or dictionary of stacks) to ensure that upstream, memoized values can be provided to downstream, scoped values

That may be it, or there may be more complications I'm not predicting.

All of the above could come with performance issues and/or introduce potential bugs. That doesn't mean it's not worth exploring, but we're not sure it is a priority for release, and there're a lot of other things on our plate right now 😅

If anyone does explore this, we're definitely interested in what they find!

PR #1399, opened 2022-09-19
Fail if testValue is invoked without providing implementation

mbrandonw:
If the `testValue` for a dependency is used in `TestStore` without an explicit value being provided we will allow it to cascade to the `previewValue` but we will also perform an `XCTFail`.

This makes it easy to get started with `DependencyKey` since you only need to provide a `liveValue`, but as soon as you try writing a test you will get a loud failure letting you know of the proper way to write tests that access dependencies.


Package.swift
```diff
@@ -60,6 +60,13 @@ let package = Package(
         .product(name: "XCTestDynamicOverlay", package: "xctest-dynamic-overlay"),
       ]
     ),
+    .testTarget(
+      name: "DependenciesTests",
+      dependencies: [
+        "ComposableArchitecture",
+        "Dependencies"
+      ]
+    ),
```
mbrandonw, line 69: Moved dependency-related tests into its own test target.

Sources/Dependencies/DependencyKey.swift
```diff
@@ -20,14 +22,6 @@ public protocol DependencyKey: TestDependencyKey {
...
 }
 
-extension DependencyKey {
-  /// A default implementation that provides the ``liveValue`` to Xcode previews.
-  public static var previewValue: Value { Self.liveValue }
-
-  /// A default implementation that provides the ``liveValue`` to tests.
-  public static var testValue: Value { Self.liveValue }
-}
-
```
mbrandonw, line 30: Moved this stuff down below since it got bigger and is just distracting if all you want to see are the protocol definitions.

Tests/DependenciesTests/DependencyValuesTests.swift
```diff
@@ -0,0 +1,33 @@
...
+final class DependencyValuesTests: XCTestCase {
+  func testMissingLiveValue() {
+    enum Key: TestDependencyKey {
+      static let testValue = 42
+    }
+
+    var line = 0
+    XCTExpectFailure {
+      $0.compactDescription == """
+        A dependency at DependenciesTests/DependencyValuesTests.swift:\(line) is being used in a \
```
mbrandonw, line 13: realized we didn't have any test coverage on this failure, so adding it.

stephencelis (approved):
Nice.

PR #1408, opened 2022-09-23
Run dependencies tests on CI.

mbrandonw:


Makefile
```diff
@@ -8,43 +8,24 @@ default: test-all
...
-		-scheme ComposableArchitecture \
-		-destination generic/platform=watchOS
+	for scheme in ComposableArchitecture Dependencies; do \
+	  for platform in "$(PLATFORM_IOS)" "$(PLATFORM_MACOS)" "$(PLATFORM_TVOS)" "generic/platform=watchOS"; do \
+			xcodebuild test \
+				-workspace ComposableArchitecture.xcworkspace \
+				-scheme $$scheme \
+				-destination platform="$$platform"; \
+	  done; \
+	done
```
mbrandonw, line 18: I'm not very good with makefile but think this is easier to deal with than listing everything out manually.

iampatbrown: @mbrandonw Can -configuration be incorporated into it as well?

mbrandonw: Yeah it def could, but we actually decided to keep debug/release as separate steps in CI so that it's easier to see when a failure happened due to a release problem vs a common test failure.

iampatbrown: Good call :)

stephencelis (approved):

PR #1443, opened 2022-10-04
Fix compiler errors for library evolution.

mbrandonw:
While we do not support library evolution in TCA, it can be helpful to build for evolution as long as you know what you're doing (see #1442). Apparently the `main` branch builds just fine with library evolution on, but there are a few errors to fix in the protocol branch. I had to do a few weird things to make things build, but maybe others have better ideas.


.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         xcode: ['13.4.1', '14.0']
-        config: ['debug', 'release']
+        config: ['debug', 'release', 'evolution']
```
mbrandonw, line 23: let's start building for evolution in CI.

Sources/ComposableArchitecture/Reducer/Reducers/BindingReducer.swift
```diff
@@ -5,7 +5,12 @@ public struct BindingReducer<State, Action>: ReducerProtocol
 where Action: BindableAction, State == Action.State {
   /// Initializes a reducer that updates bindable state when it receives binding actions.
   @inlinable
-  public init() {}
+  public init() {
+    self.init(internal: ())
+  }
+
+  @usableFromInline
+  init(internal: Void) {}
```
mbrandonw, line 13: So I guess with library evolution you can only call out to other `init`s when marking the `init` as `@inlinable`. So, to work around that I need to make an internal, `@usableFromInline` init, and then call out to that one from the public `init`. And on top of that, for `init`s that don't take any arguments I need to do something to disambiguate the initializers, so I passed a void value.

Please someone let me know if there's a better way.

Sources/ComposableArchitecture/Reducer/Reducers/CombineReducers.swift
```diff
@@ -26,7 +26,12 @@ public struct CombineReducers<Reducers: ReducerProtocol>: ReducerProtocol {
...
     @ReducerBuilderOf<Reducers> _ build: () -> Reducers
   ) {
-    self.reducers = build()
+    self.init(internal: build())
+  }
+
+  @usableFromInline
+  init(internal reducers: Reducers) {
+    self.reducers = reducers
   }
```
mbrandonw, line 35: same situation here, but since the `init` takes an argument I can disambiguate by giving it a different external argument name.

flockoffiles:
It builds fine now! You are SUPER awesome! :-)

Sources/ComposableArchitecture/Reducer/Reducers/Reduce.swift
```diff
@@ -6,12 +6,19 @@ public struct Reduce<State, Action>: ReducerProtocol {
   @usableFromInline
   let reduce: (inout State, Action) -> Effect<Action, Never>
 
+  @usableFromInline
+  static func create(
+    internal reduce: @escaping (inout State, Action) -> Effect<Action, Never>
+  ) -> Self {
+    Self(reduce)
+  }
+
```
mbrandonw, line 15: can't use the internal initializer trick I did in other spots because it causes ambiguity in tests due to `@testable import`. Still feel like there must be a better way...

mbrandonw:
There are still some things to figure out because now with the additional internal `init`s we're getting a bunch of ambiguity errors in tests. We can update all of those call sites to not use trailing closures so that we can name the argument, but I'm hoping there's a better way.

tgrapperon:
Please note that each `@usableFromInline` annotation makes the property/function part of the public ABI. It is very likely that many of them were not required to be exposed by `@inlinable` functions, so it would probably be safer to audit them before committing to support library evolution. As they directly depend on the `@inlinable` surface, this one should probably be checked too.
I think it would probably be safer to defer the official support to v1.0. It doesn't mean that one can't prepare for it though.
The workaround for binary framework providers would be to statically link the TCA version they ship with if possible.

mbrandonw:
@tgrapperon We are not making any promises of binary compatibility with this PR. All we have done is make it possible for one to build with library evolution if they choose to, and they are taking responsibility for making sure to wield it correctly. We still reserve the right to make binary incompatible changes in any future release of TCA.

tgrapperon:
@mbrandonw Ha, OK, I've read the original use-case, which is for private use and not to redistribute the binary. I have no problem with this change then.

Tests/ComposableArchitectureTests/StoreTests.swift
```diff
@@ -32,14 +32,14 @@ final class StoreTests: XCTestCase {
 
       enum Action { case start, end }
 
-      let reducer = Reduce<Void, Action> { _, action in
+      let reducer = Reduce<Void, Action>({ _, action in
```
mbrandonw, line 35: All of these changes in StoreTests.swift are just to help swift out with the ambiguity errors. This only happens because of the `@testable import`. Without that it is not ambiguous.

Sources/ComposableArchitecture/Reducer/ReducerBuilder.swift
```diff
@@ -115,7 +115,7 @@ public enum ReducerBuilder<State, Action> {
     @usableFromInline
     let wrapped: Wrapped?
 
-    @inlinable
+    @usableFromInline
```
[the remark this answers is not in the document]
stephencelis: I think we want to annotate with both if possible. Since `@usableFromInline` calls are not unlined unless they also contain the other annotation. Wanna try adding both and seeing if things still build?

Sources/ComposableArchitecture/Reducer/Reducers/Observe.swift
```diff
@@ -6,10 +6,17 @@ public struct _Observe<Reducers: ReducerProtocol>: ReducerProtocol {
   ///
   /// - Parameter build: A reducer builder that has access to the current state and action.
   @inlinable
-  public init<State, Action>(
-    @ReducerBuilder<State, Action> _ build: @escaping (State, Action) -> Reducers
+  public init(
+    @ReducerBuilderOf<Reducers> _ build: @escaping (Reducers.State, Reducers.Action) -> Reducers
```
stephencelis, line 10: We should maybe have a test to make sure this builds. I feel like I had to shadow generics like this for `CombineReducers` to work in some cases, so not sure if the same applies here or if it was just a mistake.

stephencelis, line 10: We should maybe have a test to make sure this builder builds in practice. I feel like I had to shadow generics like it was in the before diff for `CombineReducers` to work in some cases, so not sure if the same applies here or if it was just a mistake.

Sources/ComposableArchitecture/Reducer/ReducerBuilder.swift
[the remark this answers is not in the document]
mbrandonw: Unfortunately not. Adding both produces a warning in regular builds and the same error as before in library evolution builds.

stephencelis (commented):
Nice! Thanks for looking into fixes.

Just a few small thoughts. Also might want to try a pass at some of the more bizarre internal-but-usable-from-inline initializers, since those interfaces will basically be frozen for products built against library evolution binaries.

PR #1448, opened 2022-10-06
ViewStore.init(_:observe:)

mbrandonw:
A month ago we soft deprecated `WithViewStore(_:)` in favor of an explicit `WithViewStore(_:observe:)` that tries to make it more public that you are are observing data with this view. You can see the PR here #1339 and discussion here #1340.

One month later and we are still happy with this change, so we think it's appropriate to bring to `ViewStore` too, which has the same problem.

If anyone wants to share their own experiences with the change, then please do so! We would love to hear more feedback, as we plan on moving from soft deprecated to actually deprecated sometime soon.


tgrapperon:
I'm slightly concerned by the forced hard deprecation of `State` and `Action` that will generate warnings for folks who have created helpers/wrappers. If the fixes are pretty straightforward, the change `State -> ViewState` may look like a non-trivial operation at first sight. It could be worth adding a `message:` stating that no additional change is required to the deprecation annotation for these typealiases.

mbrandonw:
True, I suppose we can soft deprecate that too, though we went with [hard deprecation](https://github.com/pointfreeco/swift-composable-architecture/blob/a0065688fe66204ec150fb76c2927212158b6539/Sources/ComposableArchitecture/Internal/Deprecations.swift#L389-L397) for `WithViewStore` so theoretically someone out there has already come across this.

As far as giving a custom message, I believe the default message is pretty clear:

> ⚠️ 'State' is deprecated: renamed to 'ViewState'

Is there something else we should be calling out?

tgrapperon:
My concern is that the `(State) -> ViewState` is an operation that already exists when dealing with `(With)ViewStore`, and the renaming from `State` to `ViewState` may let think that this operation is performed, or involved in some sort. In other words, that there's more to think about than simply renaming. As the other variants are soft deprecated, it is likely that users will encounter these hard ones first and may suppose that one should now initialize the `ViewStore` with a `Store<ViewState, ViewAction>` for example.
I would have imagined something like
> ⚠️ 'State' is deprecated: renamed to 'ViewState'. This change was performed to avoid the naming collision with the `Store<State, Action>` that this `ViewStore` observes.

Of course, written in proper english. I feel that this would make clearer that this operation is merely a naming change, without further implications. But we're in the bike-shedding area, and I'm maybe over-cautious.

stephencelis (approved):

PR #1449, opened 2022-10-06
Fix missing parameter in code sample

juliensagot:
The `number` closure parameter was erroneously deleted after a migration of the code sample.
This PR adds it back.


mbrandonw:
Oops, nice catch!

Currently this PR is targeting the library-evolution PR (#1443) but the problem is actually on the protocol-beta PR (#1283). If you can retarget that branch I will get this merged in.

PR #1450, opened 2022-10-06
Fix missing parameter in code sample

juliensagot:
The `number` closure parameter was erroneously deleted after a migration of the code sample.
This PR adds it back.

Sorry I had to recreate yet another PR because I couldn't re-target the branch without adding a bunch of unrelated commits 😅


mbrandonw (approved):
Great, thanks again for catching this!

PR #1460, opened 2022-10-10
Remove dump calls

jaanussiim:
I'm guessing this is why it got quite noisy in latest commit :)


mbrandonw (approved):
Haha, oops! Good catch! Glad you saw this before we merged into main. 🙂

PR #1466, opened 2022-10-11
Loosen platform requirements for Dependencies library

stephencelis:
Dependencies is not currently usable outside Apple platforms due to unrestricted use of platform-only APIs. This PR should guard their availability a bit better and allow for compilation on Linux at the very least, allowing Dependencies code to be written in multiplatform libraries.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift
+++ b/Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift
@@ -26,7 +26,7 @@ func runtimeWarn(
           message
         )
       #else
-        fputs("\(formatter.string(from: Date())) [\(category)] \(message)", stderr)
+        fputs("\(formatter.string(from: Date())) [\(category)] \(message)\n", stderr)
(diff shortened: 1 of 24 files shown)


mbrandonw (approved):

Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift
```diff
@@ -1,15 +1,50 @@
...
+      #if canImport(os)
+        os_log(
+          .fault,
+          dso: dso,
+          log: OSLog(subsystem: "com.apple.runtime-issues", category: category),
+          "%@",
+          message
+        )
+      #else
+        fputs("\(formatter.string(from: Date())) [\(category)] \(message)", stderr)
```
pyrtsa, line 29: Missing a newline after `\(message)`?

stephencelis: Thanks!

PR #1471, opened 2022-10-12
Add a `EffectTask<Action>` typealias for `Effect<Action, Never>` and rename `Effect` to `EffectPublisher`

tgrapperon:
Specifying `Never` in `Effect<Action, Never>` is slightly inconvenient, and it happens the majority of the times `Effect` is used.

Without changing the shape of `Effect`, this PR introduces:
```swift
typealias EffectOf<Action> = Effect<Action, Never>
```
to remove the mental burden of `Never` and save a few keystrokes.

Now that TCA is slowly moving away from Combine, most dependencies are modeled using `async/await`, and `Effect` is becoming more and more focused on reducer's products. I think that a one-generic `Effect<Action>` was envisaged at some point (there are a few traces in the documentation that are fixed by this PR), but it would be a source-breaking change. If it ever happens someday, this typealias should allow codebases to have smoothly migrated toward this shape already, easing the transition.

An alternative would be to define `EffectOf<R:ReducerProtocol> = Effect<R.Action, Never>`. In both cases, they make sense, as one can see the reducer `Reducer` producing an `EffectOf<Action>` when it receives an `Action`, or an `EffectOf<Reducer>`.

My personal preference goes toward the solution proposed in this PR, which is mentally more lightweight, but the `Reducer` one also packs more information, so that could be useful at some point (although I don't see any real application right now).

What do you think about it?

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-Lifecycle.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-Lifecycle.swift
@@ -19,8 +19,8 @@ struct LifecycleReducer<Wrapped: ReducerProtocol>: ReducerProtocol {
   }
 
   let wrapped: Wrapped
-  let onAppear: Effect<Wrapped.Action, Never>
-  let onDisappear: Effect<Never, Never>
+  let onAppear: EffectOf<Wrapped.Action>
+  let onDisappear: EffectOf<Never>
 
   var body: some ReducerProtocol<Wrapped.State?, Action> {
     Reduce { state, lifecycleAction in
@@ -45,8 +45,8 @@ extension LifecycleReducer.Action: Equatable where Wrapped.Action: Equatable {}
 
 extension ReducerProtocol {
   func lifecycle(
-    onAppear: Effect<Action, Never>,
-    onDisappear: Effect<Never, Never> = .none
+    onAppear: EffectOf<Action>,
+    onDisappear: EffectOf<Never> = .none
   ) -> LifecycleReducer<Self> {
     LifecycleReducer(wrapped: self, onAppear: onAppear, onDisappear: onDisappear)
   }
--- a/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-Lifecycle.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-Lifecycle.swift
@@ -19,8 +19,8 @@ struct LifecycleReducer<Wrapped: ReducerProtocol>: ReducerProtocol {
   }
 
   let wrapped: Wrapped
-  let onAppear: EffectOf<Wrapped.Action>
-  let onDisappear: EffectOf<Never>
+  let onAppear: EffectTask<Wrapped.Action>
+  let onDisappear: EffectTask<Never>
 
   var body: some ReducerProtocol<Wrapped.State?, Action> {
     Reduce { state, lifecycleAction in
@@ -45,8 +45,8 @@ extension LifecycleReducer.Action: Equatable where Wrapped.Action: Equatable {}
 
 extension ReducerProtocol {
   func lifecycle(
-    onAppear: EffectOf<Action>,
-    onDisappear: EffectOf<Never> = .none
+    onAppear: EffectTask<Action>,
+    onDisappear: EffectTask<Never> = .none
   ) -> LifecycleReducer<Self> {
     LifecycleReducer(wrapped: self, onAppear: onAppear, onDisappear: onDisappear)
   }
--- a/Sources/ComposableArchitecture/Documentation.docc/ComposableArchitecture.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/ComposableArchitecture.md
@@ -60,7 +60,7 @@ day-to-day when building applications, such as:
 ### State management
 
 - ``ReducerProtocol``
-- ``EffectTask``
+- ``EffectPublisher``
 - ``Store``
 - ``ViewStore``
 
--- a/Sources/ComposableArchitecture/Documentation.docc/ComposableArchitecture.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/ComposableArchitecture.md
@@ -60,7 +60,7 @@ day-to-day when building applications, such as:
 ### State management
 
 - ``ReducerProtocol``
-- ``Effect``
+- ``EffectTask``
 - ``Store``
 - ``ViewStore``
 
--- a/Sources/ComposableArchitecture/Documentation.docc/Extensions/Deprecations/EffectDeprecations.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Extensions/Deprecations/EffectDeprecations.md
@@ -14,12 +14,7 @@ Avoid using deprecated APIs in your app. Select a method to see the replacement 
 
 ### Cancellation
 
-- ``EffectPublisher/cancel(ids:)-9tnmm``
-
-### Composition
-
-- ``EffectPublisher/concatenate(_:)-3awnj``
-- ``EffectPublisher/concatenate(_:)-8x6rz``
+- ``EffectPublisher/cancel(ids:)-8q1hl``
 
 ### Testing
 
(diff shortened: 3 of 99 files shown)


JacksonUtsch:
This sets in stone no evolution to add support of failure cases in reducers (which is the current direction anyhow 😕). Is TCA going to use the `Effect` type long term or opt for a strictly async return type once `Combine` is removed?

stephencelis:
> This sets in stone no evolution to add support of failure cases in reducers (which is the current direction anyhow 😕). Is TCA going to use the `Effect` type long term or opt for a strictly async return type once `Combine` is removed?

Given our current modeling of the reducer, `Effect<Action>` is definitely the most natural direction to take things. This doesn't mean we're not open to exploring how reducer "failure" could work in general, _e.g._ what it would mean to add `throws` to the signature, or what it would mean to allow effects to throw. However, previous discussion has been pretty abstract, and neither Brandon nor I understand how it could be motivated in practice yet, especially without real world uses or demonstrations via modification of the existing case studies/demos.

Your [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions/880) and [implementation](https://github.com/pointfreeco/swift-composable-architecture/pull/886) definitely _sound_ interesting, but they lacked updated sample code and real examples that demonstrate what the changes bring to the table, and so it was difficult for me to truly comprehend.

One thing that has changed since your explorations is that the new `Effect.task` and `Effect.run` modifiers _are_ throwing, giving you some flexibility especially around cancelling async work, but if they throw something other than cancellation it is up to you to handle things, or you will get runtime warnings.

JacksonUtsch:
@stephencelis That is fair. It is quite an extensive change. It is a breaking change unless you can assume a generic type unless explicitly specified.

I have the strong preference of utilizing a result type rather than `throws` for type validation.

Not sure how to best demonstrate with sample code but will see if I can come up with something. 🙏

stephencelis:
@tgrapperon Thanks for this! @mbrandonw and I decided to flesh out our current vision for 1.0 here: https://github.com/pointfreeco/swift-composable-architecture/discussions/1477

This includes a plan for simplifying the `Effect` type. In the discussion we propose `EffectTask<Action>` as the migratory placeholder instead of `EffectOf`. Would you like to make that change here? If you'd like, you can also take a stab at the associated changes for that release, like the `EffectPublisher` rename. If not, no worries! We'll try to tackle after merging this :smile:

Let us know what you think! Or if you have any questions/concerns.

stephencelis:
(Oh, and if you do have things you wanna share, @JacksonUtsch, sorry to put more pressure on you to do so before 1.0, though if you don't have time to explore, there's always 2.0 😆)

tgrapperon:
I renamed `EffectOf` to `EffectTask`, and renamed `Effect` to `EffectPublisher` according to #1477.

I had already updated all occurrences of `Effect<_, Never>` to `EffectTask` in the documentation. 

I've gone one step further by updating naked `Effect` occurrences to `EffectTask` or `EffectPublisher` according to the context. Depending on the duration of the transition to 1.0, I feel that it is better to have a documentation that is coherent with what users are experiencing, especially onboarding ones. One can easily spot these occurrences by searching for `'``EffectTask``'` and `'``EffectPublisher``'` and replace by `Effect` if needed.

I made the choice of seeing `Effect(value:)` as an `EffectTask` given the fact that it is often (over)used as the return value of a `Reducer`, even if this value's implementation is pure Combine.

I didn't update one occurrence of `Effect` in its own documentation, as well as the ones from `README`. Any general reference like "effect" was also preserved. 

I've replaced all internal uses of `Effect` to `EffectPublisher`, as if `Effect` was already hard-deprecated. When possible, `EffectTask` was used.

`Effect` is soft deprecated. It means that Xcode won't show a warning, nor a "fix" suggestion. I've opted to `EffectPublisher` as the new name for this value, as it's technically correct, but I feel that `EffectTask` would be more effective. In any case, as long as the deprecation is soft, this fix can't probably be activated, and the `renamed:` argument can probably be removed.

This is not intended to be a definitive version of the PR. I'm open to any change you would suggest!

Sources/ComposableArchitecture/Documentation.docc/Extensions/ReducerProtocol.md
```diff
@@ -7,7 +7,7 @@
 - ``reduce(into:action:)-4nzr2``
 - ``State``
 - ``Action``
-- ``Effect``
+- ``EffectTask``
```
tgrapperon, line 10: Or Maybe ``EffectPublisher`` to directly land on the page with the meaningful documentation?

stephencelis: I think so, for now:

```suggestion
- ``EffectPublisher``
```

We'll be able to clean this all up on the 1.0 release branch.

Examples/CaseStudies/SwiftUICaseStudies/04-HigherOrderReducers-Lifecycle.swift
```diff
@@ -19,8 +19,8 @@ struct LifecycleReducer<Wrapped: ReducerProtocol>: ReducerProtocol {
   }
 
   let wrapped: Wrapped
-  let onAppear: Effect<Wrapped.Action, Never>
-  let onDisappear: Effect<Never, Never>
+  let onAppear: EffectTask<Wrapped.Action>
+  let onDisappear: EffectTask<Never>
```
stephencelis, line 23: Haha, this is wonky, but so is this case study, which will probably be removed when we ship our navigation APIs.

Sources/ComposableArchitecture/Documentation.docc/ComposableArchitecture.md
```diff
@@ -60,7 +60,7 @@ day-to-day when building applications, such as:
 ### State management
 
 - ``ReducerProtocol``
-- ``Effect``
+- ``EffectTask``
```
tgrapperon, line 63:
```suggestion
- ``EffectPublisher``
```
We're changing also here to redirect on `EffectPublisher`'s page where the interesting documentation lives.

Sources/ComposableArchitecture/Effect.swift
```diff
@@ -591,3 +607,63 @@ extension Effect {
     }
   }
 }
+
+// MARK: - Transitory Deprecation
+
```
stephencelis, line 612: We don't use `MARK: -` very often, so let's get to the point earlier :smile:

```suggestion
```

tgrapperon: There was one above, so I reproduced. I'll clean the whole file

Sources/ComposableArchitecture/Documentation.docc/Extensions/Deprecations/EffectDeprecations.md
```diff
@@ -10,41 +10,41 @@ Avoid using deprecated APIs in your app. Select a method to see the replacement
 
 ### Creating an effect
 
-- ``Effect/task(priority:operation:)``
+- ``EffectPublisher/task(priority:operation:)``
 
 ### Cancellation
 
-- ``Effect/cancel(ids:)-9tnmm``
+- ``EffectPublisher/cancel(ids:)-9tnmm``
```
stephencelis, line 17: - [ ] Just a note that @mbrandonw and I should audit these after merging to make sure the identifiers haven't "drifted".

tgrapperon: I think it can't resolve on my machine, but I'm not exactly sure how it works, as Xcode doesn't suggest anything. Because they're deprecated, right?

Sources/ComposableArchitecture/Effect.swift
tgrapperon: @stephencelis, Ah, you performed the change. I'll leave other existing `MARK:` for now.

stephencelis (approved):
Thanks! Will let @mbrandonw take a look, too, but should be good to merge and cut a new release soon!

Sources/ComposableArchitecture/Documentation.docc/Extensions/Deprecations/EffectDeprecations.md
stephencelis: DocC's Xcode integration is sometimes a bit flakey. We've found that running the `swift package` command to spin up a server helps:

```sh
swift package --disable-sandbox preview-documentation --target ComposableArchitecture
```

And then we've been manually navigating to the method and copying its identifier when it drifts.

JacksonUtsch:
I'm good with the `typealias` , taking another direction for error handling. Don't believe any changes are needed for such.

stephencelis:
The reception to the change has been positive, so we're going to move forward with this. Thanks again @tgrapperon!

PR #1473, opened 2022-10-12
Fix example

qmoya:
The example implementation of `NumberFactClient.liveValue` wasn’t compiling. This PR fixes that.


mbrandonw (approved):
Thanks!

PR #1475, opened 2022-10-12
Fix typos in new Reducer Protocol README

Jager-yoo:
Fixing minor typos while studying the new Reducer Protocol. 😃


mbrandonw (approved):
Good finds, thanks!

PR #1476, opened 2022-10-12
Replace `Effect<Action>` by `Effect<Action, Never>` in documentation

tgrapperon:
Regardless of #1471 evolution, this needs to be changed.


stephencelis (approved):
🤭

PR #1479, opened 2022-10-12
docs: small rephrase

roeybiran:


Sources/ComposableArchitecture/Documentation.docc/Articles/DependencyManagement.md
```diff
@@ -316,8 +316,7 @@ extension APIClient {
 
 Unfortunately, `XCTFail` cannot be used in non-test targets, and so this instance cannot be defined
 in the same file where your dependency is registered. To work around this you can use our
-[XCTestDynamicOverlay][xctest-dynamic-overlay-gh] library that dynamically invokes `XCTFail`, and
-is already included in the Composable Architecture. It also comes with some
+[XCTestDynamicOverlay][xctest-dynamic-overlay-gh] library that dynamically invokes `XCTFail` and is already included in the Composable Architecture. It also comes with some
```
stephencelis, line 319: It's hard to tell what's different with the newlines here. Am I correct in my reading that this PR is removing the comma between "`XCTFail`" and "and"? I think we're OK with the existing punctuation, but if I'm missing something please let me know!

roeybiran: Hi, the second commit was to remove an accidental new line, and while I was at it I thought that removing the comma read better. Anyway, this edit is indeed negligible — this PR is mainly about removing the "automatically accessibly" which seemed like an error to me. 

P. S. — I wanted to mention this in the PR's description but forgot — thank you so much for everything you do for the community and for the amazing gift that is the Composable Architecture. The `ReducerProtocol` release is simply mind-blowing!

stephencelis: Ah, my brain didn't see that in the diff. I'm going to fix the word wrapping and then will merge!

Sources/ComposableArchitecture/Documentation.docc/Articles/DependencyManagement.md
```diff
@@ -315,10 +315,9 @@ extension APIClient {
 ```
 
 Unfortunately, `XCTFail` cannot be used in non-test targets, and so this instance cannot be defined
-in the same file where your dependency is registered. To work around this you can use our 
-[XCTestDynamicOverlay][xctest-dynamic-overlay-gh] library that dynamically invokes `XCTFail`, and
-it is automatically accessibly when using the Composable Architecture. It also comes with some
-helpers to make constructing these unimplemented values more easily:
+in the same file where your dependency is registered. To work around this you can use our
+[XCTestDynamicOverlay][xctest-dynamic-overlay-gh] library that dynamically invokes `XCTFail` and is already included in the Composable Architecture. It also comes with some
```
stephencelis, line 319:
```suggestion
[XCTestDynamicOverlay][xctest-dynamic-overlay-gh] library that dynamically invokes `XCTFail` and
it is automatically accessible when using the Composable Architecture. It also comes with some
```

stephencelis (approved):

PR #1483, opened 2022-10-13
Remove not used let keyword in Timer Use Example comment.

GREENOVER:
Removed the unused `let` keyword from the Timer Use Example comment.


mbrandonw (approved):
Good catch, thanks!

Issue #1489, opened 2022-10-13
Received an error after installing the library


hilalalhakani:
### Description

Hello, 

I installed the library for an SPM package and after building the project I received the following error 
`converting non-sendable function value to '@Sendable (AsyncStream<Bool>.Continuation.Termination) -> Void' may introduce data races` in OpenURL class

### Checklist

- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Expected behavior

The library builds successfully 

### Actual behavior

`converting non-sendable function value to '@Sendable (AsyncStream<Bool>.Continuation.Termination) -> Void' may introduce data races`

### Steps to reproduce

below is my Package.swift file to reproduce the issue

```
let package = Package(
    name: "HomesFoodDeliveryKit",
    defaultLocalization: "en",
    platforms: [ .iOS(.v15) ],
    products: [
        .library(name: "OnboardingFeature", targets: ["OnboardingFeature"]),
        .library(name: "StyleGuide", targets: ["StyleGuide"]),
    ],
    dependencies: [
        .package(url: "https://github.com/pointfreeco/swift-tagged", from: "0.6.0"),
        .package(
            url: "https://github.com/pointfreeco/swift-composable-architecture",
            from: "0.41.2"
        )
    ],
    targets: [
        .target(name: "StyleGuide", dependencies: [.product(name: "Tagged", package: "swift-tagged")],  resources: [.process("Resources")]),
        .target(name: "OnboardingFeature", dependencies: [
            "StyleGuide",
            .product(name: "ComposableArchitecture", package: "swift-composable-architecture")
           ]
        )
    ]
)
```

### Destination operating system

15

### Xcode version information

Xcode 13.2

hilalalhakani:
Thank you Stephen for your help

stephencelis:
@hilalalhakani On further inspection it looks like we can't easily support Swift 5.5. If you're unable to upgrade to Xcode >13.2.1, can you explain why?

PR #1491, opened 2022-10-13
Require Swift 5.6

stephencelis:
~~As pointed out in #1489, we somehow lost compatibility with Swift 5.5 (Xcode 13.2). While we hope to drop support when we hit 1.0, let's try to keep compatibility in the meantime.~~

Unfortunately, it does not appear that Swift 5.5 is up to task when it comes to the new `ReducerProtocol`. While we'd be happy to support it for folks that are stuck building on Xcode <=13.2.1, we personally don't have the bandwidth to investigate the problem. I've opened #1494 to track this.

### DIFF
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -231,7 +231,7 @@ public struct DependencyValues: Sendable {
     file: StaticString = #file,
     function: StaticString = #function,
     line: UInt = #line
-  ) -> Key.Value where Key.Value: Sendable {
+  ) -> Key.Value where Key.Value: _Sendable {
     get {
       guard let dependency = self.storage[ObjectIdentifier(key)]?.base as? Key.Value
       else {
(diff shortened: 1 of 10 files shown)


Sources/Dependencies/DependencyValues.swift
```diff
@@ -336,3 +336,9 @@ private let defaultContext: DependencyContext = {
 }()
 
 // TODO: should we have `@Dependency(\.runtimeWarningsEnabled)` and/or `@Dependency(\.treatWarningsAsErrors)`?
+
+#if swift(>=5.6)
+  public typealias _Sendable = Sendable
```
mbrandonw, line 340: Worth documenting in case people come across it?

```suggestion
  /// A typealias to Swift's `Sendable` protocol used to maintain backwards
  /// compatibility with Swift 5.5.
  public typealias _Sendable = Sendable
```

tgrapperon:
A direct consequence is that TCA can now ship SwiftPM plugins. There are probably interesting uses regarding 1.0 migration.

mbrandonw (approved):

PR #1493, opened 2022-10-13
Remove some escaping closures

mbrandonw:
We have 2 unnecessary escaping closures in `ForEachStore` and `SwitchStore`. It's not actually gonna buy us anything in practice, but still they are definitely not needed.


stephencelis (approved):

tgrapperon:
Searching for `content()` yields a few more occurrences, like overloaded `SwitchStore` initializers or its `Default`, or `IfLetStore`'s `ElseContent`. It may be worth evaluating these too.

mbrandonw:
@tgrapperon nice catch, I just removed a few more. thanks!

PR #1495, opened 2022-10-13
More docs on transformDependency and a fusion fix.

mbrandonw:
Documents some tricky behavior in `transformDependency` and fixes a bug with fusion logic.


stephencelis (approved):
Good catch!

PR #1496, opened 2022-10-13
Added Toggle action for Bool values

EvsenevDev:
More convenient to use .toggle(\.$boolValue) instead .set(\.$boolValue, !viewStore.boolValue)

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Binding.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Binding.swift
@@ -119,6 +119,17 @@ extension BindableAction {
   ) -> Self {
     self.binding(.set(keyPath, value))
   }
+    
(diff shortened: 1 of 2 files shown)


stephencelis:
@EvsenevDev This is definitely a fun exploration, but I think we want to avoid shipping too many helpers on top of these APIs, which are primarily used to interface with SwiftUI views that use the binding under the hood.

One other thing to consider is that it is possible to call `toggle` on the underlying binding's value:

```swift
viewStore.binding(…).wrappedValue.toggle()
```

I also think this is definable outside of the library, so it might be fun to share in a GitHub discussion, instead, maybe with a more concrete example of how you use it?

stephencelis:
@EvsenevDev Gonna close this for now, but thanks for sharing! And if you'd like further discussion feel free to start a post on the GitHub forum!

PR #1504, opened 2022-10-15
Fixes a small typo in concurrency docs

ts:
Noticed a simple typo on a read through of the concurrency migration docs. Hope the tiny PR is kosher!


stephencelis (approved):
Good catch, thanks!

PR #1507, opened 2022-10-15
Fixes for the 'Basic Usage' to allow project to build and run

sillygoose:
New to TCA (and Swift for that matter) and was just building the Basic Usage section and ran into some minor errors, these minor changes correct the steps so you can build and run the example.

Using the latest Xcode and Swift 5.7 and the new TCA ReducerProtocol builds.  Key changes are:
- 'number' was undefined in the numbersapi.com code and complier flagged 'using' as an error
- wrapping the FeatureView() in a WindowGroup to fix errors
- sample didn't work since my simulator flagged using HTTP, since there is no HTTPS support added an exception to allow API requests to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ struct MyApp: App {
 }
 ```
 
-You may also find that the HTTP API request to numbersapi.com is blocked by the OS as only HTTPS connections are permitted.  If this happens you can add an exception to the project settings to allow HTTP requests just for the numbersapi.com domain.
+You may also find that the HTTP API request to numbersapi.com is blocked as only HTTPS connections are permitted.  If this happens you can add an exception to the project settings to allow HTTP requests just for the numbersapi.com domain.
 
 ```
 <key>NSAppTransportSecurity</key>
--- a/README.md
+++ b/README.md
@@ -123,38 +123,38 @@ struct Feature: ReducerProtocol {
   
   func reduce(into state: inout State, action: Action) -> Effect<Action, Never> {
     switch action {
-    case .factAlertDismissed:
-      state.numberFactAlert = nil
-      return .none
-      
-    case .decrementButtonTapped:
-      state.count -= 1
-      return .none
-      
-    case .incrementButtonTapped:
-      state.count += 1
-      return .none
-      
-    case .numberFactButtonTapped:
-      return .task { [number = state.count] in
-        await .numberFactResponse(
-          TaskResult {
-            String(
-              decoding: try await URLSession.shared
-                .data(from: URL(string: "http://numbersapi.com/\(number)/trivia")!).0,
-              as: UTF8.self
-            )
-          }
-        )
-      }
-      
-    case let .numberFactResponse(.success(fact)):
-      state.numberFactAlert = fact
-      return .none
-      
-    case .numberFactResponse(.failure):
-      state.numberFactAlert = "Could not load a number fact :("
-      return .none
+      case .factAlertDismissed:
+        state.numberFactAlert = nil
+        return .none
+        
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -260,19 +261,45 @@ It is also straightforward to have a UIKit controller driven off of this store.
...
+          reducer: Feature()
+        )
       )
-    )
+    }
   }
 }
 ```
 
+You may also find that the HTTP API request to numbersapi.com is blocked as only HTTPS connections are permitted.  If this happens you can add an exception to the project settings to allow HTTP requests just for the numbersapi.com domain.
```
tgrapperon, line 281: This seems to be a lot of unrelated information to display inline in a `README`. If it stays in the final version of this PR, it would probably be better to be hidden in a `details/summary` tag:
```html
<details>
  <summary>
  About `http://numbersapi.com`
  </summary>
  You may also find that the HTTP API request to […]
</details>
```
But I think it can probably go away, so please wait for other's feedback before making drastic changes.

README.md
```diff
@@ -82,6 +82,8 @@ As a basic example, consider a UI that shows a number along with "+" and "−" b
 To implement this feature we create a new type that will house the domain and behavior of the feature by conforming to `ReducerProtocol`:
 
 ```swift
```
tgrapperon, line 85: I'm not sure this is required here, as what's follow is not intended to be a regular source file (see for example l.106 with `struct State: Equatable { … }`).

README.md
```diff
@@ -134,13 +136,13 @@ struct Feature: ReducerProtocol {
         return .none
 
       case .numberFactButtonTapped:
-        return .task { [count = state.count] in 
+        return .task { [number = state.count] in
```
mbrandonw, line 139: Good catch! I would suggest we keep the `count` naming instead of introducing another name for the variable:

```suggestion
        return .task { [count = state.count] in
```

README.md
```diff
@@ -134,13 +136,13 @@ struct Feature: ReducerProtocol {
...
           await .numberFactResponse(
-            TaskResult { 
+            TaskResult {
               String(
                 decoding: try await URLSession.shared
                   .data(from: URL(string: "http://numbersapi.com/\(number)/trivia")!).0,
```
mbrandonw, line 144: And then same here:

```suggestion
                  .data(from: URL(string: "http://numbersapi.com/\(count)/trivia")!).0,
```

README.md
sillygoose: I agree, let this go which just leaves minor changes to using the samples in a cut/paste fashion

README.md
sillygoose: But if you follow the code samples then it makes sense that the author is just saving space by not repeating and just adding new code to the struct

README.md
```diff
@@ -260,19 +261,45 @@ It is also straightforward to have a UIKit controller driven off of this store.
...
+      <true/>
+      <key>NSTemporaryExceptionAllowsInsecureHTTPLoads</key>
+      <true/>
+      <key>NSTemporaryExceptionMinimumTLSVersion</key>
+      <string>TLSv1.1</string>
+    </dict>
+  </dict>
+</dict>
+```
+
```
stephencelis, line 302: I think we should probably host our own demo from an HTTPS location to avoid this confusion. Sorry @sillygoose!

For now, though, I think we probably shouldn't bog down the README with an issue that isn't library-specific.

```suggestion
```

stephencelis (approved):
Thanks! We'll take most of these :smile: Just omitting one to keep things generally simple and with an aim to avoid the confusion entirely in the future.

PR #1517, opened 2022-10-17
Make ViewStore.yield(while:) main actor.

mbrandonw:
Considering that `viewStore.send(_:while:)` is `@MainActor`, it seems this one should be too.

### DIFF
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -406,6 +406,7 @@ public final class ViewStore<ViewState, ViewAction>: ObservableObject {
   ///
   /// - Parameter predicate: A predicate on `ViewState` that determines for how long this method
   ///                        should suspend.
+  @MainActor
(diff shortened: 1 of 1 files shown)


jshier:
I'm not sure this follows logically. While `send(_:while:)` is on main since it's primary use is to handle loading, `yield(while:)` seems like a sort of primitive that could be used for any conditional wait, including those off main. Unless, is this a move towards bringing reducers entirely onto main?

stephencelis (approved):
Nice! I think this was just an oversight and we probably should have added it when it first shipped.

jshier:
> We'll be able to revisit more aspects of the task/threading model in the future when we can abandon the rest of our pre-concurrency thread model, and should eventually get back to a space where we can better support both Swift concurrency and non-main reducers/stores, but weren't able to maintain both during these current, migratory moments.

Ah, makes sense. Non-main reducers will likely become more important as things continue to scale.

PR #1532, opened 2022-10-20
XCTest problem

Czajnikowski:
Just a quick draft of the solution.

Discussed more in [this thread](https://github.com/pointfreeco/swift-composable-architecture/discussions/1519#discussioncomment-3918382).

### DIFF
--- a/ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -72,15 +72,6 @@
           "revision": "bfb0d43e75a15b6dfac770bf33479e8393884a36",
           "version": "0.4.1"
         }
-      },
-      {
-        "package": "xctest-dynamic-overlay",
-        "repositoryURL": "https://github.com/pointfreeco/xctest-dynamic-overlay",
-        "state": {
-          "branch": null,
-          "revision": "16e6409ee82e1b81390bdffbf217b9c08ab32784",
-          "version": "0.5.0"
-        }
       }
     ]
   },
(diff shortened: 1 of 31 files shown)


mbrandonw:
Hi @Czajnikowski, thanks for giving this a shot, but unfortunately I think this is too onerous on the user of the library. They have to remember to copy-and-paste those lines into every test or test case, and if they forget they will get passing tests but that aren't actually checking anything.

It also doesn't help with the pattern of "unimplemented" dependencies, which the library highly encourages and makes use of in the base set of dependencies that ship with the library.

We agree that the problem is significant enough to warrant a solution, but I think we should focus our energy on being able to statically determine when a target is built for tests (probably requires an SPM change), or being able to dynamically access `XCTFail` in a manner that is App Store safe. If anyone wants to help on those fronts we would definitely be appreciative!

Czajnikowski:
> Hi @Czajnikowski, thanks for giving this a shot, but unfortunately I think this is too onerous on the user of the library. They have to remember to copy-and-paste those lines into every test or test case, and if they forget they will get passing tests but that aren't actually checking anything.
> 
> It also doesn't help with the pattern of "unimplemented" dependencies, which the library highly encourages and makes use of in the base set of dependencies that ship with the library.
> 
> We agree that the problem is significant enough to warrant a solution, but I think we should focus our energy on being able to statically determine when a target is built for tests (probably requires an SPM change), or being able to dynamically access `XCTFail` in a manner that is App Store safe. If anyone wants to help on those fronts we would definitely be appreciative!

Yeah, I also don't like that copy-and-pasting... This may likely be solved by a subclass of `XCTestCase`, but yeah... I'm not a fan of subclasses either.

Thanks for the discussion and consideration 🙏 

Closing the PR for now.

PR #1539, opened 2022-10-21
Add a few missing tests

mbrandonw:
While looking at code coverage reports I noticed we had a few missing tests, so just filling some of that in. There's more that could be done.


Tests/ComposableArchitectureTests/RuntimeWarningTests.swift
```diff
@@ -109,83 +109,81 @@
       Task {
         ViewStore(store).send(())
       }
-      _ = XCTWaiter.wait(for: [.init()], timeout: 2)
+      _ = XCTWaiter.wait(for: [.init()], timeout: 0.5)
     }
 
-    #if os(macOS)
```
mbrandonw, line 115: not sure why this was marked as macOS only... seems to work fine on iOS.

stephencelis (approved):
Nice finds!

Tests/ComposableArchitectureTests/EffectFailureTests.swift
```diff
@@ -13,9 +13,7 @@
       var line: UInt!
       XCTExpectFailure {
         $0.compactDescription == """
-          An "EffectTask.task" returned from \
-          "ComposableArchitectureTests/EffectFailureTests.swift:\(line+1)" threw an unhandled \
-          error. …
+          An "EffectTask.task" returned from "\(#fileID):\(line+1)" threw an unhandled error. …
```
mbrandonw, line 16: pushed a small update to make some of our test failure tests a bit more resilient by using `#fileID` instead of hard coding the file name. This makes it so that if we rename the file it does not break tests.

PR #1544, opened 2022-10-22
remove unknown suffix at documentation

JCSooHwanCho:
I found some unknown suffixes in Dependency library's documentations. OR, Are they has any functionality or meaning I did not know?


mbrandonw:
Hi @JCSooHwanCho, thanks for the PR, but these suffixes do have a purpose. The help DocC distinguish between overloaded methods. For example, `withValue` has two overloads: one that is async and one that isn't. And the `-705n` helps distinguish between the two.

tgrapperon:
@JCSooHwanCho And they're provided by Xcode when you autocomplete DocC references to overloaded functions. The exact value is not something we have to come by ourselves.

JCSooHwanCho:
@mbrandonw @tgrapperon thank you for your responses! I'll close this PR.

PR #1546, opened 2022-10-23
Make `_ForEachReducer` and `ForEachStore` work with generic collections

tgrapperon:
Here is a relatively clean proof-of-concept for a generic `_ForEachReducer` and `ForEachStore`. They each have their own requirements and the constraints for each one are slightly different.

Two protocols are thus introduced to model the constraints of each use case. Both are sub-protocol of a semi-abstract `StateContainer` protocol, that has a very simple requirement:
```swift
protocol StateContainer {
  associatedType: State
  associatedType: Tag
  func extract(tag: Tag) -> State? 
}
```
Note that no restriction is imposed over the types at this point. I've used `Tag` instead of `ID`, as the latter may implicitly suggest that it is `Hashable`, which is not required at this stage. [I've also managed to abstract all "contained states" reducers](https://github.com/pointfreeco/swift-composable-architecture/compare/main...tgrapperon:swift-composable-architecture:generic-foreach-iflet-ifcaselet?expand=1) (`IfLet`, `IfCaseLet`, `ForEach`) into one `ContainedState` reducer, but since it outgrew the purpose of this PR, I've decided to only focus on `ForEachReducer` for the moment and parked these change into their own branch. Using `Tag` instead of `ID` is a remnant of this experiment.

`ForEachReducer` works on a `MutableStateContainer`, which is a read-write `StateContainer`. `.forEach` methods of reducers have been updated to receive adopters of this type instead of `IdentifiedArray` exclusively. `IdentifiedArray` and `OrderedDictionary` both conform to this protocol and can enter `.forEach` methods without having to update existing code.

`ForEachStore` works on an `IdentifiedStatesCollection`, which is a `StateContainer` that can provide a list of its tags. Because of `SwiftUI` requirements, `Tag` must be `Hashable` and is aliased to `ID`. Furthermore, the collection of `ID`s must be a `RandomAccessCollection`. Here again, `IdentifiedArray` and `OrderedDictionary` both conform to this protocol and can enter `ForEachStore` without having to update existing code.

Please note that `IdentifiedStatesCollection` that is used for `ForEachStore` (on SwiftUI's side) doesn't need to be writable. This can open the way to efficient `lazy` variants. In the same spirit, `MutableStateContainer` that is used for reducers doesn't require any `Collection` conformance, which can allow to work on weakly coupled sets of child features. Both should allow to work more lazily, without having to form dedicated types.

Because the protocol enforces constraints itself, the new `ForEachStore` implementation loses 2 generics, from 5 to 3, which should put slightly less stress on the compiler in SwiftUI views. I've also ported temporarily deprecated `ForEachStore` variants that are working on arrays, so these continue to work without having to introduce the new `ForEachStore` as a new type.

I've also reworked and streamlined `ForEachStore` internals, and this type now offers two initializers: one that accepts a `(Store) -> Content` closure (like it is currently the case), and one that accepts an `(ID, Store) -> Content`. This second overload should work nicely in `List` of `NavigationLink`, where the `ID` is often needed to fill the `label`. It can be removed if the overload strain is too high, or the API ill-suited.

I didn't notice any build-time difference when building isowords in release mode (but this application doesn't use a lot of `ForEachStore`, so it is maybe not the most representative. Runtime performance should be equivalent to the current one if calls are correctly inlined.

I'm aware that adding two public protocols is not optimal. So this PR is currently a draft.


Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
```diff
@@ -1,6 +1,77 @@
...
+  associatedtype IDs: RandomAccessCollection where IDs.Element == ID
+  associatedtype States: Collection where States.Element == State
+
+  /// A `RandomAccessCollection` of ``ID`` which is in a 1-to-1 relationship with the ``states``
+  /// collection of ``StateContainer/State``.
+  ///
+  /// - Warning: You are responsible for keeping these values in a 1-to-1 relationship with
+  /// ``states``, which means that a ``StateContainer/State`` value should exist for any ``ID`` from
+  /// ``stateIDs``.
+  var stateIDs: IDs { get }
```
tgrapperon, line 23: There is no enforced requirement between `stateIDs` and `states`, but a the documentation warns that `stateIDs -> states` should be injective (I should probably rename 1-to-1 that can be ambiguous). We could probably require it to be bijective, but it would forbid us to define slices where the subset of `ID`s is smaller than `State`s.

Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
```diff
@@ -1,6 +1,77 @@
...
+  ///
+  /// - Warning: You are responsible for keeping these ``stateIDs`` in a 1-to-1 relationship with
+  /// these values, which means that a ``StateContainer/State`` value should exist for any ``ID``
+  /// from ``stateIDs``.
+  var states: States { get }
+
+  /// Returns `true` if this collection's ``stateIDs`` is the same as `other`'s ``stateIDs``.
+  ///
+  /// A default implementation is provided when ``IDs`` is `Equatable`.
+  func areDuplicateIDs(other: Self) -> Bool
```
tgrapperon, line 35: This shape instead of a static requirement should allow configurable witnesses.

Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
```diff
@@ -78,64 +149,69 @@ import SwiftUI
...
-  public let data: Data
-  let content: Content
+  StatesCollection: IdentifiedStatesCollection, EachAction, EachContent: View
+>: DynamicViewContent
+where StatesCollection.ID: Hashable {
+  public typealias EachState = StatesCollection.State
+  public typealias ID = StatesCollection.ID
+
+  let store: Store<StatesCollection, (ID, EachAction)>
+  @ObservedObject var viewStore: ViewStore<StatesCollection, (ID, EachAction)>
```
tgrapperon, line 159: Please note, a `ViewStore<StateCollection, _>` and not `ViewStore<StateCollection.IDs, _>` like before.

Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
```diff
@@ -78,64 +149,69 @@ import SwiftUI
...
+      removeDuplicates: { $0.areDuplicateIDs(other: $1) }
+    )
+    self.content = content
   }
 
   public var body: some View {
-    self.content
+    ForEach(viewStore.stateIDs, id: \.self) { stateID in
+      // TODO: Message if `nil`?
+      let state = viewStore.state.extract(tag: stateID)!
```
tgrapperon, line 204: As we don't have the guarantee we had with `IdentifiedArray`, it is probably better to issue a runtime warning of the user's conformance to `IdentifiedStateCollection` is deficient.

PR #1550, opened 2022-10-24
Make `Reducer` actually deprecated

konomae:
> https://github.com/pointfreeco/swift-composable-architecture/releases/tag/0.43.0
> The Reducer type alias is now hard-deprecated to be renamed to AnyReducer

But it is not actually hard-deprecated.


mbrandonw:
Good find! Thanks!

PR #1554, opened 2022-10-24
Fixes main thread check in scoped stores

ohitsdaniel:
I migrated an application to use the `.task { ... }` view modifier to setup value subscriptions and ran into some issues with the `Store`'s main thread checker. 

```swift
struct DemoView: View {
  let store: Store<State, Action>
  
  var body: some View {
      Text("Demo")
        .task { await viewStore.send(.initialise).finish() }
    }
  }
}
```

In snapshot tests, the action sent in `.task { ... }` is called on a background thread, triggering the main thread checker in a scoped ViewStore and leading to a failing test. I decided to disable the main thread checker in the snapshot tests as all state is defined in the test setup and snapshots do not depend on any subscriptions. The tests kept failing.

I investigated a little and found out that scoped stores do not inherit the parent store's `mainThreadCheckerEnabled` field. From my perspective, it makes sense to hand down this parameter to scoped stores.

Looking forward to some feedback. Maybe I'm holding it wrong.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -133,7 +133,7 @@ public final class Store<State, Action> {
   #endif
   var state: CurrentValueSubject<State, Never>
   #if DEBUG
-    private let mainThreadChecksEnabled: Bool
+    let mainThreadChecksEnabled: Bool
   #endif
 
   /// Initializes a store from an initial state and a reducer.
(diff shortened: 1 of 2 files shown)


ohitsdaniel:
Hi @Alex293,
Thanks for your reply! I marked my test class with `@MainActor` but the `.task` modifier in SwiftUI doesn't respect this and schedules on a background thread by default.

I managed to fix my tests by prepending `@MainActor` in the `.task` modifier and was able to re-enable the thread checker, even in Snapshot tests, but am still curious to see if this field should be passed down to scoped stores.

stephencelis:
@ohitsdaniel The current behavior is intentional, yup, but thanks for taking the time to investigate and propose a change.

Glad you got things sorted! The new concurrency model is not always straightforward, especially when it comes to integrations like XCTest.

Going to close this out for now, but if you'd like to discuss things further, feel free to open a GitHub discussion!

PR #1562, opened 2022-10-24
Nudge issue filers to repo in vanilla SwiftUI

gohanlon:
It's great to remind filers to attempt to reproduce issues in vanilla SwiftUI and include those findings in their initial report. But, the new checklist item may exclude the filing of issues that TCA would want to respond to. In the past, such issues have led to good discussions on GitHub, documentation enhancements, and, I think, even additions to TCA.

_Note_: Despite being wordy, I used "this bug is _also_ reproducible" rather than simply "this bug is reproducible". My intent is to clearly imply that vanilla SwiftUI issues are still welcome, rather than perhaps imply vanilla issues shouldn't be filed.

Thoughts?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -21,7 +21,7 @@ body:
   attributes:
     label: Checklist
     options:
-    - label: I have not been able to reproduce this bug in a vanilla SwiftUI project.
+    - label: I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
       required: true
     - label: If possible, I've reproduced the issue using the `main` branch of this package.
       required: false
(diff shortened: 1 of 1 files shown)


tgrapperon:
I agree with you. Being able to reproduce something in Vanilla SwiftUI requires a certain knowledge of both TCA and SwiftUI. This is maybe not the case of everyone that will stumble upon a legitimate bug.
I would probably suggest to drop this requirement.

mbrandonw:
Thanks for bringing this up, and we agree. We also agree with @tgrapperon that we could make it a non-required checklist item. Want to do that and then we can merge?

gohanlon:
> ... we could make it a non-required checklist item ...

Yep, I'll make that change in an hour.

gohanlon:
I updated the checklist item to be non-required.

PR #1565, opened 2022-10-25
Add equatable conformance to UncheckedSendable when value is equatable

maxgoedjen:
This change makes it easier to have unchecked types (eg, Color, Image) in State objects. 

- [x] `swift test` passed without issue.


tgrapperon:
It seems to me that `Equatable` is reasonable, given that one may encounter this type in `Action` cases.
Isn't the compiler capable of inferring the implementation though?

stephencelis (approved):
Thanks! This is definitely missing functionality. We'll merge as is and explore if we can simplify it per @tgrapperon's recommendation, as well as introduce other missing conditional conformances.

maxgoedjen:
Yep, right you are on the automatic synthesis @tgrapperon. Thanks for merging + extending 🙏 @stephencelis.

PR #1566, opened 2022-10-25
Add a `.stream` property to `Publisher` that returns an `Async[Throwing]Stream` of its `Output`

tgrapperon:
The [`values`](https://developer.apple.com/documentation/combine/publisher/values-v7nz) property of `Publisher` that returns an async sequence of its `Output` is not available on all OSs that TCA currently supports. While it is possible to use third party or in-house backports, this creates some friction for users willing to migrate to swift-concurrency while supporting older OSs.

This PR proposes to add a `.stream` property to any `Publisher` that returns either an `AsyncStream<Output>` if the publisher can never fail, or an `AsyncThrowingStream<Output>` otherwise. `AsyncStream` is already the _de facto_ erased async sequence in TCA, and this PR follows this pattern. 

The new `AsyncStream`'s initializers that accept a publisher are kept internal. They can be made public if preferred, and they are already documented.

I've kept buffering to its default to avoid surprising behaviors.

What do you think about this?

### DIFF
--- a/Sources/ComposableArchitecture/Effects/ConcurrencySupport.swift
+++ b/Sources/ComposableArchitecture/Effects/ConcurrencySupport.swift
@@ -398,8 +398,6 @@ public struct UncheckedSendable<Value>: @unchecked Sendable {
   }
 }
 
-// Async publishers
-
 extension Publisher where Failure == Never {
   /// This property provides an `AsyncStream`, which allows you to use the Swift
   /// `async`-`await` syntax to receive the publisher's elements.
--- a/Sources/ComposableArchitecture/Effects/ConcurrencySupport.swift
+++ b/Sources/ComposableArchitecture/Effects/ConcurrencySupport.swift
@@ -142,31 +142,6 @@ extension AsyncStream {
   }
 }
 
-extension AsyncStream {
-  /// Initializes an `AsyncStream` from any `Publisher` with `Failure == Never`.
-  ///
-  /// - Parameters:
-  ///   - publisher: A `Publisher`.
-  ///   - limit: The maximum number of elements to hold in the buffer. By default, this value is
-  ///   unlimited. Use a `Continuation.BufferingPolicy` to buffer a specified number of oldest or
-  ///   newest elements.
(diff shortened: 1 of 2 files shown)


stephencelis:
@tgrapperon Thanks for the PR! Chatted with @mbrandonw a bit about it this morning.

We definitely see the utility of these helpers in TCA apps where folks are in the process of migrating their Combine dependencies over, and they offers a small step in the direction of embracing Swift concurrency. I think we are still trying to figure out how to describe our approach to adopting general-purpose helpers like these, though, for example which make sense to include with the library, and which we can delegate to others.

Most of the helpers we ship today intend to solve a problem that comes up quickly as soon as you want to write controllable/testable async code without too much fuss, and are helpers we haven't really seen shipping elsewhere. Converting publishers to streams, on the other hand, is a helper we've seen in other popular libraries, like [AsyncCompatibilityKit](https://github.com/JohnSundell/AsyncCompatibilityKit), and so folks could technically reach for those libraries to solve the problem. We're definitely down to consider baking helpers like these in, but we try to be cautious about library surface area, and in general are probably more interested in killing helpers if they prove to be less useful than we originally thought (this may even include `ActorIsolated`, which we've removed in favor of running tests and their dependencies on `@MainActor` a bunch of places).

Now one thing to also consider is we'd probably like to extract all the more general async helpers to their own library some day, and maybe when that day comes we can expand the scope a bit with helpers like these.

Let us know what you think!

tgrapperon:
Thanks for the feedback @stephencelis. In my mind, the objective was to nudge users away from Combine, as this is ultimately be something that will create internal difficulties in the future. I agree about the increased surface, and that's why I only exposed the `.stream` property in order to be usable in `@Dependency`. These helpers should have allowed anyone to immediately convert existing reducers to Swift concurrency without having to touch their dependencies. But I think it was mostly a question in the Discussions that motivated the PR, and it maybe doesn't reflect a real need.
In any case, it can be implemented by users themselves, so they are free to copy/paste the two helpers from this PR in any case they need it.

PR #1594, opened 2022-10-28
Better handling of effect priority

mbrandonw:
We realized that we probably weren't propagating task priorities correctly when merging and concatenating effects, so this is an attempt to fix it. We're not sure it's correct, and there are some things we don't fully understand in how Swift deals with priority.

For merging, we decided to abandon `TaskGroup` because it seems that child tasks don't use their priority at all:

```swift
Task(priority: .high) {
  print("outer", Task.currentPriority.rawValue)
  await withTaskGroup(of: Void.self) { group in
    group.addTask(priority: .low) {
      print("inner", Task.currentPriority.rawValue)
    }
  }
}
```

This prints:

> outer 25
> inner 25

We're not sure if this is a bug in Swift or if there is something about this we don't understand.

So, instead of `TaskGroup` we can just use `Task { ... }` and manage the cancellation ourselves. We also make some attempts to leverage `async let` where possible, thinking that may be more efficient, but it's just a guess.

Also, the main `Effect.run` that executes the merge is given a priority that is the minimum of of the two priorities being merged. We think this is the correct decision to prevent priority inversion, although Swift already does work to upgrade low-priority tasks to avoid this.

And then, for concatenate we now set its priority to the minimum like described above, but we also do extra work to avoid spinning up new `Task`'s when not necessary.

Does anyone have any thoughts on this?

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -378,16 +378,34 @@ extension EffectPublisher {
     case (.publisher, .publisher), (.run, .publisher), (.publisher, .run):
       return Self(operation: .publisher(Publishers.Merge(self, other).eraseToAnyPublisher()))
     case let (.run(lhsPriority, lhsOperation), .run(rhsPriority, rhsOperation)):
+
+      let priority: TaskPriority?
+      if let lhsPriority = lhsPriority, let rhsPriority = rhsPriority {
+        priority = Swift.min(lhsPriority, rhsPriority)
+      } else {
+        priority = lhsPriority ?? rhsPriority
+      }
+
+      @Sendable
+      func prioritize(priority: TaskPriority?, operation: @escaping () async -> Void) async {
+        guard let priority = priority
+        else { return await operation() }
+        await Task(priority: priority) { await operation() }.cancellableValue
+      }
+
       return Self(
-        operation: .run { send in
-          await withTaskGroup(of: Void.self) { group in
-            group.addTask(priority: lhsPriority) {
-              await lhsOperation(send)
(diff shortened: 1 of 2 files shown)


tgrapperon:
@mbrandonw Nevermind, I've just looked more closely at the code, and the context priority is only used to avoid spinning new `Task`, not to run them. I have no concern then.

tgrapperon:
It means that if I'm requesting to chain `.high, .high, .background`, it will produce:
```swift
run(.background) { send in
  await Task(priority: high) { … }
  await Task(priority: high) { … }
  await …
}
```
So this task is a `.background` task and it may not immediately run. Shouldn't the `max` priority be used in this case?

mbrandonw:
> So this task is a `.background` task and it may not immediately run. Shouldn't the `max` priority be ideally used in this case?

Maybe. But on the other hand, if we take the max then it means a high priority thing can be waiting on a low priority thing, causing priority inversion. And considering the weird things I'm seeing with priority elevation in task groups and `async let`, this seems like a real possibility.

We're not going to merge this anytime soon until we get a better understanding of how priority elevation works. I started a forum post [here](https://forums.swift.org/t/task-priority-elevation-for-task-groups-and-async-let/61100).

For one thing, I couldn't even write a test to prove that priority is propagated correctly. It seems all effects just get `.high` priority no matter what.

tgrapperon:
@mbrandonw You're right about the `max` issue. I need to effectively check things because the doc is not always clear. There are special cases too, [like the doc mentions](https://developer.apple.com/documentation/swift/taskpriority), and it seems easy to step into them while implementing general utilities.

PR #1596, opened 2022-10-30
Add test for early vs late binding of @Dependency.

mbrandonw:
While answering a question about `@Dependency` I started to wonder if there really is a difference between using `@Dependency` in an early binding vs a late binding.

For example, is there a difference between these two dependencies that each use a `@Dependency`:

```swift
struct ChildDependencyEarlyBinding: DependencyKey {
  var fetch: () -> Int
  static var liveValue: Self {
    @Dependency(\.someDependency) var someDependency
    return Self { someDependency.fetch() }
  }
}

struct ChildDependencyLateBinding: DependencyKey {
  var fetch: () -> Int
  static var liveValue: Self {
    return Self {
      @Dependency(\.someDependency) var someDependency
      return someDependency.fetch()
    }
  }
}
```

Because accessing `someDependency` goes through the `@Dependency` property wrapper, which then resolves the dependency, it doesn't actually matter if bind early or late.

This is a good thing and makes it that much easier to use `@Dependency` from within other dependencies.

### DIFF
--- a/Tests/DependenciesTests/DependencyValuesTests.swift
+++ b/Tests/DependenciesTests/DependencyValuesTests.swift
@@ -149,6 +149,14 @@ final class DependencyValuesTests: XCTestCase {
       XCTAssertEqual(childDependencyEarlyBinding.fetch(), 1729)
       XCTAssertEqual(childDependencyLateBinding.fetch(), 1729)
     }
+
+    DependencyValues.withValue(\.someDependency.fetch, { 999 }) {
+      @Dependency(\.childDependencyEarlyBinding) var childDependencyEarlyBinding2;
+      @Dependency(\.childDependencyLateBinding) var childDependencyLateBinding2;
+
+      XCTAssertEqual(childDependencyEarlyBinding2.fetch(), 999)
+      XCTAssertEqual(childDependencyLateBinding2.fetch(), 999)
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


freak4pc (approved):
Thanks for addressing this 🙏

tgrapperon:
Even escaping a dependency captured from the closure will produce coherent results. For example, the following very wrong code passes while preserving the behavior:
```swift
var childDependencyEarlyBindingEscaped: ChildDependencyEarlyBinding!
var childDependencyLateBindingEscaped: ChildDependencyLateBinding!

DependencyValues.withValue(\.someDependency.fetch, { 999 }) {
  @Dependency(\.childDependencyEarlyBinding) var childDependencyEarlyBinding;
  @Dependency(\.childDependencyLateBinding) var childDependencyLateBinding;

  childDependencyEarlyBindingEscaped = childDependencyEarlyBinding
  childDependencyLateBindingEscaped = childDependencyLateBinding
  
  XCTAssertEqual(childDependencyEarlyBindingEscaped.fetch(), 999)
  XCTAssertEqual(childDependencyLateBindingEscaped.fetch(), 999)
}

XCTAssertEqual(childDependencyEarlyBindingEscaped.fetch(), 42)
XCTAssertEqual(childDependencyLateBindingEscaped.fetch(), 42)

DependencyValues.withValue(\.someDependency.fetch, { 1000 }) {
  XCTAssertEqual(childDependencyEarlyBindingEscaped.fetch(), 1000)
  XCTAssertEqual(childDependencyLateBindingEscaped.fetch(), 1000)
}
```
This is a very good sign that the implementation is sound!

mbrandonw:
@tgrapperon Nice, thanks! Just added an escaped scenario to the test case.

PR #1599, opened 2022-10-31
Non-exhaustive testing

stephencelis:
This PR introduces the ability to configure a test store's level of exhaustivity. We have always promoted TCA's exhaustive testing as a great feature, and it's helped us and others catch many bugs quickly over the years.

This thoroughness isn't without its pitfalls, though, especially when testing the integration of several reducers. We've seen this firsthand in larger applications, like [isowords](https://github.com/pointfreeco/isowords), and other TCA users have battled with it, as well.

This led @krzysztofzablocki to [propose a "non-exhaustive" test store](https://www.merowing.info/exhaustive-testing-in-tca/), which would allow you to dial in your assertions to focus on just what you care about. After working together, we're bringing this functionality into the Composable Architecture in a configurable way.


acosmicflamingo:
Nice! Is there going to be a PointFree episode dedicated to this or will I have to actually read documentation? :)

.github/workflows/ci.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-      - protocol-clocks
+      - extendable-test-store
```
mbrandonw, line 7: we could prob ditch this now:

```suggestion
```

Sources/ComposableArchitecture/Documentation.docc/Articles/Testing.md
```diff
@@ -397,10 +397,169 @@ The more time you take to control the dependencies your features use, the easier
...
+login causing the selected tab to switch to activity, without having to worry about what is 
+happening inside the login feature. To do this, we can turn off ``TestStore/exhaustivity`` in the
+test store, and then just assert on what we are interested in:
+
+```swift
+let store = TestStore(
+  initialState: App.State(),
+  reducer: App()
+)
+store.exhaustivity = .none // ⬅️
```
mbrandonw, line 488:
```suggestion
store.exhaustivity = .off // ⬅️
```

mbrandonw:
> Nice! Is there going to be a PointFree episode dedicated to this or will I have to actually read documentation? :)

No episode is planned for this change, but we may discuss it in future episodes. 😁

PR #1603, opened 2022-11-01
Add soft-deprecated warning message to Effect

yermukhanbet:
As I understand, currently used Effect is going to be deprecated, and renamed to EffectPublisher. However, after I upgraded TCA dependency in our project to latest 0.45.0, there were no soft-deprecated messages for Effect, similar to Reducer naming.

### DIFF
--- a/Sources/swift-composable-architecture-benchmark/Dependencies.swift
+++ b/Sources/swift-composable-architecture-benchmark/Dependencies.swift
@@ -25,7 +25,7 @@ let dependenciesSuite = BenchmarkSuite(name: "Dependencies") { suite in
 
 private struct BenchmarkReducer: ReducerProtocol {
   @Dependency(\.someValue) var someValue
-  func reduce(into state: inout Int, action: Void) -> Effect<Void, Never> {
+  func reduce(into state: inout Int, action: Void) -> EffectPublisher<Void, Never> {
     state = self.someValue
     return .none
   }
--- a/Tests/ComposableArchitectureTests/EffectTests.swift
+++ b/Tests/ComposableArchitectureTests/EffectTests.swift
@@ -309,7 +309,7 @@ final class EffectTests: XCTestCase {
         case response(Int)
       }
       @Dependency(\.date) var date
-      func reduce(into state: inout Int, action: Action) -> Effect<Action, Never> {
+      func reduce(into state: inout Int, action: Action) -> EffectPublisher<Action, Never> {
         switch action {
         case .tap:
           return .run { send in
--- a/Tests/ComposableArchitectureTests/TestStoreFailureTests.swift
+++ b/Tests/ComposableArchitectureTests/TestStoreFailureTests.swift
@@ -266,7 +266,7 @@
     func testExpectedStateEqualityMustModify() async {
       let reducer = Reduce<Int, Bool> { state, action in
         switch action {
-        case true: return Effect(value: false)
+        case true: return EffectPublisher(value: false)
         case false: return .none
         }
       }
--- a/Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift
+++ b/Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift
@@ -289,14 +289,14 @@
           case increment
           case loggedInResponse(Bool)
         }
-        func reduce(into state: inout State, action: Action) -> Effect<Action, Never> {
+        func reduce(into state: inout State, action: Action) -> EffectPublisher<Action, Never> {
           switch action {
           case .decrement:
             state.count -= 1
             return .none
           case .increment:
             state.count += 1
-            return Effect(value: .loggedInResponse(true))
+            return EffectPublisher(value: .loggedInResponse(true))
           case let .loggedInResponse(response):
             state.isLoggedIn = response
             return .none
@@ -331,11 +331,11 @@
           case increment
           case loggedInResponse(Bool)
         }
-        func reduce(into state: inout State, action: Action) -> Effect<Action, Never> {
+        func reduce(into state: inout State, action: Action) -> EffectPublisher<Action, Never> {
           switch action {
           case .increment:
             state.count += 1
-            return Effect(value: .loggedInResponse(true))
+            return EffectPublisher(value: .loggedInResponse(true))
           case let .loggedInResponse(response):
             state.isLoggedIn = response
             return .none
@@ -695,7 +695,7 @@
       case increment
       case decrement
     }
-    func reduce(into state: inout State, action: Action) -> Effect<Action, Never> {
+    func reduce(into state: inout State, action: Action) -> EffectPublisher<Action, Never> {
       switch action {
       case .increment:
         state.count += 1
@@ -720,7 +720,7 @@
       case response1(Int)
       case response2(String)
     }
-    func reduce(into state: inout State, action: Action) -> Effect<Action, Never> {
+    func reduce(into state: inout State, action: Action) -> EffectPublisher<Action, Never> {
       switch action {
       case .onAppear:
         state = State()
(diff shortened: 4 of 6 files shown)


Sources/swift-composable-architecture-benchmark/Dependencies.swift
```diff
@@ -25,7 +25,7 @@ let dependenciesSuite = BenchmarkSuite(name: "Dependencies") { suite in
 
 private struct BenchmarkReducer: ReducerProtocol {
   @Dependency(\.someValue) var someValue
-  func reduce(into state: inout Int, action: Void) -> Effect<Void, Never> {
+  func reduce(into state: inout Int, action: Void) -> EffectPublisher<Void, Never> {
```
tgrapperon, line 28:
```suggestion
  func reduce(into state: inout Int, action: Void) -> EffectTask<Void> {
```
This is the product of a reducer, it should be `EffectTask` for now.

Tests/ComposableArchitectureTests/EffectTests.swift
```diff
@@ -309,7 +309,7 @@ final class EffectTests: XCTestCase {
         case response(Int)
       }
       @Dependency(\.date) var date
-      func reduce(into state: inout Int, action: Action) -> Effect<Action, Never> {
+      func reduce(into state: inout Int, action: Action) -> EffectPublisher<Action, Never> {
```
tgrapperon, line 312:
```suggestion
      func reduce(into state: inout Int, action: Action) -> EffectTask<Action> {
```
Same as above.

Tests/ComposableArchitectureTests/TestStoreFailureTests.swift
```diff
@@ -266,7 +266,7 @@
     func testExpectedStateEqualityMustModify() async {
       let reducer = Reduce<Int, Bool> { state, action in
         switch action {
-        case true: return Effect(value: false)
+        case true: return EffectPublisher(value: false)
```
tgrapperon, line 269:
```suggestion
        case true: return EffectTask(value: false)
```
As this value is usually formed within a `reduce` function (or in a `Reduce` block), it should technically be `EffectTask`.

Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift
```diff
@@ -289,14 +289,14 @@
           case increment
           case loggedInResponse(Bool)
         }
-        func reduce(into state: inout State, action: Action) -> Effect<Action, Never> {
+        func reduce(into state: inout State, action: Action) -> EffectPublisher<Action, Never> {
```
tgrapperon, line 292:
```suggestion
        func reduce(into state: inout State, action: Action) -> EffectTask<Action> {
```
Same as above.

Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift
```diff
@@ -289,14 +289,14 @@
...
           switch action {
           case .decrement:
             state.count -= 1
             return .none
           case .increment:
             state.count += 1
-            return Effect(value: .loggedInResponse(true))
+            return EffectPublisher(value: .loggedInResponse(true))
```
tgrapperon, line 299:
```suggestion
            return EffectTask(value: .loggedInResponse(true))
```
Same as above.

Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift
```diff
@@ -331,11 +331,11 @@
           case increment
           case loggedInResponse(Bool)
         }
-        func reduce(into state: inout State, action: Action) -> Effect<Action, Never> {
+        func reduce(into state: inout State, action: Action) -> EffectPublisher<Action, Never> {
```
tgrapperon, line 334:
```suggestion
        func reduce(into state: inout State, action: Action) -> EffectTask<Action> {
```
Same as above.

Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift
```diff
@@ -331,11 +331,11 @@
...
           switch action {
           case .increment:
             state.count += 1
-            return Effect(value: .loggedInResponse(true))
+            return EffectPublisher(value: .loggedInResponse(true))
```
tgrapperon, line 338:
```suggestion
            return EffectTask(value: .loggedInResponse(true))
```
Same as above.

Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift
```diff
@@ -695,7 +695,7 @@
       case increment
       case decrement
     }
-    func reduce(into state: inout State, action: Action) -> Effect<Action, Never> {
+    func reduce(into state: inout State, action: Action) -> EffectPublisher<Action, Never> {
```
tgrapperon, line 698:
```suggestion
    func reduce(into state: inout State, action: Action) -> EffectTask<Action> {
```
Same as above.

Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift
```diff
@@ -720,7 +720,7 @@
       case response1(Int)
       case response2(String)
     }
-    func reduce(into state: inout State, action: Action) -> Effect<Action, Never> {
+    func reduce(into state: inout State, action: Action) -> EffectPublisher<Action, Never> {
```
tgrapperon, line 723:
```suggestion
    func reduce(into state: inout State, action: Action) -> EffectTask<Action> {
```
Same as above.

Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift
```diff
@@ -758,7 +758,7 @@
 
     @Dependency(\.mainQueue) var mainQueue
 
-    func reduce(into state: inout State, action: Action) -> Effect<Action, Never> {
+    func reduce(into state: inout State, action: Action) -> EffectPublisher<Action, Never> {
```
tgrapperon, line 761:
```suggestion
    func reduce(into state: inout State, action: Action) -> Effect<Action> {
```
Same as above.

tgrapperon (commented):
Hey @yermukhanbet! Thanks for the PR. I've made a few comments here and there. The idea is that you should very rarely need `EffectPublisher`. The only frequent places where you can find them is if you modelled your dependencies using `Combine` publishers and wrapped them to `Effect` at the `Environment`'s level. This approach is vowed to disappear in the favor async streams and functions. The type that a reducer produces is `EffectTask` right now and it will be renamed `Effect` again (but with one generic only this time) near 1.0 release.
When I made the original conversion `Effect -> Effect[Task/Publisher]`, I followed the following rule:
- If we are in some reducer's product context -> `EffectTask`
- If we are in some `Combine` context -> `EffectPublisher`
- If the `Failure` is never -> `EffectTask` (but after having excluded an `Combine's context).

As I mentioned, I'm not sure that this is already the time to deprecate `Effect`. As I'm getting it, the next objective, once the `navigation` update is released, will be to put TCA in shape for the v1.0, and I feel that this change will be more timely at this point. But this is ultimately something for the maintainers to decide.

mbrandonw:
> As I understand, currently used Effect is going to be deprecated, and renamed to EffectPublisher.

Hi @yermukhanbet, this is not quite right. We are in the middle of a multi-step renaming process and doing it slowly with deprecations to make the change easy to do for everyone. The end goal is to rename `Effect<Output, Failure>` to `Effect<Output>`, i.e. lose the `Failure` generic. We can't do that at all once without breaking everyone's code.

So we are renaming `Effect<O, F>` to `EffectPublisher<O, F>` and instantly deprecating `EffectPublisher`, and we are renaming `Effect<O, Never>` to `EffectTask<O>`. Then someday in the future we will fully rename `Effect<O, F>` to `Effect<O>`.

> However, after I upgraded TCA dependency in our project to latest 0.45.0, there were no soft-deprecated messages for Effect, similar to Reducer naming

You won't see deprecation warnings in Xcode for "soft" deprecated. The only place you will see it is in Xcode autocomplete and in documentation. This will nudge you to use `EffectTask` instead of `Effect` for newly written code.

Then, in the near future we will "hard" deprecate `Effect` and then you will start to see the warnings and should update all old code.

PR #1608, opened 2022-11-01
Rename a few new occurrences of `Effect` to `EffectTask` or `EffectPublisher`

tgrapperon:
Following #1603, this PR renames a few new internal `Effect` introduced in recent changes, as an homogenous adoption will likely make the transition easier. I've kept three `Effect` with `Never` failures as `EffectPublisher`, as they're directly using `Combine` immediately after.


stephencelis (approved):

stephencelis:
Thanks!

PR #1620, opened 2022-11-04
Test store dependency binding

mbrandonw:
This makes it so that the test context is set when constructing the `TestStore` and in the modify closures of `send`/`receive`, and provides a trailing closure initializer for `TestStore` for customizing dependencies.

This idea came from @crayment in https://github.com/pointfreeco/swift-composable-architecture/discussions/1613.

### DIFF
--- a/Sources/ComposableArchitecture/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestStore.swift
@@ -566,7 +566,7 @@ open class TestStore<State, Action, ScopedState, ScopedAction, Environment> {
   public init<Reducer: ReducerProtocol>(
     initialState: @autoclosure () -> State,
     reducer: Reducer,
-    updateDependencies: (inout DependencyValues) -> Void = { _ in },
+    prepareDependencies: (inout DependencyValues) -> Void = { _ in },
     file: StaticString = #file,
     line: UInt = #line
   )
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/TestStore.swift
```diff
@@ -564,8 +564,9 @@ open class TestStore<State, Action, ScopedState, ScopedAction, Environment> {
   ///   - initialState: The state the feature starts in.
   ///   - reducer: The reducer that powers the runtime of the feature.
   public init<Reducer: ReducerProtocol>(
-    initialState: State,
+    initialState: @autoclosure () -> State,
     reducer: Reducer,
+    updateDependencies: (inout DependencyValues) -> Void = { _ in },
```
stephencelis, line 569: I know this'll be a trailing closure without argument labels most of the time, but what do you think of:
```suggestion
    prepareDependencies: (inout DependencyValues) -> Void = { _ in },
```

mbrandonw: yep that works for me. will update

stephencelis (approved):
Good changes!

crayment:
Love it, thanks!!

crayment:
I just realized `receive` should probably be updated as well to have the test stores dependencies for `updateExpectingResult`. Totally overlooked that in my first discussion.

mbrandonw:
@crayment That should be the case by virtue of the fact that we updated the private function `expectedStateShouldMatch`, and both `send` and `receive` make use of it. Are you seeing otherwise?

We probably should update the test to get coverage on this fact though.

crayment:
I haven't verified on this branch yet. Still using my helpers. If you give me a bit I can verify, and maybe even PR an update to the test 😄

mbrandonw:
I did just verify locally with a test and it seems OK. But yes please feel free to PR a test to get coverage.

crayment:
Done: #1644

Feel free to just close if you like your test updates better though!

PR #1625, opened 2022-11-04
Disable printing by default for `DebugReducer` when running SwiftUI Previews

tgrapperon:
I had a case where one of my features was lagging a lot without any apparent reason in a SwiftUI Preview. The motive became immediately clear when I launched a dedicated app and saw the console filled by a very large amount of data from a `_printChanges()` that I forgot.

I've added a preliminary check to the default `_ReducerPrinter` so they don't generate any output when running SwiftUI Previews. As this only affects the default printers, this behavior can be overridden by users willing to provide their own `_ReducerPrinter` implementation (for example with variants capable of displaying this information in SwiftUI Previews like experimented in #1234).

This should make SwiftUI Previews a little snappier when a `._printChanges()` has been forgotten.


Sources/ComposableArchitecture/Internal/SwiftUIPreviews.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+@usableFromInline
+var isRunningForPreviews: Bool {
+  #if DEBUG
+  return ProcessInfo.processInfo.environment["XCODE_RUNNING_FOR_PREVIEWS"] == "1"
+  #else
+  return false
+  #endif
+}
```
stephencelis, line 10: Does it work as a static check that's only performed once?

```suggestion
let isRunningForPreviews = {
  #if DEBUG
  return ProcessInfo.processInfo.environment["XCODE_RUNNING_FOR_PREVIEWS"] == "1"
  #else
  return false
  #endif
}()
```

stephencelis:
Good to know! We typically don't use this modifier outside the app entry point, but know it's common enough for folks to do, so avoiding pitfalls like this sounds good to us!

stephencelis (approved):

Sources/ComposableArchitecture/Internal/SwiftUIPreviews.swift
stephencelis: Actually on second thought, how about using `@Dependency(\.context)` and check for `.preview` instead? Then we could maybe even write a test for the behavior.

tgrapperon: Ok, I'll refactor

Tests/ComposableArchitectureTests/DebugTests.swift
```diff
@@ -216,5 +216,20 @@
...
+      struct DebuggedReducer: ReducerProtocol {
+        typealias State = Int
+        typealias Action = Bool
+        func reduce(into state: inout Int, action: Bool) -> EffectTask<Bool> {
+          state += action ? 1 : -1
+          return .none
+        }
+      }
+      let store = TestStore(initialState: 0, reducer: DebuggedReducer()._printChanges())
+      store.dependencies.context = .preview
```
tgrapperon, line 231: @stephencelis We can't really test the behavior directly, as there is no way to override at distance the default `TextOuputStream` from `print()` to my knowledge. Maybe there are some way to read stdout, but that seems overkill.

Switching this value to `.live` shows that something is printed, and setting it back to `.preview` shows that nothing is printed, like for the test above that runs in `.test` context.

tgrapperon:
I refactored to use `@Dependency(\.context)` as a source of truth. This is indeed much nicer!

iampatbrown:
Is it possible to model `_ReducerPrinter` as a dependency?

tgrapperon:
@iampatbrown I think it was implemented as such at some point during the beta. But I'm not certain why it was rescinded. Likely because it wasn't worth the complexity at this stage.

iampatbrown:
@tgrapperon Ah k. That makes sense :)

Tests/ComposableArchitectureTests/DebugTests.swift
stephencelis: Ah yeah. We should prob allow more customization in the future, but the API is underscored and barebones at the moment.

iampatbrown: > there is no way to override at distance the default TextOuputStream from print()

@tgrapperon Back when you experimented with #1234 I had a go at just this :P

You can redirect stdout to `Pipe` using [dup2](https://man7.org/linux/man-pages/man2/dup.2.html)

It took a little bit of trial and error but I ended up with something that worked okay. I haven't used it though. It was just a proof of concept.

[iampatbrown/ConsoleView.swift](https://gist.github.com/iampatbrown/3034ec9756a6cdc3b9dd25d9784f3486)

tgrapperon: @iampatbrown Super interesting! I'll give it a spin if I rework on #1234 idea. I'm currently using my `Logger` abstraction to duplicate prints to the UI console, but your approach could make things even simpler.

PR #1632, opened 2022-11-07
Skip test if CI

stephencelis:
We have an existing flakey test that we want to skip on CI, but we need to pass along the env var to the scheme for it to apply to Xcode builds, simulators, etc.


Tests/ComposableArchitectureTests/Support/ProcessInfo.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+extension ProcessInfo {
+  var isCI: Bool {
+    self.environment["CI"].map { Bool($0.lowercased()) ?? ($0 == "1") } ?? false
```
stephencelis, line 5: According to GitHub, [`== "true"`](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) should probably be good enough, but doesn't hurt to generalize...

Tests/ComposableArchitectureTests/Support/ProcessInfo.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+extension ProcessInfo {
+  var isCI: Bool {
+    print("!!!")
+    dump(self.environment)
+    print("!!!")
```
mbrandonw, line 7: we wanna remove these before merging right?

```suggestion
```

mbrandonw (approved):

Tests/ComposableArchitectureTests/Support/ProcessInfo.swift
stephencelis: Still debugging.

PR #1647, opened 2022-11-09
Add a test that checks that nested dependencies are correctly overridden

tgrapperon:
This PR adds a test that covers the case where we override only a property of a `DependencyKey`. It checks that accessing the property from the dependency is producing the same result as directly exposing this property as a dependency.
It didn't seem obvious to me that it would work at first sight, but feel free to close the PR if you think this is superfluous.


mbrandonw (approved):
Ah nice, yeah that's an interesting enough edge case for us to have coverage on. Thanks!

PR #1648, opened 2022-11-09
`@Dependency()Data`

davdroman:


Alex293:
While this can be useful I think the name isn’t right

stephencelis:
@davdroman Thanks for sharing! I think controlling for data read/write access is definitely something we'd like to make easy.

We're not entirely sure how we want to approach the API design here (`Data`-focused vs. something like the `FileClient`-y approach we took in isowords), but we'll try to carve out some time to think about it soon.

davdroman:
@stephencelis I appreciate the input. I agree that "data" vs "file" is a nomenclature worth exploring.

tgrapperon:
Hey @davdroman! I took the liberty to implement this dependency in [`swift-dependencies-additions`](https://github.com/tgrapperon/swift-dependencies-additions/blob/main/Sources/DataDependency/DataDependency.swift). 
Feel free to contribute any change you'd see fit! This doesn't prevent it to eventually move in [`swift-dependencies`](https://github.com/pointfreeco/swift-dependencies) someday, but in the meantime, it has a place to grow up!

davdroman:
@tgrapperon hi Thomas. I saw it earlier today, very cool! Thank you for the shoutout on the readme too 😄 

I'll close this now that this idea has found a home.

acosmicflamingo:
@tgrapperon that repo looks amazing!!! I know what SPM package I'll be adding later :)

PR #1650, opened 2022-11-10
Inventory app added as example app

JaapWijnen:
I added a TCA version of the swiftui-navigation Inventory app to the examples folder to showcase some navigation examples. 

A few things need work in order for this to be ready to merge: 
- The duplication button of item row somehow not clickable, the app prefers the edit navigation link over triggering the duplication sheet. (any tips are very welcome!)
- One part I'm not happy with yet is in the ItemRow module where I added a .doNothing action to bypass the following:
```swift
NavigationLink(
    unwrapping: viewStore.binding(
        get: { $0 },
        send: ItemRow.Action.doNothing
    ),
    case: /ItemRow.Route.edit) { isActive in
        viewStore.send(.setEditNavigation(isActive: isActive))
    } destination: { _ in
       ...
```
Normally the send part of te binding would contain `.setRoute`, which will set `ItemRow.State.route` to `nil`
In the edit navigation case however we don't want route to be set to nil before the `.setEditNavigation(isActive: false)` action hits the reducer otherwise we can't extract the state from the route case `Route.edit(Item.State)`.

I'd love to have some general feedback! Let me know what you think.
Not sure if there's interest to merge with the upcoming navigation tools. This app might however help as an example on how to migrate to the newer tools once they are merged in!


tgrapperon:
> The duplication button of item row somehow not clickable, the app prefers the edit navigation link over triggering the duplication sheet. (any tips are very welcome!)

This is the vanilla `SwiftUI` behavior for `Button`s contained in `List`. You should be able to "opt-out" and regain individual actions by specifying an explicit `ButtonStyle` for the accessory buttons.

stephencelis:
@JaapWijnen Thanks for sharing!

Since I mentioned we'd probably tackle this ourselves when we're further along with our navigation tools, how do you feel about closing the PR and linking to it from your discussion? That way folks will be able to find it and, but it won't take up open PR real estate.

JaapWijnen:
@tgrapperon Thanks! That did the trick. It was actually in the original code just got the nesting of the button style wrong. Would not have found this without your remark thanks!

JaapWijnen:
@stephencelis Sure no problem. If it ends up being useful please do let me know :) Did run into a lot of things that I'm guessing you will be including in the new tools! Was quite a useful exercise. Let me know if you have any comments.

PR #1681, opened 2022-11-19
Make BindingAction.init public

eappel:
Exposing this initializer as public will enable folks to build more powerful `BindingAction` helpers!

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Binding.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Binding.swift
@@ -212,7 +212,7 @@ extension BindingAction {
     keyPath == bindingAction.keyPath
   }
 
-  init<Value: Equatable>(
+  public init<Value: Equatable>(
     keyPath: WritableKeyPath<Root, BindableState<Value>>,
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hi @eappel, can you explain how you plan on using this initializer if it were made public?

The reason it is not public is because it allows one to construct invalid `BindingAction`s. It is only meant to be constructed from a key path, which is why we expose the `static func set` constructor:

https://github.com/pointfreeco/swift-composable-architecture/blob/e3e33d129814e8c655196b680f1cd516c0dcc088/Sources/ComposableArchitecture/SwiftUI/Binding.swift#L185-L194

If a `BindingAction` is constructed in a way such that the `set` is not _exactly_ mutation via the key path, then the `BindingAction` is not well-defined. For one thing, its `Equatable` conformance is no longer correct, and that can lead to some subtle bugs.

stephencelis:
@eappel I'm going to close this for now given Brandon's comment, but if you follow up with more info, I'd be happy to reopen. Let us know!

eappel:
@stephencelis no worries! Opened a new PR to support more specific actions #1825

PR #1685, opened 2022-11-21
Depend on SwiftUINavigation for TextState, AlertState, etc...

mbrandonw:
We have extracted `TexState`, `AlertState` and `ConfirmationDialogState` from this library and put it in our [SwiftUINavigation library][swiftui-nav-gh] so that vanilla SwiftUI can take advantage of those tools. TCA will now depend on SwiftUINavigation, but not _all_ of it. Just on a small, "internal" library called _SwiftUINavigationState.

[swiftui-nav-gh]: http://github.com/pointfreeco/swiftui-navigation
[0_4_0]: https://github.com/pointfreeco/swiftui-navigation/releases/tag/0.4.0


Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -844,15 +857,6 @@ extension TestStore where ScopedState: Equatable, Action: Equatable {
...
 
-extension AlertState.Button {
-  @available(
-    *, deprecated, message: "Cancel buttons must be given an explicit label as their first argument"
-  )
-  public static func cancel(action: AlertState.ButtonAction? = nil) -> Self {
-    .init(action: action, label: TextState("Cancel"), role: .cancel)
-  }
-}
-
```
mbrandonw, line 855: This wasn't compiling so I just removed since it's so old. But I can debug it if we don't want to remove yet.

Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -1039,41 +1043,6 @@ extension ViewStore {
   }
 }
 
-// MARK: - Deprecated after 0.23.0:
-
-extension AlertState.Button {
```
mbrandonw, line 1044: Same with all of this.

Package.swift
```diff
@@ -28,6 +28,7 @@ let package = Package(
     .package(url: "https://github.com/pointfreeco/swift-clocks", from: "0.1.4"),
     .package(url: "https://github.com/pointfreeco/swift-custom-dump", from: "0.6.0"),
     .package(url: "https://github.com/pointfreeco/swift-identified-collections", from: "0.4.1"),
+    .package(url: "https://github.com/pointfreeco/swiftui-navigation", branch: "fixes"),
```
mbrandonw, line 31: I think we need the changes in this branch in order to integrate with TCA.

stephencelis (approved):

PR #1686, opened 2022-11-21
Don't implicitly infer test context.

mbrandonw:
Right now we infer the dependency context from whether or not we think the code is running in a test context. This causes problems when running tests on an application target, because the entry point of the application will actually execute, causing live dependencies to be used in a test context. Those failures bleed over to the test target. This is only an issue when testing application targets. Libraries and SPM modules work just fine.

Now technically it is just really bad that Xcode even runs the entry point of the application. This can cause you to accidentally execute API requests, track analytics, write data to disk, etc. without you knowing. TCA is just warning you that this is happening by yelling loudly and failing the test suite.

I'm starting to think it's not worth implicitly inferring the test context. If we stopped doing that, then the test host could go back to secretly using live dependencies without it failing the test suite.

This has come up a number of times, in both [discussions](https://github.com/pointfreeco/swift-composable-architecture/discussions/1652#discussioncomment-4112291) and on a Slack.

### DIFF
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -298,8 +298,6 @@ struct CurrentDependency {
 private let defaultContext: DependencyContext = {
   if ProcessInfo.processInfo.environment["XCODE_RUNNING_FOR_PREVIEWS"] == "1" {
     return .preview
-  } else if _XCTIsTesting {
-    return .test
   } else {
     return .live
   }
--- a/Tests/DependenciesTests/DependencyValuesTests.swift
+++ b/Tests/DependenciesTests/DependencyValuesTests.swift
@@ -66,33 +66,36 @@ final class DependencyValuesTests: XCTestCase {
   }
 
   func testDependencyDefaultIsReused() {
-    DependencyValues.withValue(\.self, .init()) {
-      @Dependency(\.reuseClient) var reuseClient: ReuseClient
+      DependencyValues.withValue(\.self, .init()) {
+        DependencyValues.withValue(\.context, .test) {
+        @Dependency(\.reuseClient) var reuseClient: ReuseClient
 
-      XCTAssertEqual(reuseClient.count(), 0)
-      reuseClient.setCount(42)
-      XCTAssertEqual(reuseClient.count(), 42)
+        XCTAssertEqual(reuseClient.count(), 0)
+        reuseClient.setCount(42)
+        XCTAssertEqual(reuseClient.count(), 42)
+      }
     }
   }
 
   func testDependencyDefaultIsReused_SegmentedByContext() {
     DependencyValues.withValue(\.self, .init()) {
-      @Dependency(\.reuseClient) var reuseClient: ReuseClient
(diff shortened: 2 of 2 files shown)


jshier:
Typically this would be handled by "turning off" your app when it's launched as a test host. Rather than breaking this nice behavior for that case, couldn't you just recommend the other solution? TCA could even help out by exposing the same "isTest" condition that is used internally.

mbrandonw:
> Typically this would be handled by "turning off" your app when it's launched as a test host. 

@jshier do you mean that this is a setting in Xcode to do this, or are you suggesting finding a way to omit your view from executing in the entry point?

If the latter, then technically this can be done using XCTestDynamicOverlay (which comes transitively with TCA) by doing this:

```swift
import XCTestDynamicOverlay

@main
struct SearchApp: App {
  var body: some Scene {
    WindowGroup {
      if !_XCTIsTesting {
        …
      }
    }
  }
}
```

> Rather than breaking this nice behavior for that case, couldn't you just recommend the other solution? TCA could even help out by exposing the same "isTest" condition that is used internally.

FWIW, the `TestStore` automatically sets up `.test` as the context. So I'm not aware of any situation this actually breaks. It _may_ break for anyone using Dependencies outside TCA, in which case we plan on extracting it to a separate library with its own test helpers very soon.

jshier:
Doing it manually is what I meant, but an official setting would be nice. It's good that the change doesn't break anything, but might this feature be a good way to audit my app startup in testing to ensure I'm not calling anything? Perhaps there could be a version that allows live instances but produces a runtime warning?

mbrandonw:
Leaving it the way it is does have the nice side benefit of proving that the launch of your app is not executing live dependencies. But, it's a bit limited since it's only proving it for the duration of how long the test takes to executes, which may be a small amount of time.

If we did decide to move forward with this API you would be able to recapture the current behavior with something like this in your entry point:

```swift
RootView(
  store: Store(
    initialState: …,
    reducer: RootReducer()
      .dependency(\.context, _XCTIsTesting ? .test : .live)
  )
)
```

stephencelis (approved):
Seems like a good change!

stephencelis:
Just as a note, we've reverted this behavior with the release of [swift-dependencies](https://github.com/pointfreeco/swift-dependencies). We believe that defaulting to the test context in tests is required for vanilla apps using the library to fail properly without extra work.

This means that if you do run your tests in an app host, the app must take care to set its context (or omit the runtime entirely during test runs).

/cc @TicTocCroc

iwt-sebastian-boldt:
@stephencelis 
What would be your preferred solution to fix the interference of the Tests with your App host running in parallel?
For me personally, it seems that omitting the launch of the runtime completely by using _XCTIsTesting, currently is the best solution, right? Overwriting the context has the disadvantage of dependencies running against a live environment so that they might call live APIs etc.

stephencelis:
@iwt-sebastian-boldt Yup! And we document this here: https://pointfreeco.github.io/swift-dependencies/main/documentation/dependencies/testing/#Testing-gotchas

Issue #1673, opened 2022-11-16
Unable to use TCA in extensions


IgorRosocha:
### Description

We are currently taking leverage of [Tuist dependencies](https://docs.tuist.io/manifests/dependencies/) to handle dependencies in our modular app. Until the release of version 0.41.0, everything worked fine, since then we're unable to update because we're getting errors in [OpenURL.swift](https://github.com/pointfreeco/swift-composable-architecture/blob/main/Sources/Dependencies/Dependencies/OpenURL.swift#L33):

```
'open(_:options:completionHandler:)' is unavailable in application extensions for iOS
'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead.
```

### Checklist

- [x] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Expected behavior

We should be able to use TCA in extensions / TCA should compile successfully.

### Actual behavior

We are unable to use TCA in extensions / TCA doesn't compile successfully.

### Steps to reproduce

1. Integrate TCA using [Tuist dependencies](https://docs.tuist.io/manifests/dependencies/) / use TCA in extension
2. Compile the project

### The Composable Architecture version information

0.45.0

### Destination operating system

iOS 16

### Xcode version information

14.0.1

### Swift Compiler version information

```shell
swift-driver version: 1.62.8 Apple Swift version 5.7 (swiftlang-5.7.0.127.4 clang-1400.0.29.50)
Target: arm64-apple-macosx12.0
```

mbrandonw:
Hi @IgorRosocha, this was fixed in 0.42.0. Have you tried that? Also there has been many releases since then, so we recommend updating to the newest.

IgorRosocha:
Hi @mbrandonw, as mentioned in the issue, we're getting the error on the latest release - 0.45.0

mbrandonw:
Oh sorry, I saw your first mention of 0.41 and thought that is what you were using. Are you able to share a sample project that reproduces the error so that we can figure out how to fix it?

IgorRosocha:
Sure, please take a look at minimal example here: https://github.com/IgorRosocha/TCA-Extension-Sample

[...]

PR #1714, opened 2022-12-02
Base `OpenURL` dependency on SwiftUI

tgrapperon:
This simplifies the implementation of the `OpenURL` dependency and should allow to build dynamic frameworks with `APPLICATION_EXTENSION_API_ONLY`.

This should address #1673.

As a consequence, this raises the minimal availability of this dependency to iOS 14 and macOS 11.


stephencelis (approved):
This looks good to us, thanks! It's relatively new, so hopefully not too many folks rely on it for iOS 13, and if they do it's hopefully easy enough to fix with a custom dependency.

PR #1715, opened 2022-12-02
Fix ButtonRole and Button deprecations

kgrigsby59:
Replaced AlertState<Action>.ButtonRole with ButtonState<Action>.Role 
Replaced AlertState<Action>.Button with ButtonState<Action>


stephencelis (approved):
Thanks!

PR #1729, opened 2022-12-10
Add Scoping & BindingState documentation

acosmicflamingo:
@tgrapperon @mbrandonw @stephencelis I basically plagiarized what @tgrapperon said in https://github.com/pointfreeco/isowords/issues/170 . Hopefully it helps prevent others from finding the issue I encountered earlier :)

### DIFF
--- a/Sources/ComposableArchitecture/Documentation.docc/Articles/Bindings.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Articles/Bindings.md
@@ -257,6 +257,34 @@ var body: some ReducerProtocol<State, Action> {
 }
 ```
 
+If a reducer is a child reducer that hosts the `BindingReducer`, the parent must embed it with
+the scope function or the program will never execute the actual binding actions:
+
+```swift
+struct Settings: ReducerProtocol {
+  struct State: Equatable { /* ... */ }
+  enum Action: BindableAction { /* ... */ }
+  var body: some ReducerProtocol<State, Action> {
+    BindingReducer()
+  }
+}
+
+struct Home: ReducerProtocol {
+  struct State: Equatable {
+    public var settings: Settings.State
+  }
+
+  enum Action: Equatable {
+    case settings(Settings.Action)
+  }
+  var body: some ReducerProtocol<State, Action> {
+    Scope(state: \.settings, action: /Action.settings) {
+      Settings()
+    }
+  }
+}
+```
+
 Binding actions can also be tested in much the same way regular actions are tested. Rather than send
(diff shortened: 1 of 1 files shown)


tgrapperon:
Hey @acosmicflamingo! I'm not 100% sure that this change is required, or at least in this form or in this spot.

`BindingReducer` (more precisely `BindingAction`s) has some internal logic that detects if the `BindingAction` is expressed (that is, if the `BindableState` is mutated). So, if an action goes through without "its" `BindingReducer` catching and expressing it, it shows a message. Most often, everything is correctly wired, but you forgot to add the `BindingReducer` to your feature's `body`. But once you've checked that the `BindingReducer` is well installed into your feature, the next logical explanation is that your feature's `Reducer` didn't run at all. Most often, it's because you forgot to embed/scope this feature's reducer into its parent. But this mistake is absolutely general and can happen for any feature of your app. For this reason, it seems rather odd to mention this issue in this section dedicated to bindings.

So I guess, if there's something that can be improved, it's probably the warning message that can suggest checking if the feature hosting the `BindingReducer` is correctly embedded into its parent. But I would prefer Brandon or Stephen to weigh in on the subject, as I'm not really sure about the correct wording.

acosmicflamingo:
@tgrapperon I suppose it's not obvious that the example I use is a parent reducer (which is why I didn't include the BindingReducer there). I amended the commit to make this more clear but even then, I do agree with you that the best place for something like this would actually be in a warning! Developers will certainly be more likely to address the problem by just reading the warning, checking to see if it applies to them, and then fixing it if it does, than to open up the BindingState documentation. In either case, I can update this PR accordingly, and Brandon or Stephen can chime in with their thoughts :)

stephencelis:
Hi @acosmicflamingo! Sorry for the delay. This came along before the holidays and we lost track of it. The isowords bug slipped in during refactoring, but we resolved it when we came back to it recently :smile:

I think we're going to hold off on `Scope`-specific documentation here, since it would apply to basically every helper in the library already. I am curious, though, if you had encountered this in a project of your own (and not a bug in our project), if the existing docs/warning would have been enough?

acosmicflamingo:
@stephencelis no need to apologize; I know you all are really busy and this isn't a high-priority issue, and thanks for responding!

To answer your question, I only encountered it in my project because I was using isowords source code as a template, so I think what you have is good enough. If I do encounter this kind of problem again in the future however, then I'll just create a new PR and reference this one.

Natchaphonwongplianphan:
ใครมาชะเมิดด้านข้อมูล

PR #1737, opened 2022-12-12
added switchstore init with 10 cases

natemann:


stephencelis:
Hey @natemann! Thanks for contributing, and sorry for the delayed response. It fell off our radar. I think we're a little worried about the overload space here already, and we think merging this could break other folks' builds as a result, so I'm going to close this for now.

We do want to revisit `SwitchStore` in the future to be a little more flexible (support more cases, like this, as well as support things like `Group`) and performant in the future, though! We had some promising sketches by introducing a dedicated builder and using Swift 5.7's new `buildPartialBlock` functionality. We're hoping to find time to polish it soon!

PR #1773, opened 2022-12-27
Fix code typo in the article 'Migrating to the reducer protocol'

dy-kim:
On my way of `migrating to the reducer protocol`, I've found this small typo in the 4th code snippet of [Optional and pullback reducers section](https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/migratingtothereducerprotocol/#Optional-and-pullback-reducers).


stephencelis (approved):
Good catch, thanks!

PR #1780, opened 2023-01-02
Make TestStore.receive actually match the action predicate

pyrtsa:
When calling `TestStore.receive(_:…)` with a predicate or a case path and exhaustivity on, the current code in `main` never actually calls the `matching` predicate. This PR fixes that, with unit tests.


pyrtsa:
While at it, I should probably also update the docs which refer to a [non-existent `matchingAction` closure](https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/teststore/receive(_:timeout:assert:file:line:)-3myco) currently. 🤔 Will see about that tomorrow.

stephencelis (approved):
Looks great, thanks for catching this! Will take a look at those docs for you :smile:

Issue #1783, opened 2023-01-03
Binding key path getter crashing


barabashd:
### Description

Hi guys, I’m practicing with TCA, not for a long time but I am excited about what you do for the community. 

Recently I stuck with a bug of using a key path for the binding getter. I see here at least 2 problems:
1. Getter could be called when it should not and if we use native @SwiftUI.State it doesn’t have such a problem. As a result, our code could be fired when it shouldn’t and trigger UI updates.
2. A getter with the key path is unsafe and as a result, it could lead to a crash.

More details in the next [example](https://github.com/barabashd/TCA_binding_keypath). Now, there are 3 examples:
1. Has a crash with TCA binding
2. Native SwiftUI State approach - works well
3. An alternative way for TCA binding - resolve possible crash (but still has useless SwiftUI updates)

My possible solution resolves a crash but for now idk the best way to prevent UI updates. What do you think?
Also, I understand that my proposal could be not the best but I'm going to create a pull request and discuss further suggestions for this issue.

### Checklist

- [X] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Expected behavior

I think we should prevent such crashes using the key path.

### Actual behavior

A crash inside ViewStore subscript's getter
`
```
private subscript<Value>(
    get state: HashableWrapper<(ViewState) -> Value>,
    send action: HashableWrapper<(Value) -> ViewAction>
) -> Value {
    get { state.rawValue(self.state) }
    set { self.send(action.rawValue(newValue)) }
}
```
`

### Steps to reproduce

1. Clone https://github.com/barabashd/TCA_binding_keypath
2. Uncomment (if needed) Example 1 in file MainView.swift
3. Run an app
4. Tap the button "Add new"
5. Tap the button "Remove last"
6. Observe the crash

### The Composable Architecture version information

0.47.2

### Destination operating system

iOS 16.2

### Xcode version information

14.2 (14C18)

### Swift Compiler version information

```shell
arm64-apple-macosx13.0
```

PR #1784, opened 2023-01-03
Resolve binding key path crash

barabashd:
[Details](https://github.com/pointfreeco/swift-composable-architecture/issues/1783)
In a nutshell: binding works differently from @SwiftUI.State binding and could lead to a crash.


stephencelis:
@barabashd Thanks for diving into the problem and your work on a solution!

Though the stack trace isn't super helpful, the crash you're encountering is that the row view computes for an element that no longer exists. This computation causes the array subscript to get called for an invalid index, so you get a "index out of range" fatal error. We're not quite sure why SwiftUI does this in this case, and we have considered it an Apple bug, but it's nice to see that there's a workaround by shuffling the domain around.

I'm going to merge this as-is and will make a few changes in a followup PR that fix the CI issue and an animation issue the original subscript intended to work around.

PR #1785, opened 2023-01-03
Simplify/fix #1784

stephencelis:
This PR works on top of #1784 and:

  * Reduces the number of moving parts
  * Restores implicit animations by directly producing the binding from the observable object
  * Strongly retains the store in the binding to avoid losing writes


stephencelis:
@barabashd Thanks again for the original fix!

mbrandonw (approved):

PR #1788, opened 2023-01-04
Deprecate 'Effect' for 'EffectTask'

stephencelis:
This is to allow reclaiming 'Effect' at a later time without the 'Failure' generic in the near future.


acosmicflamingo:
Woohoo!!!

PR #1790, opened 2023-01-04
Allow `BindingReducer` to work with `ViewState`

tgrapperon:
This PR proposes a solution to allow `BindingReducer` to work on `ViewState`. It consists in a simple protocol that you conform your `State` to, and a new property wrapper to expose `@Bindable` properties to `ViewState`. It also renames and deprecates a few types and methods in preparation for the version 1.0.

### What is changing
- `@BindableState` has been deprecated and renamed `@BindingState`, as "able" names are usually reserved for protocols.
- A `BindableStateProtocol` is introduced. You conform the `State` hosting `@BindingState` properties to it. The name of this protocol is transient and it will be renamed `BindableState` for the version 1.0.
- A `@BindingViewState` property wrapper is introduced. You use this property wrapper to annotate bindable values in your `ViewState`
- You directly create a `Binding<Value>` using `viewStore.$value`. The `viewStore.binding(\.$value)` style has been soft-deprecated (other `ViewStore.binding` methods are not concerned).
- Pattern matching in the reducer using nested KeyPath is deprecated. If you need to match a `binding` action, you should match the `KeyPath` from the `@BindingState` property directly, and not a nested property of it.

### How does it work

You conform your `State` to the `BindableStateProtocol`:
```swift
struct State: BindableStateProtocol {
  @BindingState var textFromState: String
}
```
(also note the renamed `@BindableState -> @BindingState`).

In your `ViewState`, you declare `@BindingViewState` properties in the following way:
```swift
struct ViewState: Equatable {
  @BindingViewState var text
  init(state: State) {
    self._text = state.bindings.$textFromState
  }
}
```
Note that you need to assign the private storage (with a _) of the `@BindingViewState` property wrapper. You extract one value using the `bindings` property of `BindableStateProtocol` that exposes `BindingViewState` values using dynamic member lookup from `State`.

You can then use this binding like a standard SwiftUI `Binding`:
```
TextField("Type here", text: viewStore.$text)
```
If you're not using a dedicated `ViewState` (`ViewStore(store, observe: { $0 })`), you can also use the same `$` syntax on `BindingState` properties: `viewStore.$textFromState`.

There are now two protocols (`BindableState[Protocol]` and `BindableAction`), and two wrappers types: `BindingState` and `BindingAction`, as well as a `ViewState` specific `BindingViewState`.

### Alternative implementations

I'm only exposing here alternative ways to implement this specific approach, not alternative solutions to the problem.

#### Using the projected value `$` to assign the `ViewState`'s property in the init:

The idea would be to allow to write the more familiar:
```swift
self.$text = state.bindings.$textFromState
```
This is technically possible, but it forces us to allow `BindingViewState` to be initialized without arguments. It then becomes possible to forget to link this property to a `BindingState` in the `init`. It furthermore would only work if these `BindingViewState` properties were assigned after the standard, non-wrapped, properties.
Using `_` instead of `$` doesn't seem to be a price too high to pay for the safety it brings.

#### Reusing `BindingState` instead of `BindingViewState` in `ViewState`:
This is likely technically possible, but it would allow us to perform the assignation wrongly in two different ways:
```swift
self.text = state.textFromState // This would be wrong
self.$text = state.$textFromState // This would be wrong
self._text = state.bindings.$textFromState // This would be the only correct way
```
The only recovery would be to display runtime warnings, which is very weak.

#### Using a subscript instead of `bindings` in `BindableState[Protocol]`:
This one is totally open to discussion. One would write:
```
self._text = state[\.$textFromState]
// instead of 
self._text = state.bindings.$textFromState
```
The latter approach (which is currently used in this PR) autocompletes and reads nicely, but it exposes a `bindings` property on `State` that may pop up in autocompletion through dynamic lookup, especially with `ViewStore` observing `{ $0 }`:
```swift
viewStore.b // Proposes `bindings, binding(…, etc.
```
Using a subscript would position this utility a little off-track, but the subscript is always more cryptic and less discoverable, especially for newcomers.

That's it!

### DIFF
--- a/Sources/ComposableArchitecture/Reducer/Reducers/BindingReducer.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/BindingReducer.swift
@@ -1,6 +1,46 @@
 import SwiftUI
 
 /// A reducer that updates bindable state when it receives binding actions.
+///
+/// This reducer should typically be composed into the ``ReducerProtocol/body-swift.property-97ymy``
+/// of your feature's reducer:
+///
+/// ```swift
+/// struct Feature: ReducerProtocol {
+///   struct State: BindableStateProtocol {
+///     @BindingState var isOn = false
+///     // More properties...
+///   }
+///   enum Action: BindableAction {
+///     case binding(BindingAction<State>)
+///     // More actions
+///   }
+///
+///   var body: some ReducerProtocolOf<Self> {
+///     BindingReducer()
+///     Reduce { state, action in
+///       // Your feature's logic...
+///     }
+///   }
+/// }
+/// ```
+///
+/// This makes it so that the binding's logic is run before the feature's logic, i.e. you will only
+/// see the state after the binding was written. If you want to react to the state _before_ the
+/// binding was written, you can flip the order of the composition:
+///
+/// ```swift
+/// var body: some ReducerProtocolOf<Self> {
+///   BindingReducer()
+///   Reduce { state, action in
+///     // Your feature's logic...
+///   }
+/// }
+/// ```
+///
+/// If you forget to compose the ``BindingReducer`` into your feature's reducer, then when a binding
+/// is written to it will cause a runtime purple Xcode warning letting you know what needs to be
+/// fixed.
 public struct BindingReducer<State, Action>: ReducerProtocol
 where Action: BindableAction, State == Action.State {
   /// Initializes a reducer that updates bindable state when it receives binding actions.
--- a/Sources/ComposableArchitecture/Reducer/Reducers/BindingReducer.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/BindingReducer.swift
@@ -31,10 +31,10 @@ import SwiftUI
 ///
 /// ```swift
 /// var body: some ReducerProtocolOf<Self> {
-///   BindingReducer()
 ///   Reduce { state, action in
 ///     // Your feature's logic...
 ///   }
+///   BindingReducer()
 /// }
 /// ```
 ///
--- a/Sources/ComposableArchitecture/SwiftUI/Binding.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Binding.swift
@@ -15,9 +15,21 @@ public struct BindingState<Value> {
   /// The underlying value wrapped by the bindable state.
   public var wrappedValue: Value
 
+  let file: StaticString
+  let fileID: StaticString
+  let line: UInt
+
   /// Creates bindable state from the value of another bindable state.
-  public init(wrappedValue: Value) {
+  public init(
+    wrappedValue: Value,
+    file: StaticString = #file,
+    fileID: StaticString = #fileID,
+    line: UInt = #line
+  ) {
     self.wrappedValue = wrappedValue
+    self.file = file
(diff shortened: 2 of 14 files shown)


mbrandonw:
Just pushed a small commit to make `BindingState` conditionally `Sendable` (c7c957b48). Not sure why we didn't already have that.

The same should be done with `BindingAction`, but due to the type erasure we would need to require it wholesale. Can't do it conditionally. For that reason we are going to wait a bit longer to think about it.

Sources/ComposableArchitecture/SwiftUI/Binding.swift
```diff
@@ -114,39 +177,135 @@ extension BindableAction {
...
+      self.state[keyPath: keyPath].keyPath
+      as! WritableKeyPath<ViewAction.State, BindingState<Value>>
+    return self.binding(
+      get: { $0[keyPath: keyPath].wrappedValue },
+      send: { value in
+        #if DEBUG
+          let debugger = BindableActionViewStoreDebugger(
+            value: value,
+            bindableActionType: ViewAction.self,
+            // TODO: Restore context
```
mbrandonw, line 256: Should we try to propagate at least some context to here? We could have `BindingState` capture the file/fileID/line at the moment of creation to tie the warning to the field on the `State` that was used incorrectly.

Same for the TODO down below.

tgrapperon: Capture d’écran 2023-01-04 à 20 49 41
Do you mean something like this?

mbrandonw: Yeah maybe, but I think it needs finessing. I don't think it's correct to mention `Binding.swift:357` at all since it's library code that isn't helpful. So the whole message may need to be rethought.

tgrapperon: I'm capturing each property wrapper's `init` context, and I'm also passing the one from `BindingState` to `BindingViewState` (so this one has both at hand).
In both cases, I'm pointing one the `BindingState` value in the runtime warning.
I just pushed a commit, but we can reverse if this is not what you had in mind.

tgrapperon: Ok, I just saw your comment. I'll try to improve this.

Sources/ComposableArchitecture/Reducer/Reducers/BindingReducer.swift
```diff
@@ -1,6 +1,46 @@
...
+/// This makes it so that the binding's logic is run before the feature's logic, i.e. you will only
+/// see the state after the binding was written. If you want to react to the state _before_ the
+/// binding was written, you can flip the order of the composition:
+///
+/// ```swift
+/// var body: some ReducerProtocolOf<Self> {
+///   BindingReducer()
+///   Reduce { state, action in
+///     // Your feature's logic...
+///   }
```
tgrapperon, line 37:
```suggestion
///   Reduce { state, action in
///     // Your feature's logic...
///   }
///   BindingReducer()
```

mbrandonw (approved):

stephencelis (approved):
Great stuff!

PR #1794, opened 2023-01-05
Relax constraints around bindable view state

stephencelis:
With #1790 a `BindableStateProtocol` was introduced for deriving bindings to view state, and some new APIs required this constraint. On further use, this constraint seems overly strict, as leaf nodes that don't use view state don't really need this conformance.

While we're at it, let's rename `BindableStateProtocol` to `BindableViewState`, since it is currently only used to derive view bindings. If we decide the protocol can do more in the future, we could consider renaming it to `BindableState` (once the existing property wrapper type alias is fully removed).

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Forms.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Forms.swift
@@ -16,7 +16,7 @@ private let readMe = """
 // MARK: - Feature domain
 
 struct BindingForm: ReducerProtocol {
-  struct State: BindableViewState, Equatable {
+  struct State: Equatable {
     @BindingState var sliderValue = 5.0
     @BindingState var stepCount = 10
     @BindingState var text = ""
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Forms.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Forms.swift
@@ -16,7 +16,7 @@ private let readMe = """
 // MARK: - Feature domain
 
 struct BindingForm: ReducerProtocol {
-  struct State: BindableStateProtocol, Equatable {
+  struct State: BindableViewState, Equatable {
     @BindingState var sliderValue = 5.0
     @BindingState var stepCount = 10
     @BindingState var text = ""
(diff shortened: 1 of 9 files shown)


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Forms.swift
```diff
@@ -16,7 +16,7 @@ private let readMe = """
 // MARK: - Feature domain
 
 struct BindingForm: ReducerProtocol {
-  struct State: BindableStateProtocol, Equatable {
+  struct State: BindableViewState, Equatable {
```
mbrandonw, line 19: This can be dropped since no view state is used right?

```suggestion
  struct State: Equatable {
```

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-FocusState.swift
```diff
@@ -9,7 +9,7 @@ private let readMe = """
 // MARK: - Feature domain
 
 struct FocusDemo: ReducerProtocol {
-  struct State: BindableStateProtocol, Equatable {
+  struct State: BindableViewState, Equatable {
```
mbrandonw, line 12: Same here?

```suggestion
  struct State: Equatable {
```

Tests/ComposableArchitectureTests/BindingTests.swift
```diff
@@ -48,7 +48,7 @@ final class BindingTests: XCTestCase {
 
     func testBindingState() {
       struct BindingTest: ReducerProtocol {
-        struct State: BindableStateProtocol, Equatable {
+        struct State: BindableViewState, Equatable {
```
mbrandonw, line 51:
```suggestion
        struct State: Equatable {
```

Tests/ComposableArchitectureTests/BindingTests.swift
```diff
@@ -6,7 +6,7 @@ final class BindingTests: XCTestCase {
   #if swift(>=5.7)
     func testNestedBindingStateWithNestedMatching() {
       struct BindingTest: ReducerProtocol {
-        struct State: BindableStateProtocol, Equatable {
+        struct State: BindableViewState, Equatable {
```
mbrandonw, line 9:
```suggestion
        struct State: Equatable {
```

mbrandonw (approved):

stephencelis:
Closing for #1795.

Issue #1798, opened 2023-01-06
ViewStore binding provided as `NavigationStack` path gives fatal error on back navigation with 0.48.0, not 0.47.2


sebj:
### Description

After updating from 0.47.2 to 0.48.0, a ViewStore binding provided as the `path` to a SwiftUI `NavigationStack` will now result in a crash when navigating to a screen and then pressing the back button.

The error is:
```
Swift/RangeReplaceableCollection.swift:902: Fatal error: Can't remove more items from a collection than it contains
```

I can reliably reproduce this in a [brand new minimal project](https://gist.github.com/sebj/524c207ffb4057f157eab7dc0b9d499d) with Composable Architecture as an SPM dependency, where 0.48.0 (and also `main` branch @ `30015d1`) crashes and switching to 0.47.2 (and back and forth) does not in this situation.

It's totally possible this isn't a regression and I'm doing something wrong here, would love to know if so 😅 

### Checklist

- [X] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Expected behavior

Navigating to a screen, and then pressing the Back button navigates back successfully (does not crash)

### Actual behavior

A fatal error occurs.

### Steps to reproduce

1. Add the following Gist contents to a new project with Composable Architecture as an SPM dependency: https://gist.github.com/sebj/524c207ffb4057f157eab7dc0b9d499d
2. Tap 'Go to A' or 'Go to B'
3. Tap '< Back' to go back
4. Observe fatal error

### The Composable Architecture version information

0.48.0 & `main` @ `30015d1`

### Destination operating system

iOS 16

### Xcode version information

Version 14.2 (14C18)

### Swift Compiler version information

```shell
swift-driver version: 1.62.15 Apple Swift version 5.7.1 (swiftlang-5.7.1.135.3 clang-1400.0.29.51)
Target: arm64-apple-macosx13.0
```

PR #1802, opened 2023-01-07
Wrap view store binding to avoid reading stale state

### LINKED ISSUE #1800: Find better workaround for stale ForEach rows reading from bindings
We recently got an issue #1783 of a crash in a pretty simple SwiftUI view using TCA. The problem is that stale, _removed_ rows from a ForEach will sometimes read from its binding, which means trying to read data from a row no longer in the collection, resulting in a crash.

A fix was attempted in #1784, as well as a refinement in #1785, but unfortunately those lead to a different, and even simpler crash, which was report in #1798. So, we reverted the change.

Fortunately the fix to the original issue is to just use `ForEachStore` because we have baked in [special logic][foreachstore-special-logic] to work around this issue, but it would be nice to come up with a more holistic solution.

[foreachstore-special-logic]: https://github.com/pointfreeco/swift-composable-architecture/blob/30015d13a3ca0988b4674a83eccebbea747fe1f8/Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift#L113-L119

iampatbrown:
Attempt to fix #1800

The current method creates a binding with a location type of `ObservableObjectLocation`. By wrapping the binding the location type becomes `FunctionalLocation`. Bindings that use `ObservableObjectLocation` seem to have their getter called while evaluating if a views inputs have changed. This doesn't seem to be the case with `FunctionalLocation`. Which kind of makes sense.

I've confirmed this works for #1783 and #1798 as well as checked animations work as expected.

@stephencelis @mbrandonw @tgrapperon maybe you can stress test this a bit to see if there are any unwanted effects. I imagine there is going to be some sort of optimisation that can be applied to bindings that use `ObservableObjectLocation`, but don't really think that's relevant here.

### DIFF
--- a/Sources/ComposableArchitecture/ViewStore.swift
+++ b/Sources/ComposableArchitecture/ViewStore.swift
@@ -466,8 +466,10 @@ public final class ViewStore<ViewState, ViewAction>: ObservableObject {
     get: @escaping (ViewState) -> Value,
     send valueToAction: @escaping (Value) -> ViewAction
   ) -> Binding<Value> {
-    ObservedObject(wrappedValue: self)
+    let base = ObservedObject(wrappedValue: self)
       .projectedValue[get: .init(rawValue: get), send: .init(rawValue: valueToAction)]
+    
+    return Binding(get: { base.wrappedValue }, set: { base.transaction($1).wrappedValue = $0 })
   }
 
   /// Derives a binding from the store that prevents direct writes to state and instead sends


tgrapperon:
*"Yo Dawg, I heard you like bindings, so we put bindings into your bindings so you can bind while you bind!"*

Great find! I think that most of the problems are caused by the fact that the `ObservedObject` we were using is not installed on the graph. So while it was fixing some animations issues, it likely wasn't participating as it should to the invalidation process. I would guess that function backed ones are likely eagerly evaluated, whereas `ObservedObject` based ones are maybe coalesced up to the next render pass. I haven't checked yet if a simple functional binding that would call `objectWillChange()` in its setter wouldn't have helped with animations (maybe it was like that at some point).

I'll try to give them a spin soon to check if I'm spotting issues.

barabashd:
Nice catch @iampatbrown! I'm curious about the 'transaction($1)' purpose here. Could you explain it, please? Because removing this part still fixes both issues.

iampatbrown:
> Nice catch @iampatbrown! I'm curious about the 'transaction($1)' purpose here. Could you explain it, please? Because removing this part still fixes both issues.

@barabashd `.transaction($1)` allows any animations or transactions explicitly set on the binding to work correctly. eg:

```swift
Toggle("Toggle", isOn: $isOn.animation(.easeIn))
```

Using `withTransaction(transaction) { base.wrappedValue = newValue }` as an alternative has slightly different behaviour. It causes explicit animations/transactions to take precedence over other calls to `withAnimation`/`withTransaction`.

iampatbrown:
> I think that most of the problems are caused by the fact that the `ObservedObject` we were using is not installed on the graph. So while it was fixing some animations issues, it likely wasn't participating as it should to the invalidation process.

@tgrapperon It's been a while since stumbling across the `ObservedObject` approach, can't remember having much success without it though. My guess was that something happened under the hood that tracked state changes and forwarded explicit animations/transactions correctly.

>  I haven't checked yet if a simple functional binding that would call objectWillChange() in its setter wouldn't have helped with animations (maybe it was like that at some point).

Pretty sure the main thing was forwarding explicit binding animations/transactions. Accessing the current `Transaction` to merge/override any explicit binding transaction could also work, though it's not really supported.

stephencelis (approved):
Seems to work great! We'll get a couple integration tests in place in another PR.

Issue #1811, opened 2023-01-10
Dependency override issue with 0.49.1


jaanussiim:
### Description

Dependency override that is working in 0.48.1 is failing when switching to 0.49.1

Root dependency is failable. When scoping child feature, usable dependency is extracted from the root dependency.

Calling the dependency in child reducer BestOffers is getting the root failure.

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Steps to reproduce

Sample project here: https://github.com/jaanussiim/tca-dependencies-issue

Scoping the usable dependency:
https://github.com/jaanussiim/tca-dependencies-issue/blob/main/TCADependencies/Application.swift#L29

Creating the child dependency:
https://github.com/jaanussiim/tca-dependencies-issue/blob/main/TCADependencies/OffersClient.swift#L22

### The Composable Architecture version information

0.49.1

### Destination operating system

iOS 16.2

### Xcode version information

14.2

tgrapperon:
Hey @jaanussiim!
The problem is that you modelled your dependency as a reference type. Between the moment you reduce `BestOffers.Action.loadOffer` and the moment it calls `offersClient.loadOffers()`, the `OffersClient` has changed back to the root one. You can check this out by capturing the `offersClient` reference in your `.fireAndForget`:
```swift
case .loadOffers:
  return EffectTask.fireAndForget { [offersClient = self.offersClient] in
    try await offersClient.loadOffers()
  }
}
```
This will not crash. But this will however be very prone to weird issues like these that are complex to debug. I think that you'd get warnings with the `-warn-concurrency` flag.

I'm not sure either you'll end-up sending message to the correct instances, but it looks like your dependency is a singleton, so it looks stable-ish.

I would advise to remodel your dependency into something that exposes a value type that can be safely captured by the `.fireAndforget` (and `.run` and `.task`) closures. I think that wrapping it in a struct should perform the capture trick automatically for you, and would probably be enough to fix your issue.

I'm not sure what has exactly changed between 48.1 and 49.1 (well, a lot has changed! but don't know what caused the difference of behavior), and I'm not sure this is a regression.

[...]

PR #1813, opened 2023-01-10
Bump `Dependencies` to 0.1.2

tgrapperon:
This pins the fix to #1811.


stephencelis:
@tgrapperon Thanks! Looks like the `Package.resolved` files aren't updated. Can you open the workspace to resolve them, as well as run `swift package update` from the root of the repo? That should update both files.

tgrapperon:
Done! I forgot we were doing that!

PR #1824, opened 2023-01-11
Add support for SwiftUI withTransaction API

drucelweisse:
This adds EffectTask support to perform block with SwiftUI `Transaction`. As all animation stuff happens using `Transaction`, and Animation is a concrete example of Transaction  it seems reasonable to support that, even if it has few use cases.

Changes: Instead of creating another function, I generalized it for transactions and in `.animation(_ animation: Animation? = .default)` function calls it under the hood.

Example from SwiftUI Q&A Sessions

> Is there way to show `.fullScreenCover` without animation? I don't want to user see animated cover show from bottom to top. because it still see underline view in a short time. 

> You can create a `Transaction` value, set its `disablesAnimations` property to true, then set the binding that presents the fullScreenCover inside a `withTransaction` block. Something like:
```
var t = Transaction()
t.disablesAnimations = true
withTransaction(t) {
    isCoverPresented = true
}
```

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Animation.swift
+++ b/Sources/ComposableArchitecture/Effects/Animation.swift
@@ -21,7 +21,7 @@ extension EffectPublisher {
     case let .publisher(publisher):
       return Self(
         operation: .publisher(
-          AnimatedPublisher(upstream: publisher, animation: animation).eraseToAnyPublisher()
+          TransactionPublisher(upstream: publisher, transaction: Transaction(animation: animation)).eraseToAnyPublisher()
         )
       )
     case let .run(priority, operation):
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hey @drucelweisse, thanks for the PR! 

Overall we think this is a pretty good change. We need a little more time to be able to fully look over the code, but one thing that popped out to us was that maybe even `Effect`'s [`Send`](https://github.com/pointfreeco/swift-composable-architecture/blob/a75e9dde7d4b4a43f0b3ea3742e45d704f412c42/Sources/ComposableArchitecture/Effect.swift#L377) type should take a transaction so that you can do things like:

```swift
Effect.run { send in 
  await send(.response, transaction: disableAnimations)
}
```

Further, even `ViewStore.send` could take a transaction too. Want to experiment with adding those too and we'll come back to this soon?

drucelweisse:
Hi @mbrandonw! I've definitely missed out that `Send` also have an Animation overload. Will add that into pr.

Sources/ComposableArchitecture/Effect.swift
```diff
@@ -375,11 +375,21 @@ public struct Send<Action> {
   ///   - action: An action.
   ///   - animation: An animation.
   public func callAsFunction(_ action: Action, animation: Animation?) {
-    guard !Task.isCancelled else { return }
-    withAnimation(animation) {
-      self(action)
-    }
+    callAsFunction(action, transaction: Transaction(animation: animation))
   }
+
```
stephencelis, line 380:
```suggestion
```

Sources/ComposableArchitecture/ViewStore.swift
```diff
@@ -290,7 +290,20 @@ public final class ViewStore<ViewState, ViewAction>: ObservableObject {
   ///   - animation: An animation.
   @discardableResult
   public func send(_ action: ViewAction, animation: Animation?) -> ViewStoreTask {
-    withAnimation(animation) {
+    send(action, transaction: Transaction(animation: animation))
+  }
+
```
stephencelis, line 295:
```suggestion
```

PR #1826, opened 2023-01-11
Mark popover as unavailable in watchOS (navigation)

scogeo:
Really looking forward to the new navigation APIs in TCA and started experimenting with this branch.  I recognize this branch is a work in progress, but currently it doesn't compile under watchOS due to the lack of popover APIs in watchOS.  This PR correctly marks the popover APIs as unavailable in watchOS to address the issue.


mbrandonw (approved):
Good catch! Be aware that this branch is _highly_ unstable, but this is definitely a good change for now. Thanks!

PR #1827, opened 2023-01-12
Automate UI testing of bindings animations

tgrapperon:
When bindings are activated, we can record the size of the view as it animates (or not). We can then simply extract an estimate of the time it took to perform these changes, and compare with expectations. 
It seems that we have to go back to UIKit's world to measure the changes when the animation performs, as SwiftUI is terser on the subject and doesn't produce a stream of animating sizes, even using a `TimelineView` with the `.animation`'s schedule.
This is working on top of the existing test bench. Measures are bundled in a view modifier, so it remains interactive if needed. The numbers that it now displays above the circles are the duration estimates (this is what the UITest reads).
It shouldn't be too flakey, as animation durations are rounded to the 1/10s. This can be improved relatively easily if required.


tgrapperon:
I've replaced the `UIRepresentable` by an `Animatable` view to perform the measure (an idea from @iampatbrown).

stephencelis:
@tgrapperon Sorry for sleeping on this. Is this something you think we should still merge? Care to get it all resolved before we take another look?

tgrapperon:
Sorry, I missed your notification. As far as I remember, this provided a way to check automatically that animated changes were indeed animated, promoting non-animated changes to UITest failures.
Feel free to close if you think it's not worth it!

tgrapperon:
There are a few conflicts to resolve, and I don't remember what was the exact state of affairs at the end of the PR, so I'm closing for now and we'll maybe revisit some day if we want to test animations (here or in another library).

PR #1828, opened 2023-01-12
Explicitly depend on OrderedCollections

kalupas226:
I added explicitly dependency on OrderedCollections based on https://github.com/pointfreeco/swift-composable-architecture/discussions/1818.


stephencelis (approved):
Thanks!

PR #1832, opened 2023-01-13
Add ability to specify a keypath for the enum to ifCaseLet

dahlborn:
This extends the `_IfCaseLetReducer` and the `.ifCaseLetFunction` on the `ReducerProtocol` with the option to specify a KeyPath to a property containing the enum to be checked instead of requiring the whole State to be an enum.

### Reasoning:

I recently started working with TCA on a new project and I encountered it multiple times that we might want to use a SwitchStore to handle multiple mutually exclusive children, but also might have additional state properties we might want to handle.

```
struct SomeFeature: ReducerProtocol {
  struct State: Equatable {
    enum SomeEnum: Equatable {
      case someChild(SomeChild.State)
      case someOtherChild(SomeOtherChild.State)
    }

    var someLocalState: String
    var someChildStateEnum: SomeEnum
  }

  enum Action {
    case someChild(SomeChild.Action)
    case someOtherChild(SomeOtherChild.Action)
  }

    ???
  }
}
```

The way I understand this there are currently two ways to handle this. You could do something like this:

```
    Scope(state: \.someChildStateEnum, action: /.self) {
      EmptyReducer()
        .ifCaseLet(/State.SomeEnum.someChild, action: /Action.someChild) {
          SomeChild()
        }
        .ifCaseLet(/State.SomeEnum.someOtherChild, action: /Action.someOtherChild) {
          SomeOtherChild()
        }
    }
    Reduce({ state, action in
      return .none
    })
}
```

Having the extra `Scope` and `EmptyReducer` here (It would also be possible to nest multiple `Scope`) does not feel very clean, might be hard to understand and reintroduces the ordering issue that the function is supposed to avoid.

It is also possible to extract the `SwitchStore` part of the view into an extra Feature with its own reducer and model that reducers state to be an enum. But sometimes that might feel like unnecessary extra work to just work around the `ifCaseLet` limitation, if the parent now only hold very little state and logic.

With this change it would be possible to simply use the `ifCaseLet` like in a pure `SwitchStore` view.

```
  Reduce({ state, action in
    return .none
  })
  .ifCaseLet(\.someChildStateEnum, case: /State.SomeEnum.someChild, action: /Action.someChild) {
    SomeChild()
  }
  .ifCaseLet(\.someChildStateEnum, case: /State.SomeEnum.someOtherChild, action: /Action.someOtherChild) {
    SomeOtherChild()
  }
}
```

I also saw questions regarding similar examples in the discussions, but those were often related to navigation or mentioned the first solution as a workaround. 

I think this change would make using `SwitchStore` and `ifCaseLet` more ergonomic and should hopefully have no downsides. The only issue might be that it modifies the theoretically public `_IfCaseLetReducer` so it is a breaking change. Alternatively it would be possible to introduce something like a `_IfCaseLetPullbackReducer` to avoid modifying the existing Type, but since they model pretty much functionality I thought it would be better to integrate them.

### DIFF
--- a/Sources/ComposableArchitecture/Reducer/Reducers/IfCaseLetReducer.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/IfCaseLetReducer.swift
@@ -53,10 +53,83 @@ extension ReducerProtocol {
     file: StaticString = #file,
     fileID: StaticString = #fileID,
     line: UInt = #line
-  ) -> _IfCaseLetReducer<Self, Case> {
+  ) -> _IfCaseLetReducer<State, Self, Case> {
     .init(
       parent: self,
       child: `case`(),
+      toEnum: \.self,
+      toChildState: toCaseState,
+      toChildAction: toCaseAction,
+      file: file,
+      fileID: fileID,
+      line: line
+    )
+  }
+
+  /// Embeds a child reducer in a parent domain that works on a case of an enum property in the parent state.
+  ///
+  /// For example, if a parent feature's state contains a property with an enum of multiple children
+  /// states, then `ifCaseLet` can run a child reducer on a particular case of the enum:
+  ///
+  /// ```swift
+  /// struct Parent: ReducerProtocol {
+  ///   enum State {
+  ///     enum LoginState {
+  ///       case loggedIn(Authenticated.State)
+  ///       case loggedOut(Unauthenticated.State)
+  ///     }
+  ///     var loginState: LoginState
+  ///   }
+  ///   enum Action {
+  ///     case loggedIn(Authenticated.Action)
+  ///     case loggedOut(Unauthenticated.Action)
+  ///     // ...
+  ///   }
+  ///
+  ///   var body: some ReducerProtocol<State, Action> {
+  ///     Reduce { state, action in
+  ///       // Core logic for parent feature
+  ///     }
+  ///     .ifCaseLet(\.loginState, case: /State.loggedIn, action: /Action.loggedIn) {
+  ///       Authenticated()
+  ///     }
+  ///     .ifCaseLet(\.loginState, case: /State.loggedOut, action: /Action.loggedOut) {
+  ///       Unauthenticated()
+  ///     }
+  ///   }
+  /// }
+  /// ```
+  ///
+  /// The `ifCaseLet` forces a specific order of operations for the child and parent features. It
+  /// runs the child first, and then the parent. If the order was reversed, then it would be
+  /// possible for the parent feature to change the case of the enum, in which case the child
+  /// feature would not be able to react to that action. That can cause subtle bugs.
+  ///
+  /// It is still possible for a parent feature higher up in the application to change the case of
+  /// the enum before the child has a chance to react to the action. In such cases a runtime
+  /// warning is shown in Xcode to let you know that there's a potential problem.
+  ///
+  /// - Parameters:
+  ///   - toEnum: A writable keypath from the parent state to a property containing an enum.
+  ///   - toCaseState: A case path from the picked out enum to a case containing child state.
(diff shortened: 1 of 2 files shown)


stephencelis:
@dahlborn Thanks for the PR! This may be a lack of documentation, but there is an overload of `Scope` that takes a case path to enum state that helps you avoid the `EmptyReducer`:

```swift
  Scope(state: \.someChildStateEnum, action: /.self) {
    Scope(state: /State.SomeEnum.someChild, action: /Action.someChild) {
      SomeChild()
    }
    Scope(state: /State.SomeEnum.someOtherChild, action: /Action.someOtherChild) {
      SomeOtherChild()
    }
  }
  Reduce { state, action in
    return .none
  }
}
```

While there are still things that may stand out to you (having to remember to order the child stuff up front, double-nested `Scope`-ing), it's not so so bad.

I think a lot of the nested property/case ergonomic issues that APIs like this are trying to solve for are showing that Swift could improve a lot if it shipped first-class case paths and case access, and allowed things to compose nicely out-of-the-box. Then you could flatten the nesting with something like:

```swift
  Reduce { state, action in
    // core logic...
  }
  // speculative case path composition and use syntax via `\._?`:
  .ifCaseLet(state: \.someChildStateEnum.someChild?, action: \.someChild?) {
    SomeChild()
  }
  .ifCaseLet(state: \.someChildStateEnum.someOtherChild?, action: \.someOtherChild?) {
    SomeOtherChild()
  }
}
```

At the end of the day I think we're in a bind since we can't explode every API with every kind of composition, but maybe this is reason to explore the `OptionalPath`/`WritableOptionalPath` wrapper that allows us to compose key paths and case paths directly, which would at least give us the following with the existing APIs:

```swift
  Reduce { state, action in
    // core logic...
  }
  .ifCaseLet(
    state: (\.someChildStateEnum).appending(path: /State.SomeEnum.someChild),
    action: /Action.someChild
  ) {
    SomeChild()
  }
  .ifCaseLet(
    state: (\.someChildStateEnum).appending(path: /State.SomeEnum.someOtherChild),
    action: /Action.someOtherChild
  ) {
    SomeOtherChild()
  }
}
```

Any thoughts given the above?

dahlborn:
@stephencelis Thanks for your response. Your documentation is good, I was aware of the nested `Scope` variant. But I was looking at the `.ifCaseLet` alternative for the reasons you mentioned.

I completely agree that having language support for case paths would be the optimal solution, but I think till then having a more ergonomic solution would be nice, since it seems to be a question that comes up pretty often. (Even though it often comes up regarding navigation, which you are already solving)

I like the `OptionalPath` wrapper you mentioned. I guess the question would be whether adding this complexity would be justified for the things it solves. Was there a reason you decided to have it as a helper in the `isowords` project instead of adding it to tca or case-paths?

stephencelis:
> I like the `OptionalPath` wrapper you mentioned. I guess the question would be whether adding this complexity would be justified for the things it solves. Was there a reason you decided to have it as a helper in the `isowords` project instead of adding it to tca or case-paths?

We think it works really well, but in general the topic is advanced enough that we didn't want it to be required learning for folks, so we decided that using existing tools that take key paths and case paths strikes a better balance for beginners for now. We also think there are ergonomic refinements to explore if we do ship optional paths in library form (more seamless integration with key paths, supporting a type-erased hierarchy similar to key paths, etc.), but we haven't had time to explore these ideas in full.

If the `OptionalPath` helpers we ship in isowords is useful to you, we definitely recommend bringing them into your project, though!

stephencelis:
@dahlborn OK to close this out since we can't possible add overloads for every nesting of key path and case path? In the meantime, folks with more advanced compositional needs can use the existing APIs in a nested fashion, or introduce optional paths to their code base. For reference:

https://github.com/pointfreeco/isowords/blob/main/Sources/TcaHelpers/OptionalPaths.swift

PR #1834, opened 2023-01-13
Conditionally Conform BindableState to Sendable.

jshier:
This PR makes `BindableState` conditionally conform to `Sendable`, as was done in #1790 but was subsequently lost when that PR was reverted. I can add any other desired conformances as well.


mbrandonw (approved):
Great find, thanks!

If you find any other types that are easy to make `Sendable` we'll happily take a PR.

PR #1844, opened 2023-01-18
Add `prepareDependencies` to `Store.init`

stephencelis:
Seems useful to have `Store` mirror `TestStore` here.

I've also updated all our tests to prefer this initializer, as it's a bit less noisy for declaring upfront dependencies.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-AnimationsTests.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudiesTests/01-GettingStarted-AnimationsTests.swift
@@ -7,13 +7,14 @@ import XCTest
 @MainActor
 final class AnimationTests: XCTestCase {
   func testRainbow() async {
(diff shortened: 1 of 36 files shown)


acosmicflamingo:
Oh this is nice; can't wait to clean up some of my tests using this new closure :)

stephencelis:
@acosmicflamingo This helper was actually introduced to `TestStore` awhile back! This PR just updates our tests to actually use it, and introduces the corresponding initializer on `Store`, which can be helpful to configure certain live dependencies at the root of your application.

acosmicflamingo:
> @acosmicflamingo This helper was actually introduced to `TestStore` awhile back! This PR just updates our tests to actually use it, and introduces the corresponding initializer on `Store`, which can be helpful to configure certain live dependencies at the root of your application.

Oh neat! Perhaps that was merged before I signed up to get email notifications anytime anything happens in this repo like a stalker ;)

PR #1855, opened 2023-01-20
Rename `BindableState` to `BindingState`

stephencelis:
The -`able` naming evokes protocols in Swift, and is an outlier when considered alongside the rest of TCA's binding tools:

- `BindingAction`: concrete type
- `BindableAction`: protocol
- `BindingReducer`: concrete type

So, let's make things consistent and rename `@BindableState` to `@BindingState`.

The one caveat is that Swift diagnostics for such a deprecation aren't great, so users won't get proactive warnings here for the time being:

https://github.com/apple/swift/issues/63139

We may just want to keep the deprecation around till it does...


mbrandonw (approved):

iampatbrown:
> We may just want to keep the deprecation around till it does...

@stephencelis was there a plan on using `BindableState` as a protocol in the future?

stephencelis:
@iampatbrown There have been experiments run by @tgrapperon for such a thing, so it'd be nice to free up the namespace if we need it. No concrete plans as of yet, though.

PR #1856, opened 2023-01-23
Unconstrain TestStore action for predicate/case path `receive`

stephencelis:
These methods are currently defined in a constrained extension, but it's not necessary, so let's loosen the constraint.

### DIFF
--- a/Sources/ComposableArchitecture/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestStore.swift
@@ -1203,6 +1203,111 @@ extension TestStore where ScopedState: Equatable, Action: Equatable {
     )
   }
 
+  // NB: Only needed until Xcode ships a macOS SDK that uses the 5.7 standard library.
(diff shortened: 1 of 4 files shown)


Sources/ComposableArchitecture/TestStore.swift
```diff
@@ -1626,10 +1632,10 @@ extension TestStore where ScopedState: Equatable, Action: Equatable {
       while let receivedAction = self.reducer.receivedActions.first,
         !predicate(receivedAction.action)
       {
-        actions.append(receivedAction.action)
-        self.withExhaustivity(.off) {
-          self.receive(receivedAction.action, file: file, line: line)
-        }
+        self.reducer.receivedActions.removeFirst()
+        let expectedAction = receivedAction.action
+        actions.append(expectedAction)
```
mbrandonw, line 1637: Can inline...

mbrandonw (approved):
Looks good! I just added some tests with non-equatable actions to guard against regressions.

PR #1857, opened 2023-01-23
Improve `TestStore.init` ergonomics

stephencelis:
It's a bit easy to dive into testing and be met with bad ergonomics if you forgot to conform `State` to be `Equatable`. Right now it doesn't require `Equatable` because it's possible to scope test stores and assert against view state/actions. This feature is pretty niche, though, and shouldn't make the more general case painful, so let's take a few steps to improve things:

1. We can deprecate the `TestStore.init` that takes non-equatable state
2. We can introduce a `TestStore.init(…:observe:send:…)` for testing view state/actions.
3. We can deprecate `TestStore.scope` (scoping a test store multiple times doesn't seem useful enough to support).


mbrandonw (approved):
nice

PR #1860, opened 2023-01-23
Use @StateObject for iOS 15+ alert modifier.

mbrandonw:
Just a small win I found. Eventually we want to move a lot of `@ObservedObject`s to `@StateObject`s in the library, but can't until we drop iOS 13. However, these two instances can be done now.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Alert.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Alert.swift
@@ -34,7 +34,7 @@ extension View {
 // NB: Workaround for iOS 14 runtime crashes during iOS 15 availability checks.
 @available(iOS 15, macOS 12, tvOS 15, watchOS 8, *)
 private struct NewAlertModifier<Action>: ViewModifier {
-  @ObservedObject var viewStore: ViewStore<AlertState<Action>?, Action>
+  @StateObject var viewStore: ViewStore<AlertState<Action>?, Action>
   let dismiss: Action
 
   func body(content: Content) -> some View {
--- a/Sources/ComposableArchitecture/SwiftUI/ConfirmationDialog.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/ConfirmationDialog.swift
@@ -40,7 +40,7 @@ extension View {
 // NB: Workaround for iOS 14 runtime crashes during iOS 15 availability checks.
 @available(iOS 15, macOS 12, tvOS 15, watchOS 8, *)
 private struct NewConfirmationDialogModifier<Action>: ViewModifier {
-  @ObservedObject var viewStore: ViewStore<ConfirmationDialogState<Action>?, Action>
+  @StateObject var viewStore: ViewStore<ConfirmationDialogState<Action>?, Action>
   let dismiss: Action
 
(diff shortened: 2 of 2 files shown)


stephencelis (approved):

johnrogers:
Hey guys, we just updated our project to use 0.50.1 and have started to notice that some confirmation dialogs and alerts have ceased to function. As in, they don't appear at all. This only occurs on iOS 15 and 16 -- I have tested iOS 14 and the dialogs/alerts in the affected areas still appear correctly. Could this be anything to do with this change?

tgrapperon:
@johnrogers One thing that can be useful to check if you open a discussion, is the way you're extracting your `confirmationDialog`/`alert` store. For example, something like:
```swift
isRegistered ? store.scope(state:…) : store.scope(state:…)
```
would now break. But it can also be something more subtle, or even a SwiftUI bug.

mbrandonw:
Hi @johnrogers, I just wanted to follow up with this and see if you got to the bottom of it? We're hoping to start embracing `@StateObject` more fully in the library once we can drop iOS 13 support, and so we'd like to know if there is something weird going on with them.

tgrapperon:
Hey @mbrandonw, yes we did: see https://pointfreecommunity.slack.com/archives/C04KQQ7NXHV/p1676257703922689?thread_ts=1675899910.360799&cid=C04KQQ7NXHV and below. In a few words, some store was scoped without using the `state` argument and was capturing an outer `state` value provided by a `NavigationLink` binding. `StateObject` was thus likely capturing a store that continuously returned this value. Refactoring the scope to use the `state` argument allowed to fix the issue.

alexito4:
Just commenting here to say that we also had issues with this, thanks for posting the solution, it was the same problem indeed. 

Our case was that we had state being an enum and we had some code in the UI that switched on the enum, and in one case used a scope to reach into the state but since the staet inside the scope was the base it meant that reaching for the associated type resulted in optional. Even if in theory that code would run only if the outter switch matched, I don't trust SwiftUi to not run the code in weird scenarios. So for safety we decided to take the scoped state from the associated value of the outer switch instead of using the state inside the scope. It was the easiest thing back then. 

Something like this:

```swift
            switch viewStore.stateEnum {
            case .oneCase:
                SomeView()
            case let .caseWithState(state):
                OtherViewThatNeedsState(
                    store.scope(
                        state: { _ in state }, // here, using the state in the closure is a PITA so we cheated and used the outer one!
                        action: ...n
                    )
                )
            }
        }
```

But now we have SwitchStore and CaseLet so we switched to that instead and it works quite well 🤞🏻

tgrapperon:
@alexito4 Indeed, `store.scope(state: { _ in …` is very likely pathological, and should be at the very least considered as a code smell. I'm not sure we can do things to force the argument to be used. Maybe the new ownership thing could help. I know that you're familiar with it (I'm not yet!)!

alexito4:
Thanks for confirming @tgrapperon 
Solving this with ownership will be intersting indeed, but AFAIK there are no plans (yet?) to introduce "must use" types in Swift. From reading others deal with this in other language seems like is a big complication that is not always beneficial. We shall see! ^^

PR #1863, opened 2023-01-25
Improve reducer builder inference and prepare for Swift 5.8

stephencelis:
Swift 5.8 has some changes coming to how result builders are type checked and requires us to make some changes for things to continue to compile. See the discussion [here](https://github.com/apple/swift/issues/62978) for more info.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/Deprecations.swift
+++ b/Sources/ComposableArchitecture/Internal/Deprecations.swift
@@ -5,6 +5,13 @@ import XCTestDynamicOverlay
 
 // MARK: - Deprecated after 0.49.2
 
+@available(
+  *,
+  deprecated,
+  message: "Use 'ReducerBuilder<_, _>' with explicit 'State' and 'Action' generics, instead."
+)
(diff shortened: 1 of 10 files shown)


Sources/ComposableArchitecture/Internal/Deprecations.swift
```diff
@@ -5,6 +5,13 @@ import XCTestDynamicOverlay
 
 // MARK: - Deprecated after 0.49.2
 
+@available(
+  *,
+  deprecated,
+  message: "Use 'ReducerBuilder<_, _>' with explicit 'State' and 'Action' generics, instead."
+)
```
jeffersonsetiawan, line 12: curious what the downside of using this typealias, as I use it almost in every feature.

davdroman:
Sad to see `ReducerBuilderOf` go. Is the overall inference better in Swift 5.8 though, to the point it shouldn't require explicit `Reduce<State, Action>` for big reducer bodies?

mbrandonw (approved):
Absolutely amazing! Thanks for all the help @xedin!

PR #1865, opened 2023-01-25
Update SwiftUI Navigation to support new alert mappings

stephencelis:
Alerts in SwiftUI Navigation are having a small breaking change in 0.6.0, so we need to release TCA shortly after to prevent folks from being in a broken state.

### DIFF
--- a/ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -14,8 +14,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-argument-parser",
       "state" : {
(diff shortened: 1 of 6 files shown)


ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -95,8 +95,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/pointfreeco/swiftui-navigation",
       "state" : {
-        "revision" : "46acf5ecc1cabdb28d7fe03289f6c8b13a023f52",
-        "version" : "0.4.5"
+        "branch" : "alert-mappings",
```
stephencelis, line 98: Will lock to 0.6.0 once it's cut.

mbrandonw (approved):

PR #1887, opened 2023-02-02
Reducer builder improvements

stephencelis:
A couple improvements to reducer builders:

  * Let's add explicit parameter names to builders where appropriate. It reads a bit nicer when you rarely need to call them and aren't using trailing closure syntax.

  * Let's add optional `stateType` and `actionType` to the `Reduce` initializer. It provides better discovery for improving type checking when building reducers than manually specifying the generics.

### DIFF
--- a/Sources/ComposableArchitecture/Reducer/Reducers/Reduce.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/Reduce.swift
@@ -15,6 +15,18 @@ public struct Reduce<State, Action>: ReducerProtocol {
 
   /// Initializes a reducer with a `reduce` function.
   ///
+  /// - Parameter reduce: A function that is called when ``reduce(into:action:)`` is invoked.
+  @inlinable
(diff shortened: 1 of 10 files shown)


Sources/ComposableArchitecture/Reducer/Reducers/Reduce.swift
```diff
@@ -21,6 +21,27 @@ public struct Reduce<State, Action>: ReducerProtocol {
...
+  /// - Parameters:
+  ///   - stateType: The type of state. Assists the compiler in type checking the reduce closure.
+  ///     Specify when autocompletion fails to suggest properties on state.
+  ///   - actionType: The type of action. Assists the compiler in type checking the reduce closure.
+  ///     Specify when autocompletion fails to suggest action cases.
+  ///   - reduce: A function that is called when ``reduce(into:action:)`` is invoked.
+  @inlinable
+  public init(
+    into stateType: State.Type,
+    action actionType: Action.Type,
```
stephencelis, line 39: Added an overload because specifying _both_ is required for Swift to type-check either in the body. 😕

PR #1895, opened 2023-02-08
Allow waiting for Send to finish

kabiroberai:
This PR adds a `SendTask` type returned from calling `Send`. Like `ViewStoreTask`, a `SendTask` can be awaited (`await send(.foo).finish()`) to wait for the child action to finish before continuing.

The implementation is currently slightly broken in that converting an `EffectTask.run` into an `EffectTask.publisher` (via the `publisher` property or a Combine operator) breaks things. See the currently failing `testRunFinishPublisher` for reference.

See also #1823.


Sources/ComposableArchitecture/TestStore.swift
```diff
@@ -2316,6 +2316,20 @@ class TestReducer<State, Action>: ReducerProtocol {
...
+      let effect = LongLivingEffect(action: action)
+      return EffectTask.run(priority: priority) { [effectDidSubscribe, weak self] send in
+        self?.inFlightEffects.insert(effect)
+        Task {
+          await Task.megaYield()
+          effectDidSubscribe.continuation.yield()
+        }
+        await perform(send)
+        // FIXME: is this thread-safe?
+        self?.inFlightEffects.remove(effect)
```
kabiroberai, line 2330: @mbrandonw is `TestReducer` intended to be `MainActor`-isolated? If so, should this `run` task be performed on the MainActor too? Given that `self` isn't Sendable I think this line (and the `insert` above) could introduce a race condition unless we lock/dispatch to the main thread.

Sources/ComposableArchitecture/Effects/Publisher.swift
```diff
@@ -23,7 +23,11 @@ extension EffectPublisher: Publisher {
       return .create { subscriber in
         let task = Task(priority: priority) { @MainActor in
           defer { subscriber.send(completion: .finished) }
-          let send = Send { subscriber.send($0) }
+          let send = Send {
+              subscriber.send($0)
+              // FIXME: return something reasonable
+              return nil
```
kabiroberai, line 29: This is the crux of why `testRunFinishPublisher` currently fails; I can't think of a sane way to return a `SendTask` from here.

Sources/ComposableArchitecture/Effect.swift
```diff
@@ -411,14 +418,55 @@ public struct Send<Action> {
...
+///   print("Done!")
+/// }
+/// ```
+///
+/// > Note: Unlike Swift's `Task` type, ``SendTask`` automatically sets up a cancellation
+/// > handler between the current async context and the task.
+///
+/// See ``TestStoreTask`` for the analog returned from ``TestStore``, and ``ViewStoreTask``
+/// for the analog returned from ``ViewStore``.
+public struct SendTask: Hashable, Sendable {
```
kabiroberai, line 447: I wonder whether we should unify `ViewStoreTask` and `SendTask` into a single type, just didn't want to prematurely apply DRY here.

Sources/ComposableArchitecture/TestStore.swift
```diff
@@ -2316,6 +2316,20 @@ class TestReducer<State, Action>: ReducerProtocol {
         )
         .map { .init(origin: .receive($0), file: action.file, line: action.line) }
         .eraseToEffect()
+
+    case .run(let priority, let perform):
```
kabiroberai, line 2320: we can't piggyback off of Combine operators here anymore because we lose the `SendTask` return value; conceptually speaking conversion to a `Publisher` is now lossy.

mbrandonw:
Hey @kabiroberai, I just wanted to say thanks for investigating this! We haven't had time to look at this yet, but we will make time soon and we're excited for the possibilities. 🙂

kabiroberai:
fyi I have another PR lined up to warn if `send` is called after `Effect.run` returns because it invokes all sorts of undefined behaviour. The current diff depends on this one so I haven't made a PR yet but I can do that if/when this lands, or I can rebase that one on main if you'd like to see it sooner.

You might wanna read the commit message and tests for the rationale: https://github.com/kabiroberai/swift-composable-architecture/commit/64c2d88346a0c0741c681354216b1b0a2defd62f.

kabiroberai:
@mbrandonw @stephencelis I decided to work around the publisher issue with `runtimeWarn` by basically saying "you're holding it wrong" if users run access `EffectSendTask` methods on a `.run` effect that's been converted into a `.publisher`. I'd consider this PR production-ready if you folks think that's a reasonable solution.

mbrandonw:
Hi @kabiroberai, I played around with this a few days ago and it's looking really good! I wrote some additional tests to try pushing it to its limits and it seems to have caught everything.

I do think that completely bailing when encountering a publisher-style effect is going to be a bit of a bummer. When you first pitched the idea I thought you were saying that effects run out of the store weren't going to be able to take advantage of this mechanism, and _that_ I think is a fair trade off. 

However, losing this functionality just because a Combine publisher is used somewhere is gonna be a huge gotcha and will be hard to actually track down in practice. The runtime warning lets em know what went wrong, but in order for the person to fix they would need to go through every single effect and find where a Combine interface is being used.

However, it seems to me it is possible to support this even for publisher effects. At the end of the day we just need to construct a `Task<Void, Never>` that lives as long as the effect and such that when it is canceled it cancels the underlying effect.

What do you think? Is that possible? I'm hoping to have a little bit more free time soon to explore more in depth.

kabiroberai:
Just to ensure we're on the same page, it's okay to wait _for_ the result of a publisher-type effect

```swift
// case .tap
return .run { send in
  await send(.response).finish()
}
// case .response
return Just(...)
  .delay(...)
  .eraseToEffectPublisher()
```

In the above case we're already generating a Task for .response that finishes on completion and handles cancellation appropriately, like you suggested, since that's built into `Store.send`. 

The issue only occurs when you call `send` _from_ a publisher-type effect:

```swift
// case .tap
return .run { send in
  await send(.response).finish()
}
.eraseToEffectPublisher()
```

The more I think about it the more I realise that there are many cases where calling send from a publisher-like effect could result in unexpected behaviour – for example what happens here?

```swift
return .run { send in
  await send(.action1).finish()
  await send(.action2).finish()
}
.dropFirst()
```

Or

```swift
return .run { send in
  await send(.action1).finish()
  await send(.action2).finish()
}
.collect()
.flatMap(Publishers.Sequence.init)
```

One solution I thought of is leveraging backpressure to avoid "finishing" an EffectSendTask until there's subsequent demand, but using demand as a side-channel for communication is super limiting — I'm not sure how we'd specify _which_ effect the demand corresponds to in cases such as the following

```swift
return .run { send in
  async let result1 = await send(.response1).finish()
  async let result2 = await send(.response2).finish()
  _ = await (result1, result2)
}
```

All in all I'm not too sure what it would even mean to wait for a response inside a publisher (or even an AsyncSequence for that matter). The `.run` effect type is uniquely suited to this because it isn't modelled as a sequence that you can perform time-based operations on, but rather it's modelled as a single coroutine. Transforming a `.run` effect is also constrained in that it's more or less a 1:1 mapping which means there's no dropping/merging `send`s over time, as that would make the meaning of `finish()` ambiguous. Plus if we do ever add those operations we'll have to think about what it means to support them in a call-and-response world where we can actually observe the result of our actions, which isn't the case in Combine.

These are just my thoughts on the topic from a few days of playing around with this ofc, so it's definitely possible that I'm missing something. More than happy to chat here/in the associated discussion/on Slack about pros and cons of the various approaches we can take.

mbrandonw:
> The issue only occurs when you call `send` _from_ a publisher-type effect:

Ahhh, ok I was misunderstanding. That is a much smaller limitation and far easier to document and understand when something goes wrong. Let me spend a little more time digging deeper into the code and writing some tests so that I can better understand. Thanks!

mbrandonw:
Hi @kabiroberai, I spent some time yesterday playing with the feature and I now better understand it from beginning to end. I think we can reduce some of the surface area (e.g. I don't think the mapper stuff is necessary), but that's just small details.

However, I did find a sharp edge while writing more tests. It turns out if you `eraseToEffect` in a _parent_ feature then you can accidentally mess up your ability to do `send(…).finish()` in the child effect.

Here's a test that shows it:

```swift
func testRunFinish_ParentErasesToEffect() async {
  struct Child: ReducerProtocol {
    struct State: Equatable {}
    enum Action: Equatable {
      case tap
      case response
      case responseEnd
      case tapEnd
    }
      switch action {
      case .tap:
        return .run { send in
          await send(.response).finish()
          await send(.tapEnd)
        }
      case .response:
        return .run { send in
          try await self.mainQueue.sleep(for: 1)
          await send(.responseEnd)
        }
      case .responseEnd, .tapEnd:
        return .none
      }
    }
  }

  struct Parent: ReducerProtocol {
    struct State: Equatable {
      var child = Child.State()
    }
    enum Action: Equatable {
      case child(Child.Action)
    }
      Scope(state: \.child, action: /Action.child) {
        Child()
      }
      Reduce<State, Action> { state, action in
        switch action {
        case .child(.tap):
          return .none.eraseToEffect()
        case .child:
          return .none
        }
      }
    }
  }

  let mainQueue = DispatchQueue.test
  let store = TestStore(initialState: Parent.State(), reducer: Parent()) {
    $0.mainQueue = mainQueue.eraseToAnyScheduler()
  }

  XCTExpectFailure {
    $0.compactDescription == """
      A publisher-style Effect called 'EffectSendTask.finish()'. This method no-ops if you apply \
      any Combine operators to the Effect returned by 'EffectTask.run'.
      """
  }

  await store.send(.child(.tap))
  await store.receive(.child(.response))
  await store.receive(.child(.tapEnd))
  await mainQueue.advance(by: 1)
  await store.receive(.child(.responseEnd))
}
```

It's a long test, but the important bits are that the child returns an effect that does `finish()` on sending another action:

```swift
      case .tap:
        return .run { send in
          await send(.response).finish()
          await send(.tapEnd)
        }
```

…but the parent returns a publisher effect for that same action:

```swift
        case .child(.tap):
          return .none.eraseToEffect()
```

That's a pretty big gotcha unfortunately. It means if you get the warning you need to trace up the reducer hierarchy to find where you might be returning a publisher effect, and the only way to fix is to convert them all to async/await.

I briefly looked into what it would take to support `finish()` on publisher effects, but as you said, it seems quite difficult, if not impossible.

This new gotcha does give me pause on the feature. It may be more appropriate for the 1.0 or maybe even until we fully decouple Combine from the libray. What are your thoughts?

mbrandonw:
One small update after discussing with @stephencelis. We do think this has the potential to make it into 1.0 if we decide to go as far as fully hard-deprecating the `Publisher` conformance in `Effect`. This would force anyone who wants to use publisher operators in effects to do it inside a `.run`, such as:

```swift
return .run { send in 
  for await value in self.client.updates().values {
    await send(.response(value))
  }
}
```

If we decide to go that far for 1.0 then this change works out perfectly.

kabiroberai:
I hadn't considered the case where .merge/ReducerBuilder breaks things, that definitely feels surprising. If you folks do remove the Combine interface of Effect in 1.0 then it's definitely worth including this PR imo.

Otherwise, what if we make sure that the only way to get a Combine-style effect is by using a hard-deprecated API? We could switch all explicit Effect operators (concatenate, merge, etc) to returning `.run`-style effects all the time, so that something ostensibly simple like your example doesn't end up creating a publisher.

kabiroberai:
A bit tangential but thought I'd address this point

> I think we can reduce some of the surface area (e.g. I don't think the mapper stuff is necessary)

I agree that we don't need this on `Send`, but if/when we remove the Combine interface consider that it won't be possible for API consumers to write code that "wraps" Send calls. For example, something like `Effect.animation` would be impossible to implement externally, whereas with Combine one merely needed to write a scheduler or publisher that wraps the action/`downstream.receive` in `withAnimation`. Fortunately we could simplify the mapper API a lot if we removed the Combine interface since we wouldn't need the invariant return type, and could also move the API up to the Effect level, so we could have something like

```swift
func Effect.mapSend(_ transform: (Send<Action>) -> Send<NewAction>) -> Effect<NewAction>
```

analogous to (and in fact a superset of) the existing `Effect.map`. Something something multi-layered burritos.

Just wanted to share my concerns around this, but I agree that we can discuss this further in a follow-up PR rather than the current one.

mbrandonw:
> Just wanted to share my concerns around this, but I agree that we can discuss this further in a follow-up PR rather than the current one.

Thanks for detailing everything! It's definitely a complex issue to work through.

I may be misunderstanding something, but last week I rebuilt the feature from scratch just to see what the minimal version of this would look like and to find the sharp edges. [This](https://github.com/pointfreeco/swift-composable-architecture/compare/effect-send-task) is what I came up with, and it doesn't use any send mapping. It seems to work and I didn't have to make any changes to the core effect type.

Is there something I'm missing?

kabiroberai:
The main source of complexity in my impl's `Send` was that I wanted to preserve the runtime warning if you called `Send` inside a cancelled effect. Consider this function from your branch's `Send`: https://github.com/pointfreeco/swift-composable-architecture/blob/0d7e9c97dd7931c803c0d007e20782a4fd47a9a8/Sources/ComposableArchitecture/Effect.swift#L393-L400

And consider the following example:
```swift
let effect = Effect.run { send in
  await send(.action).finish()
}
cancellable = effect.sink { _ in ... }
```

If you call `cancellable.cancel()` then the effect won't emit a runtime warning on the `send()` line. IMO it _should_ emit a warning because if the effect _hadn't_ been cancelled then calling `send` would've been UB. In order to achieve this, we need to know inside `Send` that the Effect is "detached" from a store, so that we can return `EffectSendTask(.none)` instead of `EffectSendTask(.some(.none))`. This consequently means we'd need to store the attached/detached status inside `Send` itself, at which point you end up with
```swift
enum Send {
  case attached((Action) -> StoreTask)
  case detached((Action) -> Void)
}
```
hence the `SendMapper` ceremony.

This is only necessary if you think that calling `send().finish()` inside a Publisher-based effect should warn even if the effect is cancelled. Maybe it isn't that big a deal since we'd ideally be deprecating the Publisher interface anyway, in which case your approach works identically and has a simpler API.

kabiroberai:
Also, I do want to add that an `Effect.mapSend` would be helpful regardless of whether we need `SendMapper`, since atm TCA's first-party code has privileged access to `Effect.Operation` and can modify the `Send` passed to a `.run(_, let operation)`, like we see in the implementation of `Effect.animation`. This wouldn't be possible in third-party code but it would be nice if it was.

stephencelis:
@kabiroberai After chatting with @mbrandonw I think we're going to close this out for now. It's definitely a feature we want, but we think there are probably some fundamental changes and questions that need to be answered about the future of effect management in TCA. We have some promising explorations that should get us closer to implementing this functionality in the future, but the current state of the library makes it tough to merge this as is given the caveats.

We'll keep the feature in the mind during future iterations of the library.

By the way we recently explored a PR that just adds warnings to calls to `send` in cancelled effects, but we found that it was too loud, and emitted unwanted warnings in a common case: https://github.com/pointfreeco/swift-composable-architecture/pull/2219

kabiroberai:
Sounds good, this PR definitely altered the mental model of an Effect a bit and I can imagine that may not fit the bigger picture. Eager to see what the future holds for alternative approaches 🙌

PR #1903, opened 2023-02-13
Add optional binding

wogus3602:
This PR adds optional binding.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Binding.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Binding.swift
@@ -153,6 +153,39 @@ extension ViewStore where ViewAction: BindableAction, ViewAction.State == ViewSt
       }
     )
   }
+    
+    /// Returns a binding to the resulting bindable state of a given key path.
+    ///
+    /// - Parameters:
+    ///   -  keyPath: A key path to a specific bindable state.
+    ///   - defaultValue: A default value.
+    /// - Returns: A new binding.
+    public func optionalBinding<Value: Equatable>(
+      _ keyPath: WritableKeyPath<ViewState, BindingState<Value?>>,
+      defaultValue: Value,
+      file: StaticString = #file,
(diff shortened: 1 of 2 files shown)


mbrandonw:
Hi @wogus3602, can you explain a bit more why you need this tool?

This doesn't seem to be something that needs to live in the TCA library. Instead, you can cook up your own helper for turning a `Binding<Value?>` into a `Binding<Value>`, roughly like so:

```swift
extension Binding {
  func coalesce<Wrapped>(defaultValue: Wrapped) -> Binding<Wrapped> where Value == Wrapped? {
    // ...
  }
}
```

And this helper can live outside TCA.

stephencelis:
@wogus3602 Thanks for taking the time to PR! Can you share more about what you're trying to solve with this patch, maybe with an example of how you're using it in your code base?

We try to be very careful when adding to TCA's surface area, especially if the functionality can be added at another layer. In this case, it looks like you're providing a transform from `Binding<Wrapped?>` to `Binding<Wrapped>` via the view store, but I think that can even be achieved using built-in `Binding` tools that come with SwiftUI, or with tools that ship with [SwiftUI Navigation](https://github.com/pointfreeco/swiftui-navigation).

For example:

```swift
// This `Binding.init` can be buggy, so can instead use `Binding.init(unwrapped:)`
// from SwiftUINavigation.

if let state = Binding(viewStore.binding(\.$optionalState)) {
  TextField("State", text: state)
}
```

Let us know what you think!

wogus3602:
@stephencelis @mbrandonw 
In the example below, I am trying to use binding. However, an error also occurs with optionalBinding.
I will close this PR.

```swift
struct State: Equatable {
          @BindingState var root = Root()

          struct Root: Equatable {
              var nested: Nested? = nil
          }
            
          struct Nested: Equatable {
              var field: String = ""
          }
}

ViewStore.binding(\.$root.nested?.field)
```

PR #1910, opened 2023-02-15
Force WithViewStore to be a view.

mbrandonw:
The various initializers on `WithViewStore` for dealing with scenes, toolbars, etc. have been deprecated for over 5 months now. We did that because they can cause strain on the compiler, and we are seeing that more and more, so we think it's finally time to delete them.

And once that is done we can now make `WithViewStore` always be a `View` rather than conditionally. We have found this helps type checking and compiling performance quite a bit, especially for previews where I guess due to the dynamic replacement mechanism type checking can be even slower.


stephencelis (approved):

tgrapperon:
Indeed, this is a huge improvement. I remember #1204 where locking `Content` to a `View` in a fork allowed a failing build to succeed.

DandyLyons:
What if we had a `WithSceneStore` that always returns a `Scene`?

(I didn't know WithViewStore used to work with Scenes until after I wrote this proposal.)

https://github.com/pointfreeco/swift-composable-architecture/discussions/2170

PR #1911, opened 2023-02-15
Nest `Send<Action>` as `Effect<Action>.Send`

tgrapperon:
Right now, `Send<Action>` being a top-level type makes it the preferred autocompletion choice when you type "Send…" to make some type `Sendable` (a very frequent operation). This is suboptimal.

Fortunately, this value is always provided by the library in standard operations, and the user usually doesn't have to handle its type directly. This PR relocates `Send<Action>` in `EffectTask<Action>.Send` (that is `EffectPublisher<Action, Never>.Send`, but it should transparently migrate to `Effect<Action>.Send` for the `v1.0`.). The current top-level `Send` is deprecated (which lowers its autocompletion rank and gives way to `Sendable`), but this shouldn't impact existing user code which should seamlessly resolve to `Effect<Action>.Send` instead of `Send<Action>` in the vast majority of cases.

It doesn't seem possible to provide a `rename:` argument to the deprecation annotation.

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -57,7 +57,7 @@ public struct EffectPublisher<Action, Failure: Error> {
   enum Operation {
     case none
     case publisher(AnyPublisher<Action, Failure>)
-    case run(TaskPriority? = nil, @Sendable (Send<Action>) async -> Void)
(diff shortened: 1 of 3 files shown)


stephencelis (approved):
Good idea, thanks!

PR #1916, opened 2023-02-17
Allow scoping store's action while keeping state intact

acosmicflamingo:
I noticed that there are convenience initializers and functions that allow for scoping without having to explicitly pass an action, but the same can't be said for state.

l don't know if this is intentional or not, but here's a PR for the functionality just in case.

If you're also wondering when this would ever be useful, here's a snippet of code from my FontManagerCell's `update` function where Reducer.Action utilizes [action boundaries](https://github.com/pointfreeco/swift-composable-architecture/discussions/1440):
```swift
public func update(
  id: Tagged<RadioRowItem, UUID>,
  with store: Store<RadioSection, FontManager.Action>
) {
  let viewStore = ViewStore(store, send: Reducer.Action.view)
  self.store = store
  self.viewStore = viewStore
  self.id = id

  setupBindings(on: viewStore)
}
```

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -326,6 +326,18 @@ public final class Store<State, Action> {
     self.scope(state: toChildState, action: { $0 })
   }
 
+  /// Scopes the store to one that exposes child action.
+  ///
+  /// A version of ``scope(state:action:)`` that leaves the state type unchanged.
+  ///
+  /// - Parameter fromChildAction: A function that transforms `ChildAction` into `Action`.
+  /// - Returns: A new store with its domain (state and action) transformed.
+  public func scope<ChildAction>(
+    action fromChildAction: @escaping (ChildAction) -> Action
+  ) -> Store<State, ChildAction> {
+    self.scope(state: { $0 }, action: fromChildAction)
+  }
+
   func filter(
     _ isSent: @escaping (State, Action) -> Bool
   ) -> Store<State, Action> {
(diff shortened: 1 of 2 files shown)


stephencelis:
Hey @acosmicflamingo! Thanks for the PR!

We'd love to be able to support this, and were hoping some recent features of Swift would make it easier without sacrificing performance, like [SE-0347](https://github.com/apple/swift-evolution/blob/main/proposals/0347-type-inference-from-default-exprs.md), but unfortunately it doesn't look like this feature is up to the task yet. In particular this bug seems to be blocking us: https://github.com/apple/swift/issues/63749

Our current worry is that these overloads are pretty expensive for the compiler to resolve, so the ones that we do include are the more common case: transforming state is the general use case for `scope` outside of transforming both at once, since chipping away at observed state is more important in SwiftUI than hiding some of the action domain. Even this overload may cause type checking problems, though.

I think we'll happily revisit this when the above Swift bug is resolved, but even then we're not quite sure if these default expressions are any less taxing on the compiler...

acosmicflamingo:
@stephencelis oh okay! Thanks for the explanation :) I can close this then for now if you'd like.

stephencelis:
No problem, I can close for now, and hopefully some day we can revisit in the future!

tgrapperon:
As I was reading this PR, I wondered if the `scope(state:)` variant still made sense now that `ViewStore` is exposing an `observe:` argument. I feel that creating a `ViewStore` represented the vast majority of uses of this variant. Maybe it should be deprecated to nudge folks toward `observe:`?

PR #1917, opened 2023-02-17
Re-enable `_printChanges()` for previews

tgrapperon:
For optimization reasons, `Reducer._printChanges()`'s isn't generating output for Xcode previews (#1625).
The new beta version of Xcode 14.3 now displays `Swift.print(…)` messages in the console under the "previews" tab, so we should probably revert the changes from #1625 and re-enable `_printChanges()`'s output in `preview` context.

I don't think there is a way to detect when previews are reloaded, so the output of successive reloads concatenates. It's better than nothing I guess, but this can maybe be improved in a subsequent PR.


stephencelis (approved):
Nice!

PR #1939, opened 2023-02-24
Make the new XCTModify play nicely with non-exhaustive testing.

mbrandonw:
We recently added `XCTModify` to our case paths library that makes it easy to apply an in-place mutation to a case of an enum. We employ exhaustive checking in the function so that if you apply a mutation that doesn't actually change anything it will cause a test failure. However, we need to opt out of that behavior in non-exhaustive test stores, so that is what this PR does.

### DIFF
--- a/Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift
+++ b/Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift
@@ -684,6 +684,71 @@
       await store.receive(/NonExhaustiveReceive.Action.response2)
     }
 
+    func testXCTModifyExhaustive() async {
+      struct State: Equatable {
+        var child: Int? = 0
+        var count = 0
+      }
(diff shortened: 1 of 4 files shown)


Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift
```diff
@@ -684,6 +684,71 @@
...
+            state.count += 1
+            return .none
+          }
+        }
+      )
+
+      await store.send(.tap) { state in
+        state.count = 1
+        XCTExpectFailure {
+          XCTModify(&state.child, case: /Optional.some) { _ in }
```
stephencelis, line 710: Minor take-it-or-leave-it, but we have a shorthand syntax for optionals:

```suggestion
          XCTModify(&state.child, case: /.some) { _ in }
```

stephencelis (approved):
Nice!

PR #1940, opened 2023-02-25
Adds missing documentation for the `Store` initializer

kristofferjohansson:
Noticed this was missing. I tried to somewhat match the documentation for `TestStore`.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -141,6 +141,7 @@ public final class Store<State, Action> {
   /// - Parameters:
   ///   - initialState: The state to start the application in.
   ///   - reducer: The reducer that powers the business logic of the application.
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -141,6 +141,7 @@ public final class Store<State, Action> {
   /// - Parameters:
   ///   - initialState: The state to start the application in.
   ///   - reducer: The reducer that powers the business logic of the application.
+  ///   - prepareDependencies: A closure that can be used to override dependencies that will be accessed by the reducer.
```
mbrandonw, line 144:
```suggestion
  ///   - prepareDependencies: A closure that can be used to override dependencies that will be accessed 
          by the reducer.
```

mbrandonw (approved):
Good catch thanks!

PR #1950, opened 2023-02-28
Add @warn_unqualified_access to reducer operators.

mbrandonw:
When using reducer operators (e.g. `.ifLet`, `.forEach`, etc.) it's possible to accidentally forget the `.`:

```swift
  Reduce { … }
    ifLet { … }
   ^ 
```

This will cause a mysterious infinite loop and crash. The same happens in SwiftUI, and luckily Swift has an attribute that will warn if you ever use the method without qualifying the receiver, so let's use it!


stephencelis (approved):
Nice! 😅

davdroman:
> Apple: "there's an app for that"
> Swift: "there's an attribute for that"

PR #1951, opened 2023-02-28
Deprecations only for versions supporting clock.timer

Kondamon:
Show deprecation warnings only for versions [that support](https://github.com/pointfreeco/swift-dependencies/blob/d3406a54315daccc2c86a21da04584b1fefa7136/Sources/Dependencies/DependencyValues/Clocks.swift#L2) `@Dependency(\.continuousClock) var clock`.


mbrandonw (approved):
Good catch, thanks!

stephencelis:
Just a note that we reverted this. Didn't occur to us that this was on the prerelease branch, where we want to deprecate APIs that too strongly tie `Effect` to the Combine framework. We'd like to provide a friendlier API for bridging TCA to Combine soon, though!

iampatbrown:
@Kondamon I've been using `mainQueue.timer` instead, with he intention of replacing it with `clock.timer` in the future.
```swift
.run { send in
  for await _ in mainQueue.timer(interval: .seconds(1)) {
    await send(.tick)
  }
}
```

PR #1948, opened 2023-02-28
Allow child dismissal from re-entrant effects

stephencelis:
Fixes #1947.

PR #1954, opened 2023-03-01
Do not propagate dependencies to effect actions.

### LINKED ISSUE #1947: Calling dismiss dependency fails when executed from a secondary effect (navigation beta)
### Description

In the following example, the `dismiss` dependency fails to dismiss the sheet when it is called from an effect that has been sent back into the store (for example, after saving state changes to disk).

```swift
import SwiftUI
import ComposableArchitecture

struct AppReducer: ReducerProtocol {
    
    struct State: Equatable {
        @PresentationState var sheet: MySheet.State?
    }
 
    enum Action: Equatable {
        case sheet(PresentationAction<MySheet.Action>)
        case sheetButtonTapped
    }
    
        Reduce { state, action in
            switch action {
            case .sheetButtonTapped:
                state.sheet = .init()
                return .none
                
            case .sheet:
                return .none
            }
        }
        .ifLet(\.$sheet, action: /Action.sheet) {
            MySheet()
        }
    }
}

struct MySheet: ReducerProtocol {
    
    struct State: Equatable {
    }
    
    enum Action: Equatable {
        case cancelTapped
        case saveTapped
        case saveCompleted
    }
    
    @Dependency(\.dismiss) var dismiss
    
        Reduce { state, action in
            switch (action) {
            case .cancelTapped:
                return .fireAndForget { await self.dismiss() }
        
            case .saveCompleted:
                // This has no effect
                return .fireAndForget { await self.dismiss() }
                
            case .saveTapped:
                // Simulate saving data
                return .task {
                    try await Task.sleep(nanoseconds: 100 * NSEC_PER_MSEC)
                    return .saveCompleted
                }
            }
        }
    }
    
}

struct ContentView: View {
    let store: StoreOf<AppReducer>
    
    var body: some View {
        WithViewStore(store, observe: { $0} ) { viewStore in
            Button("Open sheet") { viewStore.send(.sheetButtonTapped) }
                .sheet(store: store.scope(state: \.$sheet, action: AppReducer.Action.sheet)) { store in
                    SheetView(store: store)
                }
        }
    }
}

struct SheetView: View {
    let store: StoreOf<MySheet>
    
    var body: some View {
        WithViewStore(store, observe: { $0} ) { viewStore in
            NavigationStack {
                Text("A Sheet")
                .toolbar {
                    ToolbarItem(placement: .cancellationAction) {
                        Button("Cancel") {
                            viewStore.send(.cancelTapped)
                        }
                    }
                    ToolbarItem(placement: .primaryAction) {
                        Button("Save") {
                            viewStore.send(.saveTapped)
                        }
                    }
                }
            }
            
        }
        
    }
}

struct ContentView_Previews: PreviewProvider {
        ContentView(store: .init(initialState: .init(), reducer: AppReducer()))
    }
}
```

### Checklist

### Expected behavior

The sheet should be dismissed via an asynchronous effect.

### Actual behavior

The sheet is not dismissed via an asynchronous effect.

### Steps to reproduce

Copy the sample code into Xcode and run the preview.

navigation-beta (45b88b6f1b08c10cbd6141d9794828cb0a3f012a)

iOS 16.2

### Xcode version information

14.2

_No response_

mbrandonw:
The issue #1947 for something related to the navigation beta (#1944) has surfaced a small problem in the library that exists on `main` today.

Right now effect actions (i.e. actions sent into the system from effects) are executed with the dependencies that were active at the moment the effect was constructed. This allows some surprising behavior in which dependency changes "accumulate" when you have an effect that emits an action, and that action returns an effect that emits an action, and on and on.

Take this for example:

```swift
switch action {
case .tap:
  return withDependencies {
    $0.count.value += 1
  } operation: {
    .task { .response1(self.count.value) }
  }
case let .response1(count):
  state.count = count
  return withDependencies {
    $0.count.value += 1
  } operation: {
    .task { .response2(self.count.value) }
  }
case let .response2(count):
  state.count = count
  return withDependencies {
    $0.count.value += 1
  } operation: {
    .task { .response3(self.count.value) }
  }
case let .response3(count):
  state.count = count
  return .none
}
```

By the last `.response3` the count has become 3 because each step of the way the dependencies were carried over from the previous step.

An added benefit to this change is that it fixes a discrepancy between `Store` and `TestStore`. We were able to write a test that passed for `TestStore` but that did not actually capture what happens in reality for `Store`, which of course completely destroys the confidence one should have when writing tests against the `TestStore`. That was a huge bummer, and is now fixed.

Now, there may be some power in allowing this, but we don't think it should be the default. If it turns out there is use for this, we could (possibly) support a parameter on `Effect.run` that propagates dependencies more deeply, and then have to figure out how to integrate it with the `TestStore`.

This will fix #1947 and make #1948 unnecessary.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -402,7 +402,9 @@ public final class Store<State, Action> {
               },
               receiveValue: { [weak self] effectAction in
                 guard let self = self else { return }
-                if let task = continuation.yield({ self.send(effectAction, originatingFrom: action) }) {
+                if let task = continuation.yield({
+                  self.send(effectAction, originatingFrom: action)
+                }) {
                   tasks.wrappedValue.append(task)
                 }
               }
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -385,28 +385,29 @@ public final class Store<State, Action> {
         var didComplete = false
         let boxedTask = Box<Task<Void, Never>?>(wrappedValue: nil)
         let uuid = UUID()
-        let effectCancellable =
+        let effectCancellable = withEscapedDependencies { continuation in
           publisher
-          .handleEvents(
-            receiveCancel: { [weak self] in
-              self?.threadCheck(status: .effectCompletion(action))
-              self?.effectCancellables[uuid] = nil
-            }
-          )
-          .sink(
-            receiveCompletion: { [weak self] _ in
-              self?.threadCheck(status: .effectCompletion(action))
-              boxedTask.wrappedValue?.cancel()
-              didComplete = true
-              self?.effectCancellables[uuid] = nil
-            },
-            receiveValue: { [weak self] effectAction in
-              guard let self = self else { return }
-              if let task = self.send(effectAction, originatingFrom: action) {
-                tasks.wrappedValue.append(task)
+            .handleEvents(
+              receiveCancel: { [weak self] in
+                self?.threadCheck(status: .effectCompletion(action))
+                self?.effectCancellables[uuid] = nil
               }
-            }
-          )
+            )
+            .sink(
+              receiveCompletion: { [weak self] _ in
+                self?.threadCheck(status: .effectCompletion(action))
+                boxedTask.wrappedValue?.cancel()
+                didComplete = true
+                self?.effectCancellables[uuid] = nil
+              },
+              receiveValue: { [weak self] effectAction in
+                guard let self = self else { return }
+                if let task = continuation.yield({ self.send(effectAction, originatingFrom: action) }) {
+                  tasks.wrappedValue.append(task)
+                }
+              }
+            )
+        }
 
         if !didComplete {
           let task = Task<Void, Never> { @MainActor in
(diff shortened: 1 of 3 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -131,7 +131,7 @@ public final class Store<State, Action> {
     private let reducer: (inout State, Action) -> EffectTask<Action>
     fileprivate var scope: AnyStoreScope?
   #endif
-  var state: CurrentValueSubject<State, Never>
+  @_spi(Internals) public var state: CurrentValueSubject<State, Never>
```
mbrandonw, line 134: added this so that I could more fully test the `Store` in complete isolation (no need to interface with `ViewStore`). I just wanted to be 100% we were asserting on the full behavior inside the store without outside factors.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -385,28 +385,31 @@ public final class Store<State, Action> {
         var didComplete = false
         let boxedTask = Box<Task<Void, Never>?>(wrappedValue: nil)
         let uuid = UUID()
-        let effectCancellable =
+        let effectCancellable = withEscapedDependencies { continuation in
```
mbrandonw, line 388: easier to review this diff with whitespace turned off.

Tests/ComposableArchitectureTests/StoreTests.swift
```diff
@@ -559,4 +559,169 @@ final class StoreTests: XCTestCase {
...
+    self.reducers = reducers
+  }
+
+  @inlinable
+  public func reduce(
+    into state: inout Reducers.State, action: Reducers.Action
+  ) -> EffectTask<Reducers.Action> {
+    self.reducers(state, action).reduce(into: &state, action: action)
+  }
 }
```
tgrapperon: It may need a few adjustments to build on Swift 5.8.
```suggestion
public struct _Observe<State, Action, Reducers: ReducerProtocol>: ReducerProtocol
where Reducers.State == State, Reducers.Action == Action {
  @usableFromInline
  let reducers: (State, Action) -> Reducers

  /// Initializes a reducer that builds a reducer from the current state and action.
  ///
  /// - Parameter build: A reducer builder that has access to the current state and action.
  @inlinable
  public init(
    @ReducerBuilder<State, Action> _ build: @escaping (State, Action) -> Reducers
  ) {
    self.init(internal: build)
  }

  @usableFromInline
  init(
    internal reducers: @escaping (State, Action) -> Reducers
  ) {
    self.reducers = reducers
  }

  @inlinable
  public func reduce(
    into state: inout Reducers.State, action: Reducers.Action
  ) -> EffectTask<Reducers.Action> {
    self.reducers(state, action).reduce(into: &state, action: action)
  }
}
```

mbrandonw: ha sorry, that was accidentally committed from messing with due to that slack conversation. we are not shipping that type in this PR.

stephencelis (approved):

PR #1955, opened 2023-03-03
Fix TestStore.init with prepareDependencies.

mbrandonw:
This was reported by @oronbz in Slack. We call `prepareDependencies` immediately in the `TestStore` initializer which means accessing a dependency in the trailing closure can trigger a test failure. We need to move it into `withDependencies` so that such failures are suppressed.

### DIFF
--- a/Tests/ComposableArchitectureTests/TestStoreTests.swift
+++ b/Tests/ComposableArchitectureTests/TestStoreTests.swift
@@ -335,7 +335,7 @@ final class TestStoreTests: XCTestCase {
 
       func reduce(into state: inout Int, action: Bool) -> EffectTask<Bool> {
         _ = self.calendar
-        _ = self.client.fetch()
+        _ = self.fetch()
         _ = self.locale
         _ = self.timeZone
         _ = self.urlSession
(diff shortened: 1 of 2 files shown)


Tests/ComposableArchitectureTests/TestStoreTests.swift
```diff
@@ -328,12 +328,14 @@ final class TestStoreTests: XCTestCase {
   func testOverrideDependenciesOnTestStore_Init() {
     struct Counter: ReducerProtocol {
       @Dependency(\.calendar) var calendar
+      @Dependency(\.client) var client
```
oronbz, line 331:
```suggestion
      @Dependency(\.client.fetch) var fetch
```

Tests/ComposableArchitectureTests/TestStoreTests.swift
```diff
@@ -328,12 +328,14 @@ final class TestStoreTests: XCTestCase {
...
       @Dependency(\.locale) var locale
       @Dependency(\.timeZone) var timeZone
       @Dependency(\.urlSession) var urlSession
 
       func reduce(into state: inout Int, action: Bool) -> EffectTask<Bool> {
         _ = self.calendar
+        _ = self.client.fetch()
```
oronbz, line 338:
```suggestion
        _ = self.fetch()
```

oronbz (commented):
I believe this should better reflect the situation

mbrandonw:
> I believe this should better reflect the situation

Good catch, done!

stephencelis (approved):

PR #1957, opened 2023-03-06
Fix types, .forEach() call

finestructure:
I just went through the `ForEachStore` doc examples and noticed a couple of small inconsistencies that prevent them from compiling as is. Hope that helps!


mbrandonw (approved):
Great finds, thanks!

PR #1959, opened 2023-03-06
Remove ConcurrencySupport.swift file

darrarski:
## Description

Hello!

https://hachyderm.io/@pointfreeco/109978323469618624

This PR deletes the file with code duplicated from `swift-dependencies`. After deleting this file the test tests are passing. This should solve an issue with "ambiguous use" errors in a file that imports both `ComposableArchitecture` and `Dependencies` libraries.

## What was done

- [x] Removed `Sources/ComposableArchitecture/Effects/ConcurrencySupport.swift`


stephencelis (approved):
Huh! I'm not even sure how this file is on the prerelease branch but not `main`, when prerelease was spun off of main. Maybe some kind of wild merge issue...

PR #1962, opened 2023-03-07
Added reminder to use the .fetch() in the reduce() when using NumberF…

siliconsorcery:
Small update for the documentation. This should help beginners avoid the confusion I found myself in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -492,15 +492,18 @@ extension DependencyValues {
 ```
 
 With that little bit of upfront work done you can instantly start making use of the dependency in 
-any feature:
+any feature by using the `@Dependency` property wrapper:
 
-```swift
-struct Feature: ReducerProtocol {
-  struct State { … }
-  enum Action { … }
-  @Dependency(\.numberFact) var numberFact
-  …
-}
+```diff
+ struct Feature: ReducerProtocol {
+-  let numberFact: (Int) async throws -> String
++  @Dependency(\.numberFact) var numberFact
+   
+   …
+
+-  try await self.numberFact(count)
++  try await self.numberFact.fetch(count)
+ }
 ```
 
 This code works exactly as it did before, but you no longer have to explicitly pass the dependency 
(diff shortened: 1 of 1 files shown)


stephencelis:
@siliconsorcery Thanks for taking the time to PR! I think that code is already taken care of here, though:

https://github.com/pointfreeco/swift-composable-architecture/blob/a658e25bee0eb6d463c14dfce370d564f6dfc3b3/README.md?plain=1#L385-L399

Luckily, switching from holding onto a dependency explicitly as a property vs. implicitly through a property wrapper means the call site in `reduce` doesn't need to change!

siliconsorcery:
@stephencelis Thanks for your response.

Unless I missed a step, we are using a variable to hold our function to `fetch` a number fact. The default definition is set in:

```swift
extension NumberFactClient: DependencyKey {
  static let liveValue = Self(
    fetch: ..
 )
}
```

It can be overridden for testing

```swift
let store = TestStore(
  reducer: Feature()
) {
  $0.numberFact.fetch = { "\($0) is a good number Brent" }
}
```

Note we are referencing the `fetch` function we are replacing.

Now for usage we need to call the `fetch` function.

```swift
        case .numberFactButtonTapped:
            return .task { [count = state.count] in
                await .numberFactResponse(
                    TaskResult {
                        try await self.numberFact.fetch(count)
                    }
                )
            }
```

If I omit the `fetch` function

```swift
  try await self.numberFact(count)
```
Xcode throws an error: `Cannot call value of non-function type 'NumberFactClient'`

This had me very confused until I studied the code.

Again, unless I missed something, I think the correct function access for all uses (production, development and testing) in this case is to call the `fetch` function.

```swift
  try await self.numberFact.fetch(count)
```

Thanks four your great work.

stephencelis:
@siliconsorcery Rather than add another step and more prose, what do you think of folding the required change into this code block:

https://github.com/pointfreeco/swift-composable-architecture/blob/d73a3494bfadfec7ccbadf2f295ff4a1156153ec/README.md?plain=1#L497-L504

Basically change it to this diff instead?

```diff
 struct Feature: ReducerProtocol {
   …
-  let numberFact: (Int) async throws -> String
+  @Dependency(\.numberFact) var numberFact
   …
-  try await self.numberFact(count)
+  try await self.numberFact.fetch(count)
   …
 }
```

mbrandonw:
Hi @siliconsorcery, I have made the changes that @stephencelis mentioned and think it reads better now. Will merge this and hopefully it helps people in the future.

Thanks for bringing this up!

PR #1963, opened 2023-03-07
Fix grammar in doc

bjford:
Hi there, 

While reading the documentation for `WithViewStore` I've found one small grammar error.


stephencelis (approved):
Thanks!

PR #1970, opened 2023-03-10
Get some test coverage on child/parent effect cancellation behavior.

mbrandonw:
This discussion #1717 brings up some interesting behavior with child task cancellation that I think we want to address eventually. For now I'm getting a test in place that documents the behavior so that we can decide how to handle later.

### DIFF
--- a/Tests/ComposableArchitectureTests/StoreTests.swift
+++ b/Tests/ComposableArchitectureTests/StoreTests.swift
@@ -683,77 +683,79 @@ final class StoreTests: XCTestCase {
     XCTAssertEqual(store.state.value.count, testStore.state.count)
   }
 
-  func testChidlParentEffectCancellation() async throws {
-    struct Child: ReducerProtocol {
-      struct State: Equatable {}
-      enum Action: Equatable {
-        case task
-        case didFinish
-      }
-
-      func reduce(into state: inout State, action: Action) -> EffectTask<Action> {
-        switch action {
-        case .task:
-          return .run { send in await send(.didFinish) }
-        case .didFinish:
-          return .none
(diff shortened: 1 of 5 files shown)


Tests/ComposableArchitectureTests/StoreTests.swift
```diff
@@ -682,6 +682,78 @@ final class StoreTests: XCTestCase {
...
+    try await Task.sleep(nanoseconds: 100_000_000)
+    XCTAssertEqual(viewStore.child, nil)
+
+    await childTask.cancel()
+    await mainQueue.advance(by: 1)
+    try await Task.sleep(nanoseconds: 100_000_000)
+    XCTExpectFailure("""
+      This fails because cancelling a child task will cancel all parent effects too. If we fix
+      this we can remove the XCTExpectFailure.
+      """)
```
stephencelis, line 754: Wonder if we should define a helper or use some kind of pattern in the error message to make these tests we want to fix searchable? Like prefixing the message with "TODO"?

Tests/ComposableArchitectureTests/TestHelpers.swift
```diff
@@ -0,0 +1,10 @@
+import XCTest
+
+@available(
+  *,
+  deprecated,
+  message: "This is a test that currently fails but should not in the future."
+)
+func XCTodo(_ message: String) {
```
stephencelis, line 8: Miiinot nit, but what do you think of separating the "XCT" from "Todo" and capitalizing "TODO" for case-sensitive searches:

```suggestion
func XCTTODO(_ message: String) {
```

stephencelis (approved):

PR #1974, opened 2023-03-13
Add ifNil to ifLetUIKit

acosmicflamingo:
There may be cases when we simply want to take action whenever a state is nil (e.g. observing a Destination enum with multiple cases, where ifLet/else behavior breaks down). Much cleaner to just create an overload operation that handles this instance.

### DIFF
--- a/Sources/ComposableArchitecture/UIKit/IfLetUIKit.swift
+++ b/Sources/ComposableArchitecture/UIKit/IfLetUIKit.swift
@@ -64,4 +64,63 @@ extension Store {
         }
       }
   }
+
+  /// Calls closure when a store's optional state is `nil`, and does so
+  /// for as long as the cancellable lives.
+  ///
+  /// This method is useful for handling navigation in UIKit, but is to be used in cases when
+  /// the state is of an optional enum (i.e. Destination enum) with multiple cases. The ifLet/else
+  /// functionality is useful for when a view controller cannot push more than one view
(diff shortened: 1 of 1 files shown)


stephencelis:
Hey @acosmicflamingo! Thanks for taking the time to share and PR! You're definitely shining a light on some gaps in the new navigation tools when it comes to UIKit.

I think we probably want to think about the problem more before expanding the current surface area, though. If `ifNil` works for your app we totally recommend including it and using it in your code base :smile: And others that can benefit from it should totally bring it into their projects.

A use case we'd like to support in any tools we ship in the library is going between destination cases doing the right thing. _E.g._ if `.destination` changes from `.sheet` to `.drillDown` without ever going `nil`, we should be able to support that. But also if you go from `.drillDownA` to `.drillDownB`, we should be able to use `setViewControllers` to animate the right thing.

We have experimented a bit with UIKit navigation and have some promising, if unpolished, tools to further work on. We don't want to keep UIKit users out in the cold when it comes to navigation in TCA, we just need a little more time :smile:

acosmicflamingo:
@stephencelis thanks!! Oh wow, that is a scenario I had not really considered...which means this PR isn't that useful on its own. Shall I close this?

I'm delighted to hear that UIKit users won't be left out in the dark 🙂 you guys take all the time you need hahaha!

stephencelis:
@acosmicflamingo Yeah I'll close it for now, but thanks again for sharing and opening! :smile:

PR #1995, opened 2023-03-22
Don't disable `_printChanges()` in tests

stephencelis:
It can be handy to have this logging during a test, so let's not go out of our way to suppress it.


jshier:
How do we otherwise disable this logging during tests?

stephencelis:
@jshier The printing suppression was kinda accidentally added in the protocol version of this method. The `AnyReducer.debug()` printed always, so the behavior is kind of a regression.

I'd say in general we suggest installing `_printChanges()` in the view layer, _e.g._ at the app entry point of your app (or any "preview" apps for specific features), or in Xcode previews now that they support logging, and not check in `_printChanges()` anywhere in your general reducer tree (though it might help to temporarily install the modifier during a debug session).

At the end of the day, though, `_printChanges()` is an amalgamation of public tools, so you could copy the reducer into your project and make any changes you like if you depend on this behavior. You could also define a custom printer that recovers this behavior.

mbrandonw (approved):

PR #2002, opened 2023-03-25
Add preferredStyle to UIAlertController convenience initializer

acosmicflamingo:
This allows users to present a UIAlertController with the actionSheet style


mbrandonw:
Hi @acosmicflamingo, thanks for the PR!

However, just below the `init` that takes an `AlertState` there is an initializer that takes `ConfirmationDialogState`. That's the `init` you can use to use an `actionSheet` style UIKit alert.

It may seem a little weird, but this discrepancy is due to the fact that SwiftUI has distinct names for alerts vs dialogs, but UIKit has a single term, `UIAlertController`, with a configuration option. We've decided to mostly coalesce behind the SwiftUI style and do our best to adapt to the UIKit style.

acosmicflamingo:
D'oh! Thank you for the explanation!

PR #2010, opened 2023-03-30
Fix alert not opening issue 

ohsg0226:
A minor bug occurs in which a new alert does not appear when the Incremented button is pressed in the alert.
I think it's a timing issue.
After the first alert is closed and the alertDismissed action is called, the second alert should pop up, but the second alert doesn't appear because the alertDismissed action is called after the Incremented alert is called.

I solved the problem by deleting the statement **state.alert = nil**, but I'm not sure if this is correct.
And There are also problems with the test code not passing when the problem is solved in this way.

Can I get some advice on how to solve this problem? 
Thanks.


mbrandonw:
Hi @ohsg0226, are you suggesting that something is wrong with the case study or are you simply asking a question? AFAICT the case study is just fine, so I will be closing this PR. If you have a question, then can you open a [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions) instead?

PR #2019, opened 2023-04-03
Fix Document in BindingAction

Ryu0118:
Hi there,
I found a mistake in the documentation for BindingAction in Binding.swift and have corrected it.


mbrandonw (approved):
Good catch, thanks!

PR #2022, opened 2023-04-04
Conform ExpressibleByNilLiteral by EffectPublisher

larryonoff:
This PR adds `ExpressibleByNilLiteral` conformance for `EffectPublisher`.

Looks that this change natively fits Swift syntax sugar.

PS. I'll update other examples if this change fits for you.


mbrandonw:
Hi @larryonoff, thanks for taking the time to make this PR, but unfortunately we do not think this is something to be merged. The `ExpressibleByNilLiteral` really isn't supposed to be conformed to, as the docs say:

> nil has a specific meaning in Swift—the absence of a value. Only the Optional type conforms to ExpressibleByNilLiteral. ExpressibleByNilLiteral conformance for types that use nil for other purposes is discouraged.

It seems that it's a bit of a relic of a past Swift, but doesn't really have a role in modern day Swift.

larryonoff:
@mbrandonw thanks for explanation!

PR #2043, opened 2023-04-14
Renamed action to onTap in NavigationLinkStore

Berhtulf:
Renamed the `action` parameter to mirror other inits and differentiate itself from `action fromDestinationAction`


mbrandonw (approved):
Good catch, thanks!

PR #2045, opened 2023-04-14
Creating new pullback method which has a different destination keypath

rkreutz:
Hey there,

I've recently opened a discussion [here](https://github.com/pointfreeco/swift-composable-architecture/discussions/2032) related to the topic.

The idea here is that I would prefer to not use the `\.view` key path in the reducer, like in `case Action.binding(.set(\.view.$isToggled, true))`, since I tend to separate the Reducer and the View (they are in different files) and have the `.view` property as a private extension on the View file. I do this because this way I can reuse the same Reducer for different views (like if you are A/B testing, using different screens for different class sizes like iPhone or iPad, or whatever other reason), and so my Reducer is not coupled to a specific `.view` property (or multiple in some cases).

Currently it is not possible to do this, I've tried mapping the the ViewAction into the DomainAction directly with something like:
```
case ViewAction.binding(BindingAction<ViewState>.set(\.$isToggled, true)):
    return .binding(BindingAction<State>.set(\.$isToggled, true)) // Notice I'm not using pullback(_:)
```
But I get a runtime warning doing so, saying that the ViewAction wasn't handled.

So the idea of this PR is to do the same logic as `.pullback(_:)`, however instead of forwarding the pulled back key path (`\ DomainState.view.$isToggled`) we can specify a destination key path directly like `\DomainState.$isToggled` and have the Reducer use that instead.

I wasn't sure about `destination` for the name of the additional key path, so suggestions are welcome.

Also I haven't added any documentation as of yet, just so I can gauge what are your thoughts on this.

These changes are based on `prerelease/1.0`, but I guess we could easily change it to `master` if that would be of interest.

Thanks

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Binding.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Binding.swift
@@ -386,6 +386,41 @@ extension BindingAction {
       valueIsEqualTo: self.valueIsEqualTo
     )
   }
+
+    public func pullback<NewRoot, Value>(
+        _ keyPath: WritableKeyPath<NewRoot, Root>,
+        destination: WritableKeyPath<NewRoot, BindingState<Value>>,
+        file: StaticString = #file,
+        fileID: StaticString = #fileID,
+        line: UInt = #line
+    ) -> BindingAction<NewRoot> {
+        .init(
+            keyPath: destination as PartialKeyPath<NewRoot>,
+            set: {
+                self.set(&$0[keyPath: keyPath])
+                if let value = self.value as? Value {
+                    $0[keyPath: destination].wrappedValue = value
+                } else {
+                    runtimeWarn(
+                      """
+                      Tried to assign a new destination which didn't have matching type to the \
+                      pulled back BindinAction at "\(fileID):\(line)".
+
+                        Original value type:    \(type(of: self.value))
+                        Destination value type: \(Value.self)
+
+                      To fix this, make sure the destination in pullback(_:destination:) has the \
+                      same type as the property that was updated.
+                      """,
+                      file: file,
+                      line: line
+                    )
+                }
+            },
+            value: self.value,
+            valueIsEqualTo: self.valueIsEqualTo
+        )
+    }
 }
 
 extension BindingAction: CustomDumpReflectable {


stephencelis:
@rkreutz Thanks for taking the time to propose and PR this. We definitely know `BindingState` and `ViewState` don't work nicely together and have plans to address some of lacking ergonomics.

The most promising experiment is from a few months ago on this branch: https://github.com/pointfreeco/swift-composable-architecture/compare/main...binding-store

It adds a few concepts, such as a `@BindingViewState` property wrapper and a `BindingStore`, but we think it provides simpler ergonomics than the existing `pullback` method and work that goes into it, which we'd honestly like to deprecate if possible.

Would this branch of ours solve the problem you're trying to solve here?

rkreutz:
Hey, thanks for the reply. 

Having a look at it, it seems the `ViewAction` got removed in the cases where the `ViewStateBinding` was used, so if someone chooses to keep the `ViewAction` they'd still have the same issue? I've checked it on Mobile so I might have missed something obvious there, will check properly again later.

stephencelis:
@rkreutz I believe you could still have view actions, but you wouldn't have any need for a `binding` case, since bindings can be derived directly from `@BindingViewState`.

rkreutz:
Ooh ok, I see now. Yeah that should solve the issue I had. 

Though, is there a reason to use `ViewBindingState` instead of `Binding` directly? Also seems like `ViewBindingStore` could be redundant.

stephencelis:
> Though, is there a reason to use `ViewBindingState` instead of `Binding` directly? Also seems like `ViewBindingStore` could be redundant.

`Binding` doesn't conditionally conform to `Equatable`, so we need to use our own type.

rkreutz:
> Binding doesn't conditionally conform to Equatable, so we need to use our own type.

Should be trivial to add that conformance to `Binding`, though I guess is better to not extend it like that in the lib and provide that custom `BindingViewState` so that consumers can opt to implement that conformance differently.

> How do you mean? We're totally down to simplify if you have ideas, but this is the simplest configuration we have come up with so far :)

So I had a think about it, I had originally thought that we could use directly `ViewStore<State, Action>` and not have to introduce a new type, taking `LoginView` as an example, you could have:
```swift
struct LoginView: View {
    struct ViewState: Equatable {
        init(store: ViewStoreOf<Login>) {
            self.alert = store.state.alert
            self._email = store.binding(\.$email) // This is using binding(_:) assuming that the consumer has conformed Binding to Equatable, but could as well introduce a new method in ViewStore to get a BindingViewState instead
            self.isActivityIndicatorVisible = store.state.isLoginRequestInFlight
            self.isFormDisabled = store.state.isLoginRequestInFlight
            self.isLoginButtonDisabled = !store.state.isFormValid
            self._password = store.binding(\.$password) // Same here
            self.isTwoFactorActive = store.state.twoFactor != nil
        }
    }
}
```
You would need to add new `WithViewStore` extensions as well to be able to initialise the ViewStore accordingly with something like this:
```swift
extension WithViewStore where ViewState: Equatable, Content: View {
    public init<State>(
        _ store: Store<State, ViewAction>,
        observe toViewState: @escaping (ViewStore<State, ViewAction>) -> ViewState,
        @ViewBuilder content: @escaping (ViewStore<ViewState, ViewAction>) -> Content,
        file: StaticString = #fileID,
        line: UInt = #line
    ) where State: Equatable {
        self.init(
            store,
            observe: { (_: State) in
                toViewState(ViewStore.init(store))
            },
            send: { $0 },
            removeDuplicates: ==,
            content: content,
            file: file,
            line: line
        )
    }
}
```
Could even keep `BindingViewStore` as a `typealias` of `ViewStore<ViewState, BindingAction<ViewState>>`.

HOWEVER, the one thing that is not great with this approach IMO is that since you have access to the `ViewStore` you have access to other methods in it (like `.send(_:)`), so in the end I feel like the current approach you guys have there is the best way of doing things, it adds an extra struct there but at least the interface is cleaner.

The one thing I would add to `BindingViewStore` though would be a an extra `subscript` to get a `Binding<Value>` directly (along with the `subscript` for `BindingViewState<Value>`) so that consumers may opt to use `Binding` in their `ViewState` rather than `BindingViewState` (provided they conform `Binding` to `Equatable`) so that the View feels more like SwiftUI if you know what I mean.

Having said all that, I will close this PR, since the solution you've pointed out there seems to be the way to go, I'd just like to ask when are you guys thinking about merging these into `master`/`prerelease`? Is there something that needs to be done before this solution goes forward?

tgrapperon:
> Should be trivial to add that conformance to Binding

Without weighting on the PR, just a precision: no, it's likely not trivial: `Binding` is not merely a wrapper around a value, but rather around a way to get it and set it. They furthermore come in various flavours: functional (through `get:set:`), extracted from a `@State`, an `ObservableObject`, or through `WritableKeyPath` composition, etc. Experience shows that they behave slightly differently in SwiftUI (most likely because the `get:set:` version can't be diffed/equated as efficiently as the `KeyPath` based ones).
But then, you wouldn't be able to compare two `Binding(get:set:)` bindings, or one `Binding(get:set:)` that binds some `@ObservedObject` property with `@model.theProperty`.

That being said, conforming type we don't own to protocols we don't own is not recommended and will likely at least [raise a warning](https://github.com/apple/swift-evolution/blob/main/proposals/0364-retroactive-conformance-warning.md) in the next versions of Swift.

rkreutz:
Yeah, fair point. I agree the lib shouldn't extend types it doesn't own as well. As for it being trivial, was mostly for what `Equatable` is used in this context (checking whether the View should be updated or not) whether the get/set functions changed shouldn't be relevant for this context (though on some cases it might), but that's something consumers can weight in and decide how to do it (either try using Binding with their own Equatable conformance, or BindigViewState provided by the lib).

So is there a roadmap for including this into a future release?

rkreutz:
Hey @stephencelis sorry to bother, but just wanted to check what's missing from the roadmap to have `BindingViewStore`/`State` merged into a version of the lib? Maybe I could help on doing something that's missing there? Thanks

PR #2053, opened 2023-04-22
Missing space in Reducer constraints

freak4pc:
If there ever was a nitpick ... 🤦‍♂️

(bumped into this by accident)

### DIFF
--- a/Sources/ComposableArchitecture/ReducerProtocol.swift
+++ b/Sources/ComposableArchitecture/ReducerProtocol.swift
@@ -160,7 +160,7 @@
   /// }
   /// ```
   ///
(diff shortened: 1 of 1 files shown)


tgrapperon:
Quite interesting that `swift-format` didn't pick it up. The feature is rather new, so that's maybe a bug to report.

mbrandonw:
It's currently this way due to a past bug in swift-format (see https://github.com/apple/swift-format/issues/414), but it was fixed quite awhile ago, so as long as we are using the newest swift-format it should be OK. I will merge it and we'll see if our swift-format CI action automatically removes it. If it does then we need to forcibly upgrade our swift-format.

PR #2062, opened 2023-04-27
Rename FullscreenCover.swift to FullScreenCover.swift

Jager-yoo:
Rename the file for consistency with `FullScreenCover` nav APIs.


mbrandonw:
Hi @Jager-yoo, good catch! Could you re-target this change to navigation-beta instead of prerelease/1.0? That's where the file was first introduced. Thanks!

mbrandonw (approved):
Great, thank you!

PR #2084, opened 2023-05-05
Demonstration of compiler type resolution issues with Swift 5.8 and `CombineReducers`

mluisbrown:
This is a demonstration of code that compiles pretty quickly in Xcode 14.2 (Swift 5.7) and which doesn't compile on Xcode 14.3 (Swift 5.8). The compiler will eventually bail out saying it wasn't possible to type check the expression in reasonably time.

image

image

I created this as a PR as it seemed like the easiest way to provide you with an example. Feel free to close the PR or do whatever you want with it as it's obviously not meant to be a contribution to the project 😄

### DIFF
--- a/Tests/ComposableArchitectureTests/ReducerTests.swift
+++ b/Tests/ComposableArchitectureTests/ReducerTests.swift
@@ -4,6 +4,67 @@ import CustomDump
 import XCTest
 import os.signpost
 
+public extension ReducerProtocol {
+    @inlinable
+    func onChange<ChildState: Equatable>(
+        of toLocalState: @escaping (State) -> ChildState,
+        perform additionalEffects: @escaping (ChildState, inout State, Action) -> EffectTask<
+            Action
+        >
+    ) -> some ReducerProtocol<State, Action> {
+        self.onChange(of: toLocalState) { additionalEffects($1, &$2, $3) }
+    }
+
+    @inlinable
+    func onChange<ChildState: Equatable>(
+        of toLocalState: @escaping (State) -> ChildState,
+        perform additionalEffects: @escaping (ChildState, ChildState, inout State, Action) -> EffectTask<
+            Action
+        >
+    ) -> some ReducerProtocol<State, Action> {
+        ChangeReducer(base: self, toLocalState: toLocalState, perform: additionalEffects)
+    }
+}
+
+@usableFromInline
+struct ChangeReducer<Base: ReducerProtocol, ChildState: Equatable>: ReducerProtocol {
+    @usableFromInline let base: Base
+
+    @usableFromInline let toLocalState: (Base.State) -> ChildState
+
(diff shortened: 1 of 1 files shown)


Tests/ComposableArchitectureTests/ReducerTests.swift
```diff
@@ -97,16 +171,72 @@ final class ReducerTests: BaseTCATestCase {
     var first = false
     var second = false
 
+    let _store = Store(
+      initialState: State(counter: 0),
+      reducer: CombineReducers {
+        One(effect: { @MainActor in first = true })
+        Two(effect: { @MainActor in second = true })
+      }.onChange(of: \.counter) { _, _, _ in
```
stephencelis, line 179: The main issue here is that `Store.init(reducer:)` doesn't take a reducer as a builder context, but `CombineReducers` and other generic reducers only really work inside builder contexts in Swift 5.8 due to the new type checking limitations. So if you plan on combining reducers, you must define a custom conformance with a `body` property:

```swift
struct RootReducer: ReducerProtocol {
    CombineReducers {
      // Note: `first` and `second` must be hoisted out globally to reference here
      One(effect: { @MainActor in first = true })
      Two(effect: { @MainActor in second = true })
    }
    .onChange(of: \.counter) { _, _, _ in
      .send(.increment)
    }
  }
}
```

This compiles fine.

So we should probable add a `Store.init` to takes a builder and maybe soft-deprecate the existing initializers. It's something we've talked about but this is a good example of a problem it solves.

Tests/ComposableArchitectureTests/ReducerTests.swift
```diff
@@ -97,16 +171,72 @@ final class ReducerTests: BaseTCATestCase {
...
+          .send(.increment)
+      }.onChange(of: \.counter) { _, _, _ in
+          .send(.increment)
+      }.onChange(of: \.counter) { _, _, _ in
+          .send(.increment)
+      }.onChange(of: \.counter) { _, _, _ in
+          .send(.increment)
+      }.onChange(of: \.counter) { _, _, _ in
+          .send(.increment)
+      }
```
stephencelis, line 227: Now the sheer amount of chaining in this example does not compile even when hoisted into a `body`, and the problems is exponential:

- 9 chains: 3 seconds
- 10 chains: 6 seconds
- 11 chains: 12 seconds
- 12 chains: 30 seconds

If these compile quickly in Xcode 14.2, you may have stumbled on a big type checking regression in Swift 5.8, so it might be good to file an issue with Apple.

mluisbrown:
Will file a bug report with apple regarding the chaining issue. This PR has served its purpose, so closing it 👍

mluisbrown:
FYI: create Swift issue about this type checker regression: https://github.com/apple/swift/issues/65765

PR #2087, opened 2023-05-05
Add `Store.init` that takes reducer builder

stephencelis:
Unfinished draft for now.

#2084 brought up an issue with the existing initializer, which is since it doesn't provide a builder context, we can't use things like `CombineReducers` directly when passed to the store. While we generally wouldn't recommend composing reducers outside a conformance's `body` property, we probably shouldn't prohibit it, and instead should encourage reducers to be passed into builders for flexibility in general.

If we do go in this direction, some more to-dos:

- [x] Add a `TestStore.init`
- [x] Update remaining usage throughout code base


Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-AlertsAndConfirmationDialogs.swift
```diff
@@ -128,10 +128,9 @@ struct AlertAndConfirmationDialog_Previews: PreviewProvider {
   static var previews: some View {
     NavigationView {
       AlertAndConfirmationDialogView(
-        store: Store(
-          initialState: AlertAndConfirmationDialog.State(),
-          reducer: AlertAndConfirmationDialog()
-        )
+        store: Store(initialState: AlertAndConfirmationDialog.State()) {
+          AlertAndConfirmationDialog()
+        }
```
stephencelis, line 133: Usage is generally shorter, which is kinda nice.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -145,9 +145,10 @@ public final class Store<State, Action> {
   ///     by the reducer.
   public convenience init<R: ReducerProtocol>(
     initialState: @autoclosure () -> R.State,
-    reducer: R,
+    @ReducerBuilder<State, Action> reducer: () -> R,
     prepareDependencies: ((inout DependencyValues) -> Void)? = nil
   ) where R.State == State, R.Action == Action {
+    let reducer = reducer()
```
stephencelis, line 151: One question is if this should be evaluated using `prepareDependencies`...and if so, does that mean we can drop the `transformDependencies` we chain onto it?

Tests/ComposableArchitectureTests/EffectRunTests.swift
```diff
@@ -122,8 +122,8 @@ final class EffectRunTests: BaseTCATestCase {
   }
 
   #if DEBUG
-    func testRunEscapeFailure() async throws {
```
stephencelis, line 125: Unrelated warning that popped up. Can move to another PR if we don't merge this anytime soon.

Tests/ComposableArchitectureTests/StoreTests.swift
```diff
@@ -684,7 +684,7 @@ final class StoreTests: BaseTCATestCase {
   }
 
   #if swift(>=5.7)
-    func testChidlParentEffectCancellation() async throws {
+    func testChildParentEffectCancellation() async throws {
```
stephencelis, line 687: Unrelated typo that I spotted. Can also move if need be.

Tests/ComposableArchitectureTests/StoreTests.swift
```diff
@@ -711,7 +711,7 @@ final class StoreTests: BaseTCATestCase {
           case delay
         }
         @Dependency(\.mainQueue) var mainQueue
-        var body: Reduce<State, Action> {
+        var body: some ReducerProtocol<State, Action> {
```
stephencelis, line 714: Unrelated warning.

Tests/ComposableArchitectureTests/TestStoreTests.swift
```diff
@@ -191,7 +191,7 @@ final class TestStoreTests: BaseTCATestCase {
         predicateShouldBeCalledExpectation.fulfill()
         return action == .finished
       }
-      wait(for: [predicateShouldBeCalledExpectation], timeout: 0)
+      { wait(for: [predicateShouldBeCalledExpectation], timeout: 0) }()
```
stephencelis, line 194: Unrelated warning.

stephencelis:
New API means new chance to bike shed, too...

```diff
-Store(initialState: AppFeature.State()) {
+Store(reducing: AppFeature.State()) {
   AppFeature()
 }
```

But probably not worth changing at this point?

Sources/ComposableArchitecture/Store.swift
tgrapperon: There is also the question to forward environmental dependencies around `Store` into the reducer that was raised by @lukeredpath a few weeks ago on Slack. I don't remember exactly what was the outcome. It think this is what was behind my `isEmpty` PR in `swift-dependencies` (the property was used to detect if the environmental dependencies were empty/overridden in order to avoid a trivial `.transformDependencies`).

tgrapperon: [The discussion in question](https://pointfreecommunity.slack.com/archives/C04L2D0MNJH/p1676982334184199)

sadiklar:
Tvbn

mbrandonw (approved):

PR #2094, opened 2023-05-10
Fix invalid states count for 3 optionals and typos

Jager-yoo:
In the `# Enum state` description within TreeBasedNavigation, it is mentioned that 3 optionals lead to 3 invalid states. However, 3 optionals result in `4 invalid states`, like so:

```
O O X (where 'O' represents non-nil and 'X' represents nil)
O X O
X O O
O O O
```

The rest example is accurate! 😉 Additionally, I fixed several typos.
Please review the changes.


mbrandonw (approved):
Great finds, thanks!

PR #2100, opened 2023-05-12
Tests passing

alex-reilly-pronto:


Package.resolved
```diff
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
```
alex-reilly-pronto, line 5: Roll these back before merging

Sources/ComposableArchitecture/TestStore.swift
```diff
@@ -1463,7 +1468,6 @@ extension TestStore where ScopedState: Equatable {
   /// - Parameters:
   ///   - isMatching: A closure that attempts to match an action. If it returns `false`, a test
   ///     failure is reported.
-  ///   - nanoseconds: The amount of time to wait for the expected action.
```
alex-reilly-pronto, line 1466: No timeout in this method

Sources/ComposableArchitecture/TestStore.swift
```diff
@@ -1849,9 +1864,19 @@ extension TestStore where ScopedState: Equatable {
...
       await Task.detached(priority: .background) { await Task.yield() }.value
-
-      guard self.reducer.receivedActions.isEmpty
-      else { break }
+      
+      switch self.exhaustivity {
+      case .on:
+        guard self.reducer.receivedActions.isEmpty
+        else {
+          return
```
alex-reilly-pronto, line 1872: I think returning here is okay, because breaking and returning are equivalent since nothing happens after the while loop.

Sources/ComposableArchitecture/TestStore.swift
```diff
@@ -1890,9 +1915,6 @@ extension TestStore where ScopedState: Equatable {
         return
       }
     }
-
-    guard !Task.isCancelled
-    else { return }
```
alex-reilly-pronto, line 1895: I think this doesn't do anything since both branches returned. Let me know if you want it back.

alex-reilly-pronto:
@stephencelis @mbrandonw Wrote up an explanation for the PR and threw in some tests. Thanks for the example! Let me know if anything is unclear

Alireza6924 (approved):

stephencelis (approved):

mbrandonw (approved):
Thanks for the PR @alex-reilly-pronto! This looks good to us and we will merge soon.

alex-reilly-pronto:
Awesome! Happy to hear it. Is that CI failure expected or did I break something?

alex-reilly-pronto:
Oh nevermind this is one of mine:
```
Test Case '-[ComposableArchitectureTests.TestStoreNonExhaustiveTests testReceiveNonExhuastiveWithTimeout]' started.
/Users/runner/work/swift-composable-architecture/swift-composable-architecture/Tests/ComposableArchitectureTests/TestStoreNonExhaustiveTests.swift:743: error: -[ComposableArchitectureTests.TestStoreNonExhaustiveTests testReceiveNonExhuastiveWithTimeout] : Expected to receive a matching action, but received none after 0.003 seconds.

There are effects in-flight. If the effect that delivers this action uses a clock/scheduler (via "receive(on:)", "delay", "debounce", etc.), make sure that you wait enough time for it to perform the effect. If you are using a test clock/scheduler, advance it so that the effects may complete, or consider using an immediate clock/scheduler to immediately perform the effect instead.

If you are not yet using a clock/scheduler, or can not use a clock/scheduler, try increasing the duration of this assertion's "timeout".
Test Case '-[ComposableArchitectureTests.TestStoreNonExhaustiveTests testReceiveNonExhuastiveWithTimeout]' failed (0.010 seconds).
```

Maybe it needs a longer timeout?

mbrandonw:
I just bumped the timeouts. They are a bit flakey, but we do have some tools coming soon to improve the situation.

alex-reilly-pronto:
Ope we had the same idea. Feel free to roll mine back if you don't want all of them bumped.

PR #2103, opened 2023-05-12
Prefer `body` over `reduce` in docs and sample code

stephencelis:
Just taking a few steps from #2102.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Basics.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Basics.swift
@@ -35,24 +35,26 @@ struct BindingBasics: Reducer {
     case toggleChanged(isOn: Bool)
   }
 
-  func reduce(into state: inout State, action: Action) -> Effect<Action> {
-    switch action {
-    case let .sliderValueChanged(value):
-      state.sliderValue = value
-      return .none
-
-    case let .stepCountChanged(count):
-      state.sliderValue = .minimum(state.sliderValue, Double(count))
-      state.stepCount = count
-      return .none
-
-    case let .textChanged(text):
-      state.text = text
-      return .none
-
-    case let .toggleChanged(isOn):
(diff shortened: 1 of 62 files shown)


stephencelis:
Going to leave this as a draft to consider for now, but don't think it'll make the cut for 1.0. It's mostly a documentation concern for now, so we can always revisit it later.

stephencelis:
Think we can revisit this now!

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Basics.swift
```diff
@@ -35,24 +35,26 @@ struct BindingBasics: Reducer {
...
-      return .none
-
-    case let .textChanged(text):
-      state.text = text
-      return .none
-
-    case let .toggleChanged(isOn):
-      state.toggleIsOn = isOn
-      return .none
+  var body: some Reducer<State, Action> {
```
Muhammed9991, line 38: Are you writing `Reducer<State, Action>` because it's an example? Self can be used here as well right? If `ReducerOf` is used

```suggestion
  var body: some ReducerOf<Self> {
```

stephencelis: @Muhammed9991 We haven't 100% landed on a style here. While we like the brevity of `ReducerOf<Self>` and it can be used most of the time, it is another concept to be aware of, so we need to decide how and when to disclose it...

stephencelis:
Closing this for now, but will bring into another release branch in the future.

Muhammed9991:
Do you need help finishing this or is it another reason you want to close it?

stephencelis:
@Muhammed9991 We are working on some features that affect a bunch of docs, so we reincorporated this work into those branches instead of dealing with merge conflicts. If you join [the Slack community](https://www.pointfree.co/slack-invite) we have a #tca-development channel with more info.

PR #2109, opened 2023-05-15
Some tutorial fixes.

mbrandonw:


Makefile
```diff
@@ -52,11 +52,13 @@ benchmark:
...
 
 format:
-	swift format \
+	@mv ./Sources/ComposableArchitecture/Documentation.docc ./tmp.docc
+	@swift format \
 		--ignore-unparsable-files \
 		--in-place \
 		--recursive \
 		./Examples ./Package.swift ./Sources ./Tests
+	@mv ./tmp.docc ./Sources/ComposableArchitecture/Documentation.docc
```
mbrandonw, line 61: this is the best I could come up with to make swift-format not change the tutorial snippet files. is there something better we can do?

stephencelis (approved):

PR #2115, opened 2023-05-17
Fix  compile error  below Xcode 14.3

hj56775:
I am using Xcode 14.2.
I am getting compile errors due to the new changes to debugReducer.
It seems that @_docmentation is available in 5.8 and later.

https://www.swift.org/blog/swift-5.8-released/
https://forums.swift.org/t/swift-5-8-documentation/63346


mbrandonw (approved):
Ouch, good catch! Will merge this now to get a quick release out, and we will restore Swift 5.7 CI coverage.

PR #2116, opened 2023-05-17
Custom _ReducerPrinters to enable using loggers

ddeneka:
This change enables using custom logging for `._printChanges()` for `customDump` & `actionLabels`

Example use:

```swift
._printChanges(.actionLabels {
  logger.notice("View action: \($0)")
})
```

### DIFF
--- a/Sources/ComposableArchitecture/Reducer/Reducers/DebugReducer.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/DebugReducer.swift
@@ -42,19 +42,34 @@ public struct _ReducerPrinter<State, Action> {
 
 extension _ReducerPrinter {
   public static var customDump: Self {
-    Self { receivedAction, oldState, newState in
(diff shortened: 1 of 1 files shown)


stephencelis:
@ddeneka Thanks for taking the time to PR!

While `_printChanges()` is a super useful API for debugging, we're not totally happy with the customization it offers right now. It's very bare bones and may not work for everyone. Even though these APIs are still underscored, we'd still like to give a bit more thought to how we want to evolve them and minimize churn in the meantime, so we may hold off on changes for now. I think the good news is all these APIs are public, so you can make the same customizations in your app without waiting on us to bring changes into the library.

Let us know what you think!

ddeneka:
Thanks @stephencelis! You are right, this can be implemented using a public API + grabbing the debugCaseOutput func from the TCA code base.
The ability to use custom logging can be really helpful especially for apps which are outputting a lot of stuff to the stdout.
Thanks again for such a quick response

PR #2120, opened 2023-05-18
Fix CI: Actually pass configuration to make

stephencelis:
Make doesn't override env vars in the traditional way, so portions of our CI (library tests) were only running tests in debug mode. This should fix.


mbrandonw (approved):
🤦‍♂️

PR #2121, opened 2023-05-19
Ukrainian translation of TCA's README

barabashd:
PointFree supports new localizations. The world supports Ukraine now. So, let's add support for Ukrainian in TCA.

// PS: It would be great to automate localizations in the future to reflect updates in other Gist files as well.


stephencelis (approved):
Thanks!

PR #2124, opened 2023-05-21
Fix usage of static storage property

tomassliz:
I just fixed the compile error:
> Static member 'store' cannot be used on instance of type 'MyApp'.


mbrandonw (approved):
Great finds, thanks!

PR #2128, opened 2023-05-22
Fix assertion values

tomassliz:
It's just a small bug in tests when assertion values don't contain a comma character.


mbrandonw (approved):
Another good catch, thank you!

PR #2144, opened 2023-05-27
[Documentation] Fixed action type: `struct` to `enum`

jaesung-0o0:
There is a minor modification on the article: "Migration to the Reducer Protocol". 
The given code snippet declares the action as `struct` type, but it should be `enum`.


stephencelis (approved):
Good catch, thanks!

PR #2147, opened 2023-05-27
Add `withDependencies` to `send` and `receive`

oskarek:
The `TestStore` initializer takes an optional `withDependencies` parameter, where you can setup custom dependencies to be used during the test. This, coupled with an `unimplemented` `testValue` implementation for all dependencies, means that the test will fail if any dependency that is _not_ setup in the `withDependencies` is ever called during the test.

That's great, because you get test coverage of the fact that no unwanted dependencies are called during the test. However, it could be made even stronger. Because often when testing some user flow of actions, some dependencies (if not all) are only expected to be used _once_ during the flow, in an effect from one single action being sent into the store. But by overriding the dependency for the _entire duration_ of the test, the dependency could be called from every action being sent into the store, and you would not be notified of that (with a failing test).

This PR adds the ability to pass an optional `withDependencies` to the `TestStore.send`, and the `TestStore.receive` methods, to be even more local about when a given dependency is expected to be used.

It looks like this in a test:

```swift
let store = TestStore(initialState: MyReducer.State()) {
  MyReducer()
}

await store.send(.viewDidLoad) {
  $0.isLoading = true
} withDependencies: {
  $0.apiClient = .mock()
}

await store.receive(.initialFetchResponse(.success("dummy"))) {
  $0.isLoading = false
  $0.loadedValue = "dummy"
}

let trackedPages = LockIsolated([Tracking.Page]())
await store.send(.viewDidAppear) {
  $0.tracker.trackPage = { page in trackedPages.withValue { $0.append(page) } }
}
XCTAssertEqual(trackedPages.value, [.myPage])

let isDismissInvoked = LockIsolated(false)
await store.send(.closeButtonTapped) {
  $0.dismiss = .init { isDismissInvoked.setValue(true) }
}
XCTAssertTrue(isDismissInvoked.value)
```

With this setup, you get coverage of the fact that the `apiClient` dependency, for example, is only used in the `.viewDidLoad` action, and not in any other part of the flow being tested. and that the page tracking only happens in the `.viewDidAppear` action, and so on.

What do you think? 🙂

### DIFF
--- a/Sources/ComposableArchitecture/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestStore.swift
@@ -1067,6 +1067,20 @@ extension TestStore where ScopedState: Equatable {
     return .init(rawValue: task, timeout: self.timeout)
   }
 
+  @MainActor
+  @discardableResult
+  public func send(
+    _ action: ScopedAction,
+    assert updateStateToExpectedResult: ((inout ScopedState) throws -> Void)? = nil,
+    withDependencies updateValuesForAction: (inout DependencyValues) async -> Void,
+    file: StaticString = #file,
+    line: UInt = #line
+  ) async -> TestStoreTask {
+    await self.withDependencies(updateValuesForAction) {
+      await self.send(action, assert: updateStateToExpectedResult, file: file, line: line)
+    }
+  }
+
   /// Assert against the current state of the store.
   ///
   /// The trailing closure provided is given a mutable argument that represents the current state,
@@ -1449,6 +1463,27 @@ extension TestStore where ScopedState: Equatable, Action: Equatable {
         line: line
       )
     }
+
+    @available(iOS 16, macOS 13, tvOS 16, watchOS 9, *)
+    @MainActor
+    public func receive(
+      _ expectedAction: Action,
+      timeout duration: Duration,
+      assert updateStateToExpectedResult: ((inout ScopedState) throws -> Void)? = nil,
+      withDependencies updateValuesForAction: (inout DependencyValues) async -> Void,
+      file: StaticString = #file,
+      line: UInt = #line
+    ) async {
+      await self.withDependencies(updateValuesForAction) {
+        await self.receive(
+          expectedAction,
+          timeout: duration,
+          assert: updateStateToExpectedResult,
+          file: file,
+          line: line
+        )
+      }
+    }
   #endif
 
   /// Asserts an action was received from an effect and asserts how the state changes.
@@ -1506,6 +1541,27 @@ extension TestStore where ScopedState: Equatable, Action: Equatable {
       self._receive(expectedAction, assert: updateStateToExpectedResult, file: file, line: line)
     }()
     await Task.megaYield()
+  }
+
+  @MainActor
+  @_disfavoredOverload
+  public func receive(
+    _ expectedAction: Action,
+    timeout nanoseconds: UInt64? = nil,
+    assert updateStateToExpectedResult: ((inout ScopedState) throws -> Void)? = nil,
+    withDependencies updateValuesForAction: (inout DependencyValues) async -> Void,
+    file: StaticString = #file,
+    line: UInt = #line
+  ) async {
+    await self.withDependencies(updateValuesForAction) {
+      await self.receive(
+        expectedAction,
+        timeout: nanoseconds,
+        assert: updateStateToExpectedResult,
+        file: file,
+        line: line
+      )
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


stephencelis:
Hey @oskarek! Thanks for the PR!

I definitely think this can read pretty nice, but wanted to let you know that we do have a couple of existing tools that work here.

First, `TestStore` has a `dependencies` property that can be mutated whenever you want to in a test. So the above test could be written as follows:

```diff
 let store = TestStore(initialState: MyReducer.State()) {
   MyReducer()
 }

+store.dependencies.apiClient = .mock()
 await store.send(.viewDidLoad) {
   $0.isLoading = true
-} withDependencies: {
-  $0.apiClient = .mock()
 }

 await store.receive(.initialFetchResponse(.success("dummy"))) {
   $0.isLoading = false
   $0.loadedValue = "dummy"
 }

 let trackedPages = LockIsolated([Tracking.Page]())
+store.dependencies.tracker.trackPage = { page in trackedPages.withValue { $0.append(page) } }
+await store.send(.viewDidAppear)
-await store.send(.viewDidAppear) {
-  $0.tracker.trackPage = { page in trackedPages.withValue { $0.append(page) } }
-}
 XCTAssertEqual(trackedPages.value, [.myPage])

 let isDismissInvoked = LockIsolated(false)
+store.dependencies.dismiss = .init { isDismissInvoked.setValue(true) }
+await store.send(.closeButtonTapped)
-await store.send(.closeButtonTapped) {
-  $0.dismiss = .init { isDismissInvoked.setValue(true) }
-}
 XCTAssertTrue(isDismissInvoked.value)
```

Second, when our [Navigation beta](https://github.com/pointfreeco/swift-composable-architecture/pull/1945) is merged (soon!) we'll also have `store.withDependencies { … }` as an alternative way of solving the problem in a more scoped way. This API is admittedly a little confusing, though, and works around a more fundamental problem with the `TestStore`, which is that it's not compatible with swift-dependencies' `withDependencies` function. For 1.0 we'd like to address this shortcoming, which means rethinking a bunch of the existing machinery.

What do you think of these existing solutions for now?

oskarek:
Thank you @stephencelis for the response!

I could have mentioned this in the description, but yes, I am aware of both of those alternatives. In fact, I am actually making use of the `store.withDependencies { ... }` method in this PR under the hood in each of the overloads.

However, I don't think either of them are really ideal. The first alternative you mention, directly mutating the `store.dependencies` value, is not perfect because the dependency "leaks" to the rest of the test, rather than being local to the single `send`/`receive`. And directly using the `store.withDependencies { ... }` is more verbose then just adding on a `withDependencies` trailing closure to the `send`/`receive` method calls.

Also, even if the `TestStore` would be compatible with the global `withDependencies` function, I still don't think this:

```swift
await withDependencies {
  $0.apiClient = .mock()
} operation: {
  await store.send(.viewDidLoad) {
    $0.isLoading = true
  }
}
```
looks as good as:
```swift
await store.send(.viewDidLoad) {
  $0.isLoading = true
} withDependencies: {
  $0.apiClient = .mock()
}
```
(if that's the API you had in mind, when adding support for that function).

I think the perfect place to be able to locally set the dependencies for the scope of a particular `send`/`receive` _is_ with a trailing `withDependencies` closure, exactly matching how it's done in the `TestStore.init`, for scoping the dependencies update for _all_ assertions done via that store. It just seems like a very natural API.

In our team, we have had these overloads defined for a while, and are actually using this style almost exclusively, and are finding that to be a very nice experience for writing tests. You get even stronger assertions on how the dependencies are used, without almost any extra effort, you just move the dependency update from `withDependencies` on `TestStore.init` to `withDependencies` on the `send`/`receive` call.

Am I overlooking any downsides to this approach? Please let me know your thoughts.

oskarek:
1. It's true, this is a potential problem. But as you say, it can be possible to decipher whether it's a dependency closure or an assertion closure based on context and what's actually inside the closure. And in practice, so far, this seems to be almost always the case that they are very easily distinguishable. But yes, this isn't necessarily always the case, at least in theory. But perhaps you can just use an explicit `assert`/`withDependencies` named parameter in those (hopefully rare) occasions, instead of the trailing closure syntax?

2. Okay yes I see what you mean. Personally I think

    ```swift
    await store.send(.action, exhaustivity: .off) {
      // `withDependencies` block (or state assertion)
    }
    ```

    actually looks quite good. But if you have more than these two configurations in mind as potential additions, then yes I see why you are hesitating to go down this route.

stephencelis:
Yeah 2.) looks OK, it is mainly pointing out a problem with all these APIs spiraling out of control with configuration options. While we're not opposed to evolving the APIs to the ones you've sketched in this PR, we would like to explore a few more options before committing to an API for 1.0. Feel free to keep this PR open as a reminder in the meantime!

stephencelis:
@oskarek We're closing in on 1.0 and while we definitely appreciate some of the ergonomics of this approach, we think the ambiguity is enough to avoid it for now. We may revisit things after 1.0 whenever we make substantial changes to the test store, but for now I'm going to close this out. Thanks again for the suggestion and PR, though! We'll continue to think about it. In the meantime, we hope it's OK that you continue to define these helpers in your own code, and if you extract things to a library, please do share in the discussions!

PR #2152, opened 2023-05-31
Fix typo

hmhv:
Fixed typo in Documentation


mbrandonw (approved):
Wow, good catch! Thanks @hmhv!

PR #2099, opened 2023-05-11
Soft-deprecate `Effect.task` and `Effect.fireAndForget`

stephencelis:
See https://github.com/pointfreeco/swift-composable-architecture/discussions/1520

PR #2153, opened 2023-05-31
`Effect.fireAndForget` is soft deprecated

Ryu0118:
fireAndForget was used in the official tutorial.
Since `Effect.fireAndForget` is soft deprecated in version 0.53.0, I rewrote it to `Effect.run`.
#2099 
Is there a reason why fireAndForget is used?


stephencelis (approved):
Thanks!

PR #2163, opened 2023-06-01
Scope an actionless store

stefanosheaDeputy:
With the deprecation of `func scope<ChildState>(state toChildState: @escaping (State) -> ChildState ) -> Store<ChildState, Action>`, when we are creating an actionless store it must be written like 

```Feature(store: store.scope(state: ChildState.init, action: { $0 }).actionless```

which doesn't read very well and can be confusing. This PR aims to fix that by providing a `scope` on an actionless store.

Keen to get some feedback and this or hear about another way to solve this.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -742,3 +742,17 @@ extension ScopedReducer: AnyScopedReducer {
     return childStore
   }
 }
+
+extension Store where Action == Never {
+    /// Scopes an actionless store to one that exposes child state and actions.
+    ///
+    /// - Parameters:
+    ///   - toChildState: A function that transforms `State` into `ChildState`.
+    /// - Returns: A new store with its domain (state and action) transformed.
+    public func scope<ChildState>(
+      state toChildState: @escaping (State) -> ChildState
+    ) -> Store<ChildState, Never> {
+        func absurd<A>(_: Never) -> A { }
+        return self.scope(state: toChildState, action: absurd)
+    }
+}

### CODE REVIEW

**Comment by stephencelis** [comment]
@stefanosheaDeputy Thanks for taking the time to PR! Just a bit of context around the removal of the actionless `scope`:

We've changed the way we think about store scoping over time, and to be honest we're finding `stateless` and `actionless` scoping less and less necessary, so I'm curious if you have a few examples you can share with us to better understand.

In our experience, `scope` is more and more used exclusively for parent-to-child domain transformations, where there is both state and action types to transform. If a child is "stateless" you can leave its state struct empty, and if it is "actionless" you can leave its action enum empty:

```swift
struct Parent: ReducerProtocol {
  struct State {
    var child: Child.State
  }
  enum Action {
    case child(Child.Action)
  }
    Scope(state: \.child, action: /Action.child) {
      Child()
    }
  }
}

struct Child: ReducerProtocol {
  struct State {}  // Equivalent to `Void` but can be made equatable
  enum Action {}   // Equivalent to `Never`
    EmptyReducer()
  }
}

struct ParentView: View {
  let store: StoreOf<Parent>
  var body: some View {
    ChildView(
      store: store.scope(state: \.child, action: { .child($0) }
    )
  }
}

struct ChildView: View {
  let store: StoreOf<Child>
  var body: some View {
    EmptyView()
  }
}
```

This way, `scope` will always have explicit domain transforms for both `state` and `action`, and it makes child domains easier to refactor when child state or actions need to be added.

Because of all this, we decided to remove the overload of `scope` that doesn't take an action transformation. We also find that overloads like `scope` can slow down view compilation, especially when there's a compile error lurking in a view body.

Meanwhile, view state and action transformations are done by the view store (via the required `observe` parameter, and the optional `send` parameter), so it is still possible to whittle a store's domain into a view store's domain by transforming state, but leaving actions untouched.

Now if you really want to transform state to `Void` and actions to `Never`, you can always use a closure that returns nothing, and you can use what's been typically called the `absurd` function:

```swift
func absurd<A>(_ never: Never) -> A {}

// Instead of:
store.stateless.actionless

// Do:
store.scope(state: { _ in }, action: absurd)
```

So we may even deprecate `stateless` and `actionless` for these alternate solutions in the future...

Please let us know if our thoughts resonate with you, or if you think we're not considering certain aspects of store scoping. We're always open to better understanding how these tools are used :smile:

mbrandonw:
Hi @stefanosheaDeputy, we are going to lose this for now, but please do start a [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions) if you want to discuss more.

PR #2173, opened 2023-06-03
Fix Cannot find type 'UUID' in scope

oronbz:
image


oronbz:
I just saw that you `import SwiftUI` later on the same file for the `View` so this might be implicit, feel free to close if you want.

PR #2191, opened 2023-06-10
PoC - requiresOnDeviceRecognition for Speech Recognition Example

blueberrychopsticks:
Hey guys, huge fan. I know this is a small contribution and is not ready to be merged yet, but I was hoping to get some of your thoughts of what you would like to see to get this merged. I think this could be an opportunity to show off a little more functionality to beginners such as myself for the true power and dynamism of TCA. Let me know your thoughts! Side note: Transcription across the board is MUCH better on iOS17 so that's pretty sweet! 

Due to how the Speech engine handles emitting SpeechRecognition results, they need to be handled slightly more delicately than just setting the transcription to the resultant string.

I think it would be useful to see in this example some further usage of The Composable Architecture  to show how to dynamically switch the setting at runtime depending on NetworkConditions / etc or perhaps allow the user to dynamically enable and disable their preference for punctuation.

### DIFF
--- a/Examples/SpeechRecognition/SpeechRecognition/SpeechRecognition.swift
+++ b/Examples/SpeechRecognition/SpeechRecognition/SpeechRecognition.swift
@@ -13,12 +13,14 @@ struct SpeechRecognition: ReducerProtocol {
     @PresentationState var alert: AlertState<Action.Alert>?
     var isRecording = false
     var transcribedText = ""
+    var speechIsProcessing = false
+    var transcribedSegments: [String] = []
   }
 
   enum Action: Equatable {
     case alert(PresentationAction<Alert>)
     case recordButtonTapped
-    case speech(TaskResult<String>)
+    case speech(TaskResult<SpeechRecognitionResult>)
     case speechRecognizerAuthorizationStatusResponse(SFSpeechRecognizerAuthorizationStatus)
 
     enum Alert: Equatable {}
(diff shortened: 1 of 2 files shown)


mbrandonw:
Hi @blueberrychopsticks, I think currently we would err on the side of keeping this demo simple even if that means it's not 100% correct. The main point of the demo is just to show how to control a complex dependency, and I worry that adding all of the additional transcription processing code is going to detract from that.

I'm going to close this for now, but if you do end up experimenting with more bells and whistles (like the network connectivity stuff), do let us know. Maybe eventually this will be more appropriate for the Standups demo, which also uses a speech recognizer.

blueberrychopsticks:
> Hi @blueberrychopsticks, I think currently we would err on the side of keeping this demo simple even if that means it's not 100% correct. The main point of the demo is just to show how to control a complex dependency, and I worry that adding all of the additional transcription processing code is going to detract from that.
> 
> I'm going to close this for now, but if you do end up experimenting with more bells and whistles (like the network connectivity stuff), do let us know. Maybe eventually this will be more appropriate for the Standups demo, which also uses a speech recognizer.

Hey @mbrandonw - thanks so much for just responding! I know you guys are immensely busy so it means a lot with such a modest PR.

As for letting you guys know - I don't want to pollute your PRs - would a discussion be more appropriate?

My initial goal for the short term would be an application that mimics Otter.ai's speech recognition ability with offline support and plugable dependencies for utilizing different STT providers such as Deepgram, Symbl.ai, Google, Amazon, etc to show the full power of abstracting implementation details behind the Client structs. 

Once I get some decent functionality figured out, how would be the best way to let you guys know?

Best, bc

PR #2204, opened 2023-06-15
Fix some swiftdoc and small tidy

alexhunsley:
Note that the diff line where it looks like a no-op is actually adding a missing newline to end of file


stephencelis (approved):
All great finds, thanks!

PR #2205, opened 2023-06-15
Fix `IdentifiedCollections` import

larryonoff:
We have CocoaPods spec of swift-composable-architecture. It doesn't compile since the error
```
- ERROR | [iOS] xcodebuild:  ComposableArchitecture/Sources/ComposableArchitecture/SwiftUI/PresentationModifier.swift:220:46: error: cannot find 'Identified' in scope
```

So I've added forgotten import to fix it.


mbrandonw:
Hi @larryonoff, thanks for this. The change is innocuous enough, but I'm not sure why it is happening since we export `IdentifiedCollections` [here](https://github.com/pointfreeco/swift-composable-architecture/blob/42b01f284e9dcd06131bda03c1ac6dec8230aae9/Sources/ComposableArchitecture/Internal/Exports.swift#L6). And there are dozens of other files where we do not import the library's dependency thanks to that export file, so I would expect all of those files to have a problem too. Are you sure there isn't something else happening with Cocoapods?

stephencelis:
@larryonoff Any update here? While we're open to merging this if it solves a problem, it's not a problem we can reproduce when using the library via SPM. Are you sure this is a problem in the library and not in CocoaPods? Can you provide a sample project showing the issue?

larryonoff:
@stephencelis @mbrandonw sorry for the delay.

Our team uses both SPM and Cococapods. We have to create our podspecs for TCA dependencies (e.g. TCA, swift-collections, etc.) since CocoaPods is not officially supported. 

I don't understand the reason, why it happens for CocoaPods only. Imports sometimes behave different with CocoaPods.

Please see podspec.

```ruby
Pod::Spec.new do |s|
    s.authors = { 'pointfreeco' => 'https://github.com/pointfreeco' }

    s.name = 'ComposableArchitecture'
    s.version  = '0.54.1'
    s.homepage = 'https://github.com/pointfreeco/swift-composable-architecture'
    s.source = {
        :git => 'git@github.com:pointfreeco/swift-composable-architecture.git',
        :tag => s.version.to_s
    }

    s.summary = 'Point-Free, Inc.'
    s.description = 'Point-Free, Inc.'
    s.license = {
        :type => 'Copyright',
        :text => "Copyright #{Time.now.year} Point-Free, Inc.. All Rights Reserved."
    }
    s.requires_arc = true
    s.swift_versions = ['5.0']

    s.ios.deployment_target = '14.0'

    s.dependency '_SwiftUINavigationState', '~> 0.7.2'
    s.dependency 'CasePaths', '~> 0.14.1'
    s.dependency 'CombineSchedulers', '~> 0.10.0'
    s.dependency 'CustomDump', '~> 0.10.3'
    s.dependency 'Dependencies', '~> 0.5.1'
    s.dependency 'IdentifiedCollections', '~> 0.7.1'
    s.dependency 'XCTestDynamicOverlay', '~> 0.8.4'

    s.weak_frameworks = 'SwiftUI'

    s.source_files = [
        'Sources/ComposableArchitecture/**/*.swift'
    ]
end
```

mbrandonw:
@larryonoff Have you confirmed that adding that one import fixes everything? There are no other errors once that import is added? That would be surprising since there are lots of other files that don't explicitly import everything.

Also, you will want to update your pod spec to point to the new SwiftUINavigationCore instead of _SwiftUINavigationState. See https://github.com/pointfreeco/swiftui-navigation/pull/112 for more info.

larryonoff:
@mbrandonw I've finally found the issue. And it is not the import. I have wrong IdentifiedCollections podspec, that doesn't add Identified properly.

Sorry for confusion and thanks for help.

PR #2206, opened 2023-06-15
[prerelease/1.0] Rename ReducerProtocol.swift to Reducer.swift

Jager-yoo:
Hi, `ReducerProtocol` has already been renamed to `Reducer` in prerelease/1.0.
However, it seems that the file itself didn't get the news. 😉


mbrandonw (approved):
Good find, thank you!

Issue #2182, opened 2023-06-06
Tree-based navigation: dismiss action is not sent when navigating back on macOS (prerelease/1.0)


vadymmarkov:
### Description

I've recently updated TCA from a04be28aff42678566b76eb502ae4644bc4b36c2 to the latest commit b2815caa1727d72180836b85578272b763cc4853 from prerelease/1.0 branch and discovered that tree-based navigation stopped working as expected on macOS. 

For example, suppose we have the "featureA" screen wrapped in `NavigationStack`, and "featureB" is presented in a drill-down from "featureA" using `navigationDestination(store:state: action:destination:)`. When going back from "featureB" to "featureA" using the default "back" button the `.dismiss` action is not being sent to set the child state back to `nil`.

Strangely, this happens on macOS only, but works as expected on iOS. I don't know if it's a SwiftUI bug, but I didn't experience this behaviour with TCA from a04be28aff42678566b76eb502ae4644bc4b36c2.

### Checklist

- [X] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Expected behavior

The ``PresentationAction/dismiss`` action should be sent to `nil` out the child state when going back to the parent feature using the navigation bar "back" button.

### Actual behavior

The ``PresentationAction/dismiss`` action is not being sent to `nil` out the child state when going back to the parent feature using the navigation bar "back" button.

### Steps to reproduce

You can reproduce the incorrect behaviour in the attached project. Changing package commit to a04be28aff42678566b76eb502ae4644bc4b36c2 fixes the issue.

[NavigationTestCase.zip](https://github.com/pointfreeco/swift-composable-architecture/files/11669433/NavigationTestCase.zip)

### The Composable Architecture version information

b2815caa1727d72180836b85578272b763cc4853

### Destination operating system

macOS 13

### Xcode version information

14.3.1 

### Swift Compiler version information

```shell
swift-driver version: 1.75.2 Apple Swift version 5.8.1 (swiftlang-5.8.0.124.5 clang-1403.0.22.11.100)
Target: arm64-apple-macosx13.0
```

vadymmarkov:
[...]

PR #2210, opened 2023-06-17
Fixed tree-based navigation: dismiss action is not sent when navigating back

tplaymeow:
Fixed https://github.com/pointfreeco/swift-composable-architecture/issues/2182.

Added `.id(id)` modifier to update value of `id` in `compactSend` closure when presenting a child feature.

### DIFF
--- a/Examples/Integration/IntegrationUITests/PresentationTests.swift
+++ b/Examples/Integration/IntegrationUITests/PresentationTests.swift
@@ -288,6 +288,17 @@ final class PresentationTests: BaseIntegrationTests {
     XCTAssertEqual(self.app.staticTexts["Count: 0"].exists, false)
   }
 
+  func testNavigationDestination_BackButtonDismiss() async throws {
+    self.app.buttons["Open navigation destination"].tap()
+    XCTAssertEqual(self.app.staticTexts["Count: 0"].exists, true)
+
+    self.app.navigationBars.buttons.element.tap()
+    XCTAssertEqual(self.app.staticTexts["Count: 0"].exists, false)
+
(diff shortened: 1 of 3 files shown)


tplaymeow:
It seems to work for macOS, but breaks `navigationDestination` on iOS

stephencelis:
@tplaymeow Can you explain more what didn't work in iOS?

Also, if it works in macOS but fails in iOS, could we conditionally compile the modifier? I think the following should work:

```swift
#if os(macOS)
  .id(id)
#endif
```

tplaymeow:
@stephencelis 

It seems that this is SwiftUI bug. `navigationDestination(isPresented:)` does not update the Binding passed to the `isPresented` parameter when updating the view, although other methods for navigation (`popover(isPresented:)`, `sheet(isPresented:)`) do this. 

Because of this, when dismiss child feature, [`compacSend`](https://github.com/pointfreeco/swift-composable-architecture/blob/main/Sources/ComposableArchitecture/SwiftUI/PresentationModifier.swift#L223-L225
) contains an outdated `id` value and does not send `.dismiss` action.

I think we can not use `id` for `navigationDestination(isPresented:)` and just pass `Object Identifier(State.self)` instead of `$0.wrappedValue.map { _ in ObjectIdentifier(State.self) }`

This bug is also reproduced on some versions of iOS (16.0 - 16.3).
I have created a small project with a demonstration of this problem: [NavigationDestination.zip](https://github.com/pointfreeco/swift-composable-architecture/files/11886629/NavigationDestination.zip)

stephencelis (approved):
Looks good! Thanks for the detail deep dive into what was going wrong!

PR #2211, opened 2023-06-19
CaseStudies SharedState example update.

maksimn:
Repetitive fields from `Counter.State` and `Profile.State` have been moved into a separate `CounterData` struct. This makes the code more DRY and concise.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-SharedState.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-SharedState.swift
@@ -16,6 +16,13 @@ private let readMe = """
 
 // MARK: - Feature domain
 
+struct CounterData: Equatable {
+  var count = 0
+  var maxCount = 0
+  var minCount = 0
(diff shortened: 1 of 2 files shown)


stephencelis (commented):
Hey @maksimn, thanks for taking the time to PR! I think this is definitely a cleaner approach for the given demo, where the state can be cleanly bucketed into a data type.

For the sake of the example I think we wanted to keep things simple but also show an approach that scales with complexity. One issue that I think exists with the bucketed approach is when you begin to share more complex state, for example where shared fields are split between a parent or child or siblings, then I think we have to go back to assigning each field again. So I think we want to keep this style in the demo, but perhaps you have an idea of how to make the demo just ever so slightly more complex where the bucket approach no longer works? Then maybe the demo would be better motivated in style.

stephencelis:
Going to close this for now, but if you revisit a slightly more complex example that you'd like to propose, feel free to open another PR!

And thanks again for taking the time to contribute :smile:

maksimn:
1

PR #2219, opened 2023-06-21
Warn when sending actions from cancelled effects

stephencelis:
Today it is possible to cancel a running effect that is sending actions back into the system and, if the effect does not participate in cooperative cancellation, be none the wiser.

This PR adds a runtime warning whenever this happens. In general it shouldn't emit, as cooperatively-cancelling code will have already thrown an error, but in those cases in which you're not in cooperative code, it's nice to let the developer know.


mbrandonw (approved):

stephencelis:
I think this last failure puts the nail in this experiment's coffin. While it'd be nice to surface warnings when cooperative cancellation goes awry, we shouldn't emit warnings for a commonly acceptable case, like a cancellable task result.

While we could investigate special casing actions that hold task results...let's not go down that rabbit hole today.

PR #2226, opened 2023-06-22
Add `Reducer.onChange`

stephencelis:
We've had this reducer in isowords for a very long time, before the Dependencies library, reducer protocol, and builders, and while handy, it didn't feel right to include in the library at the time due to lack of clarity in the API itself (we had several overloads to provide various permutations of old value and new, `inout` and non-`inout` state, action, the environment, etc. But revisiting things with our more modern tools, a single signature stands out as "right":

```swift
.onChange(of: \.childState) { oldValue, newValue in
  Reduce { state, action in
    // ...
  }
}
```

So here's a PR that adds just that!

The main concern of this reducer is the potential performance cost, a caveat we've attempted to document. Basically this reducer should be preferred at leaf nodes (like `BindingReducer`) where the cost to equate the given value is low.

@krzysztofzablocki I know you've observed some issues with `onChange` in the past. Given our caveats do you have any thoughts you'd like to share?


Sources/ComposableArchitecture/Reducer/Reducers/OnChange.swift
```diff
@@ -0,0 +1,69 @@
...
+  /// > against values that are quick to equate.
+  ///
+  /// - Parameters:
+  ///   - toValue: A closure that returns a value from the given state.
+  ///   - reducer: A reducer builder closure to run when the value changes.
+  ///   - oldValue: The old value that failed the comparison check.
+  ///   - newValue: The new value that failed the comparison check.
+  /// - Returns: A reducer that performs the
+  public func onChange<V: Equatable, R: ReducerProtocol>(
+    of toValue: @escaping (State) -> V,
```
stephencelis, line 47: @mbrandonw Should this be `KeyPath<State, V>` instead to discourage doing too much work in the `toValue` block?

krzysztofzablocki:
The biggest performance cost I observed wasn't the diffing at all, it was due to protocols and higher order wrapping of reducers this lead to, the same exact code as a single `onChange` instead of N smaller `onChange` has been significantly (40%+) faster in my testing.

The other issue I saw with `onChange` was the fact that if we place it at wrong level of reducer it will never fire, where mutations happen vs where `onChange` is established is very important distinction that lots of folks didn't realize

stephencelis:
> The biggest performance cost I observed wasn't the diffing at all, it was due to protocols and higher order wrapping of reducers this lead to, the same exact code as a single `onChange` instead of N smaller `onChange` has been significantly (40%+) faster in my testing.

Did you have multiple `onChange`s being applied to the same sub-state? It seems like we could take care to fuse them together, but it also seems like the user could compose the `onChange` logic together into a single operator as it sounds like you did.

I also wonder if the version in this PR would perform differently than the version you used? Or does it look the same?

> The other issue I saw with `onChange` was the fact that if we place it at wrong level of reducer it will never fire, where mutations happen vs where `onChange` is established is very important distinction that lots of folks didn't realize

Ha yeah this is a gotcha, but we can document that it only observes changes in the reducer it's applied to.

krzysztofzablocki:
Looks the same, we had many different elements we were reacting to, simply having one reducer replace onChange and doing individual if with substate comparison was the 40% gain and it was gain on all of TCA action processing so huge difference for us

stephencelis:
@krzysztofzablocki I did a little bit of benchmarking and found that if all else is the same, inlining the work in `Reduce` vs. chaining the operators doesn't seem to incur that much of a slowdown, maybe 10% slower total with 10 chained operators:

```
name             time        std        iterations
--------------------------------------------------
Without onChange 3584.000 ns ±  15.92 %     358459
With onChange    3917.000 ns ±  10.23 %     354548
```

The biggest cost of using this operator (or any operator that combines reducers) is the `Effect.merge` that happens under the hood, so if you can avoid that by inlining the work, you'll see a much bigger difference. We do have plans to improve the performance of merging reducers much more in the future, so we think this is safe and useful enough to add to the library.

<details>
<summary>Expand for benchmark</summary>
```swift
struct R1: ReducerProtocol {
  struct State {
    var count = 0
  }
  enum Action {
    case incrementButtonTapped
  }
    Reduce { state, action in
      let oldCount = state.count
      var effect = EffectTask<Action>.none
      switch action {
      case .incrementButtonTapped:
        state.count += 1
      }
      var newCount = state.count
      if oldCount != newCount {
        state.count = newCount
        effect = .merge(effect, .none)
      }
      newCount = state.count
      if oldCount != newCount {
        state.count = newCount
        effect = .merge(effect, .none)
      }
      newCount = state.count
      if oldCount != newCount {
        state.count = newCount
        effect = .merge(effect, .none)
      }
      newCount = state.count
      if oldCount != newCount {
        state.count = newCount
        effect = .merge(effect, .none)
      }
      newCount = state.count
      if oldCount != newCount {
        state.count = newCount
        effect = .merge(effect, .none)
      }
      newCount = state.count
      if oldCount != newCount {
        state.count = newCount
        effect = .merge(effect, .none)
      }
      newCount = state.count
      if oldCount != newCount {
        state.count = newCount
        effect = .merge(effect, .none)
      }
      newCount = state.count
      if oldCount != newCount {
        state.count = newCount
        effect = .merge(effect, .none)
      }
      newCount = state.count
      if oldCount != newCount {
        state.count = newCount
        effect = .merge(effect, .none)
      }
      newCount = state.count
      if oldCount != newCount {
        state.count = newCount
        effect = .merge(effect, .none)
      }
      return effect
    }
  }
}

struct R2: ReducerProtocol {
  struct State {
    var count = 0
  }
  enum Action {
    case incrementButtonTapped
  }
    Reduce { state, action in
      switch action {
      case .incrementButtonTapped:
        state.count += 1
      }
      return .none
    }
    .onChange(of: \.count) { oldValue, newValue in
      Reduce { state, _ in
        state.count = newValue + 1
        return .none
      }
    }
    .onChange(of: \.count) { oldValue, newValue in
      Reduce { state, _ in
        state.count = newValue + 1
        return .none
      }
    }
    .onChange(of: \.count) { oldValue, newValue in
      Reduce { state, _ in
        state.count = newValue + 1
        return .none
      }
    }
    .onChange(of: \.count) { oldValue, newValue in
      Reduce { state, _ in
        state.count = newValue + 1
        return .none
      }
    }
    .onChange(of: \.count) { oldValue, newValue in
      Reduce { state, _ in
        state.count = newValue + 1
        return .none
      }
    }
    .onChange(of: \.count) { oldValue, newValue in
      Reduce { state, _ in
        state.count = newValue + 1
        return .none
      }
    }
    .onChange(of: \.count) { oldValue, newValue in
      Reduce { state, _ in
        state.count = newValue + 1
        return .none
      }
    }
    .onChange(of: \.count) { oldValue, newValue in
      Reduce { state, _ in
        state.count = newValue + 1
        return .none
      }
    }
    .onChange(of: \.count) { oldValue, newValue in
      Reduce { state, _ in
        state.count = newValue + 1
        return .none
      }
    }
    .onChange(of: \.count) { oldValue, newValue in
      Reduce { state, _ in
        state.count = newValue + 1
        return .none
      }
    }
  }
}

benchmark("Without onChange") {
  var state = R1.State()
  _ = R1().reduce(into: &state, action: .incrementButtonTapped)
}
benchmark("With onChange") {
  var state = R2.State()
  _ = R2().reduce(into: &state, action: .incrementButtonTapped)
}
```
</details>

PR #2227, opened 2023-06-22
Don't show 'expected state to change' error in non-exhaustive test stores.

mbrandonw:
Right now in non-exhaustive test stores showing skipped assertions, it is possible to get a skipped assertion complaining of "Expected state to change, but no change occurred.". That should only happen with exhaustive test stores.

### DIFF
--- a/Sources/ComposableArchitecture/Reducer/AnyReducer/AnyReducerSignpost.swift
+++ b/Sources/ComposableArchitecture/Reducer/AnyReducer/AnyReducerSignpost.swift
@@ -149,14 +149,17 @@ extension EffectPublisher where Failure == Never {
 }
 
 @usableFromInline
-func debugCaseOutput(_ value: Any) -> String {
+func debugCaseOutput(
+  _ value: Any,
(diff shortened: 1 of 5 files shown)


Sources/ComposableArchitecture/Reducer/AnyReducer/AnyReducerSignpost.swift
```diff
@@ -173,7 +176,7 @@ func debugCaseOutput(_ value: Any) -> String {
   }
 
   return (value as? CustomDebugStringConvertible)?.debugDescription
-    ?? "\(typeName(type(of: value)))\(debugCaseOutputHelp(value))"
+    ?? "\(abbreviated ? "" : typeName(type(of: value)))\(debugCaseOutputHelp(value))"
```
mbrandonw, line 179: Snuck another change in this PR to shorten the action description when you get a test failure telling you there are actions that were not yet received.

stephencelis (approved):

PR #2231, opened 2023-06-26
Add example for multiple destinations using an `Enum state`

tiagopigatto:
It took me a little bit of time to figure out all the details of navigating on multiple destinations using an `Enum state`. As in [here](https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/treebasednavigation#Enum-state).

I'm not sure if this is covered elsewhere but I figured an example would help other people.


mbrandonw (approved):
Hey @tiagopigatto, sorry for the late response, but we think this is a good addition! I made a few small tweaks and additions, but will merge once the build is green.

PR #2236, opened 2023-06-28
Test store changes

stephencelis:
Right now when asserting via `TestStore.receive`, we show a failure if the expected action doesn't match, and then we also show a failure if the expected state doesn't match. If the action doesn't match, the state assertion shouldn't be expected to match, so we can skip that failure entirely. Showing the state failure is just noise and difficult to parse.

We can also tweak the failure if the expected action _is_ found later in the array of actions.

In the future it would be nice for test stores to fully flush received actions till it finds one that matches, but that will take more work to accomplish.


Sources/ComposableArchitecture/Effect.swift
```diff
@@ -150,8 +150,8 @@ extension EffectPublisher where Failure == Never {
   /// - Returns: An effect wrapping the given asynchronous work.
   public static func run(
     priority: TaskPriority? = nil,
-    operation: @escaping @Sendable (Send<Action>) async throws -> Void,
-    catch handler: (@Sendable (Error, Send<Action>) async -> Void)? = nil,
+    operation: @escaping @Sendable (_ send: Send<Action>) async throws -> Void,
+    catch handler: (@Sendable (_ error: Error, _ send: Send<Action>) async -> Void)? = nil,
```
stephencelis, line 154: Unrelated change, but I noticed that Xcode will autocomplete `{ send in }` if we do this bit of extra work here.

mbrandonw (approved):

PR #2237, opened 2023-06-28
Add `EffectOf`

stephencelis:
Joins `ReducerOf` and `StoreOf` for simplifying some call sites.


stephencelis:
@mbrandonw Decided not to update the docs with this style to give us more time with it, but I think it's probably a non-controversial addition to the library.

mbrandonw (approved):

PR #2241, opened 2023-06-29
Added the ability to specify an animation or transaction with `Store.send`

HarshilShah:
v0.55 deprecates the `WithViewStore` initialiser that takes Void state and recommends using `Store.send` instead, however there isn't parity with the `ViewStore` APIs as those do let you specify an animation or transaction parameter, while Store.send doesn't. This PR adds those methods for full parity

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -820,10 +820,9 @@ public struct StoreTask: Hashable, Sendable {
 		self.rawValue = rawValue
 	}
 	
-	/// Cancels the underlying task and waits for it to finish.
-	public func cancel() async {
+	/// Cancels the underlying task.
+	public func cancel() {
 		self.rawValue?.cancel()
-		await self.finish()
 	}
 	
 	/// Waits for the task to finish.
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -202,7 +202,7 @@ public final class Store<State, Action> {
   ///   - action: An action.
   ///   - animation: An animation.
   @discardableResult
-  public func send(_ action: Action, animation: Animation?) -> ViewStoreTask {
+  public func send(_ action: Action, animation: Animation?) -> StoreTask {
     send(action, transaction: Transaction(animation: animation))
   }
 
@@ -214,7 +214,7 @@ public final class Store<State, Action> {
   ///   - action: An action.
   ///   - transaction: A transaction.
   @discardableResult
-  public func send(_ action: Action, transaction: Transaction) -> ViewStoreTask {
+  public func send(_ action: Action, transaction: Transaction) -> StoreTask {
     withTransaction(transaction) {
       .init(rawValue: self.send(action, originatingFrom: nil))
     }
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -814,7 +814,7 @@ extension ScopedReducer: AnyScopedReducer {
 ///
 /// See ``TestStoreTask`` for the analog returned from ``TestStore``.
 public struct StoreTask: Hashable, Sendable {
-	fileprivate let rawValue: Task<Void, Never>?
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -1,5 +1,6 @@
 import Combine
 import Foundation
+import SwiftUI
```
HarshilShah, line 3: Having to import all of SwiftUI in Store.swift is unfortunate. I'm not quite sure what the house style is so please do let me know if you'd prefer me to import just Animation and Transaction, or move these functions to an extension in a separate file instead

stephencelis:
@HarshilShah Thanks for the PR! We'll discuss when we can, but we are both going to be offline more for the next week or two.

In the meantime I believe you can work around this limitation outside the library by wrapping `store.send` in `withAnimation`, etc.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -193,6 +194,32 @@ public final class Store<State, Action> {
...
 
+  /// Sends an action to the store with a given animation.
+  ///
+  /// See ``Store/send(_:)`` for more info.
+  ///
+  /// - Parameters:
+  ///   - action: An action.
+  ///   - animation: An animation.
+  @discardableResult
+  public func send(_ action: Action, animation: Animation?) -> ViewStoreTask {
```
stephencelis, line 205: The main thing we'll want to discuss is whether or not it's appropriate to reuse `ViewStoreTask` here, or if we want to rename `ViewStoreTask` to `StoreTask` and deprecate the type alias to `ViewStoreTask`. That can probably come after the PR, though.

HarshilShah: Yeah I think especially given the long term prospects for `ViewStore` itself, `StoreTask` might be much better. Happy to go whichever way you folks prefer, and in this PR or a later one, so please do let me know which way you'd like to proceed whenever you're around!

HarshilShah: Hey guys, just checking in, any updates on this one? Would love to get it in

stephencelis: Hey @HarshilShah! We almost certainly want this in, we just haven't had enough time to sync up on it yet with our schedules this month.

I think next steps, if you're up for it, would be to rename `ViewStoreTask` to `StoreTask` and deprecate `ViewStoreTask` as a type alias rename. That should bring the PR very close to being mergable.

HarshilShah: Gotcha, will do! Just gonna push up some commits here in a bit then

HarshilShah: Just pushed up the changes. I made the deprecation message mirror that from `WithViewStore` when set up with stateless stores. I haven't deprecated the `ViewStore.send` methods that return a ViewStoreTask though, should I do that as well?

stephencelis: Thanks!

`ViewStore.send` is still fine to live, undeprecated, for a long time, and is still probably the majority use case over `Store.send`, we just want to return a consistent `StoreTask` type from both of them. The deprecation is just in case folks were referring to `ViewStoreTask` in their code bases, which is unlikely. I've pushed a small commit to address.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -771,3 +798,44 @@ extension ScopedReducer: AnyScopedReducer {
...
+	
+	internal init(rawValue: Task<Void, Never>?) {
+		self.rawValue = rawValue
+	}
+	
+	/// Cancels the underlying task and waits for it to finish.
+	public func cancel() async {
+		self.rawValue?.cancel()
+		await self.finish()
+	}
```
stephencelis, line 827:
```suggestion
	/// Cancels the underlying task.
	public func cancel() {
		self.rawValue?.cancel()
	}
```

Sources/ComposableArchitecture/Store.swift
HarshilShah: Ah gotcha, makes sense!

PR #2243, opened 2023-06-29
Weaken `BindingViewStore`'s dynamic member lookup

stephencelis:
Right now it requires a writably, equatable property, but that prevents accessing `let` constants on state.


mbrandonw (approved):
Good catch

PR #2248, opened 2023-06-30
Fixed BindingAction.set(keyPath:value:) causes equality test on value to fail

bharath2020:


mbrandonw (approved):
Hi @bharath2020, great find! I pushed a few small clean up things, but will merge as soon as tests pass.

PR #2252, opened 2023-06-30
Update code examples of Tree-based navigation documentation

mrtnlst:
# Description
I don't know if contributions just to the documentation are welcome, but I think I spotted a few instances of outdated or incorrect sample code in Tree-based navigations.

1. The store scope misses the `$` e.g:
```diff
- store: self.store.scope(state: \.destination, action: { .destination($0) }),
+ store: self.store.scope(state: \.$destination, action: { .destination($0) }),
```
2. The Destination model is missing when isolating the specific state and action, e.g:
```diff
- state: /InventoryFeature.State.addItem,
+ state: /InventoryFeature.Destination.State.addItem,
```
3. The forward slash can be removed when isolating to action
```diff
- action: /InventoryFeature.Action.addItem
+ action: InventoryFeature.Destination.Action.addItem
```

I got stuck here a little bit, while following along your excellent documentation, so updating it might help others as well ✌🏻

### DIFF
--- a/Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
@@ -265,23 +265,23 @@ struct InventoryView: View {
       // ...
     }
     .sheet(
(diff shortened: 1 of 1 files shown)


stephencelis (approved):
Great catch, thanks!

PR #2256, opened 2023-07-04
Fix WithViewStore argument label in Bindings.md

jessetipton:
This adds a missing argument label to documentation sample code.


stephencelis (approved):
Thanks!

PR #2261, opened 2023-07-05
Fix ViewStore.init that should be ViewState.init in Bindings.md

jessetipton:
Another documentation sample code fix.

Not sure how I missed this in #2256, but the `observe` parameter should be `ViewState.init`–not `ViewStore.init`.


stephencelis (approved):
Thanks!

PR #2267, opened 2023-07-06
Fix commas that should be periods in docs

jessetipton:
This fixes a recurring punctuation error I came across in documentation


stephencelis (approved):
Thanks :smile:

PR #2277, opened 2023-07-09
Add TestStoreOf typealias

brzzdev:
I kept finding myself wishing for a convenience initialiser for `TestStore` when I'm not scoping or using an environment, so I don't have to type out/copy and paste the rather verbose `TestStore<Reducer.State, Reducer.Action, Reducer.State, Reducer.Action, Void>`.

After adding it to my own project and liking it, I thought I may as well try my first PR here.

If you feel it's unnecessary, then no worries.

### DIFF
--- a/Sources/ComposableArchitecture/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestStore.swift
@@ -990,6 +990,21 @@ public final class TestStore<State, Action, ScopedState, ScopedAction, Environme
   }
 }
 
+/// A convenience type alias for referring to a test store of a given reducer's domain.
+///
+/// Instead of specifying five generics:
+///
+/// ```swift
+/// let testStore: TestStore<Feature.State, Feature.Action, Feature.State, Feature.Action, Void>
+/// ```
+///
+/// You can specify a single generic:
+///
+/// ```swift
+/// let testStore: TestStoreOf<Feature>
+/// ```
+public typealias TestStoreOf<R: ReducerProtocol> = TestStore<R.State, R.Action, R.State, R.Action, Void>
+
 extension TestStore where ScopedState: Equatable {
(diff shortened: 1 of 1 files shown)


mariano-mix:
Have the same thing in our project and it's consistent to other convenience `typealiases` the library ships with. Hope this gets merged.

stephencelis:
@brzzdev Thanks for the PR! Brandon and I are somewhat offline the next few days but will look at this soon. It definitely looks like a good idea but the main thing for us to think about is how `TestStore` has a ton of extra generics for testing view state/actions, but we're not sure this feature is even widely used. We'll discuss some things and get back to you soon!

lukeredpath:
Speaking of tyoealiases I recently added a computed property to my view for scoping a store down to its presentation state/action for passing to multiple presentation modifiers and found myself adding a PresentationStoreOf alias too, which removed the verbosity of the nested presentation state/action generics.

stephencelis:
Can anyone share some examples of _why_ they need to specify `TestStore` generics? In our experience the types can always be inferred, so it's not clear to us when you would need to use `TestStore<_, _, _, _, _>.init` (or `TestStoreOf<_>.init`) vs. a simple `TestStore.init`.

brzzdev:
I like to use `TestStore` as an implicitly unwrapped optional at the top level of the `XCTestCase`
`var store: TestStoreOf<Reducer>!`
then explicitly create it in `override func setUp()` and set it as nil in `override func tearDown()`

That way, I at least always have a standard version of my `TestStore` ready for each test case, I can always recreate it with different dependencies in specific functions if I need to, and it's always reset between tests.

stephencelis (approved):
Looks good to us! Given #2291, we'll likely lose most of the extra generics in 1.0, so this is a fine feature to have for that vision!

PR #2281, opened 2023-07-12
Add TaskResult.cancellable to allow for cancellation errors to not be wrapped as part of the result

klundberg:
In the app I'm working on at work presently, I've found that normal TaskResults aren't as ideal when used in a context where tasks can be cancelled, as those cancellation errors could be processed by the reducer at a point where I don't want them to be (like after nilling out some optional state that an ifLet reducer uses). I also would prefer to not need to filter out cancellation errors from my reducers if a taskresult that can be cancelled would have that cancellation error fed into a reducer action. 

For tasks that I know can be cancelled and that I want to keep free of cancellation errors, I've added this static function in my own codebase which has helped. Usage-wise it looks like this:
```swift
case .someAction:
  return .run { send in
    try await send(.responseAction(TaskResult.cancellable { try await someOperation() }))
  }
}
```

Cancellation errors are thrown so that they could be handled and ignored by EffectTask, while non-cancellation errors are still wrapped in the TaskResult to be processed by the reducer.

I figured this would be valuable to other folks using TaskResult, hence this PR. I would argue that this behavior should instead go in `TaskResult.init(catching:)` but I don't want to introduce a source-incompatible change that would require lots of people adding `try` everywhere, especially if they may not want this behavior yet.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/TaskResult.swift
+++ b/Sources/ComposableArchitecture/Effects/TaskResult.swift
@@ -124,6 +124,21 @@ public enum TaskResult<Success: Sendable>: Sendable {
       self = .failure(error)
     }
   }
+  
+  /// Throwing variant of `init(catching:)` that only throws `CancellationError`s if the task this is run within is cancelled.
+  /// This is useful if you don't want cancellation errors to be processed by a reducer like other errors.
+  ///
+  /// - Parameter body: An async, throwing closure.
+  @inlinable
+  public static func cancellable(catching body: @Sendable () async throws -> Success) async throws -> Self {
+    do {
+      return .success(try await body())
+    } catch let error as CancellationError {
+      throw error
+    } catch {
+      return .failure(error)
+    }
+  }
 
   /// Transforms a `Result` into a `TaskResult`, erasing its `Failure` to `Error`.
   ///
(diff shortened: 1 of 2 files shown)


stephencelis:
Hey @klundberg! Thanks for taking the time to PR. We actually considered this as the default functionality in this PR:

https://github.com/pointfreeco/swift-composable-architecture/pull/1934

The discussion there may be illuminating, but we came to the conclusion that `TaskResult` should be sugar for `do { … } catch { … }` in the same way `Result` is, and that swallowing cancellation errors should be at the `Effect.run` level only. In general, when you have an effect that expects to feed data into the system, you likely have intermediate state, like `isLoading`, that should be cleaned up, and if you swallow cancellation then you have no ability to do so.

I definitely think your helper is useful if you are careful and know what you're doing, but I think we probably don't want to ship it directly in the library, as it may lead to folks accidentally introducing bugs into their applications in some cases. What do you think?

klundberg:
I'm all for caution here, and if you think it's not appropriate as a built-in library feature due to potential misuse then I'm ok with not including it. Thanks for linking that PR as well, I hadnt seen it and the discussion was good to read, and some of the examples i hadnt thought of as well regarding possible bugs.

I don't recall if this is how things work today or not, but if automatic cancellation with the new navigation features avoids the issue of sending cancellation errors into the reducer when its state is cleared, then that's one fewer thing to need to worry about solving with a cancellation-aware task result.

I'll close the PR then and keep an eye out for any future discussion on this topic :)

PR #2283, opened 2023-07-13
Fix cancellation ID existentials in iOS 14

### LINKED ISSUE #2233: DismissEffect does not work on iOS14
### Description

DismissEffect does not work on iOS14.

Nothing happens when I run the code below on iOS14.

```swift
struct CoreReducer: ReducerProtocol {
    @Dependency(\.dismiss) private var dismiss

        switch action {
        ...
            return .fireAndForget {
                // Nothing happens.
                await dismiss()
            }
        }
    }
}
```

### Checklist

### Expected behavior

Screen closes after executing DismissEffect.

### Actual behavior

Screen does not close after executing DismissEffect.

### Steps to reproduce

1. Transition to any screen on iOS14
2. Execute DismissEffect on the transition destination screen

0.54.1

iOS 14.5

### Xcode version information

14.3

```shell
swift-driver version: 1.75.2 Apple Swift version 5.8 (swiftlang-5.8.0.124.2 clang-1403.0.22.11.100)
```

stephencelis:
Fixes #2233.

On iOS 14, `DismissEffect` is broken because effect cancellation uses `AnyHashable` throughout, which can lead to nested `AnyHashable(AnyHashable(...))` IDs, which can hash differently depending on the level of nesting. By preserving the hashable ID up to being stored in the internal cancellation token type, we can avoid a hash lookup failure that prevents a dismissal from happening.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -28,7 +28,7 @@ extension EffectPublisher {
   ///   - cancelInFlight: Determines if any in-flight effect with the same identifier should be
   ///     canceled before starting this new one.
   /// - Returns: A new effect that is capable of being canceled by an identifier.
-  public func cancellable(id: AnyHashable, cancelInFlight: Bool = false) -> Self {
+  public func cancellable<ID: Hashable & Sendable>(id: ID, cancelInFlight: Bool = false) -> Self {
     @Dependency(\.navigationIDPath) var navigationIDPath
 
     switch self.operation {
@@ -96,7 +96,7 @@ extension EffectPublisher {
   /// - Parameter id: An effect identifier.
   /// - Returns: A new effect that will cancel any currently in-flight effect with the given
   ///   identifier.
-  public static func cancel(id: AnyHashable) -> Self {
+  public static func cancel<ID: Hashable & Sendable>(id: ID) -> Self {
     @Dependency(\.navigationIDPath) var navigationIDPath
 
     return .fireAndForget {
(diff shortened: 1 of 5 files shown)


Sources/ComposableArchitecture/Effects/Cancellation.swift
```diff
@@ -182,7 +182,7 @@ extension Task where Success == Never, Failure == Never {
   /// Cancel any currently in-flight operation with the given identifier.
   ///
   /// - Parameter id: An identifier.
-  public static func cancel<ID: Hashable & Sendable>(id: ID) {
+  public static func cancel<ID: Hashable>(id: ID) {
```
stephencelis, line 185: This change is mainly to be consistent. I also tried to mark _everything_ `Hashable & Sendable` but it bled out and introduced warnings to existing code, so I held off on that for now (we could explore it again in another PR).

Sources/ComposableArchitecture/Reducer/Reducers/PresentationReducer.swift
```diff
@@ -477,8 +477,8 @@ extension Task where Success == Never, Failure == Never {
   }
 }
 extension EffectPublisher {
-  internal func _cancellable(
-    id: AnyHashable = _PresentedID(),
+  internal func _cancellable<ID: Hashable>(
+    id: ID = _PresentedID(),
```
stephencelis, line 481: I think this is safe for the versions of Swift we support, but we could also create overloads...

sadiklar:
13 Tem 2023 Per 19:19 tarihinde Stephen Celis ***@***.***>
şunu yazdı:

mbrandonw (approved):

PR #2284, opened 2023-07-13
Fix macOS dismissal

### LINKED ISSUE #2182: Tree-based navigation: dismiss action is not sent when navigating back on macOS (prerelease/1.0)

stephencelis:
Fixes #2182.

The `id` used for comparison appears to sometimes be `nil`, for example on macOS when drilling out of a `navigationDestination`. This PR adds a little nuance to the dismissal logic of `presentationModifier`:

  * It explicitly checks that something is presented
  * It always sends `dismiss` if the cached `id` was `nil`

We should maybe have some integration tests for macOS for this...

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/PresentationModifier.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/PresentationModifier.swift
@@ -221,7 +221,12 @@ public struct PresentationStore<
             : nil
         },
         compactSend: {
-          $0 == nil && self.toID(self.viewStore.state) == id ? .dismiss : nil
+          guard
+            $0 == nil,
+            self.viewStore.wrappedValue != nil,
+            id == nil || self.toID(self.viewStore.state) == id
+          else { return nil }
+          return .dismiss
         }
       ),
       DestinationContent(


sadiklar:
13 Tem 2023 Per 05:11 tarihinde Stephen Celis ***@***.***>
şunu yazdı:

mbrandonw (approved):

PR #2289, opened 2023-07-14
Fix for infinite loop when combining Stack and Tree navigation

### LINKED ISSUE #2288: Combining Stack-based and Tree-based .navigationDestination navigation leads to infinite loop
### Description

Using `NavigationStackStore`, with a destination view then presenting a third view using `.navigationDestination(store:destination)`. When the third view is presented, SwiftUI goes into an infinite loop, consuming 100% CPU and increasing amount of memory.

### Checklist

- [X] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.

### Expected behavior

The third view is pushed onto the navigation stack

### Actual behavior

The app hangs with 100% CPU usage

### Steps to reproduce

I've made a [sample project](https://github.com/benlings/TCANavigationStackStoreBugSample) that demonstrates the bug, along with demonstrations of the same functionality working correctly in plain SwiftUI. The TCA version is shown below.

In the sample project there are 3 tabs:
1. Vanilla SwiftUI - using an array-based `NavigationStack` path
2. VanillaStack - TCA, using an array-based `NavigationStack` path. Note: I had to extract `NavigationDestinationView` to stop this happening in this implementation
3. TCA - TCA, using `NavigationStackStore`

```
struct ContentFeature: ReducerProtocol {
    struct State {
      var path = StackState<Path.State>()
    }
    public enum Action {
      case path(StackAction<Path.State, Path.Action>)
    }

      Reduce { state, action in
        switch action {
        case .path:
          return .none
        }
      }
      .forEach(\.path, action: /Action.path) {
        Path()
      }
    }

    public struct Path: ReducerProtocol {
      public enum State {
        case int(IntFeature.State)
      }
      public enum Action {
        case int(IntFeature.Action)
      }
      public var body: some ReducerProtocolOf<Self> {
        Scope(state: /State.int, action: /Action.int) {
          IntFeature()
        }
      }
    }

  }

  struct ContentView: View {

    let store: StoreOf<ContentFeature>

    var body: some View {
      NavigationStackStore(store.scope(state: \.path, action: { .path($0) })) {
        VStack {
          NavigationLink("Nav link to 1", state: ContentFeature.Path.State.int(.init(num: 1)))
        }
        .padding()
      } destination: { store in
        DestinationSwitchView(store: store)
      }
    }
  }

  struct DestinationSwitchView: View {
    let store: StoreOf<ContentFeature.Path>
    var body: some View {
      SwitchStore(store) { state in
        switch state {
        case .int:
          CaseLet(
            state: /ContentFeature.Path.State.int,
            action: ContentFeature.Path.Action.int
          ) {
            IntView(store: $0)
          }
        }
      }
    }
  }

  struct IntFeature: ReducerProtocol {
    struct State: Equatable {
      var num: Int
      @PresentationState var dest: DestFeature.State? = nil
    }
    public enum Action {
      case showDest
      case dest(PresentationAction<DestFeature.Action>)
    }
      Reduce { state, action in
        switch action {
        case .showDest:
          state.dest = .init()
          return .none

        case .dest:
          return .none
        }
      }
      .ifLet(\.$dest, action: /Action.dest) {
        DestFeature()
      }
    }
  }

  struct IntView: View {
    let store: StoreOf<IntFeature>

    var body: some View {
      WithViewStore(store, observe: { $0 }) { viewStore in
        VStack {
          Text("Number \(viewStore.num)")
          Button("Show 2") {
            viewStore.send(.showDest)
          }
        }
        .navigationDestination(
          store: store.scope(state: \.$dest, action: { .dest($0) })
        ) {
          DestView(store: $0)
        }
      }
    }
  }

  struct DestFeature: ReducerProtocol {
    struct State: Equatable {
    }
    enum Action {
    }
      Reduce { state, action in
        switch action {
        }
      }
    }
  }

  struct DestView: View {
    let store: StoreOf<DestFeature>

    var body: some View {
      Text("Destination 2")
    }
  }
```

0.55.1

iOS 16.4

### Xcode version information

Version 14.3.1 (14E300c)

```shell
swift-driver version: 1.75.2 Apple Swift version 5.8.1 (swiftlang-5.8.0.124.5 clang-1403.0.22.11.100)
Target: arm64-apple-macosx13.0
```

benlings:
This fixes #2288

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/NavigationStackStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/NavigationStackStore.swift
@@ -219,9 +219,9 @@ private struct NavigationDestinationView<State, Destination: View>: View {
   let component: Component<State>
   let destination: (Component<State>) -> Destination
   var body: some View {
-    self.destination(component)
+    self.destination(self.component)
       .environment(\.navigationDestinationType, State.self)
-      .id(component.id)
+      .id(self.component.id)
   }
 }
 
--- a/Sources/ComposableArchitecture/SwiftUI/NavigationStackStore.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/NavigationStackStore.swift
@@ -105,9 +105,7 @@ public struct NavigationStackStore<State, Action, Root: View, Destination: View>
       self.root
         .environment(\.navigationDestinationType, State.self)
         .navigationDestination(for: Component<State>.self) { component in
-          self.destination(component)
-            .environment(\.navigationDestinationType, State.self)
-            .id(component.id)
+          NavigationDestinationView(component: component, destination: self.destination)
         }
     }
   }
@@ -214,6 +212,16 @@ extension NavigationLink where Destination == Never {
   )
   where Label == _NavigationLinkStoreContent<P, Text> {
     self.init(state: state, label: { Text(title) }, fileID: fileID, line: line)
+  }
+}
+
+private struct NavigationDestinationView<State, Destination: View>: View {
+  let component: Component<State>
+  let destination: (Component<State>) -> Destination
+  var body: some View {
+    self.destination(component)
+      .environment(\.navigationDestinationType, State.self)
+      .id(component.id)
   }
 }
 
--- a/Examples/Integration/Integration/NavigationStackTestCase.swift
+++ b/Examples/Integration/Integration/NavigationStackTestCase.swift
@@ -1,18 +1,8 @@
 import ComposableArchitecture
 import SwiftUI
 
-private struct DestinationFeature: ReducerProtocol {
-  struct State: Hashable {
-  }
-  enum Action {
-  }
-  var body: some ReducerProtocolOf<Self> {
-    EmptyReducer()
-  }
-}
-
 private struct DestinationView: View {
-  let store: StoreOf<DestinationFeature>
+  let store: StoreOf<EmptyReducer<Int, Never>>
   var body: some View {
     Text("Destination")
   }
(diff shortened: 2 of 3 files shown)


benlings:
I'm happy to add some tests for this, but would need some pointers on where they should go. Maybe extending `NavigationStackTestCaseView`?

mbrandonw:
Hi @benlings, thanks for the PR! This looks like a great fix, and we'd definitely like to take you up on your offer to write a test. We actually have a slim (but growing) integration test suite in the project:

https://github.com/pointfreeco/swift-composable-architecture/tree/main/Examples/Integration/Integration

How do you feel about adding a new "…TestCase.swift" file to the project that demonstrates the problem? You can even just get it most of the way and then we can do any last bits of clean up.

Thanks!

Sources/ComposableArchitecture/SwiftUI/NavigationStackStore.swift
```diff
@@ -217,6 +215,16 @@ extension NavigationLink where Destination == Never {
   }
 }
 
+private struct NavigationDestinationView<State, Destination: View>: View {
+  let component: Component<State>
+  let destination: (Component<State>) -> Destination
+  var body: some View {
+    self.destination(component)
+      .environment(\.navigationDestinationType, State.self)
+      .id(component.id)
```
benlings, line 224: Looking at this again, I'm not sure if this `id` modifier should go here or inside the `destination` block of `navigationDestination(...)`? What is this `id` used for @mbrandonw ?

benlings:
Thanks for the pointers @mbrandonw .  I looked at creating a new test case, but it would have duplicated most of the existing 'NavigationStackTestCase` for the initial setup. I've instead added the `navigationDestination` alongside the existing alert presentation. I'm happy to separate it if you prefer through 🙂.

mbrandonw:
Hi @benlings, thanks for taking a stab at the test. However, if I revert your `NavigationDestinationView` change and run the integration app, the drill down works fine. I open the nav stack sheet, then tap "Go to counter", then tap "Open navigation destination", and it drills down. No infinite loop.

Is there something else that needs to be done to trigger the bug?

stephencelis:
@mbrandonw I think the bug is fixed in iOS 17, but if you try in Xcode 14.3/iOS 16, it should happen.

mbrandonw:
Ahhh right ok. Will take another look in a minute.

mbrandonw:
Ok, I was able to confirm the bug in 14.3 and confirm that the test passes with the change. Will merge once it's green!

benlings:
Thanks @mbrandonw and @stephencelis! I hadn’t tested this on iOS 17, so wasn’t aware of the fix there. I had seen the bug on 16.2-16.4.

PR #2310, opened 2023-07-19
Remove TaskCancellableValue due to ConcurrencyExtras addition

Ryu0118:
## Overview
Since `cancelableValue` exists in `ConcurrencyExtras`, I removed TaskCancellableValue.swift and added import ConcurrencyExtras instead.


Ryu0118:
Is it better to use `@_exported import` instead?

mbrandonw:
Good catch @Ryu0118! Yeah I think doing `@_exported import` is the better approach, and is just an oversight on our part. Want to make that change?

Ryu0118:
Yes, I'll make the change.

PR #2312, opened 2023-07-19
StoreTask.cancel is not an asynchronous method

Ryu0118:
StoreTask cancel has been changed to synchronous in #2282, so await is not necessary.


mbrandonw (approved):
Good catch, thanks!

PR #2313, opened 2023-07-19
`TestStore.init(initialState:reducer:prepareDependencies:file:line:)` is deprecated

Ryu0118:
`TestStore.init(initialState:reducer:prepareDependencies:file:line:)` is deprecated, so I rewrote it to `TestStore.init(initialState:reducer:withDependencies:file:line:)`.


mbrandonw (approved):
Another good catch, thanks :)

PR #2318, opened 2023-07-24
The Composable Architecture 1.0

stephencelis:
1.0 is almost here!

This PR stages a final bit of work off the `prerelease/1.0` branch that we introduced as a preview [almost 6 months ago](https://github.com/pointfreeco/swift-composable-architecture/discussions/1905).

This work takes things one step further by fully removing the deprecated APIs that have remained in the library for a long time for backwards compatibility. We're taking 1.0 as an opportunity to shed this surface area, which should hopefully result in a leaner binary with better compilation times (especially since many of the deprecated APIs are overloads that could slow down the compiler).

If you've been targeting `prerelease/1.0`, try pointing things to `release/1.0` to make sure you're in good shape for the final release!


Alex293:
Congrats on this huge milestone ! Already eager for 2.0 with Observable shenanigans ;)

freak4pc:
Exciting times!! On my small-medium project the only changes I had to do was this:

```diff
Store(
    initialState: Tasks.State(),
-   reducer: Tasks()
+   reducer: { Tasks() }
)
```

Gotta say I'm not entirely sure why, I was sure it should work without the closure 🤔

stephencelis:
@freak4pc The non-builder initializer was soft-deprecated awhile ago, but we just added hard deprecations recently. It should also read nicely as a trailing closure:

```swift
Store(initialState: Tasks.State()) {
  Tasks()
}
```

ToineHulshof (approved):

akashsoni01:
Congrats on the huge milestone.




Just a general question are we still depending on Apple's combine framework in 1.0 ?

Iikeli:
Congrats on the milestone! Looking forward to updating to 1.0 when it is fully available!

mbrandonw:
> Just a general question are we still depending on Apple's combine framework in 1.0 ?

Hi @akashsoni01, yes the library still depends on Combine because it is still largely a SwiftUI framework and Combine is necessary for SwiftUI. We do have plans for future work to disentangle the library from Combine, but that won't be done for 1.0.

PR #2319, opened 2023-07-24
Typo in withTaskCancellation debounce documentation

brzzdev:
Noticed when trying `release/1.0` and moving from the old `EffectTask` debounce to the `.run` debounce that the comment docs don't match reality

Missing a try on the `withTaskCancellation` call, and a dot on the `send` call that shouldn't be there

Pointed this at main because it's there too, and I assume changes will be merged into the 1.0 branch


stephencelis (approved):
Thanks!

PR #2323, opened 2023-07-26
Add Store.send(_ action:) return StoreTask

hj56775:
I think the addition of Store.send is very innovative because it makes it clear that the View does not need to observe changes in State.
However, I noticed that Store.send is not returning a StoreTask, unlike ViewStore.send.

I want Store.send to also use `.task { await store.send(.task).finish() }`, just like ViewStore.

Is there any reason why Store.send isn't implemented that way?
If so, you can ignore this PR.

Thanks.


stephencelis:
Thanks!

PR #2330, opened 2023-07-28
Expose `StorePublisher` directly on `Store`

stephencelis:
Now that `ViewStore`s have a limited future, we should surface functionality that was `ViewStore`-only to the `Store` instead.

While we could probably start discouraging using `ViewStore` at all in UIKit apps by deprecating the existing API, it's pretty close to the 11th hour of shipping 1.0, and we are probably better off deprecating `ViewStore` all at once when the Observation framework is released.

### DIFF
--- a/Sources/ComposableArchitecture/Documentation.docc/Extensions/UIKit.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Extensions/UIKit.md
@@ -14,4 +14,5 @@ While the Composable Architecture was designed with SwiftUI in mind, it comes wi
 
 ### Subscribing to state changes
 
+- ``Store/publisher``
 - ``ViewStore/publisher``
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -673,6 +673,19 @@ public final class Store<State, Action> {
     #endif
     self.threadCheck(status: .`init`)
   }
+
+  /// A publisher that emits when state changes.
+  ///
+  /// This publisher supports dynamic member lookup so that you can pluck out a specific field in
+  /// the state:
+  ///
+  /// ```swift
+  /// store.publisher.alert
+  ///   .sink { ... }
+  /// ```
+  public var publisher: StorePublisher<State> {
+    StorePublisher(store: self, upstream: self.state)
+  }
 }
 
 /// A convenience type alias for referring to a store of a given reducer's domain.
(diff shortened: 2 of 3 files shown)


mbrandonw (approved):

acosmicflamingo:
I suppose what this would mean going forward in UIKit-land is that transformations from reducer's state to the view domain via `ViewState` would now have to be done using Combine's map function:
```diff
-  public struct ViewState: Hashable {
-    public var item: WebsiteBrowserItem
-    public var stage: WebsiteBrowser.State.NavigationDelegateStatus
-    public var loadedWebsite: Bool
-
-    init(state: WebsiteBrowser.State) {
-      self.item = state.item
-      self.stage = state.stage
-      self.loadedWebsite = state.stage == .didFinish
-    }
-  }
-
-  viewStore.publisher.loadedWebsite
+  store.publisher.stage
+   .map { $0 == .didFinish }
    .removeDuplicates(by: ==)
    .sink { /* */ }
    .store(in: &cancellables)
```

Is that correct?

stephencelis:
@acosmicflamingo That's one option for sure!

You could also continue to have a store with view-specific state/actions, there's just no need to use `ViewStore`. You can simply `scope` the store instead.

acosmicflamingo:
@stephencelis ohhhhhhhhhh I see. Thanks! Once this is merged to release/1.0, I'll give it a try and see how it goes with your suggestion :)

acosmicflamingo:
@stephencelis WOW so the scoping solution is perfect, if I were to assume that it would be done like this:
```diff
class FooViewController: UIViewController {
  let store: StoreOf<Foo>
- var viewStore: Store<ViewState, Foo.Action>

  /*  blah blah blah */

  func setupBindings() {
+   let viewStore = store.scope(state: ViewState.init, action: { $0 })
     viewStore.isEditing.sink {
       print($0)
     }
     .store(in: &cancellables)
  }
}
```

...and not like this:
```diff
class FooViewController: UIViewController {
-  let store: StoreOf<Foo>
-  var viewStore: Store<ViewState, Foo.Action>
+ let store: Store<ViewState, Foo.Action> 
  /*  blah blah blah */

  func setupBindings() {
-   viewStore.isEditing.sink {
+   store.isEditing.sink {
       print($0)
     }
     .store(in: &cancellables)
  }
}
```

acosmicflamingo:
@stephencelis after migrating my UIKit codebase to exclusively using `Store` and it's newly exposed publisher, I noticed some subtle animation issues were popping up. After looking a bit into it, I realized that when I transitioned to using StorePublisher instead of ViewStorePublisher, I had lost the automatic `removeDuplicates` functionality that is found in the `ViewStore` convenience initializer: https://github.com/pointfreeco/swift-composable-architecture/blob/main/Sources/ComposableArchitecture/ViewStore.swift#L619.

Looking through `Store.swift`, it appears that the `removeDuplicates` parameter is not exposed to the public: https://github.com/pointfreeco/swift-composable-architecture/blob/main/Sources/ComposableArchitecture/Store.swift#L373. Should this parameter now be exposed as well? I'm fine with slapping on `removeDuplicates(by: ==)` everywhere, but wanted you to be aware of it in case you do think that `Store` should now allow for receiving a removeDuplicates closure.

stephencelis:
@acosmicflamingo Interesting catch, thanks for sharing! I'm not 100% sure if we should handle this any differently, though. View stores are built around minimizing updates, but stores currently are not. This may change with observability, but for now I wonder if one should be explicit when using store publishers:

```diff
-store.publisher
+store.publisher.removeDuplicates()
```

Thoughts?

acosmicflamingo:
I agree with you. It does seem weird to now add this removing duplication functionality to `Store`. I also think it'd be silly to create a property like `publisherWithNoDuplicates` just to append `StorePublisher` with `removeDuplicates`. I imagine there comes a point where it just becomes over-engineering it.

I've already sprinkled `removeDuplicates` everywhere and like that I can explicitly see what is happening with the stream of events, so perhaps when you all finally decide to create the RFC: ViewStore deprecation PR, that it will need to be documented that you need to now slap on `removeDuplicates` after `store.publisher` to get the same behavior as before.

PR #2336, opened 2023-07-30
Bump version in docs

stephencelis:
Updates docs to point to 1.0.


mbrandonw (approved):
We can also update the poster+URL in the readme to the new tour if you want to knock that out now.

PR #2338, opened 2023-07-31
Adds didChange parameter to onChange reducer modifier

ohitsdaniel:
Hey TCA peepz, 👋 

We have been using our own custom version the onChange reducer for a while now and are really happy with it. We tried to get rid of our custom impl, but one tiny thing kept us from doing it: not being able to pass a custom `didChange` function to `.onChange`.

This pull request adapts the `_OnChangeReducer` to accept a `didChange: (V, V) -> Bool` closure. As laid out in the test, this is especially useful when observing child state change in a parent domain where you want to define the changes in the child domain the parent domain is interested in. It reduces the performance impact of these child state observations as the full equality check can be replaced by a simplified equality check. See the added test case where we test for ID set equality and only trigger the onChange when a new child state is added.

Looking forward to some feedback, especially on the documentation / comment side. Congrats on releasing 1.0 🙌


Sources/ComposableArchitecture/Reducer/Reducers/OnChange.swift
```diff
@@ -42,13 +96,13 @@ extension Reducer {
   ///   - reducer: A reducer builder closure to run when the value changes.
   ///   - oldValue: The old value that failed the comparison check.
   ///   - newValue: The new value that failed the comparison check.
-  /// - Returns: A reducer that performs the
```
ohitsdaniel, line 45: This looked like an unfinished sentence. What did you want to say here? 😁  @stephencelis

Sources/ComposableArchitecture/Reducer/Reducers/OnChange.swift
```diff
@@ -1,4 +1,58 @@
...
+  ///   - toValue: A closure that returns a value from the given state.
+  ///   - didChange: A closue that indicates that the value has changed.
+  ///   - reducer: A reducer builder closure to run when the value changes.
+  ///   - oldValue: The old value that failed the comparison check.
+  ///   - newValue: The new value that failed the comparison check.
+  /// - Returns: A reducer that performs
+  @inlinable
+  public func onChange<V, R: Reducer>(
+    of toValue: @escaping (State) -> V,
+    didChange: @escaping (V, V) -> Bool,
```
mbrandonw, line 50: The `didChange` naming makes it sounds like an action closure to execute when something does change. I don't know if there is a precedent for naming this kind of thing in SwiftUI, but at least Combine as `removeDuplicates`. Would you mind updating to that until we figure out something better?

Sources/ComposableArchitecture/Reducer/Reducers/OnChange.swift
mbrandonw: Not sure what @stephencelis was thinking in the moment, but something like "A reducer that performs the logic when the state changes." might work.

mbrandonw:
Hi @ohitsdaniel, overall this looks good! Thanks for taking the time to open it. We just have a few small comments.

Sources/ComposableArchitecture/Reducer/Reducers/OnChange.swift
ohitsdaniel: What's your take on `isEqual`?

mbrandonw: Stephen and I will discuss, but in some sense duplicates are being removed. Duplicate invocations of the `reducer` closure are "removed". In general we prefer to not create new naming conventions if it can be helped, and I'm not sure there is a precedent for `isEqual` in Swift or Apple's frameworks.

ohitsdaniel: `isEqual` is definitely oldschool Obj-C iOS 2.0+ naming. 😄 

Renamed to `removeDuplicates predicate: @escaping (Value, Value) -> Bool`. 

References:

```swift
func removeDuplicates(by predicate: @escaping (Self.Output, Self.Output) -> Bool
```
[RemoveDuplicates](https://developer.apple.com/documentation/combine/publishers/removeduplicates)<Self>

```swift
func isEqual(_ object: Any?) -> Bool
```
[isEqual](https://developer.apple.com/documentation/objectivec/nsobjectprotocol/1418795-isequal)

mbrandonw (approved):

mbrandonw (approved):
Thanks for this @ohitsdaniel! I'm going to merge and make a few small tweaks on main.

PR #2350, opened 2023-08-04
Documentation update in DismissEffect example

Atimca:
Hi ✌️,

Firstly, a big thank you for your remarkable work on the Composable Architecture 😀. Recently, while using a new navigation API, we referenced the `DismissEffect` documentation for a child feature dismissal test. In doing so, we spotted a small area for improvement. This PR aims to correct it for better reflecting the testing context.

Thank you for your time and consideration.

### DIFF
--- a/Sources/ComposableArchitecture/Dependencies/Dismiss.swift
+++ b/Sources/ComposableArchitecture/Dependencies/Dismiss.swift
@@ -25,7 +25,7 @@ extension DependencyValues {
 ///   func reduce(into state: inout State, action: Action) -> Effect<Action> {
 ///     switch action {
 ///     case .exitButtonTapped:
-///       return .fireAndForget { await self.dismiss() }
+///       return .run { _ in await self.dismiss() }
 ///     // ...
 ///     }
 ///   }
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Dependencies/Dismiss.swift
```diff
@@ -25,7 +25,7 @@ extension DependencyValues {
 ///   func reduce(into state: inout State, action: Action) -> Effect<Action> {
 ///     switch action {
 ///     case .exitButtonTapped:
-///       return .fireAndForget { await self.dismiss() }
+///       return .run { _ in await self.dismiss() }
```
Atimca, line 28: Here, I've replaced `fireAndForget`, which has been deprecated, with `run(priority:operation:catch:fileID:line:)` to better reflect the current state of TCA.

Sources/ComposableArchitecture/Dependencies/Dismiss.swift
```diff
@@ -60,17 +60,17 @@ extension DependencyValues {
 /// > some shared mutable state inside the `dismiss` closure to confirm that it is indeed invoked:
 /// >
 /// > ```swift
-/// > let isDismissInvoked = LockIsolated(false)
-/// > let store = Store(initialState: Child.State()) {
+/// > let isDismissInvoked: LockIsolated<[Bool]> = .init([])
```
Atimca, line 63: I've replaced `LockIsolated<Bool>` with its array-based counterpart, `LockIsolated<[Bool]>`. This isn't a correction of a typo, but rather a small improvement for reliability. It ensures that dismiss is not called more than once.

PR #2352, opened 2023-08-04
Improve TestStore failures on init.

mbrandonw:
Right now if a test failure occurs while creating a reducer/state (e.g. accessing a dependency that hasn't been overridden), the test failure doesn't show up in the best spot. This uses the new `XCTFailContext` from xctest-dynamic-overlay to improve the situation.

### DIFF
--- a/Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
@@ -363,7 +363,7 @@ case .destination(.presented(.editItem(.saveButtonTapped))):
   else { return .none }
 
   state.destination = nil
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
```diff
@@ -363,7 +363,7 @@ case .destination(.presented(.editItem(.saveButtonTapped))):
   else { return .none }
 
   state.destination = nil
-  return .fireAndForget {
+  return .run { _ in
     self.database.save(editItemState.item)
```
mbrandonw, line 367: unrelated changes but updating some docs to use non-deprecated APIs.

stephencelis (approved):

PR #2367, opened 2023-08-11
Remove unnecessary parameter of NavigationStackStore in StackBasedNavigation.md

Ryu0118:
I found a mistake in StackBasedNavigation.md and fixed it.


mbrandonw (approved):
Good find thanks!

PR #2368, opened 2023-08-12
Bring back throttle

mbrandonw:
We may have too eagerly removed this operator from 1.0 so let's bring it back.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Throttle.swift
+++ b/Sources/ComposableArchitecture/Effects/Throttle.swift
@@ -29,7 +29,6 @@ extension Effect {
     scheduler: S,
     latest: Bool
   ) -> Self {
-    print(id)
     switch self.operation {
     case .none:
       return .none
--- a/Sources/ComposableArchitecture/Effects/Throttle.swift
+++ b/Sources/ComposableArchitecture/Effects/Throttle.swift
@@ -5,6 +5,15 @@ import Foundation
 extension Effect {
   /// Throttles an effect so that it only publishes one output per given interval.
   ///
+  /// The throttling of an effect is with respect to actions being sent into the store. So, if
+  /// you return a throttled effect from an action that is sent with high frequency, the effect
+  /// will be executed at most once per interval specified.
+  ///
+  /// > Note: It is usually better to perform throttling logic in the _view_ in order to limit
+  /// the number of actions sent into the system. Only use this operator if your reducer needs to
+  /// layer on specialized logic for throttling. See <docs:Performance> for more information of why
+  /// sending high-frequency actions into a store is typically not what you want to do.
+  ///
   /// - Parameters:
   ///   - id: The effect's identifier.
   ///   - interval: The interval at which to find and emit the most recent element, expressed in
@@ -98,5 +107,4 @@ extension Effect {
 
 var throttleTimes: [AnyHashable: Any] = [:]
 var throttleValues: [AnyHashable: Any] = [:]
-var throttleTasks: [AnyHashable: Task<Void, Error>] = [:]
 let throttleLock = NSRecursiveLock()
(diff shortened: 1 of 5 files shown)


Sources/ComposableArchitecture/Effects/Throttle.swift
```diff
@@ -0,0 +1,110 @@
...
+extension Effect {
+  /// Throttles an effect so that it only publishes one output per given interval.
+  ///
+  /// The throttling of an effect is with respect to actions being sent into the store. So, if
+  /// you return a throttled effect from an action that is sent with high frequency, the effect
+  /// will be executed at most once per interval specified.
+  ///
+  /// > Note: It is usually better to perform throttling logic in the _view_ in order to limit
+  /// the number of actions sent into the system. Only use this operator if your reducer needs to
+  /// layer on specialized logic for throttling. See <docs:Performance> for more information of why
```
mbrandonw, line 14: this `<docs:Performance>` doesn't link correctly. is it not allowed to link to articles from docs of a method?

Sources/ComposableArchitecture/Effects/Throttle.swift
```diff
@@ -0,0 +1,110 @@
...
+  ///     `false`, the publisher emits the first element received during the interval.
+  /// - Returns: An effect that emits either the most-recent or first element received during the
+  ///   specified interval.
+  public func throttle<ID: Hashable, S: Scheduler>(
+    id: ID,
+    for interval: S.SchedulerTimeType.Stride,
+    scheduler: S,
+    latest: Bool
+  ) -> Self {
+    print(id)
```
iampatbrown, line 32: @mbrandonw should this one be here?

mbrandonw: Oops nope! Will remove soon.

Sources/ComposableArchitecture/Effects/Throttle.swift
stephencelis: It might need to be `<docs:/Dependencies/Performance>` to work, but cross-module linking is only supported in the Xcode 15 beta.

mbrandonw: Oh wait, it's `<doc:>` not `<docs:>`. Fixed!

stephencelis (approved):

Sources/ComposableArchitecture/Effects/Throttle.swift
```diff
@@ -0,0 +1,109 @@
...
+  /// - Returns: An effect that emits either the most-recent or first element received during the
+  ///   specified interval.
+  public func throttle<S: Scheduler>(
+    id: Any.Type,
+    for interval: S.SchedulerTimeType.Stride,
+    scheduler: S,
+    latest: Bool
+  ) -> Self {
+    self.throttle(id: ObjectIdentifier(id), for: interval, scheduler: scheduler, latest: latest)
+  }
```
stephencelis, line 104: We shouldn't bring this one back since cancellable `Any.Type` IDs were removed.

Sources/ComposableArchitecture/Effects/Throttle.swift
```diff
@@ -0,0 +1,86 @@
...
+  ///     the time system of the scheduler.
+  ///   - scheduler: The scheduler you want to deliver the throttled output to.
+  ///   - latest: A boolean value that indicates whether to publish the most recent element. If
+  ///     `false`, the publisher emits the first element received during the interval.
+  /// - Returns: An effect that emits either the most-recent or first element received during the
+  ///   specified interval.
+  public func throttle<ID: Hashable, S: Scheduler>(
+    id: ID,
+    for interval: S.SchedulerTimeType.Stride,
+    scheduler: S,
```
stephencelis, line 29: Bringing back the Combine requirement is the one bummer here. Would be nice to figure out a `Clock`-based API, but we can revisit in the future...

PR #2375, opened 2023-08-15
Add `TaskVoidResult`  since `TaskResult<Void>` is not `Equatable`

Ryu0118:
Hi there,
I propose TaskVoidResult so that functions with a `Void` return type can use a mechanism similar to `TaskResult`.

To use the `Void` function with TaskResult, We currently have to write code like this.
```Swift
TaskResult {
  try await voidAsyncThrowsFunction()
  return EmptyStruct()
}
```
or
```Swift
.run { send in
    try await voidAsyncThrowsFunction()
    await send(.success)
} catch: { error, send in
    await send(.failure(error))
}
``` 
This is not cool, so I tried to write a code in `TaskResult`  that conforms to `Equatable` where `Success` is equal to `Void`.
But this approach is already ban in Swift.
https://github.com/apple/swift-evolution/blob/master/proposals/0143-conditional-conformances.md#overlapping-conformances

So I solved this problem by creating a `TaskVoidResult`.
```Swift
TaskVoidResult { try await voidAsyncThrowsFunction() }
```

### DIFF
--- a/Sources/ComposableArchitecture/Effects/TaskVoidResult.swift
+++ b/Sources/ComposableArchitecture/Effects/TaskVoidResult.swift
@@ -1,0 +1,194 @@
+import XCTestDynamicOverlay
+
+/// A value that represents either a Void or a failure. This type differs from Swift's `Result`
+/// type in that it uses only one generic for the success case, leaving the failure case as an
+/// untyped `Error`.
+///
+/// This type is needed because Swift's concurrency tools can only express untyped errors, such as
+/// `async` functions and `AsyncSequence`, and so their output can realistically only be bridged to
+/// `Result<_, Error>`. However, `Result<_, Error>` is never `Equatable` since `Error` is not
+/// `Equatable`, and equatability is very important for testing in the Composable Architecture. By
+/// defining our own type we get the ability to recover equatability in most situations.
+///
+/// If someday Swift gets typed `throws`, then we can eliminate this type and rely solely on
+/// `Result`.
+///
+/// You typically use this type as the payload of an action which receives a response from an
+/// effect:
+///
+/// ```swift
+/// enum Action: Equatable {
+///   case factButtonTapped
+///   case factResponse(TaskVoidResult)
+/// }
+/// ```
+///
+/// Then you can model your dependency as using simple `async` and `throws` functionality:
+///
+/// ```swift
+/// struct NumberFactClient {
+///   var fetch: (Int) async throws -> Void
+/// }
+/// ```
+///
+/// And finally you can use ``Effect/run(priority:operation:catch:fileID:line:)`` to construct an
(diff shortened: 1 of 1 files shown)


stephencelis:
Hi @Ryu0118! Thanks for taking the time to PR :smile:

This is a topic that comes up occasionally, with folks also suggesting adding an `EquatableVoid` to the library. I think we'd like to better understand the problem before locking into a particular solution or adding surface are of the library, especially if that surface area can be added from the outside. Can you sketch out an example of how you use `TaskVoidResult` with a bit more context, like a small, self-contained reducer? And can you contrast that with the alternative to help motivate things? It'd be helpful to evaluate and explore things.

stephencelis:
We chatted a bit more about this today and while on the fence, we ultimately would like to encourage users to introduce solutions to this language problem at their application layer for now. [SE-0283](https://github.com/apple/swift-evolution/blob/main/proposals/0283-tuples-are-equatable-comparable-hashable.md) was accepted and the language really should consider `Void` equatable, so we'd ideally avoid the surface area of too many library-level workarounds when the language should (hopefully) fix things in the future.

Another option would be to start a `swift-composable-architecture-extras` package that collects these workarounds for folks. We'd definitely be down to link back to such a library!

> Thus, the user no longer needs to create the EquatableVoid and it's instance no longer needs to be created, which reduces the number of ARC calls and overhead.

Have you benchmarked or analyzed the generated assembly code? Would definitely be curious if there's a measurable difference here.

Ryu0118:
Indeed, I understand that it is better to avoid workarounds at the library level for future void equatable support.

Then consider developing swift-composable-architecture-extra as a temporary workaround for these!
Also, here are the benchmark results for TaskVoidResult and TaskResult.

スクリーンショット 2023-08-18 1 40 54

TaskVoidResult is slightly faster, but we see little difference.

stephencelis:
@Ryu0118 A few other ideas to think about if you do spin up an "extras" package:

- I wonder if there are better names. `DiscardingTaskResult` comes to mind and reads nicely, but isn't quite analogous to `withDiscardingTaskGroup`...so maybe not appropriate.

- Maybe this could be an alternate static on `TaskResult`? Throwing out a couple names:

    ```swift
    TaskResult.do { … }
    TaskResult.ignore { … }
    ```

    Since the benchmark showed very little difference, it might be preferable to only add a function as surface area rather than a whole new type.

Ryu0118:
That's a great idea!
 I have created [swift-composable-architecture-extras](https://github.com/Ryu0118/swift-composable-architecture-extras) and incorporated your idea.

stephencelis:
Thanks for sharing! I'm going to close this PR out for now since your library offers an alternative path. We can always revisit this in the future with more consideration.

PR #2378, opened 2023-08-16
0.x: Fix deprecation warning on TestStore

AndreaRomani:
This PR cherry-picks the [fix](https://github.com/pointfreeco/swift-composable-architecture/pull/2347) for an incorrect `TestStore` deprecation warning onto the `prerelase/1.0` branch. 

I'm opening the PR because it seemed an easy fix and should help folks getting rid of some noise while preparing for `1.0`. For instance, in highly tested codebases these incorrect warnings could hide the important deprecations that `0.57`/`0.58` should help fixing. It would be nice to include this into a `0.58.2` or `0.59` release 🙂


stephencelis (approved):
Thanks!

PR #2385, opened 2023-08-18
prevent crash on catalina from missing symbol

jaredh159:
see https://github.com/pointfreeco/swift-composable-architecture/discussions/2351


stephencelis (approved):
Seems like a good fix to the bug, thanks for digging into it!

PR #2389, opened 2023-08-21
[FIX] Fixed API description to use the recent name: `EffectTask` -> `Effect`

jaesung-0o0:
There's a very minor modification on API description of ``EffectOf``.

The description uses `EffectTask` in the code snippet, but it should be `Effect`.

I also searched for EffectTask, but there are no more results of its usage.


mbrandonw (approved):
Good find, thank you!

PR #2392, opened 2023-08-22
Undeprecate debounce and throttle overloads restored in 1.1.0

pyrtsa:


Sources/ComposableArchitecture/Effects/Publisher/Throttling.swift
```diff
@@ -80,6 +79,7 @@ extension EffectPublisher {
   ///     `false`, the publisher emits the first element received during the interval.
   /// - Returns: An effect that emits either the most-recent or first element received during the
   ///   specified interval.
+  @available(*, deprecated)
```
stephencelis, line 82:
```suggestion
  @available(*, deprecated)
  @available(
    *,
    deprecated,
    message:
      """
      Types defined for cancellation may be compiled out of release builds in Swift and are unsafe to use. Use a hashable value, instead, e.g. define a timer cancel identifier as 'enum CancelID { case debounce }' and call 'effect.debounce(id: CancelID.debounce, …)'.
      """
  )
```

stephencelis (approved):
Thanks!

PR #2394, opened 2023-08-22
Add `TestStore.bindings` for testing bindable view state

stephencelis:
Because `@BindingViewState` is populated by a live store, there is no way to easily test `ViewState` structs that use `@BindingViewState`. This adds a `bindings` property on `TestStore` (`bindings(action:)` method when using view actions) that makes it possible to write assertions against view state.

### DIFF
--- a/Tests/ComposableArchitectureTests/BindableStoreTests.swift
+++ b/Tests/ComposableArchitectureTests/BindableStoreTests.swift
@@ -167,7 +167,7 @@ final class BindableStoreTests: XCTestCase {
             return .none
           case .view(.loginButtonTapped):
             state.isRequestInFlight = true
-            return .none  // TODO: Login request
+            return .none  // NB: Login request
           }
         }
       }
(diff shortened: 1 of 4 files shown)


Tests/ComposableArchitectureTests/BindableStoreTests.swift
```diff
@@ -80,4 +80,143 @@ final class BindableStoreTests: XCTestCase {
...
+      var body: some ReducerOf<Self> {
+        BindingReducer()
+        Reduce { state, action in
+          switch action {
+          case .binding:
+            state.isFormValid = !state.email.isEmpty && !state.password.isEmpty
+            return .none
+          case .loginButtonTapped:
+            state.isRequestInFlight = true
+            return .none  // TODO: Login request
```
mbrandonw, line 105: What do you think of "NB" here so that it doesn't seem like a TODO for us to fix?

```suggestion
            return .none  // NB: Login request
```

mbrandonw (approved):

PR #2401, opened 2023-08-23
Added `Equatable` Protocol to Code Examples in Tutorial Docs.

zooxop:
I've added the missing `Equatable` protocol to the code examples in the tutorial documentation.

The behavior of the code is likely unaffected, but there might be confusion due to its deviation from the example provided in the [README](https://github.com/pointfreeco/swift-composable-architecture#basic-usage).


stephencelis:
@zooxop Thanks for taking the time to PR! I believe `Equatable` is not necessary for these code samples, though. In the interest of progressive disclosure and minimizing upfront work, we delay the `Equatable` conformances till they're needed:

- `State` doesn't need to be `Equatable` till we pass it to a `ViewStore`, which we explain midway through the first chapter here:

    image

- `Action` doesn't need to be `Equatable` till we write tests against the the action contents received by an effect, which we explain in the 3rd chapter on "testing" here:

    image

zooxop:
Thank you for your detailed explanation and quick improvements.  
I missed the content of **Step 4** and the fact that I need to add the code **`extension CounterFeature.State: Equatable {}`**.  
It was clearly my mistake. I apologize.

I will close this PR. Have a good day!

PR #2402, opened 2023-08-23
Enhance `Reducer.ifLet` child effect cancellation

stephencelis:
Right now, `ifLet` checks the presented state before and after it runs the parent reducer, and if it changes, it cancels any in-flight effects for the previous state. It will not, however, detect when child state changes _outside_ the parent reducer, _e.g._:

```swift
Reduce { state, action in
  // ...
  state.child = nil // or `= .someOther(child)`
}

Reduce { state, action in
  // ...
}
.ifLet(\.$child, action: /Action.child) {
  Child()
}
```

While we generally recommend structuring reducers in a way in which this doesn't happen, as a library we should attempt to improve the situation. In our case `@PresentationState` already uses some internal state to detect if it needs to install the child's dismiss effects, _e.g._ if it was deep-linked into at app launch. We can leverage this same state to detect if it needs to tear down the previous child's effects.

This PR's solution should improve cancellation in these edge cases, but it may not (and may not be able to) handle every case.

~~We should explore an equivalent solution for `.forEach`.~~

### DIFF
--- a/Sources/ComposableArchitecture/Internal/NavigationID.swift
+++ b/Sources/ComposableArchitecture/Internal/NavigationID.swift
@@ -12,6 +12,7 @@ private enum NavigationIDPathKey: DependencyKey {
   static let testValue = NavigationIDPath()
 }
 
+@usableFromInline
 struct NavigationIDPath: Hashable, Sendable {
   fileprivate var path: [NavigationID]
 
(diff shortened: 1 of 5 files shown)


stephencelis:
Added support for `forEach`. Both have the problem where if the internal state is fully replaced (_e.g._ a parent of a parent resets the presentation or stack state), the child reducer won't have access to the before state, but we'll need to make some more foundational changes before we can address that.

mbrandonw (approved):

PR #2405, opened 2023-08-24
Fix the typos in the tutorial reference

zooxop:
I believe that **`ContentView`** is a typo.

ContentViewTypo


mbrandonw (approved):
Hi @zooxop, good catch! Thank you very much.

PR #2410, opened 2023-08-25
Remove meeting feature.

mbrandonw:
In the tour series we don't actually implement a `MeetingFeature`, so let's not do it in the repo either.

### DIFF
--- a/Examples/Standups/Standups/AppFeature.swift
+++ b/Examples/Standups/Standups/AppFeature.swift
@@ -97,22 +97,19 @@ struct AppFeature: Reducer {
   struct Path: Reducer {
     enum State: Equatable {
       case detail(StandupDetail.State)
-      case meeting(MeetingReducer.State)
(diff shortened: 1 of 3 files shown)


Examples/Standups/Standups/AppFeature.swift
```diff
@@ -97,23 +97,20 @@ struct AppFeature: Reducer {
...
       case detail(StandupDetail.State)
-      case meeting(MeetingReducer.State)
+      case meeting(Meeting, standup: Standup)
       case record(RecordMeeting.State)
     }
 
     enum Action: Equatable {
       case detail(StandupDetail.Action)
-      case meeting(MeetingReducer.Action)
+      case meeting(Never)
```
lukeredpath, line 106: Could this just be `case meeting`? Is this case even necessary?

stephencelis (approved):

PR #2427, opened 2023-09-03
Fix Tutorials Code Diff

denil-ct:
## Overview
The file names in the tutorial for the argument `previousFiles` were missing a suffix, resulting in a wrong file name and breaking the code diffs shown. This PR fixes those wrong file names.

## Testing
All the changes were verified by building documentation in Xcode and the documentation previews in Xcode 15.


stephencelis (approved):
Good catch, thanks!

PR #2434, opened 2023-09-05
Add @warn_unqualified_access to alert modifiers

vladyslavsosiuk:
Add the @warn_unqualified_access decorator to alert modifiers to prevent silly wrong usage mistakes such as 
```swift
struct ContentView: View {
    let store: StoreOf<Feature>
    
    var body: some View {
        Text("Hello world")
        alert( // Missing dot. The @warn_unqualified_access will generate a warning `Use of 'alert' treated as a reference to instance method in protocol 'View'`
            store: store.scope(
                state: \.$destination,
                action: { .destination($0) }
            ),
            state: /Feature.Destination.State.alert,
            action: Feature.Destination.Action.alert
        )
    }
}
```

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Alert.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Alert.swift
@@ -8,6 +8,7 @@ extension View {
   /// - Parameters:
   ///   - store: A store that is focused on ``PresentationState`` and ``PresentationAction`` for an
   ///     alert.
+  @warn_unqualified_access
   public func alert<ButtonAction>(
     store: Store<PresentationState<AlertState<ButtonAction>>, PresentationAction<ButtonAction>>
   ) -> some View {
@@ -23,6 +24,7 @@ extension View {
   ///   - toDestinationState: A transformation to extract alert state from the presentation state.
   ///   - fromDestinationAction: A transformation to embed alert actions into the presentation
   ///     action.
+  @warn_unqualified_access
   public func alert<State, Action, ButtonAction>(
     store: Store<PresentationState<State>, PresentationAction<Action>>,
     state toDestinationState: @escaping (_ state: State) -> AlertState<ButtonAction>?,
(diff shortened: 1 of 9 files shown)


stephencelis:
@vladyslavsosiuk Thanks for the PR! While we use this attribute on our reducer methods, we did not add them to view modifiers since SwiftUI doesn't either. While we general follow Apple's lead, we're not against making improvements there, though I think we'd like to make them across the board on _all_ view modifiers shipped in the library rather than just `alert`. Would you be up for adding this attribute to all the view modifiers found in the library?

davdroman:
@stephencelis only reservation I have about this change is that these kinds of extensions would also throw a warning:

```swift
extension View {
	func customModifier(...) {
		alert(...) // warning
	}
}
```

`self.` would then always be required to silence them, which might go against some teams' code styleguides.

vladyslavsosiuk:
> @vladyslavsosiuk Thanks for the PR! While we use this attribute on our reducer methods, we did not add them to view modifiers since SwiftUI doesn't either. While we general follow Apple's lead, we're not against making improvements here, though I think we'd like to make them across the board on _all_ view modifiers shipped in the library rather than just `alert`. Would you be up for adding this attribute to all the view modifiers found in the library?

@stephencelis yes, will add them tomorrow

vladyslavsosiuk:
@stephencelis added to all modifiers. I also added them to the Presentation modifiers though I'm not sure those were expected. Let me know if I should revert them

stephencelis:
@vladyslavsosiuk Sorry for the churn, but @davdroman brings up a good point, and it may be a reason why Apple doesn't use this modifier in SwiftUI itself. I think in retrospect we should continue to ship APIs that intend to match Apple's in a style that matches Apple's. Thanks for exploring this, though. We personally agree that it's a much better experience.

stephencelis:
Going to close this for now. While we really do personally appreciate the problem it's solving (and thanks for taking the time to submit!), we think it's more appropriate to stick with the Apple standard for now.

PR #2437, opened 2023-09-07
[fix] Duration.seconds -> .minutes in StandupForm.swift

filblue:


mbrandonw (approved):
Good catch! Thanks for the fix.

PR #2440, opened 2023-09-08
Add missing file and line parameters to dismiss call

heiberg:
File and line parameters were not passed along to the main implementation in `Dismiss.swift`.


stephencelis (approved):
Good catch, thanks!

PR #2442, opened 2023-09-08
Fix mismatched reducer names

Jager-yoo:
Hi, while reviewing the PresentationReducer and StackReducer, I noticed that some of the documentation uses names like `ParentFeature / Parent` and `ChildFeature / Child` together, which can be a bit confusing. Throughout the library, the terms `Parent` and `Child` are more commonly used, including the tests.

For now, I've made the necessary changes, not in the test code. If you prefer me to rename every ParentFeature to Parent and ChildFeature to Child, please let me know. Additionally, I've considered changing the Reducer<State, Action> to `ReducerOf<Self>`. For the consistency, I've only modified what's in the PresentationReducer.swift. If you have any better suggestions or feedback, I'd love to hear them. Thanks!

### DIFF
--- a/Sources/ComposableArchitecture/Reducer/Reducers/PresentationReducer.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/PresentationReducer.swift
@@ -13,7 +13,7 @@ import Combine
 /// ```swift
 /// struct ParentFeature: Reducer {
 ///   struct State {
-///     @PresentationState var child: Child.State?
+///     @PresentationState var child: ChildFeature.State?
 ///      // ...
 ///   }
 ///   // ...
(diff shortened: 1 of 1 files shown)


stephencelis:
@Jager-yoo Thanks for the PR! I think our stylistic preference is for the `*Feature` versions in documentation, though. Would you be up for inverting the change in this PR?

Jager-yoo:
> @Jager-yoo Thanks for the PR! I think our stylistic preference is for the `*Feature` versions in documentation, though. Would you be up for inverting the change in this PR?

Ok, then I'm going to change some mismatched ones!

Sources/ComposableArchitecture/Reducer/Reducers/PresentationReducer.swift
```diff
@@ -212,7 +212,7 @@ extension Reducer {
   ///     // ...
   ///   }
   ///
-  ///   var body: some Reducer<State, Action> {
+  ///   var body: some ReducerOf<Self> {
```
Jager-yoo, line 215: Just for consistency with the one in `StackReducer.swift`

https://github.com/pointfreeco/swift-composable-architecture/blob/f0bae57f361b3611b48d7db0b78393e5c78be1fc/Sources/ComposableArchitecture/Reducer/Reducers/StackReducer.swift#L216

stephencelis (approved):
Thanks!

PR #2454, opened 2023-09-13
Get local value with 'let' in switch-case

jkimdev:
switch-case formatting with general with TCA code convention.


stephencelis (approved):
Thanks! 😄

acosmicflamingo:
@stephencelis Ah, you guys use Apple's swift-format. I know this would've been formatted using Nick Lockwood's swift-format, but I imagine you all prefer a 1st party solution, even if it's not as powerful

PR #2463, opened 2023-09-15
Added a failing test that demonstrates unexpected Alert dismissal

mluisbrown:
Added a failing test that demonstrates unexpected Alert dismissal when a non alert destination action is sent.

See this slack thread for reference: https://pointfreecommunity.slack.com/archives/C04KQQ7NXHV/p1694623754410579

When using `@PresentationState` with an enum destination, if an Alert is currently presented, and an action for a different destination is sent to the store, the Alert is dismissed even though no alert action was sent.  

The alert should only be dismissed if an alert action is sent to the store.

### DIFF
--- a/Tests/ComposableArchitectureTests/Reducers/PresentationReducerTests.swift
+++ b/Tests/ComposableArchitectureTests/Reducers/PresentationReducerTests.swift
@@ -2427,4 +2427,110 @@ final class PresentationReducerTests: BaseTCATestCase {
     // NB: Another action needs to come into the `ifLet` to cancel the child action
     await store.send(.tapAfter)
   }
+
+  func testPresentation_leaveAlertPresentedForNonAlertActions() async {
+    if #available(iOS 16, macOS 13, tvOS 16, watchOS 9, *) {
+      struct Child: Reducer {
+        struct State: Equatable {
+          var count = 0
+        }
+        enum Action: Equatable {
+          case decrementButtonTapped
+          case incrementButtonTapped
+        }
+        func reduce(into state: inout State, action: Action) -> Effect<Action> {
+          switch action {
+          case .decrementButtonTapped:
+            state.count -= 1
+            return .none
+          case .incrementButtonTapped:
+            state.count += 1
+            return .none
+          }
+        }
+      }
+
+      struct Parent: Reducer {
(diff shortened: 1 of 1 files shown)


mluisbrown:
Curiously, when I run the test I see this error, which is what I would have expected to see in our code, but didn't, indicating that this kind of thing is really programmer error:

```
/Users/michael/Documents/Projects/swift-composable-architecture/Tests/ComposableArchitectureTests/Reducers/PresentationReducerTests.swift:2530: error: -[ComposableArchitectureTests.PresentationReducerTests testPresentation_leaveAlertPresentedForNonAlertActions] : A "Scope" at "ComposableArchitectureTests/PresentationReducerTests.swift:2507" received a child action when child state was set to a different case. …

  Action:
    PresentationReducerTests.Parent.Destination.Action.child(.decrementButtonTapped)
  State:
    PresentationReducerTests.Parent.Destination.State.alert

This is generally considered an application logic error, and can happen for a few reasons:

• A parent reducer set "PresentationReducerTests.Parent.Destination.State" to a different case before the scoped reducer ran. Child reducers must run before any parent reducer sets child state to a different case. This ensures that child reducers can handle their actions while their state is still available. Consider using "Reducer.ifCaseLet" to embed this child reducer in the parent reducer that change its state to ensure the child reducer runs first.

• An in-flight effect emitted this action when child state was unavailable. While it may be perfectly reasonable to ignore this action, consider canceling the associated effect before child state changes to another case, especially if it is a long-living effect.

• This action was sent to the store while state was another case. Make sure that actions for this reducer can only be sent from a view store when state is set to the appropriate case. In SwiftUI applications, use "SwitchStore".
```

mbrandonw:
Hi @mluisbrown, yeah this is definitely a programmer error in general. Actions should never be sent to a destination that is not being presented. That message probably was printed to the console in your app, but you may have not noticed. We also show a purple Xcode runtime warning, but maybe you missed that too?

Now, if it's possible to _not_ dismiss the alert in this case we will happily do it just to make the experience a little nicer, but ultimately it would still be a programmer error.

stephencelis:
Closing in favor of #2468, which includes this patch. Thanks @mluisbrown!

PR #2468, opened 2023-09-19
Only dismiss alert/dialog state when receiving a domain-specific action

stephencelis:
Fixes #2463.

While receiving an action from another domain is a programmer error and will emit a runtime warning, we should avoid having this error negatively impact the end user experience by dismissing alerts and dialogs when they shouldn't be. This PR addresses things by expanding the ephemeral state protocol to describe its action type, and comparing this action type to whatever it can extract from the given action using CasePaths' enum metadata runtime APIs.

### DIFF
--- a/Tests/ComposableArchitectureTests/Reducers/PresentationReducerTests.swift
+++ b/Tests/ComposableArchitectureTests/Reducers/PresentationReducerTests.swift
@@ -2427,4 +2427,110 @@ final class PresentationReducerTests: BaseTCATestCase {
     // NB: Another action needs to come into the `ifLet` to cancel the child action
     await store.send(.tapAfter)
   }
+
+  func testPresentation_leaveAlertPresentedForNonAlertActions() async {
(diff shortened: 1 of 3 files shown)


mluisbrown (approved):
Thanks @stephencelis 👍

Tests/ComposableArchitectureTests/Reducers/PresentationReducerTests.swift
```diff
@@ -2427,4 +2427,118 @@ final class PresentationReducerTests: BaseTCATestCase {
...
+        func reduce(into state: inout State, action: Action) -> Effect<Action> {
+          switch action {
+          case .decrementButtonTapped:
+            state.count -= 1
+            return .none
+          case .incrementButtonTapped:
+            state.count += 1
+            return .none
+          }
+        }
```
jaesung-0o0, line 2450: Question: Why do you use `reduce(into:action:)` here while the `Parent` and `Destination` use `body`?

mbrandonw (approved):

PR #2474, opened 2023-09-20

vasiliyzaycev:
It seems that the PresentationAction wrapper got lost here


stephencelis (approved):
Good catch, thanks!

PR #2475, opened 2023-09-21

vasiliyzaycev:
It seems in the test cases in StackBasedNavigation.md instead of $0[id:] should be $0.path[id:]


stephencelis (approved):
Thanks!

PR #2480, opened 2023-09-23
Clean up and modernize docs/examples

stephencelis:
This PR just goes through our docs and examples and modernizes a few things, mainly:

  * Updating `action: MyFeature.Action.case` to `action: { .case($0) }`, which is often shorter and more symmetrical with `state: \.value`.
  * Removing unnecessary `<State, Action>` generic parameter list from `Reduce`.
  * General SwiftUI API modernization.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
@@ -135,7 +135,7 @@ struct AnimationsView: View {
           "Big mode",
           isOn:
             viewStore
-            .binding(get: \.isCircleScaled, send: Animations.Action.circleScaleToggleChanged)
+            .binding(get: \.isCircleScaled, send: { .circleScaleToggleChanged($0) })
             .animation(.interactiveSpring(response: 0.25, dampingFraction: 0.1))
         )
         .padding()
--- a/Examples/Standups/Standups/AppFeature.swift
+++ b/Examples/Standups/Standups/AppFeature.swift
@@ -25,7 +25,7 @@ struct AppFeature: Reducer {
     Scope(state: \.standupsList, action: /Action.standupsList) {
       StandupsList()
     }
-    Reduce<State, Action> { state, action in
+    Reduce { state, action in
       switch action {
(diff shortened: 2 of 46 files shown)


Examples/Standups/Standups/AppFeature.swift
```diff
@@ -25,7 +25,7 @@ struct AppFeature: Reducer {
     Scope(state: \.standupsList, action: /Action.standupsList) {
       StandupsList()
     }
-    Reduce<State, Action> { state, action in
```
acosmicflamingo, line 28: Should this be removed if it means that Xcode now has a very hard time auto-completing the action enums? Looks like it's still a problem in Xcode 15 and Swift 5.9. I don't have a problem with it; just keeping in mind in case this was forgotten :)

acosmicflamingo (commented):
I added a note about removing generics from Reducer and its impact on auto-completion, in case this is a priority for you guys.

Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Animations.swift
```diff
@@ -135,7 +135,7 @@ struct AnimationsView: View {
           "Big mode",
           isOn:
             viewStore
-            .binding(get: \.isCircleScaled, send: Animations.Action.circleScaleToggleChanged)
+            .binding(get: \.isCircleScaled, send: { .circleScaleToggleChanged($0) })
```
jaesung-0o0, line 138: **Question** 
Why do you use `binding(get:send:)` instead of using `BindingState`?
```swift
viewStore.$isCircleScaled
```
Is this to avoid changing the logic since you've already provided the episode video?
*OR* is there any case where it *prefers to* use `binding(get:send:)`?

Examples/Standups/Standups/AppFeature.swift
acosmicflamingo: You're right; my apologies. I am unable to reproduce with the reducers within your codebase. I am only able to reproduce in my reducers that have tons of cases (since I'm utilizing ViewAction, DelegateAction, InternalAction enums) but when I clean up a lot of it; the auto-complete is correct.

mbrandonw (approved):

PR #2481, opened 2023-09-23
correcting readme.md syntax

ictechgy:
Hi~!   
I corrected the README document because there was a part that could cause a syntax error.
Since `FeatureView` is not `Scene`, I wrapped it in `WindowGroup`.

Please check it out! 😊


stephencelis (approved):
Thanks!

PR #2482, opened 2023-09-23
Add Documentation in TreeBasedNavigation.md

Ryu0118:
Hi, I found the documentation lacking in TreeBasedNavigation.md .
The Stack-based Navigation one explained the integration from the child domain to the parent domain, but the Tree-based Navigation one did not.

### DIFF
--- a/Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
@@ -232,6 +232,23 @@ struct InventoryFeature: Reducer {
   // ...
 }
 ```
+And then we must make use of the ifLet(_:action:destination:fileID:line:) method to integrate the domain of the Destination with the domain of the parent feature:
+
+```swift
+struct InventoryFeature: Reducer {
+  // ...
+
+  var body: some ReducerOf<Self> {
+    Reduce { state, action in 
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
```diff
@@ -232,6 +232,23 @@ struct InventoryFeature: Reducer {
   // ...
 }
 ```
+And then we must make use of the ifLet(_:action:destination:fileID:line:) method to integrate the domain of the Destination with the domain of the parent feature:
```
stephencelis, line 235:
```suggestion

And then we must make use of the ``Reducer/ifLet(_:action:destination:fileID:line:)`` operator to
integrate the domain of the destination with the domain of the parent feature:
```

Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
```diff
@@ -232,6 +232,23 @@ struct InventoryFeature: Reducer {
...
+
+```swift
+struct InventoryFeature: Reducer {
+  // ...
+
+  var body: some ReducerOf<Self> {
+    Reduce { state, action in 
+      // Core logic for root feature
```
stephencelis, line 243:
```suggestion
      // ...
```

PR #2493, opened 2023-09-30
fix: documentation error for the getting started example

KyleLeneau:
I found a missing `)` in the documentation that'd be helpful to fix for others.


stephencelis (approved):
Good catch, thanks!

PR #2496, opened 2023-10-03
Use `StoreTask.cancel()` instead of `task.rawValue?.cancel()`

Ryu0118:
Rewrote `task.rawValue?.cancel()` instead of `task.rawValue?.cancel()` because `ViewStoreTask.cancel()` cancels the underlying task and also waits for the task to finish.
However, since the current `StoreTask.cancel()` only cancels the task, it is no longer necessary to write `task.rawValue?.cancel()`.
So, I changed the writing style to use `StoreTask.cancel()`.


stephencelis:
Thanks :)

PR #2499, opened 2023-10-05
[Docs/StackBasedNavigation] Fixed `self.path` to `state.path`

jaesung-0o0:
### Simple Fix
Fixed `self.path` to `state.path`
```swift
case let .path(.element(id: id, action: .editItem(.saveButtonTapped))):
-   guard case let .editItem(editItemState) = self.path[id: id]
+   guard case let .editItem(editItemState) = state.path[id: id]
    else { return .none }
```


stephencelis (approved):
Great, thanks!

PR #2503, opened 2023-10-06
Revamp integration tests

mbrandonw:
This allows us to assert on the inner workings of TCA during integration tests, such as how many times a view is computed, how many stores are created/destroyed, etc.


Examples/Integration/IntegrationUITests/Internal/BaseIntegrationTests.swift
```diff
@@ -0,0 +1,128 @@
...
+        self.logs.label,
+        expectedLogs.map(\.description).joined(separator: "\n"),
+        file: file,
+        line: line
+      )
+    }
+  }
+}
+
+enum Log: ExpressibleByStringLiteral, CustomStringConvertible {
```
mbrandonw, line 114: I think we should ditch all of this and go with unordered assertions by default, _and_ we should use inline snapshot testing to do the assertions.

Makefile
```diff
@@ -41,7 +41,7 @@ test-docs:
 		&& exit 1)
 
 test-examples:
-	for scheme in "CaseStudies (SwiftUI)" "CaseStudies (UIKit)" Integration Search Standups SpeechRecognition TicTacToe Todos VoiceMemos; do \
+	for scheme in Integration; do \
```
stephencelis, line 44:
```suggestion
	for scheme in "CaseStudies (SwiftUI)" "CaseStudies (UIKit)" Integration Search Standups SpeechRecognition TicTacToe Todos VoiceMemos; do \
```

stephencelis (approved):
All looks good to me! Just need to revert the Makefile when ready.

PR #2506, opened 2023-10-07
Fix Tree-based Navigation Bug

Ryu0118:
I have encountered a bug where the PresentationState does not turn nil when a View is popped.
This happened during a push transition using Tree-based Navigation like this.

```Swift
.navigationDestination(
    store: store.scope(state: \.$destination, action: { .destination($0) }),
    state: /MyReducer.Destination.State.screenA,
    action: MyReducer.Destination.Action.screenA
) { store in
    ScreenA(store: store)
}
.navigationDestination(
    store: store.scope(state: \.$destination, action: { .destination($0) }),
    state: /MyReducer.Destination.State.screenB,
    action: MyReducer.Destination.Action.screenB
) { store in
    ScreenB(store: store)
}
```
Currently, the following code is executed when a View is popped.
```Swift
self.viewStore.binding(
  get: {
    $0.wrappedValue.flatMap(toDestinationState) != nil
      ? toID($0).map { AnyIdentifiable(Identified($0) { $0 }) }
      : nil
  },
  compactSend: {
    guard
      $0 == nil,
      self.viewStore.wrappedValue != nil,
      id == nil || self.toID(self.viewStore.state) == id
    else { return nil }
    return .dismiss
  }
),
```
However, when I debugged it, the id was not nil.
So I erased the evaluation of `id == nil` and it works fine.
Why is it currently evaluating `id == nil || self.toID(self.viewStore.state) == id`? It seems to me that it should simply return `.dismiss` when $0 is nil.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/PresentationModifier.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/PresentationModifier.swift
@@ -213,7 +213,6 @@ public struct PresentationStore<
   }
 
   public var body: some View {
-    let id = self.toID(self.viewStore.state)
     self.content(
       self.viewStore.binding(
         get: {
@@ -224,8 +223,7 @@ public struct PresentationStore<
         compactSend: {
           guard
             $0 == nil,
-            self.viewStore.wrappedValue != nil,
-            id == nil || self.toID(self.viewStore.state) == id
+            self.viewStore.wrappedValue != nil
           else { return nil }
           return .dismiss
         }


stephencelis:
@Ryu0118 Any chance you can get a repro for us, or even UI test in the `Integration` project, so we better understand the fix and catch regressions in the future? These integration tools can be a bit thorny at times, and hopefully we have a regression test that already catches why we have these checks, but we'll see soon!

mbrandonw:
As @stephencelis mentioned, we do have an integration test in place that shows why this check is here. With your change the integration test suite fails on this assertion:

https://github.com/pointfreeco/swift-composable-architecture/blob/main/Examples/Integration/IntegrationUITests/PresentationTests.swift#L193

It looks like the second alert is not being presented after the first one is dismissed.

In order for us to best understand the problem you are trying to fix, I think it would be best if you could come up with a minimal reproducible example demonstrating the behavior. Then we can see if there is a fix that also maintains the current behavior.

Ryu0118:
Okay, I will try to write a program that can reproduce it.

And also found that it works correctly in the simulator, but not on the actual device.

In actual device, 
`self.toID(self.viewStore.state)` is
```
Optional(AnyHashable(ComposableArchitecture.(unknown context at $105d77dc0).NavigationDestinationID(objectIdentifier: ObjectIdentifier(0x0000000105f046c0), enumTag: Optional(1))))
```
`id` is
```
Optional(AnyHashable(ComposableArchitecture.(unknown context at $105d77dc0).NavigationDestinationID(objectIdentifier: ObjectIdentifier(0x0000000105f046c0), enumTag: Optional(0))))
```
Comparing the two, the enumTag has different values between self.toID(self.viewStore.state) and id.

But In simulator, 
`self.toID(self.viewStore.state)` and `id` is
```
Optional(AnyHashable(ComposableArchitecture.(unknown context at $105d77dc0).NavigationDestinationID(objectIdentifier: ObjectIdentifier(0x0000000105f046c0), enumTag: Optional(1))))
```

So it looks like the case path problem.

stephencelis:
@Ryu0118 I believe those test failures are related to why we need the identity check there.

stephencelis:
@Ryu0118 I'm going to close this for now. I'd be curious if the new non-reflective case paths fix for you, though! We also will be [streamlining](https://github.com/pointfreeco/swift-composable-architecture/discussions/2565) these modifiers soon, so potentially even less that can go wrong.

PR #2519, opened 2023-10-13
Bring back _printChanges queue.

mbrandonw:
Prior to the `ReducerProtocol` we would queue print changes in the debug operator on reducers:

https://github.com/pointfreeco/swift-composable-architecture/blob/54c384581cd0ed395d4f314dfdaca14dc70acf4c/Sources/ComposableArchitecture/Debugging/ReducerDebugging.swift#L118

We lost that in #1283, but we still used `fireAndForget` which had a sync overload, and so queuing just naturally happened on the main thread.

But, with 1.0 we got rid of `fireAndForget` and so instead used `run`, but then that introduced asynchrony and possible ordering issues with logs. So, currently it is possible for `actionA` to be sent before `actionB`, yet the logs for `actionB` will appear in the console before `actionA`.

This PR brings back the queue. Alternatively we could `@MainActor` the `run` effect. Each approach has their pros and cons. With the queue we run the risk of the logs starting to lag if too much is logged at once, and with `@MainActor` we risk slowing down the app if too much is logged (only in debug). I went with the queue for now, but open to switch to `@MainActor` too.

### DIFF
--- a/Sources/ComposableArchitecture/Reducer/Reducers/DebugReducer.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/DebugReducer.swift
@@ -1,3 +1,4 @@
+import Combine
 import Dispatch
 
 extension Reducer {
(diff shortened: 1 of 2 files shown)


davdroman:
`@MainActor` making the app lag sounds significantly worse than just the logs themselves lagging, from a DX perspective.

stephencelis (approved):

PR #2527, opened 2023-10-20
Cache store scoping when possible

stephencelis:
Currently, store scoping can happen quite often: whenever a view that scopes a store is recomputed, the old scoped store is discarded and a brand new scoped store is created. This involves creating a new object and parent/child subscription every single time a store is scoped. While stores are generally lightweight objects, this is still unnecessary work.

This PR introduces the ability to cache child stores directly in the parent via new `scope` methods. These `scope` methods take a state key path instead of state transform function. The parent store uses this key path as a hashable identifier to store the child store in a dictionary. This means that after the first time a store is scoped, scoping to it again is a simple dictionary lookup.

Anyone using key path expression syntax for their scopes will automatically benefit from this change, while the older style of scoping (soft deprecated for now and probably soft deprecated for some time) will mostly go through the older code path.

Adapting an existing benchmark to scope a store X number of times before sending an action to the scoped store shows a pretty significant improvement:

Before:

```
name                       time             std        iterations
-----------------------------------------------------------------
Store.Scoped nested tap: 1   53291.000 ns ±   6.35 %      26188
Store.Scoped nested tap: 2 1818583.000 ns ±  31.98 %       2514
Store.Scoped nested tap: 3 3289500.000 ns ±   6.57 %        469
Store.Scoped nested tap: 4 4203916.500 ns ±   5.79 %        370
Store.Scoped nested tap: 5 5205125.000 ns ±   5.25 %        296
```

After:

```
name                       time             std        iterations
-----------------------------------------------------------------
Store.Scoped nested tap: 1   52375.000 ns ±   5.75 %      26657
Store.Scoped nested tap: 2  109833.000 ns ±   6.05 %      12640
Store.Scoped nested tap: 3  165834.000 ns ±   5.53 %       8371
Store.Scoped nested tap: 4  228583.000 ns ±   4.47 %       6103
Store.Scoped nested tap: 5  299500.000 ns ±   4.11 %       4645
```


Examples/Integration/IntegrationUITests/EnumTests.swift
```diff
@@ -24,10 +24,8 @@ final class EnumTests: BaseIntegrationTests {
       StoreOf<BasicsView.Feature?>.init
       StoreOf<BasicsView.Feature?>.init
       StoreOf<BasicsView.Feature?>.init
-      StoreOf<BasicsView.Feature?>.init
```
stephencelis, line 27: These deletions in our integration suite represent scoped stores that are able to be reused rather than discarded and recreated each render.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -697,116 +813,6 @@ public final class Store<State, Action> {
 /// ```
 public typealias StoreOf<R: Reducer> = Store<R.State, R.Action>
 
-extension Reducer {
-  fileprivate func rescope<ChildState, ChildAction>(
```
stephencelis, line 701: Since stores are now being cached, we think this `rescope` machinery is no longer necessary in most cases. I also did not measure much of a performance difference in its removal in our very limited suite.

@iampatbrown Let me know your thoughts on this, especially if you have apprehensions. I think we could potentially keep `rescope` for stores that are not cached, though I'm not 100% sure what that would look like.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -388,45 +412,137 @@ public final class Store<State, Action> {
...
+        return self._isInvalidated() || isInvalid?(self.state.value) == true
+      }
+    let fromChildAction = {
+      BindingLocal.isActive && isInvalid() ? nil : fromChildAction($0, $1)
+    }
+    var isSending = false
+    let childStore = Store<ChildState, ChildAction>(
+      initialState: initialChildState
+    ) {
+      Reduce(internal: { [weak self] childState, childAction in
```
stephencelis, line 493: We should probably explore using a dedicated reducer conformance for scoping (as was done before with `rescope`), though it'd be nice to measure the difference.

iampatbrown: I could potentially contain some of the strong/weak dance too.

Sources/ComposableArchitecture/Store.swift
iampatbrown: No apprehensions :) I'm a fan of the cached approach. 
Scoping via a key path is probably preferable. Computed properties might be an alternative for people using a closure as well.

pyrtsa:
Could you explain how this affects the lifetimes of child stores? Will the parent retain them for as long as it lives, or if not then when will a child be removed from the `children` dictionary?

stephencelis:
> Could you explain how this affects the lifetimes of child stores? Will the parent retain them for as long as it lives, or if not then when will a child be removed from the children dictionary?

@pyrtsa Child stores live as long as they are "valid". Optional/presented children are removed when the associated state goes away. While this means some stores may live longer than previously, where they'd only live as long as a view is in the hierarchy, we think it won't ever be an inordinate number.

acosmicflamingo:
Many of the custom `UICollectionViewListCell` subclasses I have in my codebase are tuples and am wondering how I could leverage caching from this PR in that use-case. For example, here's how I register my supplementary views:
```swift
let footerRegistration = UICollectionView.SupplementaryRegistration<SupplementaryView>(
  elementKind: UICollectionView.elementKindSectionFooter
) { [weak store] footerView, elementKind, indexPath in
  guard let store else { return }

  store.scope(
    state: {
      let item = $0.sections[indexPath.section].footer
      return (
        item: item,
        elementKind: elementKind
      )
    },
    action: { $0 }
  ).ifLet { [weak footerView] store in
    footerView?.update(with: store)
  }
  .store(in: &footerView.cancellables)
}
```

Would there be a way I could benefit from cached scoping if I am using closures to transform a parent state to a tuple child state? I imagine the answer is no because I think 99% of the time the scoped state is a subset of original state, but that can't be said for what I'm doing in the example above (elementKind value does not come from store).

stephencelis:
@acosmicflamingo You can always define computed properties and subscripts to make new key paths. In this case I think you could add a custom subscript that takes the element kind.

Something like:

```swift
extension Feature.State {
  subscript(
    section section: Int, elementKind elementKind: ElementKind
  ) -> (item: Item, elementKind: ElementKind) {
    (item: self.sections[section].footer, elementKind: elementKind)
  }
}
```

And then:

```swift
store.scope(
  state: \.[section: indexPath.section, elementKind: elementKind],
  action: { $0 }
).ifLet { [weak footerView] store in
  footerView?.update(with: store)
}
.store(in: &footerView.cancellables)
```

acosmicflamingo:
@stephencelis BRILLIANT! Thank you for the reply AND example too!

acosmicflamingo:
@stephencelis just tried your example and am so close to get it working! However, how do I deal with extracting an enum case from an IdentifiedArray of them since extraction is always optional?
```swift
//  MARK: - Store Scoping Support

extension AppearanceSettings.State {
  // preferSystemColorSchemeCellRegistration
  subscript(
    itemId: UUID,
    sectionIndex: Int,
    prefersToggleSwitchBelowTextInLandscapeMode: Bool
  ) -> SwitchRowCell<AppearanceSettings.Action>.State {
    let item = (/AppearanceSettingsItem.preferredSystemColorSchemeRow).extract(
      from: section.items[id: itemId]
    )
    return (
      item: item,
      sectionIndex: sectionIndex,
      prefersToggleSwitchBelowTextInLandscapeMode: true
    )
  }
}
```

Integrating that to my cell registration logic, I also get this:
image

Removing the extraneous labels gives me this:
image

stephencelis:
@acosmicflamingo I believe you want the subscript to return an optional since it can fail:

```swift
extension AppearanceSettings.State {
  // preferSystemColorSchemeCellRegistration
  subscript(
    itemId itemId: UUID,
    sectionIndex sectionIndex: Int,
    prefersToggleSwitchBelowTextInLandscapeMode prefersToggleSwitchBelowTextInLandscapeMode: Bool
  ) -> (
    item: SwitchRowItem<AppearanceSettings.Action>,
    sectionIndex: Int,
    prefersToggleSwitchBelowTextInLandscapeMode: Bool
  )? {
    guard
      let item = (/AppearanceSettingsItem.preferredSystemColorSchemeRow).extract(
        from: section.items[id: itemId]
      )
    else { return nil  
    return (
      item: item,
      sectionIndex: sectionIndex,
      prefersToggleSwitchBelowTextInLandscapeMode: prefersToggleSwitchBelowTextInLandscapeMode
    )
  }
}
```

(I've also doubled up the parameter names in the subscript so you don't lose them at the call site.)

acosmicflamingo:
@stephencelis oh, of COURSE! Yes, there's certainly nothing stopping me from returning an optional tuple. I guess I didn't think about that because when I read this in the Slack channel, it made it sound like I no longer have to do some of those long guard unwrapping lines like this:
image
 
Did I take away the wrong information from that?
 
Also, IT WORKS WOOHOO!!! That's AMAZING!!! My cell registration closure is MUCH cleaner now :D thanks a bunch for the help!
```swift
let preferSystemColorSchemeCellRegistration = UICollectionView.CellRegistration<
  SwitchRowCell<Action>, UUID
> { [weak store] cell, indexPath, itemId in 
    guard let store else { return }

    store.scope(
      state: \.[
        itemId: itemId,
        sectionIndex: indexPath.section,
        prefersToggleSwitchBelowTextInLandscapeMode: true
      ],
      action: \.self
    ).ifLet { [weak cell] store in
      cell?.update(with: store)
    }
    .store(in: &cell.cancellables)
  }
}
```

acosmicflamingo:
LOL swift-format thinks this is an array and keeps adding a comma at the end:
```swift
state: \.[
  itemId: itemId,
  sectionIndex: indexPath.section,
  prefersToggleSwitchBelowTextInLandscapeMode: true,
],
```

stephencelis:
> Did I take away the wrong information from that?

I just didn't fully digest your example. In my code snippet I was channeling the optionality of a simple `IdentifiedArray.subscript(id:)` :smile:


> Also, IT WORKS WOOHOO!!! That's AMAZING!!! My cell registration closure is MUCH cleaner now :D thanks a bunch for the help!

🥳

Sources/ComposableArchitecture/Documentation.docc/Articles/MigratingTo1.5.md
```diff
@@ -0,0 +1,165 @@
...
+```swift
+ChildView(
+  store: self.store.scope(
+    state: \.child, 
+    action: { .child($0) }
+  )
+)
+```
+
+However, as of version 1.5 of the Composable Architecture, the version of
```
Muhammed9991, line 41: Sneak peak 🫣

Sources/ComposableArchitecture/Documentation.docc/Articles/MigratingTo1.5.md
```diff
@@ -0,0 +1,165 @@
...
+)
+```
+
+In practice you could typically use key paths for the `state` transformation since key path literals
+can be promoted to closures. That means often scoping looked something like this:
+
+```swift
+ChildView(
+  store: self.store.scope(
+    state: \.child,
```
Muhammed9991, line 35: Shouldn't this be: 
```suggestion
    state: \.$child, 
```

Similar instance in docs id the above is valid

stephencelis: The `$` is for features using property wrappers like `PresentationState`, so I don't think it's needed for this example.

acosmicflamingo:
I don't know if this is a bug or I need to treat my specific use-case a little differently. I have a MediaLibraryViewController whose collection view displays songs sorted either by "recently played", "song name", "artist name", or "album name". Tapping the UISegmentedControl button will refresh the snapshot accordingly.

image

The issue I'm finding is that when I have adopted the subscripting solution to handle turning a state to a tuple, having the most up-to-date header view is not guaranteed. The items are certainly sorted correctly so this is purely an issue with the way the header view is being updated. I don't know if this is a race condition or if something like this should just work and I need to investigate further on my end, but I thought I'd ask in the off-chance that someone knows something off the top of their head.

For reference, this is what my CollectionViewDataSource file looks like:
```swift
// swiftlint:disable empty_line_after_super function_body_length
// swiftformat:disable braces indent large_tuple trailingCommas

import CFKCollectionViews
import ComposableArchitectureExtensions
import Foundation
import LocalDatabaseClient
import UICollectionViewModels
import UIKit

extension MediaLibraryViewController {
  public typealias Section = MediaLibrarySection
  public typealias Item = UUID

  public class CollectionViewDataSource: TCACollectionViewDiffableDataSource<Section, Item> {
    public init(
      store: StoreOf<MediaLibrary>,
      collectionView: UICollectionView,
      bindingOptions: DataSourceReloadingBindingOptions = .none,
      dynamicSnapshot: Bool = false
    ) {
      let headerRegistration = UICollectionView.SupplementaryRegistration<
        MediaLibrarySectionCell
      >(
        elementKind: UICollectionView.elementKindSectionHeader
      ) {
        [weak store] headerView, _, indexPath in
        guard let store else { return }

        store.scope(
          state: \.[sectionIndex: indexPath.section],
          action: \.self
//          state: {
//            let sectionIndex = indexPath.section
//            guard let section = $0.snapshotData.sections[safeIndex: sectionIndex]
//            else { return nil }
//
//            return (
//              text: section.text,
//              placeholder: nil
//            )
//          },
//          action: { $0 }
        ).ifLet { [weak headerView] store in
          headerView?.update(with: store)
        }
        .store(in: &headerView.cancellables)
      }

      let mediaLibraryItemCellRegistration = UICollectionView.CellRegistration<
        MediaLibraryItemCell, UUID
      > { [weak store] cell, _, itemId in
        guard let store else { return }

        store.scope(
          state: \.[itemId: itemId],
          action: \.self
        ).ifLet { [weak cell] store in
          cell?.update(with: store)
        }
        .store(in: &cell.cancellables)
      }

      super.init(
        collectionView: collectionView,
        bindingOptions: .all,
        dynamicSnapshot: dynamicSnapshot
      ) { collectionView, indexPath, itemId in
        collectionView.dequeueConfiguredReusableCell(
          using: mediaLibraryItemCellRegistration,
          for: indexPath,
          item: .init(itemId)
        )
      }

      supplementaryViewProvider = {
        collectionView, _, indexPath -> UICollectionReusableView? in
        collectionView.dequeueConfiguredReusableSupplementary(
          using: headerRegistration,
          for: indexPath
        )
      }
    }
  }
}

// MARK: - Store Scoping Support

extension MediaLibrary.State {
  // headerRegistration
  subscript(sectionIndex sectionIndex: Int) -> MediaLibrarySectionCell.State? {
    guard let section = snapshotData.sections[safeIndex: sectionIndex]
    else { return nil }

    return (
      text: section.text,
      placeholder: nil
    )
  }

  // mediaLibraryItemCellRegistration
  subscript(itemId itemId: UUID) -> MediaLibraryItemCell.State? {
    let mediaLibraryItem = data[id: itemId]
    var itemForSegment: LocalDatabaseClient.MediaItem?
    // TODO: see if this brace style can be improved by moving first 'case let' next to 'if'
    if
      case let .segment(segment) = mediaLibraryItem,
      case let .item(item) = data[id: segment.ofItemId]
    {
      itemForSegment = item
    }
    let sortMode = segmentedControl.selectedSegment
    let isSearching = searchBar.isSearching
    let searchBarText = searchBar.textField.text
    let displaySeparator = itemId != snapshotData.first(where: {
      $0.value.contains(itemId)
    })?.value.last

    return (
      mediaLibraryItem: mediaLibraryItem,
      itemForSegment: itemForSegment,
      sortMode: sortMode,
      isSearching: isSearching,
      searchBarText: searchBarText,
      displaySeparator: displaySeparator
    )
  }
}

// swiftformat:enable braces indent large_tuples trailingCommas
// swiftlint:enable empty_line_after_super function_body_length

```

acosmicflamingo:
How strange...using print statements within following line, I confirm that the values are indeed correct but the supplementary view does not reflect that: https://github.com/pointfreeco/swift-composable-architecture/blob/store-tree/Sources/ComposableArchitecture/Store.swift#L390.
image

mbrandonw:
@acosmicflamingo Just to confirm, this is new behavior you are seeing with this branch? This does not happen in 1.4 and you made no other changes to your app to support the new `scope` operation other than using case paths? I just want to make sure there are not other changes in your app that are unrelated to this PR that could be causing the problem.

Also if you can share a minimal project that demonstrates the behavior that would be best.

acosmicflamingo:
@mbrandonw that is correct. The commit where the regression occurs only includes changes where action parameters are using the keyPath syntax over the closure. I also verified that the scope where the issue occurs only exists in store-tree and not 1.4.2. I'll try to create a reproducer soon.

acosmicflamingo:
@mbrandonw this might be quite a corner case that only impacts me because I am doing something a little different in order to work around an issue that has to do with how UICollectionView itself handles applying snapshots when the snapshot sectionIdentifiers are dynamic. There was a workaround that an Apple engineer in the UICollectionView team recommended years ago where to work around this issue I should apply an empty snapshot first and then the actual snapshot to a data source...so it might not be a coincidence that the instances where I'm getting these issues are the same ones that involve this workaround. Any time a UICollectionReusableView instance is brought into view, it contains the right data as well.

I'll have to do further investigation but I do not think that this is a blocker for the store-tree release. In all other cases in my app where I utilize diffable data source snapshot API, everything works perfectly.

mbrandonw:
@acosmicflamingo Thanks for following up. Please do let us know if you uncover anything fishy happening.

acosmicflamingo:
Hmm...how does a `Store` know when to clear out the `Store` cache whose key comprises the keyPath and casePath of the state and action, respectively? With some print statements, I have found that a store instance theoretically has the right value, but the scoped store does not reflect that:
```swift
  if
    let item = store.withState(\.snapshotData.sections[safeIndex: indexPath.section])
  {
    print("Store.withState:", item)
  }
  if let scopedState = store.scope(
    state: \.snapshotData.sections[safeIndex: indexPath.section],
    action: \.self
  ).withState({ $0 }) {
    print("ScopedStore.withState", scopedState)
  }
  
Prints:
Store.withState: string("A")
ScopedStore.withState string("Next Week")
Store.withState: string("B")
ScopedStore.withState string("Today")
```

When I asked this question, I thought that there was some race condition going on (and maybe there still is), but now I can definitely say that a `Store` instance does certainly have the value I expect, but scoping to that with the cache gives me the wrong value. I know you asked for a reproducer but it's going to take a whole lot more time than I anticipated to do that.

acosmicflamingo:
I'm pretty confident I have found the root of the problem and it looks to be a race condition. In my scenario, I'm extracting some scoped state within a cell registration closure and it appears that by getting the cached `Store`, it's so absolutely fast that my `SupplementaryView` instance is extracting some state before the stateSubject publisher has even had a chance to emit a new stream and update itself with a new value: https://github.com/pointfreeco/swift-composable-architecture/blob/50b5cc6fffc09bf2c2a1ecca3058aa4b8e61b4da/Sources/ComposableArchitecture/Store.swift#L559C10-L559C10. 

I can confirm this with some trusty print statements:
```swift
ChildState has updated in scoped subscription closure with: Optional(MediaLibraryFeature.MediaLibrarySection.string("A"))
Store.withState: string("A")
Returning childStore
ScopedStore.withState string("A")
MediaLibrarySectionCell updated with A
Store.withState: string("B")
Returning childStore
ScopedStore.withState string("Today")
MediaLibrarySectionCell updated with Today
ChildState has updated in scoped subscription closure with: Optional(MediaLibraryFeature.MediaLibrarySection.string("B"))
ChildState has updated in scoped subscription closure with: Optional(MediaLibraryFeature.MediaLibrarySection.string("A"))
ChildState has updated in scoped subscription closure with: Optional(MediaLibraryFeature.MediaLibrarySection.string("B"))
```
image

Notice that the first header view is correct, and the logs show that the subscription for the letter "A" is executed before any of the `Store.scope` functions within the cell registration closure have printed anything. Now notice that the order is flipped on the incorrect "Today" header at the bottom (which should be the letter "B"). It's very interesting that the logs show that `MediaLibrarySectionCell` have updated with the out-of-date state before "ChildState has updated in scoped subscription closure with: Optional(MediaLibraryFeature.MediaLibrarySection.string("B"))" ever appeared.

mbrandonw:
Thanks for the details, but without a minimal reproducing project there really isn't much help we can offer. I think that would be the best path forward.

acosmicflamingo:
@mbrandonw Yep, I understand. Since I've narrowed the problem down, it'll be easier for me to create a reproducer. In addition, there are things I can do on my end to work around this (like slapping on a `receive(on: DispatchQueue.main)`. Thanks!

acosmicflamingo:
Here's a reproducer: https://github.com/acosmicflamingo/StoreScopingCacheReproducer.git. To reproduce, simply tap on the second segmented control button and you'll see that sometimes the header does not reflect what should be shown. These logs that I have also confirm what I'm seeing:
```swift
/*
Initial:
Store.withState: first("9")
ScopedStore.withState first("9")
MusicLibrarySectionCell updated with 9
Store.withState: first("A")
ScopedStore.withState first("A")
MusicLibrarySectionCell updated with A
Store.withState: first("B")
ScopedStore.withState first("B")
MusicLibrarySectionCell updated with B
Store.withState: first("C")
ScopedStore.withState first("C")
MusicLibrarySectionCell updated with C
Store.withState: first("D")
ScopedStore.withState first("D")
MusicLibrarySectionCell updated with D
<img width="368" alt="image" src="https://github.com/pointfreeco/swift-composable-architecture/assets/67525430/dc2aa161-0db4-490d-8e2a-632feeee1452">


After tapping second segment:
Store.withState: first("E")
ScopedStore.withState first("E")
MusicLibrarySectionCell updated with E
Store.withState: first("W")
ScopedStore.withState first("9")
MusicLibrarySectionCell updated with 9
Store.withState: first("S")
ScopedStore.withState first("S")
MusicLibrarySectionCell updated with S
Store.withState: first("H")
ScopedStore.withState first("H")
MusicLibrarySectionCell updated with H
Store.withState: first("D")
ScopedStore.withState first("D")
MusicLibrarySectionCell updated with D
Store.withState: first("E")
ScopedStore.withState first("E")
MusicLibrarySectionCell updated with E
Store.withState: first("S")
ScopedStore.withState first("S")
MusicLibrarySectionCell updated with S
Store.withState: first("H")
ScopedStore.withState first("H")
MusicLibrarySectionCell updated with H
Store.withState: first("W")
ScopedStore.withState first("9")
MusicLibrarySectionCell updated with 9
Store.withState: first("D")
ScopedStore.withState first("D")
MusicLibrarySectionCell updated with D
Store.withState: first("W")
ScopedStore.withState first("9")
MusicLibrarySectionCell updated with 9
Store.withState: first("H")
ScopedStore.withState first("H")
MusicLibrarySectionCell updated with H
Store.withState: first("S")
ScopedStore.withState first("S")
MusicLibrarySectionCell updated with S
Store.withState: first("E")
ScopedStore.withState first("E")
MusicLibrarySectionCell updated with E
<img width="363" alt="image" src="https://github.com/pointfreeco/swift-composable-architecture/assets/67525430/d7d90991-2d5b-49fa-8a64-72e06a524812">
*/
```

It is a race condition so sometimes it will work, in which case you just have to rebuild and try tapping again. Any of the headers could have the issue too; it's just a matter of looking in the log and seeing that `Store.withState` does not coincide with `ScopedStore.withState`. Sorry to dump cold water on the potential store-tree merge but maybe we'll find what I'm doing wrong :)

mbrandonw:
@acosmicflamingo It's hard to say. In order for us to best look into the issue we think we need a smaller reproducing example. The current one is quite complex and it's hard to tell if it's an issue with the library or in UIKit. 

Ideally the problem could be expressed as a simple unit test, but if that's not possible then an integration test with fewer moving parts would be best.

acosmicflamingo:
@mbrandonw how is this latest commit (07930c863d1a69b850cbc667caebd0220defcdea)? If you check it out, you'll notice that I've removed all possible models, extensions, and view/controllers that makes the app more complex than it needs to be (and hopefully makes it easier to debug this issue). I've also added an integration test that will fail about half the time:
```
final class StoreScopingCacheReproducerUITests: XCTestCase {
  override func setUpWithError() throws {
    continueAfterFailure = false
  }

  func testSectionHeaderText() throws {
    let app = XCUIApplication()
    app.launch()

    // Always passes
    app.buttons["A to Z"].tap()
    let headerA = app.collectionViews.element(boundBy: 0).cells.element(boundBy: 0)
    XCTAssertEqual(headerA.label, "9")

    // Does not always pass
    app.buttons["Z to A"].tap()
    let headerZ = app.collectionViews.element(boundBy: 0).cells.element(boundBy: 0)
    XCTAssertEqual(headerZ.label, "W")
  }
}
 
Result: Executed 100 tests, with 58 failures (0 unexpected) in 527.966 (528.031) seconds
```

PR #2549, opened 2023-11-07
[FIX] Updated Swift compiler version to 5.7.1 to follow Apple's policy

jaesung-0o0:
Since April, the apps and SDKs must be built in Xcode 14.1 + to submit to the AppStore. To support Xcode 14.1+, I modified the swift version to **5.7.1** from 5.7

> Reference: [developer.apple.com/news](https://developer.apple.com/news/?id=jd9wcyov)

After this changes, it no longer needs to worry about the apple-side bug that caused `ReducerOf` to specify support for versions 5.7.1+ : https://github.com/apple/swift/issues/60550

```swift
- #if swift(>=5.7.1)
public typealias ReducerOf<R: Reducer> = Reducer<R.State, R.Action>
- #endif
```
---
In addition, since April 2024, the apps and SDKs must be built in Xcode 15+ to submit to the AppStore.  Therefore, versions under 5.9 will no longer be needed to be supported after that date

> Reference: [developer.apple.com/ios/submit](https://developer.apple.com/ios/submit/#:~:text=Please%20note%20that,iOS%2017%20SDK.)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.7
+// swift-tools-version:5.7.1
 
 import PackageDescription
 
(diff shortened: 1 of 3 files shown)


stephencelis (approved):
Good call, thanks!

PR #2550, opened 2023-11-07
[Suggestion] Modify description to be clearer

jaesung-0o0:
I modified the description of `Reducer`: `"current state given an action"` -> `"current state based on the given action."`

However, because English is not my first language, so please consider this PR as a suggestion rather than a correction. Thanks

```
- The `reduce` method's first responsibility is to mutate the feature's current state given an action.
+ The `reduce` method's first responsibility is to mutate the feature's current state based on the given action.
```

### DIFF
--- a/Sources/ComposableArchitecture/Reducer.swift
+++ b/Sources/ComposableArchitecture/Reducer.swift
@@ -41,7 +41,7 @@
 /// }
 /// ```
 ///
-/// The `reduce` method's first responsibility is to mutate the feature's current state given an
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hi @jaesung-0o0, thank you for the PR, but we think the current wording is correct. The `reduce` method is handed arguments from the outside, and so saying "given an action" should be clear enough.

I am going to close this for now, but feel free to reply if you have more to add to the topic.

PR #2556, opened 2023-11-09
Short circuit == when ordered sets don't have same size.

mbrandonw:
Just something we discussed today.


Examples/Integration/Integration.xcodeproj/project.pbxproj
```diff
@@ -838,8 +838,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/pointfreeco/swift-snapshot-testing.git";
 			requirement = {
-				kind = revision;
-				revision = bbc9ec6e41ba964a5d04371af0acfb574cadafbd;
+				kind = upToNextMajorVersion;
+				minimumVersion = 1.0.0;
```
mbrandonw, line 842: The integration target was depending on a SHA of snapshot testing, so fixed it to point to an actual version.

stephencelis (approved):

PR #2558, opened 2023-11-09
Add effect delay

Finemas:
Add a delay effect next to the debounce and throttle functions because we need to delay emitted values without debouncing them. There are two options for this function: one is deprecated and uses the combine operator, and the other uses an asynchronous clock from async&await.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Delay.swift
+++ b/Sources/ComposableArchitecture/Effects/Delay.swift
@@ -17,6 +17,10 @@ extension Effect {
   ///     canceled before starting this new one.
   /// - Returns: An effect that publishes events  after a specified time.
   ///
+  @available(macOS, deprecated: 13.0, message: "use delay with a clock and tolereance parameters insted (`.delay(id:for:tolerance:clock:cancelInFlight)`).")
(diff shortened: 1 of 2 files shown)


stephencelis:
@Finemas Can you share a use case for this? Is there any reason you can't call `clock.sleep` in an `Effect.run` instead?

FWIW Ideally we would eventually get rid of `Effect.debounce` and `Effect.throttle`, but there's no easy way to achieve throttling in Swift concurrency at the moment and we haven't tackled the problem ourselves...yet.

Finemas:
Well, it's simply a convenient function.

We extensively use the delay function in our calling application e.g. in the auto-pickup feature, using debounce is not practical because we want to accept all calls. This is merely the first use case that comes to mind and I suspect it's not the only situation where the delay would be useful.

Unfortunately, in our SDK, we still use a publisher, and the operation in Effect is internal, so I couldn't directly integrate this extension into our project.

stephencelis:
@Finemas We do have a goal to make `Effect` more publicly extensible, but that will take some time as we further decouple its internals from Combine. When we do it'll make it possible to define effect operators like this.

I do think we want to avoid adding library surface area for this, though. I think our recommendation for now would be to push the `.delay` call into the `.publisher`. Hope that's not too inconvenient!

I'm going to close this for now, but if you'd like to continue discussing the problem, please open a discussion.

PR #2562, opened 2023-11-12
Update Korean Translation of README.md

Achoo-kr:
Former Korean Translation was the pre-release version which is talking about Environment. I updated the Korean translation with current TCA 1.0


stephencelis:
/cc @pilgwon 

We'll take this change, thanks! If the original submitter wants to PR an updated version, we'll take that in the future.

PR #2563, opened 2023-11-12
Update Japanese Translation of README.md

Achoo-kr:
The former version of Japanese translation was outdated. You can check it in the answer of FAQ part. It still talks about pullback which is now gone.


stephencelis:
/cc @kalupas226 

We'll take this change, thanks! If the original submitter wants to PR an updated version, we'll take that in the future.

PR #2570, opened 2023-11-15
Re-adds `Reducer.forEach` method removed in 1.4

rhysm94:
TCA 1.4 perhaps unintentionally removed the `Reducer.forEach` code which allowed creating a `ForEachReducer` with a `AnyCasePath<Action, (ID, ElementAction)>` as the `action` parameter.
Unfortunately, it seems that:

`AnyCasePath<Action, (ID, ElementAction)>`, and 
`AnyCasePath<Action, (id: ID, action: ElementAction)>`

are not (always) considered equivalent by the Swift compiler.

The helper library, [TCACoordinators](https://github.com/johnpatrickmorgan/TCACoordinators), relies on the previous behaviour by passing an `AnyCasePath<Action, (ID, ElementAction)>`. I have tried updating TCACoordinators to use `AnyCasePath<Action, (id: ID, action: ElementAction)>`, but this has the effect of breaking TCACoordinators when used with TCA 1.3. Without re-adding this particular method, TCACoordinators cannot compile in a way that works with both TCA 1.3 and TCA 1.4. With this small change, TCACoordinators compiles and works just fine!

### DIFF
--- a/Sources/ComposableArchitecture/Reducer/Reducers/ForEachReducer.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/ForEachReducer.swift
@@ -146,50 +146,6 @@ extension Reducer {
   @warn_unqualified_access
   public func forEach<ElementState, ElementAction, ID: Hashable, Element: Reducer>(
     _ toElementsState: WritableKeyPath<State, IdentifiedArray<ID, ElementState>>,
-    action toElementAction: CaseKeyPath<Action, (id: ID, action: ElementAction)>,
-    @ReducerBuilder<ElementState, ElementAction> element: () -> Element,
-    fileID: StaticString = #fileID,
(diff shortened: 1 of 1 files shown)


stephencelis:
@rhysm94 Thanks for pointing out the problem! I think we introduced more churn here than we should have. I've updated your PR to fix the old `AnyCasePath` API and require the new `CaseKeyPath` API go through `IdentifiedAction`. While it's technically a small breaking change, the API was introduced already deprecated and we don't think folks should have adopted it.

rhysm94:
@stephencelis Thanks for that! That's great, I'm glad it was possible to make it simpler!
Tested the change on my own mirror branch, and it worked perfectly. Happily, TCACoordinators works without issue with these changes in!

PR #2572, opened 2023-11-15
Add empty public init to ReducerMacro

natemann:
This adds an empty public init `public init() {}` to the ReducerMacro. If the reducer struct is marked 'public', and an existing public initializer does not exist, it will add one.

### DIFF
--- a/Sources/ComposableArchitectureMacros/ReducerMacro.swift
+++ b/Sources/ComposableArchitectureMacros/ReducerMacro.swift
@@ -8,6 +8,20 @@ import SwiftSyntaxMacros
 public enum ReducerMacro {
 }
 
+extension ReducerMacro: MemberMacro {
+  public static func expansion<D: DeclGroupSyntax,C: MacroExpansionContext>(
+    of node: AttributeSyntax,
+    providingMembersOf declaration: D,
+    in context: C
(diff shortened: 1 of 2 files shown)


stephencelis:
@natemann Thanks for exploring this and submitting a PR. I think it is probably functionality we do not want, though. Remember that reducers can have properties, like explicitly configured dependencies, and so this `init() {}` would break their builds. While it may be possible for the macro to detect these properties, I'm not sure if there are other edge cases that could break things.

At the end of the day, luckily the need to write a public initializer only affects a small portion of users, those who modularize, and the need to be explicit with these initializers is how Swift was designed. It is also just a few characters you need to manually add, so hopefully not a dealbreaker to adopting modular TCA :wink:

lukeredpath:
Another thing to consider in a modularised codebase is that because reducers form part of the type signature of the entire reducer tree, you need to make reducers public even if they are only intended to be used from within a feature module - adding or not adding a public initializer is the main way of controlling which reducers can be used from outside a module so making this opt-in makes sense.

stephencelis:
@lukeredpath Ideally modularized code bases could choose to hide that type information using opaque `some` types, though we've found that release builds can't always resolve those symbols unless they're public.

PR #2574, opened 2023-11-15
Fix syntax error in Migrate To 1.4.

Ryu0118:
I found and fixed an error in the syntax of enum.


mbrandonw (approved):
Good catch, thanks!

PR #2577, opened 2023-11-15
Fix `swift-case-paths` being pinned to 1.1.0 the wrong way

bdolewski-intellias:
I think that `Package.swift` and `Package@swift-5.9.swift` should be the same and we should not pin `swift-case-paths` to given release by `branch:`.

Also it kinda breaks SPM in my project so that's how we were able to spot this issue.


stephencelis (approved):
Thanks! Not sure how it was working for us.

PR #2586, opened 2023-11-21
Depend on and export DependenciesMacros

tgrapperon:
This PR exports `DependenciesMacros`, which should ease `DependencyClient`'s adoption.
I don't know if we should update sample code. We can maybe refresh `SpeechRecognition` to adopt this macro?


stephencelis (approved):
Looks good to us, thanks for the idea!

PR #2593, opened 2023-11-26
Observable Architecture

stephencelis:
This PR introduces Observation tools to the Composable Architecture.

Discussion with more details: https://github.com/pointfreeco/swift-composable-architecture/discussions/2594


Sources/ComposableArchitecture/Documentation.docc/Articles/MigratingTo1.6.md
```diff
@@ -0,0 +1,638 @@
...
+```swift
+.sheet(
+  item: $store.scope(
+    state: \.destination?.editForm,
+    action: \.destination.editForm
+  )
+) { store in
+  ChildView(store: store)
+}
+```
```
0xcefaedfe, line 368: The above currently doesn't seem to work with the `@Bindable` property wrapper as its type will be `Bindable<Store<State, Action>>` and that type doesn't have a `scope` method defined that would return a `Binding`. 

Copying the implementation of the corresponding method from the `Binding` extension sort of does the job, but there's no `transaction` property defined on `Bindable` and I am not familiar enough with this topic to suggest a proper solution.

```
@available(iOS 17.0, *)
extension Bindable {
  public func scope<State: ObservableState, Action, ChildState, ChildAction>(
    state: KeyPath<State, ChildState?>,
    action: CaseKeyPath<Action, PresentationAction<ChildAction>>
  ) -> Binding<Store<ChildState, ChildAction>?>
  where Value == Store<State, Action> {
    let isInViewBody = PerceptionLocals.isInPerceptionTracking
    return Binding<Store<ChildState, ChildAction>?>(
      get: {
        // TODO: Is this right? Should we just be more forgiving in bindings?
        PerceptionLocals.$isInPerceptionTracking.withValue(isInViewBody) {
          self.wrappedValue.scope(
            state: state,
            action: action.appending(path: \.presented)
          )
        }
      },
      set: {
        if $0 == nil, self.wrappedValue.stateSubject.value[keyPath: state] != nil {
//          Bindable doesn't have a transaction property --> self.transaction($1).wrappedValue.send(action(.dismiss))
        }
      }
    )
  }
}
```

mbrandonw: Hey @0xcefaedfe, thanks for bringing this up. It is true that `@Bindable` doesn't work quite as seamlessly as we hoped (or as the docs suggest). For now you can just use `@State` and we will look into `@Bindable` in more depth soon, or perhaps introduce a different tool.

stephencelis: @0xcefaedfe We've pushed helpers for `@Bindable` to this branch now. Thanks for letting us know!

Sources/ComposableArchitecture/Documentation.docc/Articles/MigrationGuides/MigratingTo1.7.md
```diff
@@ -0,0 +1,862 @@
...
+    ActivityView(store: store)
+  }
+case .settings:
+  if let store = store.scope(state: \.settings, action: \.settings) {
+    SettingsView(store: store)
+  }
+}
+```
+
+## Replacing @PresentationState with @Presentation
```
tevelee, line 302:
```suggestion
## Replacing @PresentationState with @Presents
```

PR #2603, opened 2023-11-30
Store Invalidation

stephencelis:
Still working out the subtleties and details here, but let's get a PR going that can be checked with CI.


acosmicflamingo:
Could it be that the missing `AnyStore` conformance is the root of the problems you're seeing? Without that conformance, any grandchildren or great grandchildren, etc is not going to be called, right?

stephencelis:
@acosmicflamingo That may cause other issues for sure, but not one of the issues this PR intends to address.

PR #2604, opened 2023-12-01
Add `@Presents` macro for observable presentation

stephencelis:
While it would be nice for the `@PresentationState` property wrapper to "just work" with TCA's upcoming observable tools, sadly that does not seem to be the case. Adding observation directly to `@PresentationState`, as we have done with the beta so far, can break existing projects due to the additional observation. This primarily manifests itself in projects that present navigation stacks, where the `@PresentationState` observation can cause the navigation hierarchy to recompute and trigger SwiftUI bugs.

The best we've come up with so far is introducing a brand new macro that automatically wraps a property with `@PresentationState` _and_ instruments it with observation.

We're open to other ideas, and we do have future plans to eliminate the need for a property wrapper or macro at all, but till then this offers a non-breaking upgrade path!

### DIFF
--- a/Tests/ComposableArchitectureMacrosTests/PresentsMacroTests.swift
+++ b/Tests/ComposableArchitectureMacrosTests/PresentsMacroTests.swift
@@ -1,0 +1,163 @@
+import ComposableArchitectureMacros
+import MacroTesting
+import SwiftSyntaxMacros
+import SwiftSyntaxMacrosTestSupport
+import XCTest
+
+final class PresentsMacroTests: XCTestCase {
+  override func invokeTest() {
+    withMacroTesting(
+      isRecording: true,
+      macros: [PresentsMacro.self]
+    ) {
+      super.invokeTest()
+    }
+  }
+
+  func testBasics() {
+    assertMacro {
+      """
+      struct State {
+        @Presents var destination: Destination.State?
+      }
+      """
+    } expansion: {
+      #"""
+      struct State {
+        var destination: Destination.State? {
+          @storageRestrictions(initializes: _destination)
+          init(initialValue) {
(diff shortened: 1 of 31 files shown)


Tests/ComposableArchitectureMacrosTests/PresentsMacroTests.swift
```diff
@@ -0,0 +1,163 @@
...
+        ┬─────────────────
+        ╰─ 🛑 '@PresentationState' property wrapper cannot be used directly in '@ObservableState'
+           ✏️ Use '@Presents' instead
+      }
+      """
+    } fixes: {
+      """
+      @ObservableState
+      struct State: Equatable {
+        @Presents
```
stephencelis, line 134: The `FixItApplier` we took from the SwiftSyntax repo does not appear to be up to task here. We should check to see if the latest upstream version has any bug fixes would address this.

Tests/ComposableArchitectureMacrosTests/PresentsMacroTests.swift
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        var $destination: ComposableArchitecture.PresentationState<Destination.State> {
+          get {
+            _destination.projectedValue
+          }
+          set {
+            _destination.projectedValue = newValue
+          }
+        }
```
stephencelis, line 54: Not sure if we should be doing any `access` or `withMutation` here...

oliverfoggin (approved):
This fixes the navigation issue I had previously reported.

Also, this fixed an issue I was having with a TextField that kept getting reset. I assume because the view hierarchy was getting inadvertently re-rendered and the TextField was involved in that somehow.

🔥

mbrandonw (approved):

PR #2605, opened 2023-12-01
Short-circuit the store cache from uncached stores

stephencelis:
1.5 introduced the store tree, which works well when the entire tree is cached, but can produce invalid stores when an uncached store caches a child store and is then reinitialized. While folks upgrading to 1.5 should be in a good state, adopting the new `scope` operator can introduce this problem. To mitigate, we should only cache from the root to nodes as far as the cache resolves.

This is a temporary limitation to avoid breaking existing apps. We should be able to remove this workaround when uncached scopes are fully deprecated and removed from the library.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -133,6 +133,7 @@ import SwiftUI
 /// of the store are also checked to make sure that work is performed on the main thread.
 public final class Store<State, Action> {
   private var bufferedActions: [Action] = []
+  fileprivate var canCacheChildren = true
   fileprivate var children: [AnyHashable: AnyObject] = [:]
   @_spi(Internals) public var effectCancellables: [UUID: AnyCancellable] = [:]
   var _isInvalidated = { false }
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -1092,7 +1096,11 @@ extension ScopedStoreReducer: AnyScopedStoreReducer {
         Logger.shared.log("\(storeTypeName(of: store)).scope")
       }
     if let id = id {
-      store.children[id] = childStore
+      if store.canCacheChildren {
+        store.children[id] = childStore
+      } else {
+        // runtimeWarn("Can't cache child store at '\(id)' in an uncached store")
```
stephencelis, line 1102: This runtime warning isn't super helpful while the library still has `WithViewStore`s that scope in an uncache-able way.

```suggestion
      } else {
        // runtimeWarn("Can't cache child store at '\(id)' in an uncached store")
```

Tests/ComposableArchitectureTests/StoreLifetimeTests.swift
```diff
@@ -0,0 +1,106 @@
...
+    XCTAssertEqual(2, childStore.withState(\.count))
+
+    parentStore = nil
+
+    childStore.send(.tap)
+    XCTAssertEqual(3, grandparentStore.withState(\.child.child.count))
+    XCTAssertEqual(3, childStore.withState(\.count))
+    grandparentStore.send(.incrementGrandchild)
+    XCTAssertEqual(4, grandparentStore.withState(\.child.child.count))
+    XCTAssertEqual(4, childStore.withState(\.count))
```
stephencelis, line 45: Both `childStore` assertions fail on `1.5.0...main`.

mbrandonw (approved):

PR #2608, opened 2023-12-02
Fix typo in MigratingTo1.6.md

Jimmy-Prime:
There is a typo in section `## Replacing navigation view modifiers with SwiftUI modifiers`


mbrandonw (approved):
Good catch, thanks!

PR #2610, opened 2023-12-02
Introduce @BindableStore for bindings in pre-iOS 17

mbrandonw:
Currently in #2594 we don't have a nice story for deriving bindings in pre-iOS 17. We recommend just holding onto stores with `@State`, but that can be problematic. SwiftUI has a nasty habit of over caching those stores for too long (see [this feedback](https://gist.github.com/mbrandonw/db904f2ad544da594eecd8c56ce86212)). Also, we just don't need the power of `@State`. We do our own store caching, and so there is no benefit.

However, the `@Bindable` property wrapper for deriving bindings _without_ `@State` only works on iOS 17+. So we now provide our own tool, `@BindableStore`, that can be used in pre-iOS 17 to deriving bindings, and is deprecated in iOS 17+.


Sources/ComposableArchitecture/Observation/BindableStore.swift
```diff
@@ -0,0 +1,57 @@
...
+public struct BindableStore<State, Action> {
+  public let wrappedValue: Store<State, Action>
+  public init(wrappedValue: Store<State, Action>) {
+    self.wrappedValue = wrappedValue
+  }
+
+  public var projectedValue: Binding<Store<State, Action>> {
+    Binding(
+      get: { self.wrappedValue },
+      set: { _ in }  // TODO: Should this technically allow assignment?
```
mbrandonw, line 54: I forget where we landed on this... we do want to support this right?

stephencelis: Yeah I think if SwiftUI allows it on `@Bindable` objects we should also allow it.

Sources/ComposableArchitecture/Observation/BindableStore.swift
```diff
@@ -0,0 +1,57 @@
...
+@available(tvOS, deprecated: 17, renamed: "Bindable")
+@available(watchOS, deprecated: 10, renamed: "Bindable")
+@propertyWrapper
+public struct BindableStore<State, Action> {
+  public let wrappedValue: Store<State, Action>
+  public init(wrappedValue: Store<State, Action>) {
+    self.wrappedValue = wrappedValue
+  }
+
+  public var projectedValue: Binding<Store<State, Action>> {
```
stephencelis, line 51: A small difference here is that `@Bindable` returns a `Bindable` here, so the equivalent would be projecting to `BindableStore` here...I think that's a pain and would require more overloads, though, but just in case we want to document or consider the difference.

stephencelis:
This does make me wonder again if `Perception` should backport `Bindable` instead...

stephencelis (approved):

PR #2612, opened 2023-12-04
Introduce @ViewAction(for:) macro.

mbrandonw:
This makes it possible to embrace the "view action" pattern in features.


Sources/ComposableArchitectureMacros/ViewActionMacro.swift
```diff
@@ -0,0 +1,158 @@
...
+  public static func expansion<Declaration, Context>(
+    of node: SwiftSyntax.AttributeSyntax,
+    providingMembersOf declaration: Declaration,
+    in context: Context
+  ) throws -> [SwiftSyntax.DeclSyntax]
+  where Declaration: SwiftSyntax.DeclGroupSyntax, Context: SwiftSyntaxMacros.MacroExpansionContext {
+    guard declaration.hasStoreVariable
+    else {
+      // TODO: Fix it to add `let store: StoreOf<<#Feature#>>`?
+      context.diagnose(
```
mbrandonw, line 16: I don't think this fix-it is actually possible since we need a node insertion rather than replacement. But maybe you know some fancy tricks!

Sources/ComposableArchitectureMacros/ViewActionMacro.swift
```diff
@@ -0,0 +1,158 @@
...
+        memberAccess.declName.baseName.text == "send"
+      {
+        context.diagnose(
+          Diagnostic(
+            node: decl,
+            message: MacroExpansionWarningMessage(
+              """
+              Do not use 'store.send' directly when using @ViewAction. Instead, use 'send'.
+              """
+            ),
```
mbrandonw, line 86: I went with a warning for these just in case we mis-diagnose something so it does not break people's builds. We could strengthen in the future if we are more comfortable with the diagnostics.

Sources/ComposableArchitectureMacros/ViewActionMacro.swift
```diff
@@ -0,0 +1,169 @@
...
+        store.send(.view(action))
+      }
+      @discardableResult
+      fileprivate func send(_ action: \(raw: rawType).Action.View, animation: Animation?) -> StoreTask {
+        store.send(.view(action), animation: animation)
+      }
+      @discardableResult
+      fileprivate func send(_ action: \(raw: rawType).Action.View, transaction: Transaction) -> StoreTask {
+        store.send(.view(action), transaction: transaction)
+      }
```
mbrandonw, line 63: Adding the 3 main variations of `send` for now. Can reconsider `send(_:while:)` later.

Sources/ComposableArchitecture/Documentation.docc/Articles/MigratingTo1.6.md
```diff
@@ -639,8 +639,101 @@ struct State {
...
+
+```swift
+@Reducer
+struct Feature
+  // ...
+  enum Action {
+    // ...
+
+    enum View: ViewAction {  // 👈
+      case loginButtonTapped
```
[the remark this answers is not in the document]
mbrandonw: Good catch, thanks!

Sources/ComposableArchitectureMacros/ViewActionMacro.swift
mbrandonw: Oh, well I guess we could replace the entire `declaration` node with a new version of itself that has a member added...

oliverfoggin:
Some quick Find-and-Replace-fu and it looks like this is all working AFAICT.

Sources/ComposableArchitecture/Observation/ViewAction.swift
```diff
@@ -0,0 +1,30 @@
...
+
+/// Defines the actions that can be sent from a view.
+///
+/// See the ``ViewAction(for:)`` macro for more information on how to use this.
+public protocol ViewAction<ViewAction> {
+  associatedtype ViewAction
+  static func view(_ action: ViewAction) -> Self
+}
+
+public protocol ViewActionable<State, Action> {
```
stephencelis, line 11: What do you think of `ViewActionSending` instead, since it describes that the view can send view actions?

stephencelis (approved):
Just one naming thing.

PR #2614, opened 2023-12-05
Don't show perception warnings in action closures.

mbrandonw:
I'm sure there is better ways to do this, but it seems to work. Unfortunately it's not possible to test without an integration test.

### DIFF
--- a/Sources/Perception/PerceptionRegistrar.swift
+++ b/Sources/Perception/PerceptionRegistrar.swift
@@ -222,9 +222,15 @@ extension PerceptionRegistrar: Hashable {
       else {
         continue
       }
+      let demangledParts = demangled.split(separator: " ")
       guard
         demangled.hasPrefix("protocol witness for SwiftUI.View.body.getter : ")
-          || demangled.contains(" SwiftUI.") && demangled.contains("body.getter : some")
+          || demangled.contains(" SwiftUI.")
+          && demangled.contains("body.getter : some")
+          && !(
+            demangledParts.first == "closure"
+            && demangledParts.dropFirst(2).prefix(3).joined(separator: " ") == "() -> ()"
+          )
       else {
         continue
       }


Sources/Perception/PerceptionRegistrar.swift
```diff
@@ -222,9 +222,15 @@ extension PerceptionRegistrar: Hashable {
...
+      let demangledParts = demangled.split(separator: " ")
       guard
         demangled.hasPrefix("protocol witness for SwiftUI.View.body.getter : ")
-          || demangled.contains(" SwiftUI.") && demangled.contains("body.getter : some")
+          || demangled.contains(" SwiftUI.")
+          && demangled.contains("body.getter : some")
+          && !(
+            demangledParts.first == "closure"
+            && demangledParts.dropFirst(2).prefix(3).joined(separator: " ") == "() -> ()"
+          )
```
mbrandonw, line 233: This tries to match things like `closure #1 () -> ()`.

Sources/Perception/PerceptionRegistrar.swift
```diff
@@ -222,9 +222,15 @@ extension PerceptionRegistrar: Hashable {
...
       }
+      let demangledParts = demangled.split(separator: " ")
       guard
         demangled.hasPrefix("protocol witness for SwiftUI.View.body.getter : ")
-          || demangled.contains(" SwiftUI.") && demangled.contains("body.getter : some")
+          || demangled.contains(" SwiftUI.")
+          && demangled.contains("body.getter : some")
+          && !(
+            demangledParts.first == "closure"
+            && demangledParts.dropFirst(2).prefix(3).joined(separator: " ") == "() -> ()"
```
stephencelis, line 232: Let's beef this up to support `onDelete: (IndexSet) -> ()` and other argument-taking actions.

stephencelis:
It's still slow but seems to work...

stephencelis (approved):
I'm down to merge as is and we can address performance later.

Sources/Perception/PerceptionRegistrar.swift
```diff
@@ -207,24 +206,13 @@ extension PerceptionRegistrar: Hashable {
 
   var isInSwiftUIBody: Bool {
     for callStackSymbol in Thread.callStackSymbols {
+      let mangledSymbol = callStackSymbol.utf8
+        .drop(while: { $0 != .init(ascii: "$") })
+        .prefix(while: { $0 != .init(ascii: " ") })
```
mbrandonw, line 211: I was able to get a small performance boost by doing all the string manipulation in UTF8 world instead of doing string splitting.

Sources/Perception/PerceptionRegistrar.swift
```diff
@@ -207,24 +206,13 @@ extension PerceptionRegistrar: Hashable {
...
-      guard
-        let demangled = String(symbol).demangled
-      else {
-        continue
-      }
-      guard
-        demangled.hasPrefix("protocol witness for SwiftUI.View.body.getter : ")
-          || demangled.contains(" SwiftUI.") && demangled.contains("body.getter : some")
+        let demangled = String(Substring(mangledSymbol)).demangled,
+        demangled.contains("body.getter : "),
```
mbrandonw, line 214: I had to pretty significantly weaken this condition in order for this simple test to emit a warning:

```swift
struct FeatureView: View {
  let model = Model()
  var body: some View {
    Text(self.model.count.description)
  }
}
```

It's probably not very common, but it seems that if there's just a single, simple view at the root of the `body` then the call stack is simplified a bit and there's no mention of `SwiftUI.` or even `some`. The generics are made concrete in the call stack.

Tests/PerceptionTests/PerceptionTests.swift
```diff
@@ -30,6 +30,34 @@ final class PerceptionTests: XCTestCase {
...
 
+  func testRuntimeWarning_NotInPerceptionBody_InSwiftUIBody_Wrapper() {
+    self.expectFailure()
+
+    struct FeatureView: View {
+      let model = Model()
+      var body: some View {
+        Wrapper {
+          Text(self.model.count.description)
+        }
```
mbrandonw, line 41: I added a new test to make sure things work when accessing state inside a view not defined in the `SwiftUI` module.

Tests/PerceptionTests/PerceptionTests.swift
```diff
@@ -227,6 +255,19 @@ final class PerceptionTests: XCTestCase {
...
     self.render(FeatureView())
   }
+
+  func testRuntimeWarning_ActionClosure() {
+    struct FeatureView: View {
+      @State var model = Model()
+      var body: some View {
+        Text("Hi")
+          .onAppear { _ = self.model.count }
+      }
```
mbrandonw, line 265: I was able to get some basic test coverage on action closures by using `onAppear`. This test fails if the `!demangled.isActionClosure` check is removed from `perceptionCheck`.

PR #2623, opened 2023-12-06
Add async throwing closure initializer for Result

Ryu0118:
The CasePathable macro eliminates the need to make the Action conform to Equatable, and the TaskResult is now deprecated.
However, the current TCA does not provide a complete replacement for TaskResult, and an extension to the Result must be created in each individual project.
I believe that this initializer should be added to TCA since it is most often used in an async effect being returned from a reducer.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/TaskResult.swift
+++ b/Sources/ComposableArchitecture/Effects/TaskResult.swift
@@ -332,3 +332,20 @@ extension TaskResult {
     try self.value
   }
 }
+
+public extension Result where Failure == any Error {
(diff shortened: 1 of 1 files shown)


stephencelis:
@Ryu0118 Thanks for exploring and taking the time to PR :smile: We actually snuck this initializer into our ConcurrencyExtras module here: https://github.com/pointfreeco/swift-concurrency-extras/pull/15

This module should be exported by TCA, so long as you're on version 1.1+, so I think it should already be available. Is it not working for you?

Ryu0118:
Thanks!
I missed it 😅. I will close this PR.

PR #2625, opened 2023-12-06
Fix uncached warning when using Store.ifLet

mbrandonw:
Using `Store.ifLet` jettisons one out of cached stores, but I don't think it's necessary. 

This fixes the issue brought up in https://github.com/pointfreeco/swift-composable-architecture/discussions/2624.

### DIFF
--- a/Sources/ComposableArchitecture/UIKit/IfLetUIKit.swift
+++ b/Sources/ComposableArchitecture/UIKit/IfLetUIKit.swift
@@ -53,16 +53,13 @@ extension Store {
       .publisher
       .removeDuplicates(by: { ($0 != nil) == ($1 != nil) })
       .sink { state in
-        if var state = state {
+        if state != nil {
           unwrap(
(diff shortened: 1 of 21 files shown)


Sources/ComposableArchitecture/UIKit/IfLetUIKit.swift
```diff
@@ -60,7 +60,7 @@ extension Store {
                 state = $0 ?? state
                 return state
               },
-              id: nil,
+              id: \State.self,
```
mbrandonw, line 63: I'm thinking we may want to do this for other scopes that were changed to `id: nil` in #2618, for example:

https://github.com/pointfreeco/swift-composable-architecture/pull/2618/files#diff-6b5c0df263281f9af5d6ea852817a00b3bf3cd5a08dcf7c56e4200226996ccbeR405

stephencelis (approved):

PR #2626, opened 2023-12-06
Use `ScopeID` for all cached child stores

stephencelis:
This PR forces child stores to be cached more strictly with actual scope ids.


Sources/ComposableArchitecture/SwiftUI/ForEachStore.swift
```diff
@@ -204,7 +197,7 @@ public struct ForEachStore<
               element = $0[id: id] ?? element
               return element
             },
-            id: id,
+            id: nil,
```
stephencelis, line 200: We would need to fully erase `ScopeID`'s generics to provide an `id` here, but since it's the old, non-identified action version of `ForEachStore` maybe it's OK to not cache, since historically it hasn't needed it...

PR #2627, opened 2023-12-06
Child store caching improvements

stephencelis:
This PR supersedes #2626. It:

- Only caches stores via `ScopeID` instead of the more nebulous `AnyHashable`.
- Uses store caching with more view modifiers, _e.g._ various presentation modifiers.
- Leverages store caching/invalidation for the last non-nil value (rather than managing a local copy of state).


acosmicflamingo:
LGTM!

mbrandonw (approved):

PR #2630, opened 2023-12-07
Implement memoization for perception checks

scogeo:
This PR adds memoization to the calculation of `isInSwiftUIBody` during perception checks as originally discussed [here](https://github.com/pointfreeco/swift-composable-architecture/discussions/2620#discussioncomment-7768616).

This substantially increases performance while exercising an app in a simulator running iOS 16 (or older). In my standard test case of text input in a back-ported SyncUps app, typing a single character now only takes 0.0005 secs when the memoization cache has been pre-warmed.  It was previously taking approximately 0.15 secs after the fixes in #2622.

The memoization cache is built on top of an `OrderedDictionary` and caps the capacity to a fixed number of entries. As new entries are added, the oldest entry is evicted. In my experiments with SyncUps, fully exercising the app results in about 500 unique stack backtraces across the various views. When the cache is fully primed, the app performance on iOS 16 is indistinguishable from iOS 17 from a user interaction perspective.

Based on my experiments with SyncUps I have set the default cache max size to 500, which seems to be large enough to hold the working set of a moderately complex app under development/test. A typical stack backtrace has about 80-90 frames (for math purposes, we will round up to 100).  So the rough memory cost of this cache is ~400k (500 entries * 100 frames per entry * 8 bytes per frame). This seems like a reasonable tradeoff for most apps under development/test.

Note, I have left some instrumentation code in the memoization cache. It should probably be removed at some point, but it was useful in testing this feature so I left it there for now.

### DIFF
--- a/Package@swift-5.9.swift
+++ b/Package@swift-5.9.swift
@@ -81,6 +81,7 @@ let package = Package(
       name: "Perception",
       dependencies: [
         "PerceptionMacros",
+        .product(name: "OrderedCollections", package: "swift-collections"),
         .product(name: "XCTestDynamicOverlay", package: "xctest-dynamic-overlay"),
       ]
     ),
--- a/Sources/Perception/Internal/Locking.swift
+++ b/Sources/Perception/Internal/Locking.swift
@@ -49,3 +49,12 @@ extension _ManagedCriticalState: Identifiable {
     ObjectIdentifier(buffer)
   }
 }
+
+extension NSLock {
+  @inlinable @discardableResult
+  @_spi(Internals) public func sync<R>(work: () -> R) -> R {
+    self.lock()
+    defer { self.unlock() }
+    return work()
+  }
+}
--- a/Sources/Perception/Internal/Memoization.swift
+++ b/Sources/Perception/Internal/Memoization.swift
@@ -50,14 +50,12 @@ final class MemoizedCache<Key: Hashable, Value>: @unchecked Sendable {
 
 func memoize<Input: Hashable, Result>(
   maxCapacity: Int = 500,
-  _ apply: @escaping (_ key: Input) -> Result
+  _ apply: @escaping (Input) -> Result
 ) -> (Input) -> Result {
   let cache = MemoizedCache<Input, Result>(maxCapacity: maxCapacity)
   
   return { input in
-    //    defer {
-    //       cache.printStats()
-    //    }
+//    defer { cache.printStats() }
     if let memoizedResult = cache[input] {
       return memoizedResult
     }
--- a/Sources/Perception/Internal/Memoization.swift
+++ b/Sources/Perception/Internal/Memoization.swift
@@ -1,0 +1,70 @@
+import Foundation
+import OrderedCollections
+
+final class MemoizedCache<Key: Hashable, Value>: @unchecked Sendable {
+  private var dictionary = OrderedDictionary<Key, Value>()
+  private var lock = NSLock()
+  private let maxCapacity: Int
+  
+  // TODO possibly remove these stats later
+  private var totalEvictions = 0
+  private var totalCalls = 0
+  private var totalHits = 0
+  
+  init(maxCapacity: Int = 500) {
+    self.maxCapacity = maxCapacity
+  }
+  
+  func printStats() {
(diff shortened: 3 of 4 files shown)


mbrandonw:
Hey @scogeo, thanks again for taking the time to look into this.

However, I'm not able to really measure a significant speed bump with memoization over what is on `observation-beta` right now. We included a few performance improvements in #2614, which was merged after your PR. Perhaps that closed the gap a bit?

To measure I put the following code at the top of `perceptionCheck`:

```swift
let start = Date()
defer {
  if #available(iOS 15.0, *) {
    print(
      "perceptionCheck",
      Date().timeIntervalSince(start)
        .formatted(.number.precision(.significantDigits(1...10)))
    )
  }
}
```

And then I ran the app with the memoization condition:

```swift
if #unavailable(iOS 17, macOS 14, tvOS 17, watchOS 10),
  !PerceptionLocals.isInPerceptionTracking,
  !PerceptionLocals.isInWithoutPerceptionChecking,
  isInSwiftUIBody  // ⬅️
```

…and without:

```swift
if #unavailable(iOS 17, macOS 14, tvOS 17, watchOS 10),
  !PerceptionLocals.isInPerceptionTracking,
  !PerceptionLocals.isInWithoutPerceptionChecking,
   determineIfInSwiftUIBody  // ⬅️
```

And it just seems that on average they are about the same.

Can you confirm that you see a measurable performance improvement over the most recent commit on observation-beta?

scogeo:
@mbrandonw I synced my fork and merged the latest from the `observation-beta` branch (as of about an hour ago) and I definitely still see substantial performance improvements with memoization in my SyncUp experiments.

Using your instrumentation when calling with `determineIfInSwiftUIBody` and typing a key after clicking the plus button, the first 10 calls are (note I disabled _printChanges() for these tests):

> perceptionCheck 0.01766705513
> perceptionCheck 0.01663196087
> perceptionCheck 0.01708400249
> perceptionCheck 0.01737201214
> perceptionCheck 0.01763093472
> perceptionCheck 0.01731300354
> perceptionCheck 0.01727092266
> perceptionCheck 0.01709699631
> perceptionCheck 0.01724898815
> perceptionCheck 0.01633000374

Using your instrumentation when calling with `isInSwiftUIBody`.  I first warm up the cache by typing in the title text field a few times in the add sheet.  Then close the add sheet and reopen.  After typing a key, the first 10 calls are:

> perceptionCheck 0.00007092952728
> perceptionCheck 0.00003397464752
> perceptionCheck 0.00003397464752
> perceptionCheck 0.0000319480896
> perceptionCheck 0.00002789497375
> perceptionCheck 0.00002992153168
> perceptionCheck 0.00002896785736
> perceptionCheck 0.00003397464752
> perceptionCheck 0.00002992153168
> perceptionCheck 0.00002801418304

So there is still a substantial improvement.  But the qualitative improvement in using the simulator to track down perception issues is more of my concern.  The qualitative improvement is very obvious from just using the app with and without memoization.

I'm going to create a branch in my fork that includes my SyncUp iOS 16 backport merged with my memoization changes so that we are both using the same branch/app for our tests. Will post an update when it's ready.

scogeo:
@mbrandonw You should be able to clone my fork and checkout the `syncups-ios16-memoized` [branch](https://github.com/scogeo/swift-composable-architecture/tree/syncups-ios16-memoized).

You can then just run SyncUps on an iOS 16 simulator.  I happen to be running on an iPhone 14 Pro with iOS 16.4, Xcode 15.0.1.

The easiest test is to just add a new SyncUp and start typing in the title field (make edits, delete, etc), both with and without memoization.

Hopefully, you will be able to see the improvement.

mbrandonw:
Hey @scogeo, thanks for following up. I was just running the case studies, so maybe that was too simplistic to see the difference. I'm traveling today, but we will get to this soon.

Thanks again for pushing this forward!

mbrandonw:
Hey @scogeo, ok was able to look into this more and I do see more significant changes in SyncUps than just using the Integration app target. So we will be merging this soon!

scogeo:
@mbrandonw Great, I'm glad you were able to see what I was experiencing.

mbrandonw (approved):

PR #2633, opened 2023-12-08
Don't replay changes to invalidated child stores.

### LINKED ISSUE #2632: Crash in 1.5.2 - Unwrapping in IfLetStore, line 88
### Description

Noticed a crash when upgrading from 1.5.1 to 1.5.2

### Checklist

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

```
import ComposableArchitecture
import SwiftUI

struct ContentView: View {
    var body: some View {
        FeatureView(
            store: Store(
                reducer: Feature.init
            )
        )
    }
}

#Preview {
    ContentView()
}

public struct FeatureView: View {
    let store: StoreOf<Feature>
    
    public var body: some View {
        ZStack {
            Button(action: { store.send(.tappedShowOverlay) }) {
                Text("Show overlay")
            }
            IfLetStore(
                store.scope(state: \.$destination.overlay, action: \.destination.overlay),
                then: OverlayView.init(store:)
            )
        }
    }
}

@Reducer
public struct Feature {
    public struct State {
        @PresentationState var destination: Destination.State?
    }
    
    public enum Action {
        case tappedShowOverlay
        
    }
    
    public var body: some ReducerOf<Self> {
        Reduce {
            state, action in
            
            switch action {
            case .destination(.presented(.overlay(.delegate(let action)))):
                switch action {
                case .dismiss:
                    state.destination = nil
                    return .none
                }
                
            case .tappedShowOverlay:
                state.destination = .overlay(Overlay.State())
                return .none
                
            case .destination:
                return .none
            }
        }
        ._printChanges()
        .ifLet(\.$destination, action: \.destination) {
            Destination()
        }
    }
    
    @Reducer
    public struct Destination {
        public enum State {
            case overlay(Overlay.State)
        }
        
        public enum Action {
            case overlay(Overlay.Action)
        }
        
        public var body: some ReducerOf<Self> {
            Scope(state: \.overlay, action: \.overlay) {
                Overlay()
            }
        }
    }
}

public struct OverlayView: View {
    let store: StoreOf<Overlay>
    
    public var body: some View {
        Color.blue
            .overlay(content: { Text("Tap me") })
            .onTapGesture {
                store.send(.tapped)
            }
    }
}

@Reducer
public struct Overlay {
    public struct State {
        
    }
    
    public enum Action {
        case tapped
        
        case delegate(Delegate)
        
        public enum Delegate {
            case dismiss
        }
    }
    
    public var body: some ReducerOf<Self> {
        Reduce {
            state, action in
            
            switch action {
            case .tapped:
                return Effect.send(.delegate(.dismiss))
                
            case .delegate:
                return .none
            }
        }
    }
}

```

1.5.2

iOS17

### Xcode version information

Version 15.1 (15C65)

```shell
swift-driver version: 1.87.3 Apple Swift version 5.9.2 (swiftlang-5.9.2.2.56 clang-1500.1.0.2.5)
```

mbrandonw:
Fixes #2632.

I was able to get a unit test and integration test to verify the fix.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -863,6 +863,7 @@ private protocol AnyStore {
 
 private protocol _OptionalProtocol {}
 extension Optional: _OptionalProtocol {}
+extension PresentationState: _OptionalProtocol {}
 
 func storeTypeName<State, Action>(of store: Store<State, Action>) -> String {
   let stateType = typeName(State.self, genericsAbbreviated: false)
@@ -1002,12 +1003,13 @@ private final class ScopedStoreReducer<RootState, RootAction, State, Action>: Re
 
   @inlinable
   func reduce(into state: inout State, action: Action) -> Effect<Action> {
-    if self.isInvalid() {
-      self.onInvalidate()
-    }
     self.isSending = true
     defer {
-      if !self.isInvalid() {
+      let isInvalid = self.isInvalid()
+      if isInvalid {
+        self.onInvalidate()
+      }
+      if !isInvalid || state is _OptionalProtocol {
         state = self.toState(self.rootStore.stateSubject.value)
       }
       self.isSending = false
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -463,7 +463,7 @@ public final class Store<State, Action> {
     )
   }
 
-  func scope<ChildState, ChildAction>(
+  @_spi(Internals) public func scope<ChildState, ChildAction>(
     state toChildState: @escaping (State) -> ChildState,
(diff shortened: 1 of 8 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -463,7 +463,7 @@ public final class Store<State, Action> {
     )
   }
 
-  func scope<ChildState, ChildAction>(
+  @_spi(Internals) public func scope<ChildState, ChildAction>(
```
mbrandonw, line 466: Made some things SPI so that I could test without a `@testable` import.

stephencelis:
Turns out we already had this fix on the observation beta for similar reasons. There were a few small differences so I updated it to more closely match observation. Will merge and release when it's green!

PR #2635, opened 2023-12-10
Cache initial state in store collections

stephencelis:
A crash was [reported on Slack](https://pointfreecommunity.slack.com/archives/C062TFT9M54/p1702182857138549) related to deleting rows and the laziness of store collections. This PR offers one fix, though we could consider others.

I'd also like to better understand the crash and see if we can get some test coverage for it.

### DIFF
--- a/Sources/ComposableArchitecture/Observation/IdentifiedArray+Observation.swift
+++ b/Sources/ComposableArchitecture/Observation/IdentifiedArray+Observation.swift
@@ -71,25 +71,23 @@ extension Store where State: ObservableState {
 
 public struct _StoreCollection<ID: Hashable, State, Action>: RandomAccessCollection {
   private let store: Store<IdentifiedArray<ID, State>, IdentifiedAction<ID, Action>>
-  // TODO: Should this be an entire snapshot of store state? `IdentifiedArray<ID, State>`?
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Observation/IdentifiedArray+Observation.swift
```diff
@@ -71,25 +71,23 @@ extension Store where State: ObservableState {
 
 public struct _StoreCollection<ID: Hashable, State, Action>: RandomAccessCollection {
   private let store: Store<IdentifiedArray<ID, State>, IdentifiedAction<ID, Action>>
-  // TODO: Should this be an entire snapshot of store state? `IdentifiedArray<ID, State>`?
-  // Could return inert stores that warn when sending actions or accessing state without crashing.
-  private let ids: OrderedSet<ID>
+  private let data: IdentifiedArray<ID, State>
```
stephencelis, line 74: This does keep a copy of collection state around longer, but seems necessary for how SwiftUI's `ForEach` lifecycle works.

The alternative fix would be to make `_StoreCollection` eagerly scope all its children, thus hydrating state automatically, but this is a lot of work to create stores, especially for large collections, so I think this PR's fix is probably the better way to go.

mbrandonw (approved):

PR #2641, opened 2023-12-11
Fix archiving problem since 1.4.0

rhysm94:
After discussion with Jon Shier (@jshier) on the Slack, he suggested making this change.
It appears to fix the key path getter error that was appearing when trying to archive a project using the `BindingReducer`.
 I've tested a project which compiled but failed to archive on TCA 1.4 onwards, when macros were introduced, and this change allows it to compile and archive successfully.

For context, the Slack discussion is here: https://pointfreecommunity.slack.com/archives/C04KQQ7NXHV/p1702313629767679

This fix is entirely thanks to @jshier's suggestions! I take no credit for this!

### DIFF
--- a/Sources/ComposableArchitecture/Reducer/Reducers/BindingReducer.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/BindingReducer.swift
@@ -70,6 +70,8 @@ where State == ViewAction.State {
 
   @inlinable
   public func reduce(into state: inout State, action: Action) -> Effect<Action> {
+    // NB: Using a closure and not a `\.binding` key path literal to avoid a bug with archives:
+    //     https://github.com/pointfreeco/swift-composable-architecture/pull/2641
     guard let bindingAction = self.toViewAction(action).flatMap({ $0.binding })
     else { return .none }
 
--- a/Sources/ComposableArchitecture/Reducer/Reducers/BindingReducer.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/BindingReducer.swift
@@ -70,7 +70,7 @@ where State == ViewAction.State {
 
   @inlinable
   public func reduce(into state: inout State, action: Action) -> Effect<Action> {
-    guard let bindingAction = self.toViewAction(action).flatMap(\.binding)
+    guard let bindingAction = self.toViewAction(action).flatMap({ $0.binding })
     else { return .none }
 
     bindingAction.set(&state)


rhysm94:
> @rhysm94 Thanks for looking into this and taking the time to PR! Any chance you can share a failing repro for us so we can look into capturing the issue in CI somehow?

Hi! This is a really tricky one – I cannot find an obvious reproduction case for this to occur.

In the particular app I've encountered it with, it only happens on two of our feature modules. We have _other_ feature modules which use the `BindingReducer` family of features without any kind of issue, but for some reason, two of our feature modules have this issue. One of them is built completely normally – Reducer, State, Action all in one file. It has a `BindingViewState` – but so do other modules, which don't flag up this error. The other is a bit more funky for legacy reasons, but given the more regular feature also has this issue, I don't think this feature's funkiness is the issue.

I've tested commenting out the Binding functionality in both these modules, and it allows me to archive a build again, which showed me that there is some change here.

I cannot, for the life of me, figure out how to reliably replicate it in a small test app, nor why changing from a key path to a closure stops this error occurring. Sorry I can't currently be more help in replicating the bug. I will take another look tomorrow, but for now, I can't find anything in these two features that stands out as causing this.

For more context, I've included the error in the archive log for the particular app I'm working on:

<details><summary>Details</summary>

>ld: warning: Could not find or use auto-linked framework 'CoreAudioTypes': framework 'CoreAudioTypes' not found
>ld: Undefined symbols:
>  key path getter for ComposableArchitecture.BindableAction.binding : ComposableArchitecture.BindingAction<A.State>? : <A, B, C where A == C.State, C: ComposableArchitecture.BindableAction>C, serialized, referenced from:
>      generic specialization <LicenceInfoFeature.LicenceInfo.State, LicenceInfoFeature.LicenceInfo.Action, LicenceInfoFeature.LicenceInfo.Action.LicenceInfoViewAction> of ComposableArchitecture.BindingReducer.reduce(into: inout A, action: B) -> ComposableArchitecture.Effect<B> in LicenceInfoFeature.o
>      generic specialization <LicenceInfoFeature.LicenceInfo.State, LicenceInfoFeature.LicenceInfo.Action, LicenceInfoFeature.LicenceInfo.Action.PersonalInfoViewAction> of ComposableArchitecture.BindingReducer.reduce(into: inout A, action: B) -> ComposableArchitecture.Effect<B> in LicenceInfoFeature.o
>      generic specialization <PersonalInfoFeature.PersonalInfo.State, PersonalInfoFeature.PersonalInfo.Action, PersonalInfoFeature.PersonalInfo.Action> of ComposableArchitecture.BindingReducer.reduce(into: inout A, action: B) -> ComposableArchitecture.Effect<B> in PersonalInfoFeature.o

</details>

Sources/ComposableArchitecture/Reducer/Reducers/BindingReducer.swift
```diff
@@ -70,7 +70,7 @@ where State == ViewAction.State {
 
   @inlinable
   public func reduce(into state: inout State, action: Action) -> Effect<Action> {
-    guard let bindingAction = self.toViewAction(action).flatMap(\.binding)
+    guard let bindingAction = self.toViewAction(action).flatMap({ $0.binding })
```
stephencelis, line 73:
```suggestion
    // NB: Using a closure and not a `\.binding` key path literal to avoid a bug with archives:
    //     https://github.com/pointfreeco/swift-composable-architecture/pull/2641
    guard let bindingAction = self.toViewAction(action).flatMap({ $0.binding })
```

Issue #2508, opened 2023-10-07
StackState/StackReducer related leak


bioche:
### Description

For the last 2 days I'v been trying to understand a leak & I have been able to reproduce it in this sample project : [NavigationStackLeak.zip](https://github.com/pointfreeco/swift-composable-architecture/files/12838891/NavigationStackLeak.zip)

The issue occurs when creating a `StoreOf<FlowCore>` from scratch, where `FlowCore` is a Reducer wrapping a `StackState` / `StackAction` combined using forEach operator.

It seems like pushing a node in `StackState` causes effects to be added by `_StackReducer` that will cause `Store` & all its properties to not being deallocated once the `NavigationStack` gets dismissed causing a leak

Thanks in advance for looking into this :)

### Checklist

- [X] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Expected behavior

I would expect Store to be removed from memory once the view/controller using them has been dismissed

### Actual behavior

Stores are stacking up after each present/dismiss of FlowView

### Steps to reproduce

1. Launch attached project app : [NavigationStackLeak.zip](https://github.com/pointfreeco/swift-composable-architecture/files/12838891/NavigationStackLeak.zip)
2. Tap "present stack" button
3. Tap "go to next screen"
4. Dismiss modal using cross or swipe down gesture
5. Repeat 2 or 3 times previous steps
6. Open memory inspector & filter with `FlowCore` : you will see as many instance of `StoreOf<FlowCore>` as modal opened

<img width="1249" alt="Screenshot 2023-10-07 at 22 16 25" src="https://github.com/pointfreeco/swift-composable-architecture/assets/4895727/c6772895-b2b4-44cf-b384-7e5b252d477b">

It's important to note that the issue is not there when `FlowCore` is embedded in another Reducer like `MainCore` in example project (try same scenario using "present stack using TCA" button instead). probably because effects built in `_StackReducer` get properly cancelled in that case.

A quick fix on my side would be to send action to pop all nodes of `StackState` when receiving dismiss action. However a fix on the lib would be preferable as others could stumble upon this issue & like me waste hours trying to figure out what is happening ;)

[...]

PR #2643, opened 2023-12-11
Fix Store leak when async effect is in flight

stephencelis:
A [Slack conversation](https://pointfreecommunity.slack.com/archives/C04KQQ7NXHV/p1702049135565039) pointed to a potential leak of the `Store` when it was executing an effect. This PR allows cancellation to properly propagate during deinitialization, and ensures that async effects don't strongly capture the store.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -488,6 +488,7 @@ public final class Store<State, Action> {
     self.effectCancellables.values.forEach { cancellable in
       cancellable.cancel()
     }
+    self.effectCancellables.removeAll()
   }
 
   fileprivate func invalidateChild(id: ScopeID<State, Action>) {
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -485,6 +485,9 @@ public final class Store<State, Action> {
     for id in self.children.keys {
       self.invalidateChild(id: id)
     }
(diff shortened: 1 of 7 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -485,6 +485,9 @@ public final class Store<State, Action> {
     for id in self.children.keys {
       self.invalidateChild(id: id)
     }
+    self.effectCancellables.values.forEach { cancellable in
+      cancellable.cancel()
+    }
```
mbrandonw, line 490: Could this even be:

```suggestion
    self.effectCancellables = [:]
```

?

stephencelis:
Also fixes #2508!

Tests/ComposableArchitectureTests/StoreTests.swift
```diff
@@ -931,6 +931,84 @@ final class StoreTests: BaseTCATestCase {
     store.send(.child(.dismiss))
     _ = (childViewStore1, childViewStore2, childStore1, childStore2)
   }
+
+  func testStoreDeinit() {
+    Logger.shared.isEnabled = true
+    defer { Logger.shared.isEnabled = false }
+    Logger.shared.clear()
```
mbrandonw, line 938: down to move this into the base test case like you were about to do, or move these 3 tests to their own test file.

mbrandonw (approved):

PR #2649, opened 2023-12-13
Perception caching updates

mbrandonw:
Just a few small updates to @scogeo's work on caching in the perception checks (see #2630). I've mainly just removed the stats and moved things around. I considered removing the `maxCapacity` entirely to keep things simple, but I do think it's necessary. The cache does get quite big over time.

### DIFF
--- a/Sources/Perception/Internal/Memoization.swift
+++ b/Sources/Perception/Internal/Memoization.swift
@@ -1,43 +1,45 @@
-import Foundation
-import OrderedCollections
+#if DEBUG
+  import Foundation
+  import OrderedCollections
 
-func memoize<Result>(
-  maxCapacity: Int = 500,
-  _ apply: @escaping () -> Result
-) -> () -> Result {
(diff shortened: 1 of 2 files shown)


Sources/Perception/Internal/Memoization.swift
```diff
@@ -1,68 +1,43 @@
...
-  private var dictionary = OrderedDictionary<Key, Value>()
-  private var lock = NSLock()
-  private let maxCapacity: Int
-  
-  // TODO possibly remove these stats later
-  private var totalEvictions = 0
-  private var totalCalls = 0
-  private var totalHits = 0
-  
+func memoize<Result>(
```
stephencelis, line 4: Should we move all of this in a `#if DEBUG`?

mbrandonw: yep, done

stephencelis (approved):

PR #2650, opened 2023-12-13
Better support for observing copies of values

stephencelis:
This PR instruments the observation machinery a bit more in order to better track changes to objects that are copied around and replaced wholesale.

Thanks @kabiroberai for coming up with the test cases.


Sources/ComposableArchitectureMacros/PresentsMacro.swift
```diff
@@ -191,22 +187,17 @@ extension PatternBindingListSyntax {
           ),
           typeAnnotation: binding.typeAnnotation,
           accessorBlock: AccessorBlockSyntax(
+            // TODO: Should this use the same, non-projected key path for access/mutation to work
```
stephencelis, line 190:
```suggestion
```

Sources/ComposableArchitecture/Observation/ObservableState.swift
```diff
@@ -9,31 +9,81 @@ import Perception
...
+  public mutating func _$willModify() {
+    self.location = UUID()
+  }
+
+  private final class Storage: Sendable {
+    fileprivate let location: LockIsolated<UUID>
+    fileprivate let tag: Int?
+
+    init(location: UUID = UUID(), tag: Int? = nil) {
+      self.location = LockIsolated(location)
```
kabiroberai, line 83: I might be missing something but if Storage has true CoW semantics it should be unnecessary to lock-isolate it (given that it'll be mutated in place iff you already have exclusive access.)

mbrandonw: Yeah I think you're right. Pushed a commit here https://github.com/pointfreeco/swift-composable-architecture/pull/2650/commits/f10c8d445aa8c7db73d1606fddf9cc6b1b28e14a.

PR #2664, opened 2023-12-17
Overhaul store scope communication

stephencelis:
With the recent introduction of child store caching (#2527) the performance of re-scoping stores dramatically improved, but the lifetime of scoped stores has also been extended (sometimes to the lifetime of the application), which can lead to performance issues of its own. In particular, the current model of scoping spins off a child store that maintains its own state and reducer, subscribes to its parent for state updates, and communicates actions back up to its parent when its reducer receives one. Because child stores are no longer created and destroyed regularly alongside views as they appear and disappear, more and more active subscriptions to parent stores can begin to bog down an application.

This PR attempts to alleviate this issue by rethinking the parent/child relationship of scoped stores. Rather than thinking of scoping as an operation that synchronizes child with parent, we can think of all scoped stores as simple views on a root store, where accessing state is a matter of key-pathing from some root state to some child state, and sending actions is a matter of embedding some child action in a root before sending it along. This means we can eliminate all child store subscriptions: only view stores need to subscribe to updates in order to ping their `objectWillChange` publisher.

Note that store scoping was previously a flat stack operation due to the fact that all child stores maintained their own local copy of state. This is no longer the case, but store scoping can remain flat via key path-scoping, which is stackless: we can append key paths at each layer of scoping rather than accumulate another stack frame in a nested closure. This does mean that closure-based scopes will incur additional stack frames when adopting this code, so applications with large root states should update to use key path-based scoping wherever possible before adopting these changes.


Sources/ComposableArchitecture/Store.swift
```diff
@@ -455,30 +448,59 @@ public final class Store<State, Action> {
...
       removeDuplicates: { $0.sharesStorage(with: $1) }
     )
   }
 
-  @_spi(Internals) public func scope<ChildState, ChildAction>(
-    state toChildState: @escaping (State) -> ChildState,
-    id: ScopeID<State, Action>?,
-    action fromChildAction: @escaping (ChildAction) -> Action,
-    isInvalid: ((State) -> Bool)?,
-    removeDuplicates isDuplicate: ((ChildState, ChildState) -> Bool)?
```
stephencelis, line 471: - [ ] We only supply an `isDuplicate` argument when scoping to presentation state/actions, but in this PR the argument is now effectively ignored. Before merging, we should figure out if it is really doing much and if we want to continue to support it. If we don't want to continue to support it we should remove it entirely from the internal APIs.

Sources/ComposableArchitecture/Store.swift
```diff
@@ -742,7 +559,10 @@ public final class Store<State, Action> {
   ///   .sink { ... }
   /// ```
   public var publisher: StorePublisher<State> {
-    StorePublisher(store: self, upstream: self.stateSubject)
+    StorePublisher(
+      store: self,
+      upstream: self.rootStore.didSet.map { self.currentState }
```
stephencelis, line 564: Not sure if it matters, but maybe this should be:

```suggestion
      upstream: self.rootStore.didSet.compactMap { [weak self] in self?.currentState }
```

Tests/ComposableArchitectureTests/StoreTests.swift
```diff
@@ -153,38 +189,40 @@ final class StoreTests: BaseTCATestCase {
...
-    XCTAssertEqual(numCalls1, 1)
-    XCTAssertEqual(numCalls2, 1)
-    XCTAssertEqual(numCalls3, 1)
-
-    viewStore4.send(())
-
-    XCTAssertEqual(numCalls1, 2)
-    XCTAssertEqual(numCalls2, 2)
+    XCTAssertEqual(numCalls1, 6)
+    XCTAssertEqual(numCalls2, 4)
```
stephencelis, line 200: We expect these numbers to creep up, since scopes are composed, but since they closure-based scoping is deprecated, it should almost always be a simple key path access soon.

acosmicflamingo:
image

With the changes in this PR, the issue in https://github.com/acosmicflamingo/StoreScopingCacheReproducer disappears.
```
[StoreScopingCacheReproducerUITests.StoreScopingCacheReproducerUITests testSectionHeaderText]' passed (5.405 seconds).
Test Suite 'StoreScopingCacheReproducerUITests' passed at 2023-12-17 18:30:04.010.
	 Executed 100 tests, with 0 failures (0 unexpected) in 532.213 (532.288) seconds
```

Additionally, the section header issue that was going on in my own music app also goes away with this change (and I really tried my best to break it). Now this is a real Christmas miracle!

Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
```diff
@@ -627,19 +627,15 @@ as an enum instead of multiple optionals. In order to assert on state changes wh
 state you must be able to extract the associated state from the enum, make a mutation, and then
 embed the new state back into the enum.
 
-The library provides a tool to perform these steps in a single step, and it is called `XCTModify`:
+The library provides a tool to perform these steps in a single step. It's the
+``PresentationState/subscript(case:)-7uqte`` defined on ``PresentationState`` which allows you to
+modify the data inside a case of the destination enum:
```
mbrandonw, line 632: This was unrelated to this PR, but I just realized we should probably be pushing people towards the `[case:]` modify subscript instead of `XCTModify` by default. `XCTModify` is really only handy if you want to perform a bunch of mutations on the same case.

Sources/ComposableArchitecture/Reducer/Reducers/PresentationReducer.swift
```diff
@@ -92,7 +92,14 @@ public struct PresentationState<State> {
 
   /// Accesses the value associated with the given case for reading and writing.
   ///
-  /// > Note: Accessing the wrong case will result in a runtime warning.
+  /// You can access the projected value of ``PresentationState`` and subscript into a particular
```
mbrandonw, line 95: same here

mbrandonw (approved):

PR #2669, opened 2023-12-19
Receive action with value checking.

mbrandonw:
This adds a new `receive` method to `TestStore` that allows you to use  case key paths to specify which action is received, but also allows you to assert on the data in the action:

```swift
await store.receive(\.delegate.success, 42)
```

This shouldn't be needed typically, but it can be handy for testing delegate actions.


lukeredpath:
Referring back to my original discussion:
https://github.com/pointfreeco/swift-composable-architecture/discussions/2646

My main feedback here is:
* I think the assertion reads better with an explicit label for the second parameter - I think “with:” reads more fluently. 
* Allowing for a closure-based API, whilst less elegant, does also allow for this type of assertion to be made against non-equatable values and also delegate actions that have multiple associated values, due to Swift’s automatic destructuring of tuple values as closure arguments.

mbrandonw:
> * I think the assertion reads better with an explicit label for the second parameter - I think “with:” reads more fluently.

We'll think about it more, but it seems a little superfluous.

> * Including a closure-based API, whilst less elegant, does also allow for this type of assertion to be made against non-equatable values and also delegate actions that have multiple associated values, due to Swift’s automatic destructuring of tuple values as closure arguments.

We'll think about this too, but personally I would like to minimize the overloads. With the release of variadic types I would hope that tuples are becoming equatable soon.

stephencelis:
@lukeredpath I tend to agree with @mbrandonw. I think the `with:` probably isn't necessary, and also I hope that Swift will close the gap on equatable tuples rather than force overloads.

If you think it's possible to use parameter packs here, though, and it doesn't break the compiler, I think we'd consider a single overload to support all tuple types.

Alternately, is it too much of a pain to use equatable structs instead of tuples for payloads you want to test?

Tests/ComposableArchitectureTests/TestStoreTests.swift
```diff
@@ -530,6 +530,22 @@ final class TestStoreTests: BaseTCATestCase {
...
+        case .tap:
+          return .send(.delegate(.success(42)))
+        case .delegate:
+          return .none
+        }
+      }
+    }
+
+    await store.send(.tap)
+    await store.receive(\.delegate.success, 42)
```
stephencelis, line 547: Should we do an `XCTExpectFailure` too?

mbrandonw: yep, done

PR #2672, opened 2023-12-19
Migration guide: `@CasePathable` tweaks

stephencelis:
Came up on Slack. Tweaks the 1.4 migration guide a bit to note that `@CasePathable` is explicitly needed in some cases.

### DIFF
--- a/Sources/ComposableArchitecture/Documentation.docc/Articles/MigratingTo1.4.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Articles/MigratingTo1.4.md
@@ -110,7 +110,7 @@ To form a case key path for any other enum, you must apply the `@CasePathable` m
 ```swift
 @CasePathable
 enum DelegateAction {
-  didFinish(success: Bool)
+  case didFinish(success: Bool)
 }
 ```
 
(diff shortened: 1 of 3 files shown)


Sources/ComposableArchitecture/Documentation.docc/Articles/MigratingTo1.4.md
```diff
@@ -83,20 +83,42 @@ Reduce { state, action in
...
+```
+
+Further, if the feature's `State` is an enum, `@CasePathable` will also be applied.
+
+To form a case key path for any other enum, you must apply the `@CasePathable` macro explicitly:
+
+```swift
+@CasePathable
+enum DelegateAction {
+  didFinish(success: Bool)
```
mbrandonw, line 113:
```suggestion
  case didFinish(success: Bool)
```

Sources/ComposableArchitecture/Documentation.docc/Articles/MigratingTo1.4.md
```diff
@@ -114,6 +136,17 @@ store.receive(.child(.presented(.response(.success("Hello!")))))
...
 ```
 
+> Note: Case key path syntax requires that every nested action is `@CasePathable`. Reducer actions
+> are typically `@CasePathable` automatically via the ``Reducer()`` macro, but other enums must be
+> explicitly annotated:
+>
+> ```swift
+> @CasePathable
+> enum DelegateAction {
+>   didFinish(success: Bool)
```
mbrandonw, line 146:
```suggestion
>   case didFinish(success: Bool)
```

mbrandonw (approved):

PR #2674, opened 2023-12-20
Fix ownership between parent and child stores

mbrandonw:
This is a continuation of our work in #2664.

This PR cleans up a few of the internals of the revamped store scoping, including:

- It avoids a retain cycle between parent and child store. This retain cycle does not generally matter, as cached stores live as long as the lifetime of the app at this time, but in the future when we more proactively prune invalidated stores, it could prevent stores from being reinitialized.
- It brackets some debug codes to debug builds only.
- It cleans up some unused code.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -454,7 +454,7 @@ public final class Store<State, Action> {
     let childStore = Store<ChildState, ChildAction>(
       rootStore: self.rootStore,
       toState: self.toState.appending(state.base),
-      fromAction: { self.fromAction(fromChildAction($0)) }
+      fromAction: { [fromAction] in fromAction(fromChildAction($0)) }
     )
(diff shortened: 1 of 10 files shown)


stephencelis (approved):

stephencelis (approved):

Sources/ComposableArchitecture/Store.swift
```diff
@@ -443,39 +441,30 @@ public final class Store<State, Action> {
...
-      self.invalidateChild(id: id)
-    }
-  }
-
   fileprivate func invalidateChild(id: ScopeID<State, Action>) {
-    guard self.children.keys.contains(id) else { return }
-    (self.children[id] as? any AnyStore)?.invalidate()
+    guard self.children.keys.contains(id)
+    else { return }
     self.children[id] = nil
```
mbrandonw, line 468:
```suggestion
    self.children[id] = nil
```

PR #2679, opened 2023-12-22
Update .spi.yml

finestructure:
Not sure what the intent was for these changes but I believe they're what made the 5.7...5.8 builds being flagged as incompatible on SPI.

https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2798#issuecomment-1867634649

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -3,15 +3,10 @@ builder:
   configs:
   - platform: ios
     scheme: ComposableArchitecture
-    swift_version: 5.9
   - platform: macos-xcodebuild
     scheme: ComposableArchitecture
-    swift_version: 5.9
   - platform: tvos
     scheme: ComposableArchitecture
-    swift_version: 5.9
   - platform: watchos
     scheme: ComposableArchitecture
-    swift_version: 5.9
   - documentation_targets: [ComposableArchitecture]
-    swift_version: 5.9


.spi.yml
```diff
@@ -3,15 +3,10 @@ builder:
...
     scheme: ComposableArchitecture
-    swift_version: 5.9
   - platform: tvos
     scheme: ComposableArchitecture
-    swift_version: 5.9
   - platform: watchos
     scheme: ComposableArchitecture
-    swift_version: 5.9
   - documentation_targets: [ComposableArchitecture]
-    swift_version: 5.9
```
finestructure, line 17: This instance of `swift_version` was only necessary to make docs being built with 5.9 when 5.8 was the default. By removing it, docs will still be built with 5.9 and will automatically move to use 5.10 etc once those become available.

.spi.yml
```diff
@@ -3,15 +3,10 @@ builder:
   configs:
   - platform: ios
     scheme: ComposableArchitecture
-    swift_version: 5.9
```
finestructure, line 6: What `swift_version` does here is contains the `scheme` config to `ios` and `5.9` where previously it applied to _all_ iOS builds. You almost certainly don't want this here or in the other configs below.

finestructure:
Ok, this isn't the whole story.  `xcrun xcodebuild -resolvePackageDependencies` also fails when running it with 5.8 against 1.5.6 and I need to investigate why.

stephencelis:
@finestructure This good to go now?

finestructure:
Yes! I won't fix things by itself but is required to make the other fix work.

PR #2680, opened 2023-12-22
Optimization: Key path subscript presentation scopes

stephencelis:
When scoping a store over a "presentation" boundary (optional, stack element, list element), we maintain a local, memoized value so that during dismissal, views still have access to the last element should SwiftUI render.

This PR introduces private subscripts so that these scopes can be done via key path.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -379,8 +379,8 @@ public final class Store<State, Action> {
     action: CaseKeyPath<Action, ChildAction>
   ) -> Store<ChildState, ChildAction> {
     self.scope(
-      state: ToState(state),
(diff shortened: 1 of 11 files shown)


Sources/ComposableArchitecture/Store.swift
```diff
@@ -427,8 +427,8 @@ public final class Store<State, Action> {
   @_spi(Internals)
   public
     func scope<ChildState, ChildAction>(
-      state: ToState<State, ChildState>,
       id: ScopeID<State, Action>?,
+      state: ToState<State, ChildState>,
```
stephencelis, line 431: Reordering `id` to be first while I'm in here so that `state` and `action` remain grouped together.

mbrandonw (approved):

PR #2685, opened 2023-12-27
Updates @Reducer to no longer apply @CasePathable macro to enums with explicit conformance.

scogeo:
I have a macro library that I'm developing for TCA that helps generate members of `Action` and `State`. This also requires that I manage the `AllCasePaths` struct when implementing `CasePathable` conformance.  The current `@Reducer` macro automatically applies the `@CasePathable` macro to the `Action` and `State` enums, even though my implementation has explicitly conformed to `CasePathable`.

For example the following code produces multiple compiler errors due to the application of `@CasePathable` to `Action` by the `@Reducer` macro, resulting in duplicate declarations of `AllCasePaths` and `allCasePaths`.

```swift
@Reducer
struct Feature {
  // ...

  enum Action: CasePathable {
    struct AllCasePaths {}
    static var allCasePaths: AllCasePaths { AllCasePaths() }
  }

  // ...
}
```

This PR changes the behavior of `@Reducer` to check if either `State` or `Action` enums provide explicit conformance to `CasePathable`, and if so, the `@CasePathable` macro is not applied.

### DIFF
--- a/Sources/ComposableArchitectureMacros/ReducerMacro.swift
+++ b/Sources/ComposableArchitectureMacros/ReducerMacro.swift
@@ -50,6 +50,14 @@ extension ReducerMacro: MemberAttributeMacro {
       default:
         break
       }
+      if let inheritanceClause = enumDecl.inheritanceClause,
+        inheritanceClause.inheritedTypes.contains(
+          where: {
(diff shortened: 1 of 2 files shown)


stephencelis (approved):
Looks like a good change to us! Thanks for tackling it!

PR #2692, opened 2024-01-04
Prevent unnecessary diffs in the tutorials

bricklife:
Just removing all trailing whitespaces 😄 

### Before:
before

### After:
after


mbrandonw (approved):
Good finds, thanks!

PR #2698, opened 2024-01-06
Fix "Composing features" tutorial

bricklife:
I would have liked to add `@testable import XxxApp` as well, but I couldn't think of a good app name 😅


stephencelis (approved):
Thanks!

PR #2704, opened 2024-01-10
Add stack action path element subscript to testing notes

lukeredpath:
Updates the stack navigation documentation to include information on how to test that a specific element action is received using case key paths.

Wording could maybe be improved but its a starting point.


mbrandonw (approved):
Thanks!

PR #2735, opened 2024-01-29
Fix typos in documents

yimajo:
Is this possibly a typo?


Sources/ComposableArchitecture/Reducer/Reducers/IfLetReducer.swift
```diff
@@ -37,7 +37,7 @@ extension Reducer {
   ///   * It automatically cancels all child effects when it detects the child's state is `nil`'d
   ///     out.
   ///
-  ///   * Automatically `nil`s out child state when an action is sent for alerts and confirmation
```
stephencelis, line 40: This is actually what we want. We're using `nil` as a verb:

"It automatically nils out the child state" meaning: it automatically sets the child state to nil.

PR #2736, opened 2024-01-29
Fix Navigation tutorial

bricklife:
The new Observation tools in 1.7.0 are incredible..! Thanks for the release 🚀 

While re-doing the latest version of the tutorial, I found a few things that need to be fixed.


stephencelis (approved):
Thank you!

PR #2738, opened 2024-01-29
Documentation: Fix a small typo in the 1.7 migration guide

zvona031:
> You are most likely going to want to incrementally your application

should be:

> You are most likely going to want to incrementally **migrate** your application

otherwise, the sentence does not make sense.

Another proposal could also be:

> You are most likely going to incrementally migrate your application


stephencelis:
Thanks!

PR #2740, opened 2024-01-30
Move binding warning suppression to dynamic member lookup

stephencelis:
In TCA it is considered a bug to send an action to a child feature when it is dismissed.

In SwiftUI, a focused text field binding will write to itself when the view it is presented in is dismissed.

So if you derive a TCA binding to a text field in a dismissable child feature, dismissing said feature will write to the text field and cause a warning to be emitted.

To work around this, we set a task local, and are currently setting it in debug-only methods that construct a binding from scratch. Creating bindings from scratch unfortunately leads to issues with animations, so we should avoid doing so.

Instead, we can leverage properties on store that bindings are derived from and suppress the warnings from there.

### DIFF
--- a/Sources/ComposableArchitecture/Observation/Binding+Observation.swift
+++ b/Sources/ComposableArchitecture/Observation/Binding+Observation.swift
@@ -100,7 +100,11 @@
       dynamicMember keyPath: WritableKeyPath<State, Value>
     ) -> Value {
       get { self.state[keyPath: keyPath] }
-      set { self.send(.binding(.set(keyPath, newValue))) }
+      set {
+        BindingLocal.$isActive.withValue(true) {
+          self.send(.binding(.set(keyPath, newValue)))
+        }
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Observation/Binding+Observation.swift
```diff
@@ -145,129 +118,110 @@
...
-              _ = self.wrappedValue.send(
-                .view(.binding(.set(keyPath, newValue))), transaction: transaction
-              )
-            }
-          }
-        )
-      }
+  @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
+  @dynamicMemberLookup
+  public struct _StoreBindable_SwiftUI<State: ObservableState, Action, Value> {
```
stephencelis, line 154: While I think it might be possible to abstract these 3 "store binding" types, I'm not sure it's worth it or a good idea.

mbrandonw (approved):

PR #2741, opened 2024-01-30
Support `package` access modifier for Observation architecture

nnsnodnb:
`@ObservableState` macro was not support `package` access modifier.

error

| Before | After |
| :----: | :----: |
| before | after |


stephencelis (approved):
Looks good, thanks!

PR #2743, opened 2024-01-30
Add missing `var` keyword to Migrating to 1.7 doc

shgew:
I noticed that there is no `var` keyword on the line where `isOn` is declared 🙂


mbrandonw (approved):
Good catch, thank you!

PR #2746, opened 2024-01-30
Workaround for the build break on visionOS

peterbohac:
This is likely not the proper solution, but it does get my projects building.


stephencelis:
Thanks! Going to make sure we have visionOS CI set up and then will merge this as a fix.

stephencelis (approved):

PR #2755, opened 2024-01-31
Try out M1 CI

stephencelis:
Let's see how quickly this runs...


jshier:
I've found that the macOS-14 runners are 30 - 50% faster than the older ones, which is still about 100% slower than a bare M1 machine.

stephencelis:
@jshier Yeah, the builds themselves are quite impressive, the Examples is down from 70ish minutes to 14ish. But waiting for the job takes awhile, and at the end we might be about the same build time at the end of it. It's hard to tell without keeping an eye on the build since the "waiting for job" period doesn't seem to be recorded anywhere.

stephencelis:
This was already done elsewhere.

PR #2759, opened 2024-02-01
Some cleanup

stephencelis:
- `if/guard let x = x` -> `if/guard let x`
  - Remove unneeded `self` from a few demos
  - Fix up some presentation logic in TicTacToe


Examples/TicTacToe/tic-tac-toe/Sources/LoginUIKit/LoginViewController.swift
```diff
@@ -110,16 +110,16 @@ public class LoginViewController: UIViewController {
       }
 
       if let store = store.scope(state: \.twoFactor, action: \.twoFactor.presented),
-        twoFactorViewController == nil
+         twoFactorController == nil
```
stephencelis, line 113:
```suggestion
        twoFactorController == nil
```

mbrandonw (approved):

PR #2765, opened 2024-02-02
Fix navigationDestinationWrapper in docs.

### LINKED ISSUE #2761: Nesting a Cased reducer > Tabbed reducer > Navigation Stack > `navigationDestination(item:destination:)` breaks navigating to the same item twice.
### Description

Currently following the [1.7 Migration Guide](https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/migratingto1.7) and having an issue with `navigationDestination(item:destination:)`.

My app supports iOS 16, and so I'm using the `Perception` tools. This happens on iOS 17.

My app's set up is that the root reducer uses a `switch / case` to switch between the onboarding and the tab view, there are a number of tabs, each with a navigation stack, etc.

Within this hierarchy, the reducer within the tab presents a leaf state on a button press through `navigationDestinationWrapper(item:destination:)` (using the wrapper suggested in the migration)

Tapping the button once navigates, then pressing back and navigating again does not navigate forward.

If you remove the `Switching` reducer or the `Tabbed` reducer/view from the hierarchy, it works as expected

### Checklist

- [X] If possible, I've reproduced the issue using the `main` branch of this package.

### Expected behavior

Pushing and popping the view should always work.

In my example, tapping `First Child`, dismissing, then `First Child` again, the view will not navigate the second time.

### Actual behavior

The view is pushed on the first attempt, and won't get pushed again on subsequent attempts.

In my example, if you tap `First Child`, dismissing, then `Second Child`, dismissing, then `First Child` again, everything works as expected,

### Steps to reproduce

[A basic repro](https://gist.github.com/autoreleasefool/5bb5a5582e0a89decc51342f8dca147d)

1.7.2

iOS 17

### Xcode version information

Xcode 15.2

```shell
swift-driver version: 1.87.3 Apple Swift version 5.9.2 (swiftlang-5.9.2.2.56 clang-1500.1.0.2.5)
Target: arm64-apple-macosx14.0
```

mbrandonw:
Fixes #2761.

### DIFF
--- a/Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
@@ -392,7 +392,9 @@ extension View {
     @ViewBuilder destination: @escaping (D) -> C
   ) -> some View {
     navigationDestination(isPresented: item.isPresented) {
-      item.wrappedValue.map(destination)
+      if let item = item.wrappedValue
+        destination(item)
+      }
     }
   }
 }
--- a/Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
@@ -382,29 +382,25 @@ project:
 
 ```swift
 extension View {
-  @available(iOS, introduced: 13, deprecated: 17)
-  @available(macOS, introduced: 10.15, deprecated: 14)
-  @available(tvOS, introduced: 13, deprecated: 17)
-  @available(watchOS, introduced: 6, deprecated: 10)
+  @available(iOS, introduced: 16, deprecated: 17)
+  @available(macOS, introduced: 13, deprecated: 14)
+  @available(tvOS, introduced: 16, deprecated: 17)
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/Documentation.docc/Articles/TreeBasedNavigation.md
```diff
@@ -382,31 +382,27 @@ project:
...
 ```swift
 extension View {
-  @available(iOS, introduced: 13, deprecated: 16)
-  @available(macOS, introduced: 10.15, deprecated: 13)
-  @available(tvOS, introduced: 13, deprecated: 16)
-  @available(watchOS, introduced: 6, deprecated: 9)
+  @available(iOS, introduced: 13, deprecated: 17)
+  @available(macOS, introduced: 10.15, deprecated: 14)
+  @available(tvOS, introduced: 13, deprecated: 17)
+  @available(watchOS, introduced: 6, deprecated: 10)
```
stephencelis, line 388: The underlying API wasn't introduced till iOS 16...

```suggestion
  @available(iOS, introduced: 16, deprecated: 17)
  @available(macOS, introduced: 13, deprecated: 14)
  @available(tvOS, introduced: 16, deprecated: 17)
  @available(watchOS, introduced: 9, deprecated: 10)
```

PR #2769, opened 2024-02-04
Add deep-linking support to UIKit navigation for modal views

acosmicflamingo:
One of the benefits of tree-based navigation is its deep-linking capabilities, but it seems like all the UIKit examples will only support deep-linking for view controllers that will be pushed or popped into UINavigationController's view controller stack and not modal view controllers that reside within the presented view controller. This is because it's only when `viewDidAppear` has been called that the now-presented view controller is ready to present modal view controllers.

The initial commit for this PR is one way of treating modal navigation a bit differently. @mbrandonw @stephencelis if you think this is the right approach, I can go ahead and make changes elsewhere in the codebase since right now, I just started with `LoginViewController`.

### DIFF
--- a/Examples/TicTacToe/tic-tac-toe/Sources/LoginUIKit/LoginViewController.swift
+++ b/Examples/TicTacToe/tic-tac-toe/Sources/LoginUIKit/LoginViewController.swift
@@ -87,7 +87,6 @@ public class LoginViewController: UIViewController {
       divider.heightAnchor.constraint(equalToConstant: 1),
     ])
 
-    var alertController: UIAlertController?
     var twoFactorController: TwoFactorViewController?
 
     observe { [weak self] in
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hi @acosmicflamingo, this is definitely something we should address in the case study, but currently this PR isn't quite right. This will _re_-observe everything if you navigate away and back to `LoginViewController`. You would only want to do this observation once, meaning some additional state needs to be tracked in the controller.

We will discuss more if there is a more robust tool that can be provided in the library, and in the meantime you can always `DispatchQueue.main.async` it.

acosmicflamingo:
@mbrandonw ah, this is certainly more complicated than I anticipated, not just for the reasons you stated but because `DispatchQueue.main.async` cannot be called until `viewDidAppear` is called. If it's called any earlier, not only do I get an error (`UIKit Internal bug: Unbalanced call to _endOcclusion, please file a feedback report with any information you have that helps reproduce this bug!`), but the view controller being presented is being dismissed when the "parent" view controller executes the following in its `viewDidAppear` function:
```swift
if !isMovingToParent {
  store.send(.destination(.dismiss))
}
```

It looks like the solution I'll be going with is adding a `allowModalNavigation` property state that I will set to `true` in a `store.send(.viewDidAppear)` and then in the `observe` closure I will be checking whether that property is true before pushing anything (for destination enum cases representing a modal view).

PR #2786, opened 2024-02-08
Properly tag observable enums

stephencelis:
It's possible for nested, inert enums to have the same "tag" and thus identity when they are distinct structures, which prevents cases changes from being observed. Let's allow tagging to happen recursively, which will allow for distinct identity in these cases.

### DIFF
--- a/Sources/ComposableArchitecture/Observation/ObservableState.swift
+++ b/Sources/ComposableArchitecture/Observation/ObservableState.swift
@@ -74,7 +74,10 @@
 
     @inlinable
     public func _$tag(_ tag: Int?) -> Self {
-      Self(location: self.location, tag: tag)
+      Self(
+        location: self.location,
+        tag: self.tag ?? tag
+      )
     }
 
     @inlinable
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/Observation/ObservableState.swift
```diff
@@ -23,41 +23,31 @@
...
     }
 
-    @usableFromInline
-    var tag: Int? {
-      self.storage.tag
-    }
-
     private var storage: Storage
 
-    @usableFromInline
```
stephencelis, line 43: Losing the inlinability here, but think it's fine.

Sources/ComposableArchitecture/Observation/ObservableState.swift
```diff
@@ -72,9 +62,8 @@
 
     public static let _$inert = Self()
 
-    @inlinable
```
stephencelis, line 75: Losing the inlinability here, but think it's fine.

Sources/ComposableArchitecture/Observation/ObservableState.swift
```diff
@@ -72,9 +62,8 @@
...
-    public func _$tag(_ tag: Int?) -> Self {
-      Self(location: self.location, tag: tag)
+    public func _$tag(_ tag: Int) -> Self {
```
stephencelis, line 65: Slight API change here, but couldn't find an instance where `nil` was being passed.

mbrandonw (approved):

PR #2789, opened 2024-02-10
[macro] Make _$id and _$willModify() public

freak4pc:
Opening this PR per our discussion on Slack :) 

When using the following syntax:
```swift
public extension Feature {
    @ObservableState
    struct State { } 
}
```

A macro isn't able to determine `State` is public since the macro isn't aware of its surrounding context.
A solution to this is to always give the generated `_$id` and `_$willModify()` a public ACL, which should have no negative effect on other use cases.


stephencelis (approved):
Thanks!

PR #2800, opened 2024-02-14
Support `#if` branching in enum reducers/state

stephencelis:
This adds support for `#if` branching across enum cases in observable state and reducer enums.

### DIFF
--- a/Sources/ComposableArchitectureMacros/ObservableStateMacro.swift
+++ b/Sources/ComposableArchitectureMacros/ObservableStateMacro.swift
@@ -248,9 +248,6 @@ extension ObservableStateMacro: MemberMacro {
 
     var declarations = [DeclSyntax]()
 
-    let access = declaration.modifiers.first {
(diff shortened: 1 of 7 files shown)


stephencelis:
This is updated for #2827. I don't think there is a reasonable way to parse `#if` conditions and accumulate the proper `body` type per branching leaf, especially when it comes to reusing the same condition more than once, and in slightly different ways:

```swift
#if os(macOS) && DEBUG
  case macFeature1(MacFeature1)
#endif
… // other features
#if DEBUG && os(macOS)
  case macFeature2(MacFeature2)
#endif
```

As such, as soon as a case is contained in an `#if` we will erase the `body` to be a `Reduce`, instead.

In the future when the Swift bug is fixed we can remove this erasure, but it shouldn't be a big deal in the meantime, and isn't much different from the `buildAvailability` erasure that happens with SwiftUI views.

mbrandonw (approved):

PR #2820, opened 2024-02-17
Replace usage of 'viewStore' with 'store' in UIKit code

juliensagot:
Hi folks 👋 
Just a small fix to the `README.md` file where `viewStore` was still being used within the UIKit code.


mbrandonw (approved):
Good catch, thanks!

PR #2821, opened 2024-02-17
[Enhancement] - Scope init overload for enum reducers

thebarndog:
Added an overload for `Scope` to work with enum reducers. This is really useful for when you want to utilize the `Destination` pattern but non-optionally. I use this pattern when I want to have a top-level `enum` state but also need to include additional information:

```swift
@Reducer 
struct App {
    
    @ObservableState
    struct State {
        var flow: Flow.State = .onboarding(.init())
        var firstTimeLaunching = false
        var networkIsConnected = true
    }
    
    enum Action {
        case flow(Flow.Action)
    }
}

@Reducer
enum Flow {
    case onboarding(Onboarding)
    case dashboard(Dashboard)
}
```

Before this change, I couldn't use enum reducers for `Flow` because I was unable to create an instance of it in `Scope`

```swift
Scope(state: \.flow, action: \.flow) { 
    Flow()
}
```

Now, I can integrate this sub-reducer into my domain like `Scope(state: \.flow, action: \.flow)`

### DIFF
--- a/Sources/ComposableArchitecture/CaseReducer.swift
+++ b/Sources/ComposableArchitecture/CaseReducer.swift
@@ -61,3 +61,21 @@ extension Store where State: CaseReducerState, State.StateReducer.Action == Acti
     State.StateReducer.scope(self)
   }
 }
+
+extension Scope {
+    
+    /// A special overload of ``Scope/init(state:action:child:)`` for enum reducers.
+    @inlinable
+    public init<ChildState: CaseReducerState, ChildAction>(
+        state toChildState: WritableKeyPath<ParentState, ChildState>,
+        action toChildAction: CaseKeyPath<ParentAction, ChildAction>
+    )
+    where
+    ChildState == Child.State,
+    ChildAction == Child.Action,
+    ChildState.StateReducer.Body == Child {
+        self.init(state: toChildState, action: toChildAction) {
+            ChildState.StateReducer.body
+        }
+    }
+}


stephencelis:
@thebarndog Thanks for taking the time to PR! @mbrandonw and I will discuss soon.

One thing we worry about with the `ifLet` and `forEach` overloads is that they can strain the compiler, so adding a third one for `Scope` does have us wondering if it would have been more reasonable to have less magic and overloads, and instead enforce a new calling convention with the existing reducers:

```swift
.ifLet(\.$destination, action: \.destination) {
  Destination.body  // or `Destination.self`
}
```

Thoughts on these trade-offs? It'd be a breaking change so we'd have to reconsider with 2.0, but we thought it'd be good to start the conversation now.

Alex293:
Just my two cents, I find Destination.self to be more discoverable

hmhv:
@stephencelis 

```swift
.ifLet(\.$destination, action: \.destination) {
  Destination.body  // or `Destination.self`
}
```

I think this should be added to Migrating to 1.8.

stephencelis:
@thebarndog While originally we thought omitting the closure in `ifLet` and `forEach` was a nice benefit of the new syntax, it's proven to mostly confuse folks, and it's also expanded the API overload surface area, which can slow down the type checker. Because of this I think we're going to close this out, and we plan on deprecating the new `ifLet` and `forEach` operators in favor of explicit `Destination.body` in the future.

stephencelis:
Thanks again for taking the time to submit, though, and for starting the conversation! If you do end up making an "extras" library with these helpers, we'll link to it in the README.

PR #2837, opened 2024-02-20
Add missing opening brace in a TreeBasedNavigation example

imjn:
Just adding a missing opening brace in one of the examples in TreeBasedNavigation.md


mbrandonw (approved):
Good catch, thanks!

PR #2844, opened 2024-02-21
fix unnecessary diff

oka-yuji:
Unnecessary diff were created by spaces, so spaces were added and diff were eliminated


stephencelis (approved):
Thanks!

PR #2853, opened 2024-02-25
fix document in PresentationReducer.swift

ozumin:
Hi, I thought this part should be `@Presents`, not `@PresentationState`.
The document just before says
> This version of `ifLet` requires the usage of the ``Presents()`` macro

so I thought maybe it should be `@Presents`.

### DIFF
--- a/Sources/ComposableArchitecture/Reducer/Reducers/PresentationReducer.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/PresentationReducer.swift
@@ -341,7 +341,7 @@ extension Reducer {
   /// @Reducer
   /// struct Parent {
   ///   struct State {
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Reducer/Reducers/PresentationReducer.swift
```diff
@@ -341,7 +341,7 @@ extension Reducer {
   /// @Reducer
   /// struct Parent {
   ///   struct State {
```
mbrandonw, line 343:
```suggestion
  ///   @ObservableState
  ///   struct State {
```

mbrandonw (approved):
Good catch, thanks!

PR #2856, opened 2024-02-25
Remove DEBUG in tests to get more coverage in release

mbrandonw:
We removed DEBUG checks in XCTestDynamicOverlay over a month ago, and it seems to be going fine so far. So, I think it's safe to start asserting on more things in TCA, in particular we can now do `XCTExpectFailure` in release builds.

I'm having trouble running these tests locally in RELEASE for some reason, so curious to see what CI does.

### DIFF
--- a/Sources/ComposableArchitecture/Observation/IdentifiedArray+Observation.swift
+++ b/Sources/ComposableArchitecture/Observation/IdentifiedArray+Observation.swift
@@ -69,11 +69,9 @@
       state: KeyPath<State, IdentifiedArray<ElementID, ElementState>>,
       action: CaseKeyPath<Action, IdentifiedAction<ElementID, ElementAction>>
     ) -> some RandomAccessCollection<Store<ElementState, ElementAction>> {
-      #if DEBUG
-        if !self.canCacheChildren {
-          runtimeWarn(uncachedStoreWarning(self))
-        }
-      #endif
+      if !self.canCacheChildren {
+        runtimeWarn(uncachedStoreWarning(self))
+      }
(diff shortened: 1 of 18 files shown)


Sources/ComposableArchitecture/Observation/IdentifiedArray+Observation.swift
```diff
@@ -69,11 +69,9 @@
...
       action: CaseKeyPath<Action, IdentifiedAction<ElementID, ElementAction>>
     ) -> some RandomAccessCollection<Store<ElementState, ElementAction>> {
-      #if DEBUG
-        if !self.canCacheChildren {
-          runtimeWarn(uncachedStoreWarning(self))
-        }
-      #endif
+      if !self.canCacheChildren {
+        runtimeWarn(uncachedStoreWarning(self))
+      }
```
mbrandonw, line 74: This `#if DEBUG` _technically_ doesn't need to be removed, but `runtimeWarning` is already wrapped in `#if DEBUG` so I thought it might be nice to just clean up some unnecessary noise at call sites.

Tests/ComposableArchitectureTests/EffectFailureTests.swift
```diff
@@ -1,34 +1,32 @@
-#if DEBUG
-  import Combine
-  @_spi(Internals) import ComposableArchitecture
-  import XCTest
+import Combine
+@_spi(Internals) import ComposableArchitecture
```
mbrandonw, line 2: You should probably review this PR with whitespace turned off.

stephencelis (approved):

PR #2858, opened 2024-02-26
Shared State

stephencelis:
This PR introduces all new tools for sharing state in the Composable Architecture.

For more information and for questions and comments, please see this discussion: https://github.com/pointfreeco/swift-composable-architecture/discussions/2857


Sources/ComposableArchitecture/Documentation.docc/Articles/SharingState.md
```diff
@@ -0,0 +1,539 @@
...
+@Shared(.appStorage("count")) var count = 0
+```
+
+That small change will guarantee that all changes to `count` are persisted and will be 
+automatically loaded the next time the application launches.
+
+This form of persistence only works for simple data types because that is what works best with
+`UserDefaults`. This includes strings, booleans, integers, doubles, URLs, data, and more. If you
+need to store more complex data, such as custom data types serialized to JSON, then you will want
+to use the <doc:SharingState#File-storage> strategy or a <doc:SharingState#Custom-persistence>
```
Muhammed9991, line 163: This is showing up like this:
Screenshot 2024-02-27 at 15 46 14

mbrandonw: Is that screenshot from GitHub or from Xcode docs? It’s only meant to render properly in Xcode as it’s Docc syntax.

Muhammed9991: Ahh. My bad. That was GitHub

Muhammed9991: Just noticed this says "then you will want to use the Sharing state strategy or a Sharing state strategy." Should it not say.

```
then you will want to use the File storage strategy or a custom persistence strategy
```

Screenshot from Xcode Docs 

Screenshot 2024-02-28 at 10 44 28

mbrandonw: What version of Xcode are you using and how are you viewing the docs? When the docs are fully built it will show correctly. For example, here is what it looks like for me when I build docs (cmd+ctrl+shift+D):

Screenshot 2024-02-28 at 4 20 28 PM

But for whatever reason the preview of docs in Xcode does not expand those names. So I don't think this is actually an issue when docs are generated.

Muhammed9991: I was using Xcode 15.2. And yes, it was previews.

Sources/ComposableArchitecture/Documentation.docc/Articles/SharingState.md
```diff
@@ -0,0 +1,731 @@
...
+And this technique also works on [custom persistence](<doc:SharingState#Custom-persistence>)
+strategies.
+
+## Shared state in pre-observation apps
+
+It is possible to use [`@Shared`](<doc:Shared>) in features that have not yet been updated with
+the observation tools released in 1.7, such as the ``ObservableState()`` macro. In the reducer
+you can use `@Shared` regardless of your use of the observation tools. 
+
+However, if you are deploying to iOS 16 or earlier, then you must use `WithObservationTracking`
```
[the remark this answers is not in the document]
mbrandonw: Yep good catch, thanks!

mbrandonw (approved):

Sources/ComposableArchitecture/SharedState/PersistenceKey/FileStorageKey.swift
```diff
@@ -0,0 +1,303 @@
...
+      id: AnyHashableSendable(queue),
+      async: { queue.async(execute: $0) },
+      asyncAfter: { queue.asyncAfter(deadline: .now() + $0, execute: $1) },
+      createDirectory: {
+        try FileManager.default.createDirectory(at: $0, withIntermediateDirectories: $1)
+      },
+      fileExists: { FileManager.default.fileExists(atPath: $0.path) },
+      fileSystemSource: {
+        let source = DispatchSource.makeFileSystemObjectSource(
+          fileDescriptor: open($0.path, O_EVTONLY),
```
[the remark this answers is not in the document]
stephencelis: @Matejkob We'd take a PR that bails out on a negative file descriptor if you're down to submit! Since it's generally a developer-specified URL then failure to access will generally be a programmer error, so a crash is appropriate there, but there is the edge case where the URL could come in from the user, so we shouldn't make the assumption.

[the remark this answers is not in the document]
Matejkob: Sure thing, can do. The question is how do you wanna handle such an error? 
The `createDirectory` endpoint can throw an error as well and I can see that you handled it by ignoring it: https://github.com/pointfreeco/swift-composable-architecture/blob/8624a8e2c35f78e5e61d5f12c69110e9020e80c8/Sources/ComposableArchitecture/SharedState/PersistenceKey/FileStorageKey.swift#L73 
Do you think it would be appropriate solution here? Or maybe we should emit error down to the user land?

PR #2860, opened 2024-02-26
Clean up some typos in SharingState.md

NFulkerson:
There were a few minor typos in the latter half of the Sharing State documentation I thought I could clean up.


mbrandonw (approved):
Good catches, thanks!

PR #2869, opened 2024-02-29
chore: Add a sheet example for ReducerCaseIgnored

vladyslavsosiuk:
This PR aims to improve the documentation example for the ReducerCaseIgnored Macro by adding a demonstration of how to scope a store for the sheet modifier


mbrandonw:
Hi @vladyslavsosiuk, thanks for taking the time to contribute! However, we do not think this is the most appropriate place for this information. We think it'd be better to get an example of this in a case study and update the navigation docs. I'm going to close this for now, but we do have plans to improve things soon.

PR #2880, opened 2024-03-01
Give persistence keys a synchronous update interface

stephencelis:
Adding an async sequence around persistence updates caused a bunch of synchronization issues, so let's change things to use a synchronous API instead.

### DIFF
--- a/Sources/ComposableArchitecture/SharedState/PersistenceKey.swift
+++ b/Sources/ComposableArchitecture/SharedState/PersistenceKey.swift
@@ -19,15 +19,36 @@
     /// Saves a value to storage.
     func save(_ value: Value)
 
-    /// An async sequence that emits when the external storage is changed from the outisde.
+    /// An async sequence that emits when the external storage is changed from the outside.
     ///
     /// An "empty" sequence that does not emit and finishes immediately is provided by default.
     var updates: Updates { get }
+
+    func subscribe(didSet: @escaping (Value?) -> Void) -> Shared<Value>.Subscription
   }
 
   extension PersistenceKey where Updates == _Empty<Value?> {
     public var updates: _Empty<Value?> {
       _Empty()
+    }
+
+    public func subscribe(didSet: @escaping (Value?) -> Void) -> Shared<Value>.Subscription {
+      Shared.Subscription {}
+    }
+  }
+
+  extension Shared {
+    public class Subscription {
+      let onCancel: () -> Void
+      init(onCancel: @escaping () -> Void) {
+        self.onCancel = onCancel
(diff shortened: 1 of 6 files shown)


stephencelis:
@oliverfoggin I think you were seeing some issues with app storage. Want to point to this branch instead of the beta and see if it clears things up?

oliverfoggin:
Perfect! Will do and I'll report back. Thanks 👍🏻

oliverfoggin:
Hey @stephencelis I've updated to this branch but the compiler seems to be having difficulties with it.

I'm getting a few...

> /.../BlahView.swift:15:29 The compiler is unable to type-check this expression in reasonable time; try breaking up the expression into distinct sub-expressions

That I haven't seen before with no code changes.

oliverfoggin:
I'll have a look in a bit.

stephencelis:
@oliverfoggin Nothing in this PR should have affected a view builder, but we did just merge some updates from `main` that include an overload that could have an impact on some views, especially those with lots of case key path scoping in them.

We can also chat in Slack if it helps get to the bottom of things quicker.

327Fffs (approved):

327Fffs (approved):

Sources/ComposableArchitecture/SharedState/PersistenceKey.swift
```diff
@@ -11,32 +11,35 @@
...
-  public struct _Empty<Element>: AsyncSequence {
-    public struct AsyncIterator: AsyncIteratorProtocol {
-      public func next() async throws -> Element? { nil }
-    }
-    public func makeAsyncIterator() -> AsyncIterator {
-      AsyncIterator()
+  extension Shared {
+    public class Subscription {
+      let onCancel: () -> Void
+      init(onCancel: @escaping () -> Void) {
```
hallee, line 34: Should this initializer be public since custom `PersistenceKey`s need to be able to return this from `func subscribe(didSet:)`?

hallee, line 34:
```suggestion
      public init(onCancel: @escaping () -> Void) {
```

Should this initializer be public since custom `PersistenceKey`s need to be able to return this from `subscribe(didSet:)`?

stephencelis: Yup, I think so, thanks!

stephencelis:
@oliverfoggin Going to merge this back into the beta since we think it's the right direction to go, but would still love to know if the issues you were seeing earlier are now fixed!

PR #2885, opened 2024-03-04
Leverage dynamic case writability in example code

stephencelis:
This PR simplifies various examples to use dynamic writability of case pathable enums.

### DIFF
--- a/ComposableArchitecture.xcworkspace/contents.xcworkspacedata
+++ b/ComposableArchitecture.xcworkspace/contents.xcworkspacedata
@@ -2,9 +2,6 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "group:Sources/ComposableArchitecture/Documentation.docc/Articles/MigrationGuides/MigratingTo1.9.md">
-   </FileRef>
-   <FileRef
       location = "group:">
    </FileRef>
    <FileRef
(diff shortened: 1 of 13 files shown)


Sources/ComposableArchitecture/TestStore.swift
```diff
@@ -529,43 +529,6 @@ public final class TestStore<State, Action> {
...
-  /// See <doc:Testing> and the documentation of ``TestStore`` for more information on how to best
-  /// use a test store.
-  ///
-  /// - Parameters:
-  ///   - initialState: The state the feature starts in.
-  ///   - reducer: The reducer that powers the runtime of the feature.
-  ///   - prepareDependencies: A closure that can be used to override dependencies that will be
-  ///     accessed during the test. These dependencies will be used when producing the initial
-  ///     state.
-  @available(*, deprecated, message: "State must be equatable to perform assertions.")
```
stephencelis, line 543: Unrelatedly, this deprecated overload of `TestStore.init` has started to cause diagnostics problems in Xcode. It was introduced before 1.0.0 when test stores could be scoped, but is no longer necessary or useful, so let's drop it entirely.

ComposableArchitecture.xcworkspace/contents.xcworkspacedata
```diff
@@ -1,9 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Workspace
    version = "1.0">
-   <FileRef
-      location = "group:Sources/ComposableArchitecture/Documentation.docc/Articles/MigrationGuides/MigratingTo1.9.md">
```
stephencelis, line 5: An Xcode bug adds these to the root of the workspace for some reason.

mbrandonw (approved):

PR #2892, opened 2024-03-06
Bump swift-perception

takehilo:


stephencelis:
Thanks for taking the time to PR! Our general stance, however, is to avoid explicitly bumping dependency versions unless there is a mission-critical fix or new feature we depend on. This gives consumers of our package more control and allows them to update at a more gradual pace.

Because of this I'm going to close for now, but if you'd like to start a discussion about this, we'd be down to chat more about it!

takehilo:
Thanks for your reply!

PR #2900, opened 2024-03-07
Bump swift-custom-dump to allow archiving with Xcode 15.3

HarshilShah:


Package.swift
```diff
@@ -23,7 +23,7 @@ let package = Package(
     .package(url: "https://github.com/pointfreeco/combine-schedulers", from: "1.0.0"),
     .package(url: "https://github.com/pointfreeco/swift-case-paths", from: "1.3.0"),
     .package(url: "https://github.com/pointfreeco/swift-concurrency-extras", from: "1.1.0"),
-    .package(url: "https://github.com/pointfreeco/swift-custom-dump", from: "1.0.0"),
```
stephencelis, line 26: `from: "1.0.0"` should allow targeting `1.2.1`, but you may need to manually update your packages to lock the resolve file.

HarshilShah: Yep, I just thought it might be worth including the bump explicitly in the next release so folks just get the update automatically

stephencelis: We've unfortunately gotten folks into bad spots by being overly strict with these bumps in the past, so our general stance has been to avoid explicit bumps unless there's a mission critical (security) bug fix, or a feature of the dependency we require a bump to use.

We will require a new version of Custom Dump for the sharing tools, though, so the bump is already on that branch. For now I think we'll hold off on the bump, though. Thanks for taking the time to PR it, though!

PR #2904, opened 2024-03-08
Pass initial shared value to strategy / register app storage

stephencelis:
This PR modifies the `PersistenceKey` protocol so that its `load` and `subscribe` endpoints are handed the initial value and must return a non-optional value. By feeding this value in, we can ensure that `.appStorage` declared in a feature is registered outside the feature.

### DIFF
--- a/Sources/ComposableArchitecture/SharedState/References/ValueReference.swift
+++ b/Sources/ComposableArchitecture/SharedState/References/ValueReference.swift
@@ -22,7 +22,7 @@
     }
     
     public init(
-      wrappedValue value: @autoclosure @escaping () -> Value,
+      wrappedValue value: Value,
       _ persistenceKey: some PersistenceKey<Value>,
       fileID: StaticString = #fileID,
       line: UInt = #line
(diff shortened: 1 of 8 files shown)


Sources/ComposableArchitecture/SharedState/References/ValueReference.swift
```diff
@@ -114,18 +114,20 @@
     #endif
 
     init(
-      initialValue: @autoclosure @escaping () -> Value,
+      initialValue: Value,
```
stephencelis, line 117: I think we were being a little too precious here with the autoclosure, which needed to be escaped. I think simply passing the value along is probably better.

Tests/ComposableArchitectureTests/AppStorageTests.swift
```diff
@@ -14,6 +14,16 @@ final class AppStorageTests: XCTestCase {
     XCTAssertEqual(defaults.integer(forKey: "count"), 1)
   }
 
+  func testDefaultsRegistered() {
+    @Dependency(\.defaultAppStorage) var defaults
+    @Shared(.appStorage("count")) var count = 42
+    XCTAssertEqual(defaults.integer(forKey: "count"), 42)
```
stephencelis, line 20: Previously this would fail because `defaults.integer(forKey: "count")` would return `0`.

mbrandonw (approved):

PR #2905, opened 2024-03-08
Add support for Date and Optional Date in AppStorage

rurza:
I noticed a potential oversight where support for Date and Optional<Date> was missing in the appStorage implementation. This PR rectifies that by introducing support for these types, enhancing the library's usability and flexibility.

### DIFF
--- a/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
+++ b/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
@@ -9,7 +9,7 @@
     /// wrapper:
     ///
     /// ```swift
-    /// @Shared(.inMemory("appLaunchedAt")) var appLaunchedAt = Date()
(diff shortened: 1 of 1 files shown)


stephencelis:
@rurza Thanks for taking the time to PR. This actually wasn't an oversight! We designed `.appStorage` to mostly have the same APIs as [`@AppStorage`](https://developer.apple.com/documentation/swiftui/appstorage), which doesn't have `Date` initializers, either.

Our current thought is that folks can use other types by defining their own `RawRepresentable` wrappers, or by extending `UserDefaults` with properties and using the key path interface, instead:

```swift
extension UserDefaults {
  @dynamic @objc var openedAt: Date {
    get { object(forKey: "opened-at") }
    set { set(newValue, forKey: "opened-at") }
  }
}

@Shared(.appStorage(\.openedAt)) var openedAt = Date()
```

How's that sound?

rurza:
Sounds good :)

PR #2906, opened 2024-03-09
Add `StackActionOf`

d-date:
It turns out that this typealias is not real, and now real.

https://github.com/pointfreeco/swift-composable-architecture/blob/a0acf426f7e8e794563b6bcbe4fcf5a9fd6cebef/Sources/ComposableArchitecture/Observation/NavigationStack%2BObservation.swift#L22

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/03-NavigationStack.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/03-NavigationStack.swift
@@ -22,7 +22,7 @@ struct NavigationDemo {
   enum Action {
     case goBackToScreen(id: StackElementID)
     case goToABCButtonTapped
-    case path(StackAction<Path.State, Path.Action>)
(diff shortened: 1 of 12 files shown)


schrismartin:
I was literally just wondering why this didn't exist. +1

ddanilyuk:
Also, there is similar feature request for `IdentifiedStateOf` in [this discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions/2645).

stephencelis (approved):
Thanks! Given the typo and `IdentifiedActionOf`, we think it's probably good to add this.

stephencelis:
@ddanilyuk We think `IdentifiedStateOf` is just off the beaten path enough that it's probably not a good idea to include. `IdentifiedAction` could also be used outside of `IdentifiedArray` in the future, so we don't want to lock down the idea of "identified state" to a single thing.

Issue #2907, opened 2024-03-11
Warnings when using observable state in sheet presented from presentation state


lukeredpath:
### Description

A purple runtime warning appears complaining that observation is not compatible with uncached stores even though I'm not aware that I am using uncached stores.

### Checklist

- [X] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Expected behavior

I am able to use observation tools without a warning.

### Actual behavior

I'm a bit confused about the coupling between store caching and the newer observation tools. I was under the impression the two things were independent of each other, but there still seems to be an issue when using the old presentation tools to present a feature that uses observation state.

I am aware that it is preferable to migrate to `ObservationState` from the root towards leaf features but that isn't always possible - it would be a shame if we couldn't selectively opt in to using observation on newer features without having to massively refactor a lot of other code first. 

However, when attempting this I am getting a purple runtime warning about observation not being compatible with unscoped stores. This confused me because I didn't realise I was using unscoped stores - I had already taken the time to convert my `scope` calls over to using key paths instead of closures.

Digging into this, it seems the problem occurs when you use any of the old navigation APIs such as `.sheet(store:)` - even if you hand this a store scoped using key paths (e.g. `\.$destination.feature`), it internally creates a new store that has a nil `id` which means the feature being presented has a store at its root that does not support store caching. This means any calls to `store.scope` within the new feature result in a purple runtime warning.

Is there a reason why the old navigation APIs do this? Can they not be made to work with store caching when given a store that is scoped correctly?

### Steps to reproduce

I have uploaded a full repro app of this here:
https://gist.github.com/lukeredpath/f895504cc16fd11f9c0f541f6bf4573e

In this app, the `RootFeature` uses `@ObservableState` and presents a "legacy" feature that has not been updated to use `@ObservableState` - it still uses `WithViewStore` and the `.sheet(store:)` APIs. It uses the legacy `.sheet` API to present a `NewFeature` that uses `@ObservableState`. This new feature allows you to toggle a child section in its list that uses a child feature using `if let child = store.scope(...)` - when you toggle this section the runtime warning appears.

### The Composable Architecture version information

1.9.2

### Destination operating system

[...]

PR #2908, opened 2024-03-11
Give internal presentation store an id

lukeredpath:
Addresses the problem raised in #2907 where using legacy presentation APIs (e.g. `sheet(store:)`) result in the presented destination always receiving a store that does not support store caching due to the hardcoded `nil` ID given to the internal store that is created.

This prevents you from presenting features that uses observation tools using these APIs without lots of purple runtime warnings even though there's no reason why you shouldn't be able to do this.

Note: An [earlier implementation](https://github.com/pointfreeco/swift-composable-architecture/commit/087b6b5730df5494ec7cefa8add8ced89c645429#diff-a7aa3773363760502a84ef57acd54a1b745cee8cedad7a01a065bd9083546c7eL220) in of this logic used to have a comment stating that a `\.self` ID could not be used because it would prevent dismissal. In my [example app](https://gist.github.com/lukeredpath/f895504cc16fd11f9c0f541f6bf4573e) I am able to present and dismiss sheets using both the SwiftUI `dismiss` environment value, or by sending an action to the store and setting the destination state to `nil`.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/PresentationModifier.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/PresentationModifier.swift
@@ -215,7 +215,7 @@ public struct PresentationStore<
     ) -> Content
   ) where State == DestinationState, Action == DestinationAction {
     let store = store.scope(
-      id: nil,
+      id: store.id(state: \.self, action: \.self),
       state: ToState(\.self),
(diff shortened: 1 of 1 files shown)


stephencelis (approved):
Thanks!

PR #2923, opened 2024-03-18
Swift bug workaround for `TestStore` completion

stephencelis:
Due to a Swift bug, stores and test stores initialized with type inference are broken when it comes to autocomplete:

https://github.com/apple/swift/issues/72399

For example, this test store cannot be completed from:

```swift
let store = TestStore(initialState: Feature.State()) {
  Feature()
}
```

While this can affect regular stores, as well, it is less likely outside of root stores, since stores are typically declared in views and view controllers as properties, and inference isn't required:

```swift
let store: StoreOf<Feature>
```

While the user can employ a workaround in their project by making the types very explicit:

```diff
-let store = TestStore(initialState: Feature.State()) {
+let store: TestStoreOf<Feature> = TestStore(initialState: Feature.State()) {
   Feature()
 }
```

This PR takes the approach of removing the builder context from `TestStore.init`, since the issue is more common in tests, and because composing a reducer for a test store is less common.

### DIFF
--- a/Sources/ComposableArchitecture/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestStore.swift
@@ -499,13 +499,17 @@ public final class TestStore<State, Action> {
   ///
   /// - Parameters:
   ///   - initialState: The state the feature starts in.
-  ///   - reducer: The reducer that powers the runtime of the feature.
+  ///   - reducer: The reducer that powers the runtime of the feature. Unlike
+  ///     ``Store/init(initialState:reducer:withDependencies:)``, this is _not_ a builder closure
+  ///     due to a [Swift bug](https://github.com/apple/swift/issues/72399) that is more likely to
+  ///     affect test store initialization. If you must compose multiple reducers in this closure,
+  ///     wrap them in ``CombineReducers``.
   ///   - prepareDependencies: A closure that can be used to override dependencies that will be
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/TestStore.swift
```diff
@@ -499,13 +499,17 @@ public final class TestStore<State, Action> {
...
+  ///     ``Store/init(initialState:reducer:withDependencies:)``, this is _not_ a builder closure
+  ///     due to a [Swift bug](https://github.com/apple/swift/issues/72399) that is more likely to
+  ///     affect test store initialization. If you must compose multiple reducers in this closure,
+  ///     wrap them in ``CombineReducers``.
   ///   - prepareDependencies: A closure that can be used to override dependencies that will be
   ///     accessed during the test. These dependencies will be used when producing the initial
   ///     state.
   public init<R: Reducer>(
-    initialState: @autoclosure () -> R.State,
-    @ReducerBuilder<State, Action> reducer: () -> R,
```
stephencelis, line 508: We unfortunately can't add a deprecated, disfavored overload of this initializer.

> Invalid redeclaration of 'init(initialState:reducer:withDependencies:file:line:)'

mbrandonw (approved):

PR #2930, opened 2024-03-19
Added Privacy Manifest file

DandyLyons:
### "Blank"ish Privacy manifests
Food for thought: GRDB recently, added a Privacy manifest file which is explicitly shows no tracking. That PR is [here](https://github.com/groue/GRDB.swift/commit/1920482158af699c6b3a55860b7697f91a318bf8). This may seem pointless, but I suspect it will be best practice going forward. By explicitly saying that our library does not track the user, we reduce the surface area that devs need to search for compliance. 

### User Defaults
Thankfully it looks like Apple added a privacy reason which perfectly matches our use case: 
- `C56D.1`: https://arc.net/l/quote/aqxlpvkx
> Declare this reason if your third-party SDK is providing a wrapper function around user defaults API(s) for the app to use, and you only access the user defaults APIs when the app calls your wrapper function.

My guess is that this reason will cover the privacy compliance of TCA, but if the app dev uses `.appStorage` for example, they will still be responsible for declaring if their reason is legitimate.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -42,6 +42,9 @@ let package = Package(
         .product(name: "OrderedCollections", package: "swift-collections"),
         .product(name: "SwiftUINavigationCore", package: "swiftui-navigation"),
         .product(name: "XCTestDynamicOverlay", package: "xctest-dynamic-overlay"),
+      ],
+      resources: [
+        .process("PrivacyInfo.xcprivacy")
       ]
     ),
     .testTarget(
--- a/Package@swift-5.9.swift
+++ b/Package@swift-5.9.swift
@@ -49,6 +49,9 @@ let package = Package(
         .product(name: "Perception", package: "swift-perception"),
         .product(name: "SwiftUINavigationCore", package: "swiftui-navigation"),
         .product(name: "XCTestDynamicOverlay", package: "xctest-dynamic-overlay"),
+      ],
+      resources: [
+        .process("PrivacyInfo.xcprivacy")
       ]
     ),
     .testTarget(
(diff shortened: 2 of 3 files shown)


DandyLyons:
I thought it made most sense to add this manifest to `shared-state-beta` first, and if need be we can merge it into `main` later.

stephencelis (approved):
Thanks!

Package.swift
```diff
@@ -42,6 +42,9 @@ let package = Package(
         .product(name: "OrderedCollections", package: "swift-collections"),
         .product(name: "SwiftUINavigationCore", package: "swiftui-navigation"),
         .product(name: "XCTestDynamicOverlay", package: "xctest-dynamic-overlay"),
+      ],
+      resources: [
+        .process("PrivacyInfo.xcprivacy")
```
stephencelis, line 47: Does this need to be added to [Package@swift-5.9.swift](https://github.com/pointfreeco/swift-composable-architecture/blob/main/Package%40swift-5.9.swift), too?

DandyLyons: I believe so?

For any TCA users publishing an app using 5.9 after the App Store rules change. 

(Sorry, I don't have experience with @ files for Swift Packages, but I assume it's a Package.swift file used for people still developing in a Swift 5.9 environment).

stephencelis: @DandyLyons Can you add those lines there, as well? Basically anyone on a modern Xcode will be using the -5.9.swift version (the other version you updated will only be used by <5.9).

stephencelis: I pushed it quick to your branch, will merge assuming CI looks generally good! I think the shared state branch CI isn't totally green right now, so no worries if it partially fails.

DandyLyons: Thanks. Sorry. I was out of the house. Let me know if I can help with anything else.

PR #2939, opened 2024-03-22
Add package access modifier support to @Reducer on enums

samrayner:
Addresses https://github.com/pointfreeco/swift-composable-architecture/discussions/2936

I'm not sure why the access modifier inserts a leading line break but that seems to have been the case already for `public` so I've left it as is.

Shout if anything needs tweaking!


stephencelis (approved):
Thanks!

stephencelis:
> I'm not sure why the access modifier inserts a leading line break but that seems to have been the case already for `public` so I've left it as is.

I think it's taking the trivia between `@Reducer` and the modifier. Seems like a good thing to fix but we can do that in another commit.

PR #2942, opened 2024-03-23
Remove force unwrap!

johnno1962:
Hi,

I was wondering if you'd be prepared to replace a force unwrap that has been problematic for [InjectionIII](https://github.com/johnno1962/InjectionIII) users with an explanatory message suggesting they could set an environment variable during development to avoid the problem. This arose out of [this GitHub issue](https://github.com/krzysztofzablocki/Inject/issues/87#issuecomment-2015514768) where injection and key paths were not happy bed fellows. 

Perhaps you don't want to clutter your code with something so specific to InjectionIII but it would seem better than a force unwrap failure my users are encountering!

Cheers,

John

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -683,7 +683,20 @@ private enum PartialToState<State> {
   func appending<ChildState>(_ state: PartialToState<ChildState>) -> PartialToState<ChildState> {
     switch (self, state) {
     case let (.keyPath(lhs), .keyPath(rhs)):
+      #if DEBUG
+      guard let appended = lhs.appending(path: rhs) else {
+        fatalError("""
+          ℹ️ Key path has failed to append. This can occur if the \
+          root or value types have been replaced. If you are using \
+          InjectionIII.app v4.8.4+, you can create an environment \
+          variable INJECTION_KEYPATHS in your scheme to avoid this \
+          problem.
+          """)
+      }
+      return .keyPath(appended)
+      #else
       return .keyPath(lhs.appending(path: rhs)!)
+      #endif
     case let (.closure(lhs), .keyPath(rhs)):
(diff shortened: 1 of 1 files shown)


stephencelis:
Hey @johnno1962! Thanks for bringing this to our attention. We're traveling at the moment for a conference but will try to take a look soon

johnno1962:
👋 Thanks for getting back. Perhaps you might be concerned some of your users might find the message confusing but I'd anticipate only people who are already using InjectionIII would ever see it. No hurry, I haven't even fully released 4.8.4 yet.

stephencelis:
@johnno1962 Revisiting this now that we're back from traveling!

I wonder, if we extended `AnyKeyPath` with our own method, let's say it were called `_pf_unsafeAppending` and it simply invoked `!` under the hood, would this be enough for Injection to hook into and provide its own error messaging, or is there some other method symbol name we can use to let Injection know that it's effectively a force unwrap?

I think we'd prefer to avoid introducing messaging into the library itself for an external/unrelated library, since only a subsection of our users will be using Injection, and it may be confusing for everyone else that encounters it.

johnno1962:
Not to worry, I anticipated this might be your response and have modified the latest release candidate of InjectionIII to turn on the INJECTION_KEYPATH processing automatically if the composable architecture is present so the message is less important. I'm wondering though if in general you might be best replacing the force unwrap with a message if we could agree on the text. The first part seems OK to me so I've added another commit.

stephencelis:
@johnno1962 I think the main thing for us to think about is that this key path appending is an implementation detail, so we're not sure the proper messaging. We'll chat about it and get back to you!

stephencelis:
@johnno1962 We chatted about this and aren't sure there's much messaging we can provide, especially if this code path won't get hit by InjectionIII. Going to close for now, but if we're missing any context, please let us know!

PR #2944, opened 2024-03-24
Fix StackBasedNavigation.md

takehilo:
Case name should be specified but missed.


mbrandonw (approved):
Good catches, thank you!

PR #2948, opened 2024-03-28
Alert with binding

randomeizer:
Added Case Study for using new `alert` and `confirmationDialog` view modifiers.


randomeizer:
I ended up renaming the old `AssertAndConfirmationDialogs` to `AssertAndConfirmationDialogState` and added the new API in an `AssetAndConfirmationDialogs` case study. Not sure if these examples are a good idea?

stephencelis:
@randomeizer Thanks for showing a demo of these! I think we'll probably hold off on adding dedicated case studies for now, but we'll consider it in the future whenever we do a larger revamp.

randomeizer:
No worries. It's a bit of an awkward fit right now to be honest.

PR #2960, opened 2024-03-31
Add AppStorage persistence keys for types that are raw representable as Data

AlexKobachiJP:
AppStorage (UserDefaults) supports persisting, among other types, Int, String and Data.  Further, the library can persist types that are raw representable as Int or String, but doesn’t do the bridging for types that are raw representable as Data.

Having support for this, one can then easily persist any type in AppStorage that is Codable (because Codeable types can be coded to and decoded from Data).  While I think the library could/should add support for persisting Codable types out of the box, if that is not desirable for some reason, clients can work around this easily as long as raw representing Data itself is persistable.

A workaround I have been using in my project is to define a property wrapper (this is preferable to conforming types, esp. those one does not own, to RawRepresentable directly):

```swift
import Foundation

@propertyWrapper 
public struct RawRepresentableFromCodable<Value: Codable> {
  public var wrappedValue: Value

  public init(wrappedValue: Value) {
    self.wrappedValue = wrappedValue
  }
}

extension RawRepresentableFromCodable: Equatable where Value: Equatable {}

extension RawRepresentableFromCodable: RawRepresentable {
  public init?(rawValue: Data) {
    guard let value = try? JSONDecoder().decode(Value.self, from: rawValue) else {
      return nil
    }
    self = RawRepresentableFromCodable(wrappedValue: value)
  }

  public var rawValue: Data {
    try! JSONEncoder().encode(wrappedValue)
  }
}
```

and then add add that before `@Shared`:

```swift
@Shared(.appStorage(“someKey”))
@RawRepresentableFromCodable
public var someProperty: SomeCodableType
```

### DIFF
--- a/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
+++ b/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
@@ -76,6 +76,16 @@
       AppStorageKey(key)
     }
 
+    /// Creates a persistence key that can read and write to a data user default, transforming
+    /// that to a `RawRepresentable` data type.
+    ///
+    /// - Parameter key: The key to read and write the value to in the user defaults store.
+    /// - Returns: A user defaults persistence key.
+    public static func appStorage<Value: RawRepresentable>(_ key: String) -> Self
+    where Value.RawValue == Data, Self == AppStorageKey<Value> {
+      AppStorageKey(key)
+    }
+
     /// Creates a persistence key that can read and write to an optional boolean user default.
     ///
     /// - Parameter key: The key to read and write the value to in the user defaults store.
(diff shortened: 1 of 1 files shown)


stephencelis:
@AlexKobachiJP Thanks for taking the time to PR this. For now we would like to stick with providing the same APIs on `.appStorage` as Apple's native `@AppStorage`, and to our knowledge Apple does _not_ provide such APIs and instead would also require the workaround you post above. Unless we're mistaken, do you agree it's a good idea to remain consistent with Apple here and close this PR for now?

AlexKobachiJP:
@stephencelis thanks for the review!  If staying close to Apple's implementation is a goal---it _does_ make it easier to switch back if necessary---then not merging this is agreeable.  P.S. It was good to meet you guys at try! Swift Tokyo last month!

----

Just in case this helps anyone:  Without merging the current PR, the workaround is simply to use `String` instead of `Data` as the raw representing type, and one can still put any `Codable` type into `Shared` (the use of Base64-encoding just increase storage size vs. raw bytes, but that should not cause concern):

```swift
import Foundation

@propertyWrapper 
public struct RawRepresentableFromCodable<Value: Codable> {
  public var wrappedValue: Value

  public init(wrappedValue: Value) {
    self.wrappedValue = wrappedValue
  }
}

extension RawRepresentableFromCodable: Equatable where Value: Equatable {}

extension RawRepresentableFromCodable: RawRepresentable {
  public init?(rawValue: String) {
    guard
      let base64EncodedData = rawValue.data(using: .utf8),
      let data = Data(base64Encoded: base64EncodedData),
      let value = try? JSONDecoder().decode(Value.self, from: data)
    else {
      return nil
    }
    self = RawRepresentableFromCodable(wrappedValue: value)
  }

  public var rawValue: String {
    try! JSONEncoder().encode(wrappedValue).base64EncodedString()
  }
}
```

stephencelis:
Nice to meet you too! And thanks for understanding, and sharing the alternative for folks interested.

PR #2962, opened 2024-04-01
Fix Favorites case study by using episode mocks

beribas:
The Favorites case study doesn't list any favorites. 
Here is a small fix for that.


mbrandonw (approved):
Good catch @beribas! I made a few small tweaks, but otherwise looks good. Thanks!

PR #2968, opened 2024-04-05
Make Effect.publisher eager

mbrandonw:
Makes the trailing closure of `Effect.publisher { … }` non-escaping, and hence the publisher is created eagerly when the effect is constructed. This means you can access `state` inside the closure, which will be handy in the shared state tools coming to TCA soon:

```swift
return .publisher { 
  state.$shared.publisher
    .map(Action.sharedUpdated)
}
```

Note, however, the publisher is not _run_ eagerly. That is still delayed until state changes have been processed and all effects are started together.

We don't think there is a big use case for needing the construction of a publisher to be lazy since the publisher itself can encapsulate lazy work, but if for whatever reason one needs to lazily create the publisher then you can use `Deferred`:

```swift
return .publisher {
  Deferred {
    // create publisher
  }
}
```


stephencelis (approved):
Seems like a general improvement! Let's just be careful to mention the change in the logs for folks that may be impacted by the change.

PR #2976, opened 2024-04-09
Shared state beta task snaps

stephencelis:
This PR moves `@Shared`'s snapshotting into the change tracking system itself. This makes it possible for multiple change trackers to interleave (_e.g._ a reducer printing changes shouldn't interfere with a test store assertion), and it prevents snapshots from leaking into reference state (since they live in the tracker now).


mbrandonw (approved):

juliensagot:
Hi folks 👋 
I just wanted to let you know that those changes caused a regression with the `observe` method in UIKit. 
The closure is now called **before** the state has changed, whereas it used to be properly called **after** the state had effectively changed in the previous commit (7fd4e89ff891b5664bc4fc1d1d091dd3624c88b8).

mbrandonw:
Hi @juliensagot, this PR did not make any changes to `observe`, nor does the shared-state-beta PR. Also we have tests on this that pass and show that `observe` is called _after_ state changes.

Are you able to provide a project that reproduces the problem you are seeing?

juliensagot:
@mbrandonw Here you go! Apologies for the number of lines, it's... UIKit 😅
With TCA on commit 7fd4e89ff891b5664bc4fc1d1d091dd3624c88b8 it will display "Banana" on the screen, meanwhile on commit 8b841181012a43ce6048f9c14e6925c4a39f44f7, it will display nothing.

[TCAObserveUIKit.zip](https://github.com/pointfreeco/swift-composable-architecture/files/14952339/TCAObserveUIKit.zip)

stephencelis:
@juliensagot Can you try pointing to the `shared-state-beta-fixes` branch I just pushed and see if it fixes things for you? Can you also try to make sure it doesn't introduce any new problems? :smile:

juliensagot:
@stephencelis Thanks for taking the time to investigate this! The `shared-state-beta-fixes` branch does seem to fix the issue on my side 👏 From what I can tell, it doesn't introduce new problems, but I'll keep my eyes wide open and report anything weird.

PR #2978, opened 2024-04-10
Use notification center instead of KVO for user defaults observation.

mbrandonw:
This was brought up by @jegnux in [Slack](https://pointfreecommunity.slack.com/archives/C04L56P2VEF/p1712753624317759?thread_ts=1712534913.585879&cid=C04L56P2VEF). Right now observation of user defaults is broken if someone uses a "." in their key, e.g. `@Shared(.appStorage("pointfree.isOn"))`. To fix we will use `NotificationCenter` to subscribe to the whole firehose of _any_ change to `UserDefaults`. 

Now technically this does mean we could be playing back changes to `@Shared` when the key does not actually change. We can't easily de-dupe the `didSet` from within `AppStorageKey` because we don't have access to the current value. So we could explore doing some de-duping in `ValueReference`, but then we'd need to cast to `any Equatable` and open the existential. That might be worth it though.

### DIFF
--- a/Tests/ComposableArchitectureTests/AppStorageTests.swift
+++ b/Tests/ComposableArchitectureTests/AppStorageTests.swift
@@ -94,6 +94,13 @@ final class AppStorageTests: XCTestCase {
     XCTAssertEqual(count, 42)
   }
 
+  func testChangeUserDefaultsDirectly_KeyWithPeriod() {
+    @Dependency(\.defaultAppStorage) var defaults
+    @Shared(.appStorage("pointfreeco.count")) var count = 0
+    defaults.setValue(count + 42, forKey: "pointfreeco.count")
+    XCTAssertEqual(count, 42)
+  }
+
   func testDeleteUserDefault() {
     @Dependency(\.defaultAppStorage) var defaults
     @Shared(.appStorage("count")) var count = 0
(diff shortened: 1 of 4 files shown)


Tests/ComposableArchitectureTests/AppStorageTests.swift
```diff
@@ -94,6 +94,13 @@ final class AppStorageTests: XCTestCase {
     XCTAssertEqual(count, 42)
   }
 
+  func testChangeUserDefaultsDirectly_KeyWithPeriod() {
+    @Dependency(\.defaultAppStorage) var defaults
+    @Shared(.appStorage("pointfreeco.count")) var count = 0
+    defaults.setValue(count + 42, forKey: "pointfreeco.count")
+    XCTAssertEqual(count, 42)
+  }
```
mbrandonw, line 102: this test was failing before the change.

stephencelis (approved):

mbrandonw:
> Now technically this does mean we could be playing back changes to @Shared when the key does not actually change.

Looks like `@AppStorage` behaves the same in SwiftUI, so we aren't going to try to do any de-duping.

stephencelis (approved):

PR #2979, opened 2024-04-10
Introduce @SharedReader

mbrandonw:
This introduces a read-only version of `@Shared`. It works much like `@Shared`, except you are not allowed to write to it. This can be useful for allowing features to hold onto shared state that it shouldn't be allowed to write to, as well as using it with external systems for which it does not make sense to "save" data to the system (e.g. remote config, feature flags, notifications, etc.).


Examples/CaseStudies/SwiftUICaseStudies/02-SharedState-Notifications.swift
```diff
@@ -0,0 +1,130 @@
...
+      name: MainActor.assumeIsolated {
+        UIApplication.userDidTakeScreenshotNotification
+      }
+    ) { value, _ in
+      value += 1
+    }
+  }
+}
+
+struct NotificationReaderKey<Value: Sendable>: PersistenceReaderKey {
```
mbrandonw, line 86: I think something like this could be moved into library code, but for now I've kept it in this case study.

Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
```diff
@@ -288,7 +288,8 @@ extension AppStorageKey: PersistenceKey {
   }
 
   public func subscribe(
-    initialValue: Value?, didSet: @escaping (_ newValue: Value?) -> Void
+    initialValue: Value?,
+    didSet: @Sendable @escaping (_ newValue: Value?) -> Void
```
mbrandonw, line 292: Making `didSet` sendable since it will typically be used in concurrent situations.

Sources/ComposableArchitecture/SharedState/References/ValueReference.swift
```diff
@@ -120,7 +177,13 @@ final class ValueReference<Value>: Reference, @unchecked Sendable {
         initialValue: initialValue
       ) { [weak self] value in
         guard let self else { return }
-        self.lock.withLock { self._value = value ?? initialValue }
+        #if canImport(Perception)
+          self._$perceptionRegistrar.willSet(self, keyPath: \.value)
+          defer { self._$perceptionRegistrar.didSet(self, keyPath: \.value) }
+        #endif
```
mbrandonw, line 183: this is a fix that is independent of this PR. I also added a failing test that this fixes.

stephencelis (approved):

PR #2981, opened 2024-04-11
Fix a few sendable warnings

mbrandonw:
The majority of sendability problems in TCA right now is in APIs that will be deleted in 2.0, but there are a few very simple warnings we can fix right now just to be a little proactive.

### DIFF
--- a/Package@swift-5.9.swift
+++ b/Package@swift-5.9.swift
@@ -84,14 +84,13 @@ let package = Package(
   ]
 )
 
-//for target in package.targets where target.type != .system {
-//  target.swiftSettings = target.swiftSettings ?? []
-//  target.swiftSettings?.append(
-//    .unsafeFlags([
-//      "-c", "release",
-//      "-emit-module-interface", "-enable-library-evolution",
-//      "-Xfrontend", "-warn-concurrency",
-//      "-Xfrontend", "-enable-actor-data-race-checks",
-//    ])
-//  )
-//}
+for target in package.targets where target.type != .system {
+  target.swiftSettings = target.swiftSettings ?? []
+  target.swiftSettings?.append(.enableExperimentalFeature("StrictConcurrency"))
+  //  target.swiftSettings?.append(
+  //    .unsafeFlags([
+  //      "-c", "release",
+  //      "-emit-module-interface", "-enable-library-evolution",
+  //    ])
+  //  )
+}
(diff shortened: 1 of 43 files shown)


Package@swift-5.9.swift
```diff
@@ -87,7 +96,6 @@ let package = Package(
 //    .unsafeFlags([
 //      "-c", "release",
 //      "-emit-module-interface", "-enable-library-evolution",
-//      "-Xfrontend", "-warn-concurrency",
 //      "-Xfrontend", "-enable-actor-data-race-checks",
```
stephencelis, line 99:
```suggestion
```

stephencelis (approved):

Sources/ComposableArchitecture/Effects/Cancellation.swift
```diff
@@ -215,8 +215,15 @@ extension Task where Success == Never, Failure == Never {
   }
 }
 
-@_spi(Internals) public var _cancellationCancellables = CancellablesCollection()
-private let _cancellablesLock = NSRecursiveLock()
+#if swift(>=5.10)
+  @_spi(Internals)
+  nonisolated(unsafe)
+  public var _cancellationCancellables = CancellablesCollection()
+  private let _cancellablesLock = NSRecursiveLock()
```
mbrandonw, line 222: since we do our own synchronization of these variables we can turn off isolation checking. of course that means the responsibility is on us to make sure everything is correct, but luckily all of this goes away somewhat soon!

Sources/ComposableArchitecture/Internal/MainActorAssumeIsolated.swift
```diff
@@ -0,0 +1,8 @@
+@MainActor(unsafe)
+func assumeMainActorIsolated<T>(
+  _ operation: @MainActor () throws -> T,
+  file: StaticString = #fileID,
+  line: UInt = #line
+) rethrows -> T {
+  try operation()
+}
```
mbrandonw, line 8: The `Actor.assumeIsolated` function is supposed to be back deployed to iOS 13, and that does seem to be the case locally, but CI says iOS 17+. If we can figure out CI we can get rid of this.

pyrtsa: It was back deployed only in the Xcode 15.3 toolchain with Swift 5.10 AFAICT, not in 5.9 yet.

mbrandonw: Ahhh ok, so I can work around that with some `#if`s. Thanks @pyrtsa!

stephencelis:
Going to reopen this in smaller steps.

PR #2983, opened 2024-04-11
Make Shared.Subscription.cancel public

hiltonc:
This makes `Shared.Subscription.cancel` public so that we can wrap a `PersistenceKey` and correctly implement `subscribe(initialValue:didSet:)`.

See https://github.com/pointfreeco/swift-composable-architecture/discussions/2857#discussioncomment-9087925


stephencelis (approved):
Thanks!

PR #2986, opened 2024-04-13
Fix dismissal in UIKit demos.

mbrandonw:
This was brought up by @acosmicflamingo in [Slack](https://pointfreecommunity.slack.com/archives/C04KQQ7NXHV/p1711312252812529). We should be checking that the state that drives navigation is `nil`/`false` before sending the `.dismiss` action so that we don't accidentally send it twice. This will cause a runtime warning if the child feature programmatically dismissed itself, causing the parent to then dismiss again.

### DIFF
--- a/Examples/TicTacToe/tic-tac-toe/Sources/TwoFactorCore/TwoFactorCore.swift
+++ b/Examples/TicTacToe/tic-tac-toe/Sources/TwoFactorCore/TwoFactorCore.swift
@@ -33,6 +33,7 @@ public struct TwoFactor: Sendable {
   }
 
   @Dependency(\.authenticationClient) var authenticationClient
+  @Dependency(\.dismiss) var dismiss
 
   public init() {}
 
@@ -57,6 +58,8 @@ public struct TwoFactor: Sendable {
         return .none
 
       case .view(.submitButtonTapped):
+        return .run { _ in await self.dismiss() }
+
         state.isTwoFactorRequestInFlight = true
         return .run { [code = state.code, token = state.token] send in
           await send(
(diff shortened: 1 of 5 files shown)


acosmicflamingo (approved):
Oh yeah I forgot about that! Lg2m 🥇

Examples/TicTacToe/tic-tac-toe/Sources/TwoFactorCore/TwoFactorCore.swift
```diff
@@ -57,6 +58,8 @@ public struct TwoFactor: Sendable {
         return .none
 
       case .view(.submitButtonTapped):
+        return .run { _ in await self.dismiss() }
```
stephencelis, line 61: Old code?

Examples/TicTacToe/tic-tac-toe/Sources/TwoFactorUIKit/TwoFactorViewController.swift
```diff
@@ -63,7 +63,7 @@ public final class TwoFactorViewController: UIViewController {
       guard let self else { return }
       activityIndicator.isHidden = store.isActivityIndicatorHidden
       codeTextField.text = store.code
-      loginButton.isEnabled = store.isLoginButtonEnabled
+      //loginButton.isEnabled = store.isLoginButtonEnabled
```
stephencelis, line 66: Comment back in?

stephencelis (commented):
Just a couple small things.

Examples/TicTacToe/tic-tac-toe/Sources/TwoFactorCore/TwoFactorCore.swift
mbrandonw: oops yes, just test code. reverted this and the stuff below.

PR #2989, opened 2024-04-14
Shared change tracking enhancements

stephencelis:
In #2976, shared state snapshots were moved off the `Shared` reference and into the change tracker, which allowed multiple change trackers to observe changes independently, but not concurrently, so when change trackers observed changes in a nested fashion, the outer trackers would fail to observe some changes. This was mainly a problem for testing reducers that also had `_printChanges()` attached to them.

This PR makes the following changes:

- Change tracking is now done by a set of trackers that are added/removed from a task local pool for the duration of their scoped tracking
- Change tracker assertions are synchronously scoped to a single tracker and the changes it observed

### DIFF
--- a/Examples/SyncUps/SyncUps/AppFeature.swift
+++ b/Examples/SyncUps/SyncUps/AppFeature.swift
@@ -71,18 +71,13 @@ struct AppView: View {
 }
 
 #Preview {
-  AppView(
-    store: Store(
-      initialState: AppFeature.State(
-        syncUpsList: SyncUpsList.State(
-          syncUps: [
-            .mock,
-            .productMock,
-            .engineeringMock,
-          ]
-        )
-      )
-    ) {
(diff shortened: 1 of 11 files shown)


Examples/SyncUps/SyncUps/AppFeature.swift
```diff
@@ -71,18 +71,13 @@ struct AppView: View {
...
-        syncUpsList: SyncUpsList.State(
-          syncUps: [
-            .mock,
-            .productMock,
-            .engineeringMock,
-          ]
-        )
-      )
-    ) {
+  @Shared(.syncUps) var syncUps: IdentifiedArrayOf<SyncUp> = [
```
stephencelis, line 74: I had some sync-ups cleanup on the shared branch that snuck into this. The PR review can stay focused on the `Sources` directory, I can also cherry-pick the sync-ups changes back into the beta branch if that makes it easier to review.

PR #2991, opened 2024-04-15
Add support for storing Date

rcasula:
This PR aims adds a convenience initialiser for AppStorageKey that stores a Date


mbrandonw:
Hi @rcasula, thanks for the PR, but for right now we are going to hold off on convenience APIs like this. Currently we are just mimicking the current API of `UserDefaults`, which does not allow directly storing a date. There are some techniques described in [your thread here](https://github.com/pointfreeco/swift-composable-architecture/discussions/2857#discussioncomment-9058683) for using a date in user defaults though, so that should help you.

PR #2992, opened 2024-04-15
Save to file storage when app is about to be terminated

mbrandonw:
This was brought up a few weeks ago and then again today in discussions and we just never got around to it.

### DIFF
--- a/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
+++ b/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
@@ -300,16 +300,23 @@ extension AppStorageKey: PersistenceKey {
       else { return }
       didSet(self.store.value(forKey: self.key) as? Value ?? initialValue)
     }
-    let willEnterForeground = NotificationCenter.default.addObserver(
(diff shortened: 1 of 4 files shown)


Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
```diff
@@ -300,16 +300,23 @@ extension AppStorageKey: PersistenceKey {
...
       didSet(self.store.value(forKey: self.key) as? Value ?? initialValue)
     }
-    let willEnterForeground = NotificationCenter.default.addObserver(
-      forName: willEnterForegroundNotificationName,
-      object: nil,
-      queue: nil
-    ) { _ in
-      didSet(self.store.value(forKey: self.key) as? Value ?? initialValue)
+    let willEnterForeground: (any NSObjectProtocol)?
+    if let willEnterForegroundNotificationName {
```
mbrandonw, line 304: Since these notification names are `nil` for platforms that don't support the notification we should unwrap them. Otherwise we subscribe to notifications with a `nil` name, which basically means subscribe to _all_ notifications.

stephencelis (approved):

PR #2995, opened 2024-04-17
Better shared state change tracking and `TestStore` interactions

stephencelis:
This tweaks shared change tracking a bit further:

1. Test stores now maintain their own, isolated change trackers instead of sharing the default test change tracker
2. Test stores now emit state-related failures on `deinit` when shared state changes before the end of the test but wasn't asserted on (_e.g._ in an effect)

### DIFF
--- a/Sources/ComposableArchitecture/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestStore.swift
@@ -422,7 +422,7 @@ import XCTestDynamicOverlay
 /// [merowing.info]: https://www.merowing.info
 /// [exhaustive-testing-in-tca]: https://www.merowing.info/exhaustive-testing-in-tca/
 /// [Composable-Architecture-at-Scale]: https://vimeo.com/751173570
-public final class TestStore<State, Action> {
+public final class TestStore<State: Equatable, Action> {
 
   /// The current dependencies of the test store.
   ///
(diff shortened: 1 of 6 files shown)


Sources/ComposableArchitecture/TestStore.swift
```diff
@@ -422,7 +422,7 @@ import XCTestDynamicOverlay
 /// [merowing.info]: https://www.merowing.info
 /// [exhaustive-testing-in-tca]: https://www.merowing.info/exhaustive-testing-in-tca/
 /// [Composable-Architecture-at-Scale]: https://vimeo.com/751173570
-public final class TestStore<State, Action> {
+public final class TestStore<State: Equatable, Action> {
```
stephencelis, line 425: The `Equatable` constraint was only absent here when we supported `TestStore.scope`, but now it's effectively required everywhere, so we can just force the conformance up front here, which improves compiler errors, but also allows us to invoke an assertion helper in the `deinit`.

Sources/ComposableArchitecture/TestStore.swift
```diff
@@ -488,13 +488,9 @@ public final class TestStore<State, Action> {
   public var timeout: UInt64
 
   private let file: StaticString
-  private var line: UInt
+  private let line: UInt
```
stephencelis, line 491: We previously advanced the test store's line whenever it was interacted with so that failures for unasserted things occur toward the end of the test. This is actually a little confusing, though, as failure messages can be tied to assertion lines that passed just fine. So instead we can maintain a constant `line` for where the store was initialized, and emit failures there.

mbrandonw (approved):

PR #2997, opened 2024-04-18
Remove `@unchecked` from `StackState`'s `Sendable` conformance

rhysm94:
I was digging around in the internals of the library, and I noticed this comment about removing this `@unchecked` when Swift Collections 1.1 is released.
I checked, and it has been released now. So I've updated TCA's dependency version, removing the `@unchecked` bit of `StackState`'s `Sendable` conformance.

This does have the potentially unfortunate side effect of needing to update the minimum Swift version to 5.7.2 from 5.7.1, as Swift Collections 1.1 bumps its minimum Swift version: https://github.com/apple/swift-collections/releases/tag/1.1.0

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.7.2
+// swift-tools-version:5.7.1
 
 import PackageDescription
 
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.7.1
+// swift-tools-version:5.7.2
 
 import PackageDescription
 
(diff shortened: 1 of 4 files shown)


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.7.1
+// swift-tools-version:5.7.2
```
stephencelis, line 1: While the release notes say it requires 5.7.2, their tools is still 5.7, so I think we can keep things as they are and let the compiler fail where it may fail.
```suggestion
// swift-tools-version:5.7.1
```

stephencelis (approved):
Thanks!

PR #2999, opened 2024-04-19
fix typo in examples: `let` -> `var`

andtie:
There is a small typo in the shared-state examples. This PR fixes that. A computed property property should use the `var`-keyword instead of `let`.

Keep up the good work!


mbrandonw (approved):
good catch thanks!

PR #3008, opened 2024-04-24
Updated swift-custom-dump to latest version

y-mimura:
The `swift-custom-dump@1.2.1` contains a fix to prevent crashes during release builds that occurred with Xcode 15.3.

https://github.com/pointfreeco/swift-custom-dump/releases/tag/1.2.1

I noticed that this fix was already included in the Shared State's PR, but I created this PR because I thought this fix might should be released earlier.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -68,8 +68,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/pointfreeco/swift-custom-dump",
       "state" : {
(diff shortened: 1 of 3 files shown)


stephencelis:
@y-mimura The shared state beta actually depends on new functionality, which is why it is pinned to a newer version. We're down to merge this, but also we plan on releasing shared state soon.

y-mimura:
@stephencelis Got it, Thank you for merging. I'm looking forward to meet the shared state.

PR #3011, opened 2024-04-24
Don’t assume CastableLookup

AlexKobachiJP:
If the observed value in `UserDefaults` is a `RawRepresentable` type, simply casting to `Value` might fail because the raw representing type could be different from `Value`.  Instead of querying `UserDefaults` directly here, we should call `loadValue()` which will do the right thing for raw representable types (esp. try to call `Value.init(rawValue:)` with the value obtained from `UserDefaults`).

Or am I missing something and there is a reason this is using `UserDefaults.value(forKey:)` directly?  I am especially curious because `CastableLookup` on line 374 is using `UserDefaults.object(forKey:)` instead, so it does look somewhat deliberate.

### DIFF
--- a/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
+++ b/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
@@ -298,7 +298,7 @@ extension AppStorageKey: PersistenceKey {
     ) { _ in
       guard !SharedAppStorageLocals.isSetting
       else { return }
-      didSet(self.store.value(forKey: self.key) as? Value ?? initialValue)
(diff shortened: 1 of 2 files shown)


stephencelis (approved):
Good catch, thanks! Are you interested in pushing a test that confirms the fix? If not we can merge and write a quick test ourselves.

AlexKobachiJP:
@stephencelis added the test.  This illustrates nicely how the type stored in user defaults is a `String` and the old code would have tried to directly cast (and fail) that string to the type of the initial value, that is, the `Direction` enum.

stephencelis:
Ah that test failure was fixed on another PR. Will merge this now, thanks again!

Issue #3021, opened 2024-04-29
Missing an expected key: 'NSPrivacyCollectedDataTypes'


OguzYuuksel:
### Description

Missing an expected key: 'NSPrivacyCollectedDataTypes'
xxx.app/swift-composable-architecture_ComposableArchitecture.bundle/PrivacyInfo.xcprivacy

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Actual behavior

Error

### Steps to reproduce

- Open your project in Xcode.
- Choose Product > Archive. Xcode creates the archive and reveals it in the organizer.
- Control-click the archive in the organizer and choose Generate Privacy Report.
- Choose a location to save the privacy report.
- Switch to Finder.
- Navigate to the location where you saved the privacy report, and double-click to open the report in Preview.

### The Composable Architecture version information

1.10.0

### Xcode version information

15.3.0

### Swift Compiler version information

```shell
swift-driver version: 1.90.11.1 Apple Swift version 5.10 (swiftlang-5.10.0.13 clang-1500.3.9.4)
Target: arm64-apple-macosx14.0
```

mbrandonw:
Hi @OguzYuuksel, thanks for the report. Can you clarify whether or not this is preventing you from submitting your app to the App Store? Or is it just a warning?

@DandyLyons You added this file in #2930. Is there something missing from the plist that we need to add?

If we can't get this sorted in the next day or so we will have to remove the privacy file so that it does not cause issues for others.

jshier:
Apple's latest update on the manifests seems to have removed the May 1st deadline, which hasn't passed yet. I'm not sure whether the lack of this key *would* be a blocking issue or whether it's just part of the PDF generation, but you can fix it by adding the proper empty root key.

```xml
<key>NSPrivacyCollectedDataTypes</key>
<array/>
```

DandyLyons:
Thanks. I'll work on it now.

OguzYuuksel:
Hello,
[...]

PR #3027, opened 2024-04-29
Add missing NSPrivacyCollectedDataTypes to PrivacyInfo.xcprivacy

DandyLyons:
Revision of: https://github.com/pointfreeco/swift-composable-architecture/pull/2930
Per: https://github.com/pointfreeco/swift-composable-architecture/issues/3021

The Xcode behavior appears to be different from when I submitted PR 2930. I originally created the manifest without NSPrivacyCollectedDataTypes (since TCA is not actually collecting anything) and Xcode generated an empty Privacy Report. 

Now it generates the error `Missing an expected key: 'NSPrivacyCollectedDataTypes'`

I've now added the key in however, it no longer generates an empty Privacy Report, so I think it needs to be a library decision how we proceed. Here's the new generated report. 

Screenshot 2024-04-29 at 1 10 41 PM

According to this we are collecting "other" data even though TCA isn't collecting anything and even though we already have an Apple-approved exemption (C56D.1). I believe that this is misleading to library users, and Apple may force app makers to include this in their final App Store nutrition label declaration, which would be further misleading to customers.

### DIFF
--- a/Sources/ComposableArchitecture/Resources/PrivacyInfo.xcprivacy
+++ b/Sources/ComposableArchitecture/Resources/PrivacyInfo.xcprivacy
@@ -3,20 +3,7 @@
 <plist version="1.0">
 <dict>
 	<key>NSPrivacyCollectedDataTypes</key>
-	<array>
-		<dict>
-			<key>NSPrivacyCollectedDataType</key>
-			<string>NSPrivacyCollectedDataTypeOtherDataTypes</string>
-			<key>NSPrivacyCollectedDataTypeLinked</key>
-			<false/>
-			<key>NSPrivacyCollectedDataTypeTracking</key>
-			<false/>
-			<key>NSPrivacyCollectedDataTypePurposes</key>
-			<array>
-				<string>NSPrivacyCollectedDataTypePurposeAppFunctionality</string>
-			</array>
-		</dict>
-	</array>
+	<array/>
 	<key>NSPrivacyAccessedAPITypes</key>
 	<array>
 		<dict>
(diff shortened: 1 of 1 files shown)


DandyLyons:
It's not an array. It's a string. 

Also the NSPrivacyCollectedDataType is not allowed to be an empty string. And it must be one of the approved strings. 
[TransformFeature-PrivacyReport 2024-04-29 12-06-51.pdf](https://github.com/pointfreeco/swift-composable-architecture/files/15153922/TransformFeature-PrivacyReport.2024-04-29.12-06-51.pdf)

Sources/ComposableArchitecture/Resources/PrivacyInfo.xcprivacy
```diff
@@ -2,6 +2,21 @@
...
+			<key>NSPrivacyCollectedDataTypeLinked</key>
+			<false/>
+			<key>NSPrivacyCollectedDataTypeTracking</key>
+			<false/>
+			<key>NSPrivacyCollectedDataTypePurposes</key>
+			<array>
+				<string>NSPrivacyCollectedDataTypePurposeAppFunctionality</string>
+			</array>
+		</dict>
+	</array>
```
stephencelis, line 19: @DandyLyons This is the array I was referring to:

```suggestion
	<array/>
```

Does it not work to pass in an empty array?

DandyLyons:
Looks good with the empty array now. It generates an empty Privacy Report again.

stephencelis (approved):
Thanks!

PR #3051, opened 2024-05-06
Precondition: persistence keys must match value

stephencelis:
We currently soft-fail if a persistence key is reused across different types. At runtime a persistence key will write over the other's value, and vice versa, and data is lost.

Reusing the same file storage URL or app storage key with conflicting types is a programmer error and ideally should crash over data corruption.


Sources/ComposableArchitecture/Internal/TypeName.swift
```diff
@@ -1,15 +0,0 @@
-@usableFromInline
-func typeName(_ type: Any.Type) -> String {
```
stephencelis, line 2: We have another helper that does a better job that I was utilizing, so deleted this one.

Sources/ComposableArchitecture/SharedState/PersistenceKey.swift
```diff
@@ -7,10 +7,14 @@
...
 /// See the article <doc:SharingState> for more information, in particular the
 /// <doc:SharingState#Custom-persistence> section.
-public protocol PersistenceReaderKey<Value>: Hashable {
+public protocol PersistenceReaderKey<Value> {
   /// A type that can be loaded or subscribed to in an external system.
   associatedtype Value
 
+  associatedtype ID: Hashable = Self
+
+  var id: ID { get }
```
stephencelis, line 16: - [ ] Docs

Sources/ComposableArchitecture/SharedState/PersistenceKey.swift
```diff
@@ -7,10 +7,18 @@
 ///
 /// See the article <doc:SharingState> for more information, in particular the
 /// <doc:SharingState#Custom-persistence> section.
-public protocol PersistenceReaderKey<Value>: Hashable {
```
stephencelis, line 10: Losing this can be a very slight breaking change to custom persistence strategies, so we could keep the requirement for now, it just isn't useful when using a custom ID type...

mbrandonw (approved):

PR #3058, opened 2024-05-08
Enable customDump only in debug mode

heoblitz:
Reduce unnecessary code execution by enabling customDump only in debug mode, as in .run Effect. 

https://github.com/pointfreeco/swift-composable-architecture/blob/main/Sources/ComposableArchitecture/Effect.swift#L101

### DIFF
--- a/Sources/ComposableArchitecture/Reducer/Reducers/IfCaseLetReducer.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/IfCaseLetReducer.swift
@@ -191,38 +191,32 @@ public struct _IfCaseLetReducer<Parent: Reducer, Child: Reducer>: Reducer {
     guard let childAction = self.toChildAction.extract(from: action)
     else { return .none }
     guard var childState = self.toChildState.extract(from: state) else {
-      #if DEBUG
-        var actionDump = ""
-        customDump(action, to: &actionDump, indent: 4)
-        var stateDump = ""
-        customDump(state, to: &stateDump, indent: 4)
-        runtimeWarn(
-          """
-          An "ifCaseLet" at "\(self.fileID):\(self.line)" received a child action when child state \
-          was set to a different case. …
+      runtimeWarn(
+        """
+        An "ifCaseLet" at "\(self.fileID):\(self.line)" received a child action when child state \
+        was set to a different case. …
 
-            Action:
-          \(actionDump)
-            State:
-          \(stateDump)
+          Action:
+        \(String(customDumping: action).indent(by: 4))
+          State:
+        \(String(customDumping: state).indent(by: 4))
 
-          This is generally considered an application logic error, and can happen for a few reasons:
+        This is generally considered an application logic error, and can happen for a few reasons:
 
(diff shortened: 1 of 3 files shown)


Sources/ComposableArchitecture/Reducer/Reducers/IfCaseLetReducer.swift
```diff
@@ -191,36 +191,38 @@ public struct _IfCaseLetReducer<Parent: Reducer, Child: Reducer>: Reducer {
...
+          • An in-flight effect emitted this action when child state was unavailable. While it may \
+          be perfectly reasonable to ignore this action, consider canceling the associated effect \
+          before child state changes to another case, especially if it is a long-living effect.
+
+          • This action was sent to the store while state was another case. Make sure that actions \
+          for this reducer can only be sent from a view store when state is set to the appropriate \
+          case. In SwiftUI applications, use "SwitchStore".
+          """
+        )
+      #endif
```
stephencelis, line 225: We can use some helpers to inline the work:

```suggestion
      runtimeWarn(
        """
        An "ifCaseLet" at "\(self.fileID):\(self.line)" received a child action when child state \
        was set to a different case. …

          Action:
        \(String(customDumping: action).indent(by: 4))
          State:
        \(String(customDumping: state).indent(by: 4))

        This is generally considered an application logic error, and can happen for a few reasons:

        • A parent reducer set "\(typeName(Parent.State.self))" to a different case before this \
        reducer ran. This reducer must run before any other reducer sets child state to a \
        different case. This ensures that child reducers can handle their actions while their \
        state is still available.

        • An in-flight effect emitted this action when child state was unavailable. While it may \
        be perfectly reasonable to ignore this action, consider canceling the associated effect \
        before child state changes to another case, especially if it is a long-living effect.

        • This action was sent to the store while state was another case. Make sure that actions \
        for this reducer can only be sent from a view store when state is set to the appropriate \
        case. In SwiftUI applications, use "SwitchStore".
        """
      )
```

stephencelis (commented):
Thanks for taking the time to PR! The `runtimeWarn` helper is already lazy, but will still emit failures in release tests. In order to retain that behavior, let's not `#if DEBUG` stuff entirely. Instead, can you move that work into the `runtimeWarn` message argument?

Sources/ComposableArchitecture/Reducer/Reducers/IfCaseLetReducer.swift
heoblitz: Thank you for checking! I've removed the EXP flag and modified it to pass as an autoclosure parameter directly.

heoblitz: BindableActionViewStoreDebugger implementation is within the EXP flag, and since maxDepth configuration is not possible, I've excluded it for now.

heoblitz: BindableActionViewStoreDebugger implementation is within the EXP flag, and since maxDepth configuration is not possible, I've excluded it for now.

stephencelis (approved):
Thanks!

PR #3060, opened 2024-05-09
Lazy loading of Shared initial values

seanmrich:
Extends PR #3057 to include initial values in addition to default values.

Persistent Shared values are often declared with an initial value.
```swift
@Shared(.appStorage("key")) var value = 0
```

This initial value is eagerly evaluated and passed to the Shared initializer. When there's an existing reference, this initial value is thrown away. For simple types like the one above, this isn't a problem. However, when the initial value is more complex, it can be wasteful if the Shared value is declared in many features of an app.

More significantly, the initial value may have side effects. For example, consider a `Folder` model that creates a `UUID` id using the `uuid` dependency. If you want to persist the `Folder`, you would declare a `PersistenceKey`.
```swift
extension PersistenceKey where Self == FileStorageKey<Folder> {
  public static var rootFolder: Self {
    fileStorage(.documentsDirectory.appendingPathComponent("rootFolder", conformingTo: .json))
  }
}
```

Your feature declares the shared folder in its state.
```swift
@ObservableState
struct State: Equatable {
  @Shared(.rootFolder) var root = Folder(…)
}
```

Now you want to override that root folder in a test.
```swift
func testAddFolder() async {
  let store = TestStore(
    reducer: { Feature() },
    withDependencies: {
      $0.uuid = .incrementing
      @Shared(.rootFolder) var root = Folder()
    }
  )
  await store.send(.addFolderButtonTapped) {
    $0.root.subfolders = [
      Folder(id: UUID(???))
    ]
  }
}
```

Because the `Feature.State` initializer evaluated the `Folder` declaration, the `uuid` was incremented. The value was thrown away because the Shared value in the dependencies block already created the reference, but the side effect still happened.

This PR makes that initial value load lazily so there are no side effects unless the value is actually needed. State initializers and local declarations are more efficient, and tests are deterministic.

### DIFF
--- a/Sources/ComposableArchitecture/SharedState/References/ValueReference.swift
+++ b/Sources/ComposableArchitecture/SharedState/References/ValueReference.swift
@@ -10,7 +10,7 @@ import Foundation
 
 extension Shared {
   public init(
-    wrappedValue value: Value,
+    wrappedValue value: @autoclosure @escaping () -> Value,
     _ persistenceKey: some PersistenceKey<Value>,
     fileID: StaticString = #fileID,
     line: UInt = #line
@@ -31,7 +31,7 @@ extension Shared {
             return reference
           } else {
             let reference = ValueReference(
-              initialValue: value,
+              initialValue: value(),
               persistenceKey: persistenceKey,
               fileID: fileID,
               line: line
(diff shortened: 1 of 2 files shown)


stephencelis (approved):
Thanks for taking this feature to the finish line!

PR #3061, opened 2024-05-09
Add array support for Shared appStorage

dadalar:
This PR adds support for the following Shared states:
```swift
// First use case: Store an array of strings
@Shared(.appStorage("stringArray")) var texts: [String]

// Second use case: Store an array of RawRepresentables
enum Direction: String {
  case north, south, east, west
}
@Shared(.appStorage("directions")) var directions: [Direction] = [.north]
```

### DIFF
--- a/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
+++ b/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
@@ -38,6 +38,15 @@ extension PersistenceReaderKey {
     AppStorageKey(key)
   }
 
+  /// Creates a persistence key that can read and write to a string array user default.
+  ///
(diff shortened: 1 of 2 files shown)


stephencelis:
Hi @dadalar! Thanks for taking the time to PR. For now we want to keep `appStorage` more or less limited to the APIs Apple provides for `@AppStorage` in SwiftUI. Apple seems to want to encourage sending only simple data to app storage, and they decided to omit the support for arrays, even though `UserDefaults` does have some array support. If this changes in the future we for sure would revisit.

For now if you need support for this we recommend defining it directly in your project.

We closed a similar PR extending `appStorage`'s APIs here: https://github.com/pointfreeco/swift-composable-architecture/pull/2905

I'm going to close this for now, as well, but feel free to open a discussion if you have more you'd like to chat about!

PR #3079, opened 2024-05-13
Some improvements to throttling

mbrandonw:
I've moved all of the state for throttling to a single struct that is isolated rather than a bunch of isolated fields. I also fixed an edge case bug where one could delete the persisted file while a throttling task is inflight. We now cancel that task.

### DIFF
--- a/ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -14,8 +14,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-argument-parser",
       "state" : {
-        "revision" : "0fbc8848e389af3bb55c182bc19ca9d5dc2f255b",
-        "version" : "1.4.0"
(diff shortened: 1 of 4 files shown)


stephencelis (approved):

ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -14,26 +14,26 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-argument-parser",
       "state" : {
-        "revision" : "0fbc8848e389af3bb55c182bc19ca9d5dc2f255b",
-        "version" : "1.4.0"
+        "revision" : "14c7622293ffbb7a235e896e37adf7eb089884a3",
+        "version" : "0.0.3"
```
pyrtsa, line 18: Are you sure about these package downgrades?

mbrandonw: Since it's just the local Package.resolved it isn't really that big of a deal, but I've gone ahead and updated them.

PR #3089, opened 2024-05-14
Add `ifLet` SwiftUI integration warning

stephencelis:
We recently added a warning when we detect that `forEach` wasn't integrated with a reducer when path elements are pushed or popped from the stack, and I realized that we could do similarly for `ifLet` during dismissal.

### DIFF
--- a/Sources/ComposableArchitecture/Observation/Store+Observation.swift
+++ b/Sources/ComposableArchitecture/Observation/Store+Observation.swift
@@ -152,10 +152,18 @@
     /// - Returns: A binding of an optional child store.
     public func scope<State: ObservableState, Action, ChildState, ChildAction>(
       state: KeyPath<State, ChildState?>,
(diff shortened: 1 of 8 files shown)


Sources/ComposableArchitecture/Observation/Store+Observation.swift
```diff
@@ -152,10 +152,18 @@
     /// - Returns: A binding of an optional child store.
     public func scope<State: ObservableState, Action, ChildState, ChildAction>(
       state: KeyPath<State, ChildState?>,
-      action: CaseKeyPath<Action, PresentationAction<ChildAction>>
+      action: CaseKeyPath<Action, PresentationAction<ChildAction>>,
+      fileID: StaticString = #fileID,
+      line: UInt = #line
```
mbrandonw, line 157: there's probably some DocC references to fix after this...

mbrandonw (approved):

PR #3097, opened 2024-05-16
Add @_SPI(Internal) to some more navigation apis

Alex293:
I'm working on back-porting iOS 16 NavigationStack and every TCA related apis. I managed to make a demo [here](https://github.com/Alex293/swift-composable-architecture/tree/observable-navigation-backport) but I needed to `@testable  import ComposableArchitecture` 

This branch contains the apis required but keep them behind `@_SPI(Internals)` as the rest of those apis.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/StackIDGenerator.swift
+++ b/Sources/ComposableArchitecture/Internal/StackIDGenerator.swift
@@ -11,7 +11,8 @@ extension DependencyValues {
   public let next: @Sendable () -> StackElementID
   public let peek: @Sendable () -> StackElementID
 
-  func callAsFunction() -> StackElementID {
+  @_spi(Internals)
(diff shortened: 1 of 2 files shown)


Alex293:
Related to #1674

stephencelis (approved):
@Alex293 We chatted about this today and think the impact is minimal enough to take. We do want to caveat that `@_spi(Internals)` APIs are something we may change/remove at any time, so be aware that your downstream project will need to be aware of changes.

Will merge when green!

Alex293:
Yeah I know but TCA + SwiftUI apis are too good to ignore and we are stuck on iOS 15 and might be for a while as it dropped devices.

Alex293:
Also I’d gladly switch to a TCA provided uinavigationcontroller solution if it could support both tree and stack navigation. I couldn’t make your branch work before iOS 16.

PR #3098, opened 2024-05-16
Fix exception when using AppStorageKey with URL Value

pwszebor:
When using AppStorageKey persistence strategy with the Value associated type `URL` or `URL?`, saving values to UserDefaults causes a NSInvalidArgumentException exception to be thrown. To avoid it, dedicated APIs for saving and retrieving URLs to/from UserDefaults need to be used.

### DIFF
--- a/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
+++ b/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
@@ -193,7 +193,7 @@ public struct AppStorageKey<Value> {
 
   public init(_ key: String) where Value == URL {
     @Dependency(\.defaultAppStorage) var store
-    self.lookup = CastableLookup()
+    self.lookup = URLLookup()
     self.key = key
     self.store = store
   }
@@ -251,7 +251,7 @@ public struct AppStorageKey<Value> {
 
   public init(_ key: String) where Value == URL? {
     @Dependency(\.defaultAppStorage) var store
-    self.lookup = OptionalLookup(base: CastableLookup())
+    self.lookup = OptionalLookup(base: URLLookup())
     self.key = key
     self.store = store
   }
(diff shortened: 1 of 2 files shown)


Tests/ComposableArchitectureTests/AppStorageTests.swift
```diff
@@ -24,6 +24,33 @@ final class AppStorageTests: XCTestCase {
     XCTAssertEqual(defaults.integer(forKey: "count"), 43)
   }
 
+  func testDefaultsReadURL() {
+    @Dependency(\.defaultAppStorage) var defaults
+    defaults.set(URL(string: "https://pointfree.co"), forKey: "url")
+    @Shared(.appStorage("url")) var url: URL?
+    XCTAssertEqual(url, URL(string: "https://pointfree.co"))
```
pwszebor, line 31: without the changes, this test would fail because UserDefaults.objectForKey would return Data which is not castable to URL

Tests/ComposableArchitectureTests/AppStorageTests.swift
```diff
@@ -24,6 +24,33 @@ final class AppStorageTests: XCTestCase {
...
+  }
+
+  func testDefaultsRegistered_URL() {
+    @Dependency(\.defaultAppStorage) var defaults
+    @Shared(.appStorage("url")) var url: URL = URL(string: "https://pointfree.co")!
```
pwszebor, line 36: without the changes, this line crashes with NSInvalidArgumentException

Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
```diff
@@ -364,13 +364,36 @@ private protocol Lookup<Value> {
 }
 
 private struct CastableLookup<Value>: Lookup {
+  private let getValue: (_ key: String, _ store: UserDefaults) -> Value?
+  private let setValue: (Value?, _ key: String, _ store: UserDefaults) -> Void
```
stephencelis, line 368: Instead of introducing these escaping closures, is it possible to add a new `Lookup` conformance just for `URL`?

stephencelis (commented):
Thanks for submitting a fix! Just one question about the implementation.

Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
pwszebor: sure, see updated code

PR #3155, opened 2024-06-07
Add conditional Hashable conformance to Shared

Jackstone92:
Here's a small PR that adds conditional `Hashable` conformance to `Shared` in a similar fashion to `BindingState`.

### DIFF
--- a/Sources/ComposableArchitecture/SharedState/Shared.swift
+++ b/Sources/ComposableArchitecture/SharedState/Shared.swift
@@ -298,6 +298,12 @@ extension Shared: Equatable where Value: Equatable {
   }
 }
 
+extension Shared: Hashable where Value: Hashable {
+  public func hash(into hasher: inout Hasher) {
+    self.wrappedValue.hash(into: &hasher)
+  }
+}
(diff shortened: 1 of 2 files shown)


stephencelis:
@Jackstone92 Thanks for taking the time to PR!

Unfortunately the `@Shared` type cannot be conditionally hashable because it's reference-y, and that means its hash could change over time. We document this as a "gotcha" here: https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/sharingstate/#Hashability

One example problem is that if you used a `Shared` value as a key in a dictionary, and its hash changed after insertion, the dictionary would be in an invalid state, which I believe would lead to a precondition/crash.

Generally reference-y things are hashable by their object identifier, but we can't do that here, either, because the equatable conformance would then be wrong. So we're kinda stuck here.

Please let us know the scenario you ran into this problem, though! Maybe there are other solutions to it.

Jackstone92:
Ah yeah that is a "gotcha"! Thanks, will close this and keep digging!

PR #3157, opened 2024-06-07
Note Swift bug in documentation

stephencelis:
It'd be good to have a place to point folks to when encountering the issue.


stephencelis:
When https://github.com/apple/swift/issues/74203 is fixed we can get rid of this note.

mbrandonw (approved):

PR #3161, opened 2024-06-13
Fixed extension typos in the Building SyncUps tutorial

meltsplit:
This PR fixes extension typos in 01-WhatIsSyncUps.


stephencelis (approved):
DocC seems to be smart enough to figure things out, but we can definitely be consistent here, thanks! :smile:

PR #3170, opened 2024-06-15
Disfavor shared optional dynamic member lookup

stephencelis:
This makes the default work as expected, avoiding the issue brought up in #3169.

Technically this breaks the following invocation:

```swift
if let wrapped = $shared.optional { /* ... */ }
```

And limits the warning to:

```swift
func share<T>(_: Shared<T>?) {}
share($shared.optional)
```

We consider this lookup to be a bug, though, which is already deprecated in 1.11.0.

Since we encourage folks to migrate by one minor version at a time, we should only merge this when we plan on releasing 1.12.0.


mbrandonw (approved):

acosmicflamingo:
What would be the proper way to do something like this in newer versions of TCA? I'm seeing the `subscript(dynamicMember:)' is deprecated` warning: 
```swift
 guard let row = state.$presentedData[id: itemId] else { return .none }
```

stephencelis:
@acosmicflamingo The warning should give a hint. Is it not helpful?

You can use `Shared.init` to explicitly unwrap an optional shared, just like `Binding.init`.

PR #3174, opened 2024-06-17
Share state appstorage nullable

fdzsergio:
This PR allows deleting elements from an `AppStorageKey` when it loads with a **non-null** value and we attempt to delete that value in persistence.


mbrandonw (approved):
Hi @fdzsergio, thanks for the report and the PR!

The test definitely shows a problem to be fixed, but I think the change you made isn't quite right. I have just pushed another fix that seems to do the job. Will merge once it goes green.

PR #3180, opened 2024-06-18
Leverage UIKitNavigation

stephencelis:
This draft PR demonstrates how to use the new [UIKitNavigation beta](https://github.com/pointfreeco/swiftui-navigation/discussions/168) tools in a Composable Architecture application.

We can merge when the tools are finalized and released.


acosmicflamingo:
Have you encountered any scoping issues when using `UIBinding`? The weirdest thing, I have two identical scopes but one works and the other gives me an error like below:
```
let binding = $store.scope(  // 1
  state: \.self,  // 2
  action: \.self // 3
)

// 1 error
Cannot call value of non-function type '_StoreUIBinding<AppMain.State, AppMain.Action, _>'
Value of type 'UIBinding<StoreOf<AppMain>>' (aka 'UIBinding<Store<AppMain.State, AppMain.Action>>') has
no dynamic member 'scope' using key path from root type 'AppMain.State'

// 2 error
Cannot infer key path type from context; consider explicitly specifying a root type

// 3 error
Cannot infer key path type from context; consider explicitly specifying a root type
```

PR #3190, opened 2024-06-19
Add ConfirmationDialogState support to UIBindable via UIAlertController(store:) initializer

acosmicflamingo:


acosmicflamingo:
RIP Point-Free repo subscriber inboxes

acosmicflamingo:
Converted to draft since I'm debugging an issue where `state.destination` seems to be nil-ed out earlier than expected, resulting in a purple runtime warning.

stephencelis:
@acosmicflamingo Just curious if you have any update here? Still troubleshooting anything or is this ready for review?

acosmicflamingo:
@stephencelis uh oh, looks like we both thought the other was working on fixing this LOL in Slack, I had said "So, if the logic [here](https://github.com/pointfreeco/swiftui-navigation/blob/uikit-navigation-beta/Sources/UIKitNavigation/Navigation/Presentation.swift) is going to handle presenting and dismissing modal view controllers, then AlertState and ConfirmationDialogState no longer have to conform to _EphemeralState , which explains the purple runtime warnings I was seeing in this reproducer", and then you said "We still need those conformances for SwiftUI, but we should avoid emitting those warnings on UIKit bindings. Thanks for the repro". I haven't touched it since then.

acosmicflamingo:
So in other words, I stopped because this seems like the kind of thing that requires refactoring how some models will conform to `_EphemeralState`, and thought it'd be best to leave it to you to handle this, but at least use a reproducer I provided to validate your solution.

PR #3192, opened 2024-06-20
Add Shared AppStorage support for [Bool], [Int], [Double] and [String]

johankool:
My project needed to access an array of integers from user defaults.

### DIFF
--- a/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
+++ b/Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
@@ -76,6 +76,42 @@ extension PersistenceReaderKey {
     AppStorageKey(key)
   }
 
+  /// Creates a persistence key that can read and write to a string array user default.
+  ///
+  /// - Parameter key: The key to read and write the value to in the user defaults store.
+  /// - Returns: A user defaults persistence key.
+  public static func appStorage(_ key: String) -> Self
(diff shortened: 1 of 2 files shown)


mbrandonw:
Hi @johankool, we have made the decision to not support things like this in `.appStorage` because neither SwiftUI's `@AppStorage` nor `UserDefaults` support it. Based on Apple's recommendation, one should only put small bits of simple data in user defaults, and so it seems they discourage putting arrays in a store. You still can do it, but you will need to encode it to data and then decode when reading. Or you could consider using `.fileStorage` instead.

I am going to close this PR for now, but do feel free to open a discussion if you have questions.

johankool:
Hm, that's disappointing. I get your point about `@AppStorage`  but `UserDefaults` quite specifically supports arrays and arrays of strings in particular. That the data should be simple/smallish is on the consumer of the API. Storing a few indices this way doesn't seem unreasonable to me. 

Unfortunately it seems not enough is exposed of the `AppStorageKey` and the `Lookup`s to be able to add this extension in my own code.

mbrandonw:
You are right, user defaults does support arrays. However `@AppStorage` does not. I'm not sure why, perhaps just due to the overload space of trying to provide a type-safe interface to user defaults. Maybe that leads to compiler issues.

We could discuss making more of the internals public, but I'm still not sure why it isn't better to use `.fileStorage`?

PR #3198, opened 2024-06-21
Convert all ActorIsolated to LockIsolated.

mbrandonw:


Alex293:
Can you briefly explain the reason behind this ?

mbrandonw:
> Can you briefly explain the reason behind this ?
> 
> will actorisolated be deprecated ?

Yeah, we will be deprecating `ActorIsolated` in swift-concurrency-extras. It really just has no advanced over `LockIsolated` (as far as we can tell), yet it's less ergonomic to use.

Do you have an example of where it is better to use `ActorIsolated` over `LockIsolated`?

stephencelis:
@Alex293 For reference, here's the PR deprecating: https://github.com/pointfreeco/swift-concurrency-extras/pull/29

stephencelis (approved):

Alex293:
I don’t have any specific use case, I was just wondering why, thanks !

PR #3201, opened 2024-06-24
Fix Shared publisher docs

larryonoff:


mbrandonw (commented):
Good catch, thanks @larryonoff!

PR #3207, opened 2024-06-25
Update Package.swift for SwiftSyntax 601-prerelease

mbrandonw:


stephencelis:
Oops, we lost track of this. I also took care of this in #3234, so going to close this in favor of that.

PR #3220, opened 2024-07-02
Improve UIKit case studies: ListsOfState

takehilo:
I have improved the ListsOfState example.
The `tableView(_:cellForRowAt:)` delegate method can be called multiple times for the same indexPath, and in the current code, the `observe(_:)` method is executed multiple times for the same indexPath.
When the `observe(_:)` method is executed multiple times, the closure is called unnecessarily many times when the Counter model at the corresponding index is changed.
To prevent this, I have added the observation cancellation logic.


stephencelis (approved):
Thanks!

PR #3221, opened 2024-07-02
Add `final` to `UIViewController`s in UIKit CaseStudies

junbangg:
While checking out the UIKit CaseStudies, I noticed that a couple of ViewControllers were missing the `final` keyword.
I didn't see any subclassing, and all the other ViewControllers seem to have `final`, so I assumed it was an oversight. Let me know if I'm missing something 👍


mbrandonw (approved):
Thanks!

PR #3225, opened 2024-07-07
Add support for custom decoding/encoding to `fileStorage`

oskarek:
This PR generalizes the `fileStorage` persistence key so that it can be used with values that are not `Codable`, by adding the ability to provide custom decode/encode closures as parameters. Also, for `Codable` values, the json decoder/encoder can now be customized.


Alex293:
I had a recent need to do some cache validation before retrieving the value (drop the file if it was created before some date). I wonder if instead of just focusing on codable this should go as far for loading/writing from/to URL

stephencelis:
@oskarek Thanks for taking the time to PR! We do think that supporting more than JSON is a good idea but haven't had time to review yet and think about the API. Hopefully soon!

stephencelis (approved):
Just discussed and this looks good to us, thanks @oskarek!

PR #3230, opened 2024-07-11
Drop Swift <5.9 support

stephencelis:
Many of TCA's dependencies have dropped support, so this brings it in line with the rest.


stephencelis:
I was in a confusing state. This was already done in #3203 but something was preventing my local copy to see!

PR #3233, opened 2024-07-11
Improve store scoping in benchmark

woxtu:
This PR resolves the following warning message:

```
'scope(state:action:)' is deprecated: Pass 'state' a key path to child state and 'action' a case key path to child action, instead. For more information see the following migration guide: https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/migratingto1.5#Store-scoping-with-key-paths
```


stephencelis:
@woxtu Thanks for looking into this! While we do want to revisit updating our benchmarks in the future, for now these benchmarks explicitly intend to test the old, uncached stores, so those deprecations are OK for now. We appreciate you taking the time to try to fix, though!

PR #3238, opened 2024-07-14
Hide UIViewController method/properties in UIKit Example code

junbangg:
I noticed that a few ViewControllers in the UIKit example code had some methods and properties that could be hidden. 
I also thought the stores would be better off hidden as well. (Please give me your thoughts if you think they should be left alone. )
Cheers🙂


mbrandonw (approved):
Hi @junbangg, thanks for contributing this. We are happy to take this change, but overall we are not too worried about these kinds of details for the case studies.

junbangg:
Got it, I'll keep that in mind👌Thanks

PR #3250, opened 2024-07-21
Fix a typo in Tutorial

yaslab:
This PR fixes a typo in Tutorial. According to the documentation for [String.init(decoding:as:)](https://developer.apple.com/documentation/swift/string/init(decoding:as:)), this initializer never returns `nil`.


mbrandonw (approved):
Good catch @yaslab, thanks!

PR #3258, opened 2024-07-24
Shared: Update `XCTFail` to `reportIssue`

stephencelis:
Another fix for a symbol mentioned in https://github.com/pointfreeco/swift-issue-reporting/issues/100

### DIFF
--- a/ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/pointfreeco/combine-schedulers",
       "state" : {
-        "revision" : "487a4d151e795a5e076a7e7aedcd13c2ebff6c31",
-        "version" : "1.0.1"
+        "revision" : "9fa31f4403da54855f1e2aeaeff478f4f0e40b13",
+        "version" : "1.0.2"
       }
     },
     {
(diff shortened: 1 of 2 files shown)


mbrandonw (approved):

ryanmeisters:
is there a planned release for this one? 🙏🏼 😇

mbrandonw:
Hi @ryanmeisters, this was already released as 1.12, but be sure to target 1.12.1 because we had a small fix right after the 1.12 release.

ryanmeisters:
Oh I think Stephen added this after the 1.12.1 release in response to [this comment](https://github.com/pointfreeco/swift-issue-reporting/issues/100#issuecomment-2249001057), right?

mbrandonw:
Oh sorry, no this specific commit has not yet been released yet. We can cut a 1.12.2 soon.

stephencelis:
@ryanmeisters This PR should be mostly cosmetic, as issue reporting's `XCTFail` unconditionally reports issues to both XCTest _and_ Testing. Are you experiencing a particular issue?

ryanmeisters:
Ah yeah I think this was the fix four our issue [here](https://github.com/pointfreeco/swift-issue-reporting/issues/100#issuecomment-2249005134).

Honestly not a huge deal if it's high effort to make a patch release. To solve it for us I temporarily pinned all the tca dependencies to the XCTestDyanmicOverlay versions. I was just trying to keep track of when I could undo that and get back to current versions.

stephencelis:
Ah, sorry, I forgot about the symbol-related issue. We're planning a minor release with the new UIKit navigation tools this week, so when that's out it will get this patch. Thanks for your patience!

PR #3267, opened 2024-08-02
fix some fypo of syncup tutorial

hmhv:


mbrandonw (approved):
Good catches, thanks @hmhv!

PR #3277, opened 2024-08-10
MainActor Store Isolation

stephencelis:
Stores have been required to be on the main thread for a long time, but because they predate Swift concurrency, we adopted a model that was standard at the time: pushing that responsibility on the user, and warning them when we detect access off the main thread.

This PR adopts a more modern sensibility by adding `@preconcurrency @MainActor` to APIs that we require to be on the main thread. We'll use `@preconcurrency` to avoid breaking changes, but the main advantages:

- It will prevent stores from being created or interacted with off the main thread, and it will warn the user at compile time if they attempt to do so without `await`ing.
- Tests no longer need to be marked `@MainActor` for the `TestStore`.

This PR should not prevent us from providing stores the ability to work on non-main actors, and should actually help us get closer to that goal.


Sources/ComposableArchitecture/Internal/DispatchQueue.swift
```diff
@@ -12,6 +12,18 @@ func mainActorASAP(execute block: @escaping @MainActor @Sendable () -> Void) {
   }
 }
 
+func mainActorNow(execute block: @escaping @MainActor @Sendable () -> Void) {
```
stephencelis, line 15: I added this helper to call from `TestStore.deinit`, since `deinit`s are currently non-isolated and so we can't otherwise invoke some of the teardown logic the test store performs otherwise.

This seems to work fine in practice, but if we're worried about deadlocks we could consider removing this helper and record an issue if `deinit` is ever called off the main thread instead.

Sources/ComposableArchitecture/RootStore.swift
```diff
@@ -64,15 +62,10 @@ public final class RootStore {
           let uuid = UUID()
           let effectCancellable = withEscapedDependencies { continuation in
             publisher
-              .handleEvents(
-                receiveCancel: { [weak self] in
-                  threadCheck(status: .effectCompletion(action))
-                  self?.effectCancellables[uuid] = nil
-                }
-              )
+              .receive(on: UIScheduler.shared)
```
stephencelis, line 65: While the compiler ensures access on the main thread when using modern Swift concurrency, older APIs like Combine do allow for escape hatches. I've added this `receive(on:)` to this internal Combine code to ensure the store code is really performed on the main actor, and as we de-combine-ify the library in the future we won't have to worry about edge cases like these as much.

stephencelis:
Some 5.10 warnings we should consider before merging, but generally things are looking good in 6.0!

Examples/TicTacToe/tic-tac-toe/Sources/GameSwiftUI/GameView.swift
```diff
@@ -2,6 +2,7 @@ import ComposableArchitecture
 import GameCore
 import SwiftUI
 
+@MainActor
```
mbrandonw, line 5: is this needed or just left over test code?

stephencelis: It's needed in 5.10, or we need to annotate the helper methods.

mbrandonw (approved):

Issue #3247, opened 2024-07-17
Crash when writing to user defaults in background thread


lukeredpath:
### Description

After updating a feature to use `@Shared` with the `.appStorage` persistence key, our app has started crashing when entering the background.

### Checklist

- [X] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Expected behavior

View updates are triggered on the main actor and the app does not crash.

### Actual behavior

The app crashes when user defaults are updated from a background thread.

### Steps to reproduce

I have traced this back to a custom view property wrapper that implements `DynamicProperty` - its `update()` method is marked as `nonisolated` to fix a concurrency warning but uses `MainActor.assumeIsolated` in this function because this should only be called on the main actor as part of a view update, however because the view is updating on a background thread it crashes.

The background view update is being triggered by a third party SDK that writes to the user defaults when the app enters the background. This in turns triggers the `NotificationCenter` observer in the app storage property key on the same background thread, which calls `didSet`.

<img width="661" alt="Screenshot 2024-07-18 at 00 05 34" src="https://github.com/user-attachments/assets/931ff97d-4f8c-4b0b-9406-b1a8ab6701c3">

I updated TCA to point to main as I thought #3206 would fix this issue however the final fix in that PR did not make the entire `willSet` isolated to the main actor, only the part the updates the current value but as you can see from the backtrace above, the view update is triggered before that (or possibly, IIRC the `update()` function on a `DynamicProperty` is called _before_ the view is about to be recomputed which might be why its crashing when calling `willSet` on the registrar).

I think the entire `didSet` closure needs to be called on the main actor, so `mainActorASAP` should wrap the entire closure contents.

Alternatively (or additionally), would it make more sense to ensure that all notifications are received on the main queue in the first place by passing `.main` to the queue when calling `NotificationCenter.default.addObserver`?

### The Composable Architecture version information

54eb417336fa849270671b8d2b73180835f9975a

### Destination operating system

iOS 17

### Xcode version information

Xcode 15.3

### Swift Compiler version information

```shell
[...]

PR #3285, opened 2024-08-14
Fix: Crash when writing to user defaults in background thread

larryonoff:
Potential fix for the issue https://github.com/pointfreeco/swift-composable-architecture/issues/3247

### DIFF
--- a/Tests/ComposableArchitectureTests/AppStorageTests.swift
+++ b/Tests/ComposableArchitectureTests/AppStorageTests.swift
@@ -261,7 +261,7 @@ final class AppStorageTests: XCTestCase {
   func testWillEnterForegroundFromBackgroundThread() async throws {
     @Shared(.appStorage("count")) var count = 0
 
-    let publisherExpectation: XCTestExpectation = expectation(description: "publisher")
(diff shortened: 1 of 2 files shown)


stephencelis:
@larryonoff Can you add a test that would have failed/crashed before the fix?

larryonoff:
> @larryonoff Can you add a test that would have failed/crashed before the fix?

@stephencelis Updated the tests to include perception. Previously, the tests only covered updates from publisher (that handled in the main thread), leaving out perception updates.

Tests/ComposableArchitectureTests/AppStorageTests.swift
```diff
@@ -253,13 +261,21 @@ final class AppStorageTests: XCTestCase {
   func testWillEnterForegroundFromBackgroundThread() async throws {
     @Shared(.appStorage("count")) var count = 0
 
-    let publisherExpectation = expectation(description: "publisher")
+    let publisherExpectation: XCTestExpectation = expectation(description: "publisher")
```
mbrandonw, line 264: can this be deleted?

```suggestion
    let publisherExpectation = expectation(description: "publisher")
```

PR #3289, opened 2024-08-18
Add ability to send actions / return effects using `UITransaction`

juliensagot:
Adds methods to send actions and return effects using `UITransaction`:

```swift
return .run { send in
   await send(.buttonTapped, uiKitAnimation: .easeInOut)
}
```

```swift
return .run { send in
   await send(.buttonTapped)
}
.uiKitAnimation(.easeInOut)
```

```swift
return .run { send in
  await send(.buttonTapped, transaction: UITransaction(animation: .easeInOut))
}
```

```swift
var transaction = UITransaction()
transaction.uiKit.disablesAnimations = true

return .run { send in
  await send(.buttonTapped)
}
.transaction(transaction)
```

```swift
store.send(.buttonTapped, uiKitAnimation: .easeInOut)
store.send(.buttonTapped, transaction: UITransaction())

// ViewActionSending
send(.buttonTapped, uiKitAnimation: .easeInOut)
send(.buttonTapped, transaction: UITransaction())
```

### DIFF
--- a/Sources/ComposableArchitecture/UIKit/Animation+UIKit.swift
+++ b/Sources/ComposableArchitecture/UIKit/Animation+UIKit.swift
@@ -1,0 +1,107 @@
+#if canImport(UIKit) && !os(watchOS)
+  import Combine
+  import UIKitNavigation
+  
+  extension Effect {
+    /// Wraps the emission of each element with `withUIKitAnimation`.
+    ///
+    /// ```swift
+    /// case .buttonTapped:
+    ///   return .run { send in
+    ///     await send(.activityResponse(self.apiClient.fetchActivity()))
+    ///   }
+    ///   .uiKitAnimation()
+    /// ```
+    ///
+    /// - Parameter animation: An animation.
+    /// - Returns: A publisher.
+    public func uiKitAnimation(_ animation: UIKitAnimation? = .default) -> Self {
+      self.transaction(UITransaction(animation: animation))
+    }
+    
+    /// Wraps the emission of each element with `withUITransaction`.
+    ///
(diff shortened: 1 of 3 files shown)


stephencelis (commented):
@juliensagot Thanks for taking the time to PR!

For some context on the original APIs, we added them because TCA was even more heavily SwiftUI-focussed at the time, and they were convenient helpers to have. Over time, though, we do kind of regret adding the surface area when the core problem it's solving is avoiding a single nested function call:

```diff
-withAnimation { _ = store.send(.action) }
+store.send(.action, animation: .default)
```

Another consideration is that these APIs predate a lot of modern Swift concurrency, and so they are incorrect: SwiftUI's `Transaction` type is not `Sendable` nor do we know if it's safe to send across boundaries, so we're not sure that API is even legitimate anymore.

One thing to worry about is the number of overloads and API surface area and how it affects the compiler. And also there are even more platforms/paradigms out there, which could potentially explode things even more in the future should we add support for AppKit, WindowsSDK, Wasm, and so on and so on and so on... 

So I think we'd actually consider removing these APIs in the next major version of TCA instead of allowing them to grow, and maybe even hard-deprecate them sooner, but it hasn't been an area we've been focussed on.

That said we do think the APIs are convenient enough for a community package to fill in the gaps. Something like a ComposableArchitectureSwiftUIExtras or ComposableArchitectureUIKitExtras. What do you think?

juliensagot:
@stephencelis That definitely makes sense! I hadn’t thought about creating an external package, but I think it’s a good idea. The more platform-agnostic TCA is, the better.

PR #3296, opened 2024-08-22
Add missing parameter in NavigationLink docs

woohyunjin06:
I found that the `onNavigate` closure isn’t being passed a parameter and is only being called when `isActive` is true.
I believe the `onNavigate` closure should also be called when isActive becomes false 🤔


stephencelis (approved):
Thanks!

PR #3299, opened 2024-08-23
Improve annotations in Effect.swift

MaraMincho:


stephencelis (approved):
Ah good catch, thanks!

PR #3303, opened 2024-08-25
Adds inspector view extension for usage with iOS 17 and macOS 14 inspector API

mmllr:


mbrandonw:
Hi @mmllr, thanks for the PR! This does seem like a good addition, however we think it should go in our [SwiftUINavigation library](http://github.com/pointfreeco/swift-navigation) instead. If we build this tool for vanilla SwiftUI, then TCA will get to leverage it immediately for free using the `$store.scope` syntax.

Would you be interested in creating a PR against that repo instead? If not, then we will take a look sometime this week.

mmllr:
Hi @mbrandonw, great - should have done that right away. [PR for SwiftUINavigation library](https://github.com/pointfreeco/swift-navigation/pull/215).

PR #3313, opened 2024-08-27
Fix SharingState doc

takehilo:
Fix code in SharingState.md.

### DIFF
--- a/Sources/ComposableArchitecture/Documentation.docc/Articles/SharingState.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Articles/SharingState.md
@@ -935,7 +935,7 @@ extension PersistenceReaderKey
 where Self == PersistenceKeyDefault<FileStorageKey<IdentifiedArrayOf<User>>>
 {
   static var users: Self {
-    PersistenceKeyDefault(.fileStorage(.users), [])
+    PersistenceKeyDefault(.users, [])
   }
 }
 ```


Sources/ComposableArchitecture/Documentation.docc/Articles/SharingState.md
```diff
@@ -935,7 +935,7 @@ extension PersistenceReaderKey
 where Self == PersistenceKeyDefault<FileStorageKey<IdentifiedArrayOf<User>>>
 {
   static var users: Self {
-    PersistenceKeyDefault(.fileStorage(.users), [])
```
takehilo, line 938: `fileStorage` is not needed here.

mbrandonw: Hi @takehilo, are you sure this compiles? It seems like there's a chance that `.users` would be ambiguous here since it can be either the one defined on `PersistenceKeyDefault` or `FileStorageKey`. I didn't try it myself, but seems like it could be a problem.

But even besides that, I don't think we need to make this change because the way it is currently is the most explicit. In the paragraphs preceding this we demonstrate a few examples of `static var users`, including one defined on `AppStorageKey`. And so to prevent any confusion we think it's best to just be super explicit and use `.fileStorage(.users)`.

takehilo: I thought I got a compilation error when I tried it before, but it seems I was mistaken. Both approaches can be compiled. However, as you mentioned, I think it should be explicit here.

I'll close this pr.

PR #3315, opened 2024-08-27
Avoid unsafe extraction in `BindingReducer`

stephencelis:
We currently have a single case of `AnyCasePath(unsafe:)` in the code base, which uses reflection for extraction, and is something we would like to remove in a future major version of Case Paths.

This explores a refactoring of `BindingReducer` that takes an explicit case key path toe the binding action case, instead:

```swift
BindingReducer(action: \.binding)
```

`BindableAction` as a protocol is still necessary for bindable stores to know how to embed binding actions in a short syntax, _e.g._ `$store.text`.

It's a slight bummer that the reducer gets slightly more verbose, especially when there _is_ something abstractable here, I'm just not sure Swift provides the tools yet to truly abstract over a case path in an ergonomic fashion.

The alternative would be to introduce a special protocol (`BindableCasePath`?) that the `@Reducer` macro would need to apply correctly when detecting a `BindableAction`, but this is one-off logic that seems to take things in the wrong direction.

This is a draft for now that does not clean up all the existing occurrences, as I'm not sure we want to complete and merge it.

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Forms.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-Bindings-Forms.swift
@@ -28,7 +28,7 @@ struct BindingForm {
   }
 
   var body: some Reducer<State, Action> {
-    BindingReducer()
+    BindingReducer(action: \.binding)
     Reduce { state, action in
       switch action {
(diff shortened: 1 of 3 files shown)


Wortason (approved):

stephencelis:
Closing in favor of upcoming changes to the binding infrastructure in 2.0.

PR #3323, opened 2024-08-29
Fix typo and deprecated content in `Performance.md`

MaraMincho:
I found some deprecated content in the `Performance.md` file and have made corrections. However, I'm concerned that this change might potentially alter the original intent of the document. I would appreciate your feedback. Thank you!


stephencelis (approved):
Thanks!

PR #3325, opened 2024-08-29
Address `Effect.throttle` sendability

stephencelis:
Addressing some more sendability warnings.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Throttle.swift
+++ b/Sources/ComposableArchitecture/Effects/Throttle.swift
@@ -1,8 +1,8 @@
-import Combine
+@preconcurrency import Combine
 import Dispatch
 import Foundation
 
-extension Effect {
+extension Effect where Action: Sendable {
   /// Throttles an effect so that it only publishes one output per given interval.
   ///
   /// The throttling of an effect is with respect to actions being sent into the store. So, if
@@ -23,12 +23,13 @@ extension Effect {
   ///     `false`, the publisher emits the first element received during the interval.
   /// - Returns: An effect that emits either the most-recent or first element received during the
   ///   specified interval.
-  public func throttle<S: Scheduler>(
+  public func throttle<S: Scheduler & Sendable>(
     id: some Hashable & Sendable,
     for interval: S.SchedulerTimeType.Stride,
     scheduler: S,
     latest: Bool
-  ) -> Self {
+  ) -> Self
+  where S.SchedulerTimeType.Stride: Sendable {
     switch self.operation {
     case .none:
       return .none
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Effects/Throttle.swift
```diff
@@ -1,8 +1,8 @@
-import Combine
+@preconcurrency import Combine
 import Dispatch
 import Foundation
 
-extension Effect {
+extension Effect where Action: Sendable {
```
stephencelis, line 5: Because the action flows through Combine across potential concurrency boundaries, I think we're forced to require a sendable action here. It's possible we could weaken this with a pure Swift concurrency throttle.

Sources/ComposableArchitecture/Effects/Throttle.swift
```diff
@@ -23,12 +23,13 @@ extension Effect {
   ///     `false`, the publisher emits the first element received during the interval.
   /// - Returns: An effect that emits either the most-recent or first element received during the
   ///   specified interval.
-  public func throttle<S: Scheduler>(
+  public func throttle<S: Scheduler & Sendable>(
```
stephencelis, line 26: The scheduler also travels across concurrency boundaries, so it needs to be sendable. Dispatch queues _are_ sendable, but run loops are not. Combine's `ImmediateScheduler` is also not sendable, but our `.immediate` scheduler is, as is our `TestScheduler`, so I think this is safe to do, it'll just mean folks shouldn't throttle with a run loop.

Sources/ComposableArchitecture/Effects/Throttle.swift
```diff
@@ -23,12 +23,13 @@ extension Effect {
...
     id: some Hashable & Sendable,
     for interval: S.SchedulerTimeType.Stride,
     scheduler: S,
     latest: Bool
-  ) -> Self {
+  ) -> Self
+  where S.SchedulerTimeType.Stride: Sendable {
```
stephencelis, line 32: Another requirement. Dispatch queue's stride type is sendable.

mbrandonw (approved):

Sources/ComposableArchitecture/Effects/Throttle.swift
AgapovOne: Hello. Is dispatchqueue stride sendable? I don't see how I could use throttle with DispatchQueue.main

In swift public interface I see this:
```
public struct Stride : SchedulerTimeIntervalConvertible, Comparable, SignedNumeric, ExpressibleByFloatLiteral, Hashable, Codable {
```

Docs don't mention Stride.

Also I get a warning:

```
Type 'DispatchQueue.SchedulerTimeType.Stride' does not conform to the 'Sendable' protocol; this is an error in the Swift 6 language mode
```


I don't see examples with throttle in ComposableArchitecture examples. I wanted to see if I'm missing something. Should it be other Scheduler than DispatchQueue.main? UIScheduler uses DispatchQueue.Stride too

PR #3326, opened 2024-08-29
Address effect cancellation sendability

stephencelis:
Some more changes to avoid sendability warnings in effect cancellation.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Cancellation.swift
+++ b/Sources/ComposableArchitecture/Effects/Cancellation.swift
@@ -1,4 +1,4 @@
-import Combine
+@preconcurrency import Combine
 import Foundation
 
 extension Effect {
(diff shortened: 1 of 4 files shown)


Sources/ComposableArchitecture/Effects/Cancellation.swift
```diff
@@ -101,7 +102,7 @@ extension Effect {
   /// - Parameter id: An effect identifier.
   /// - Returns: A new effect that will cancel any currently in-flight effect with the given
   ///   identifier.
-  public static func cancel<ID: Hashable>(id: ID) -> Self {
+  public static func cancel(id: some Hashable & Sendable) -> Self {
```
stephencelis, line 105: Cancellation IDs must now be `Sendable`. Technically we can avoid this with:

```swift
@autoclosure @escaping @Sendable () -> some Mashable
```

But I'm not sure it's worth it.

mbrandonw (approved):

PR #3327, opened 2024-08-30
Edit `04-Navigation-Lists-NavigateAndLoad` file in SwiftUI CaseStudies

MaraMincho:
I found the `if let` code in the `04-Navigation-Lists-NavigateAndLoad` file and made significant modifications to improve it. However, I'm concerned that these changes might unintentionally alter the original intent.

I would appreciate your feedback. Thank you!"

### DIFF
--- a/Examples/CaseStudies/SwiftUICaseStudies/04-Navigation-Lists-NavigateAndLoad.swift
+++ b/Examples/CaseStudies/SwiftUICaseStudies/04-Navigation-Lists-NavigateAndLoad.swift
@@ -17,8 +17,7 @@ struct NavigateAndLoadList {
       Row(count: 42, id: UUID()),
       Row(count: 100, id: UUID()),
     ]
-    var selectedRowID: UUID?
-    var selection: Counter.State?
+    var selection: Identified<Row.ID, Counter.State?>?
 
     struct Row: Equatable, Identifiable {
       var count: Int
@@ -42,7 +41,7 @@ struct NavigateAndLoadList {
         return .none
 
       case let .setNavigation(selection: .some(id)):
-        state.selectedRowID = id
+        state.selection = Identified(nil, id: id)
         return .run { send in
           try await self.clock.sleep(for: .seconds(1))
           await send(.setNavigationSelectionDelayCompleted)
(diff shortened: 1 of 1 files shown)


Examples/CaseStudies/SwiftUICaseStudies/04-Navigation-Lists-NavigateAndLoad.swift
```diff
@@ -10,13 +10,15 @@ private let readMe = """
...
+  @ObservableState
   struct State: Equatable {
     var rows: IdentifiedArrayOf<Row> = [
       Row(count: 1, id: UUID()),
       Row(count: 42, id: UUID()),
       Row(count: 100, id: UUID()),
     ]
-    var selection: Identified<Row.ID, Counter.State?>?
+    var selectedRowID: UUID?
+    var selection: Counter.State?
```
stephencelis, line 21: While this makes the `EmptyReducer` stuff less awkward, it unfortunately makes the domain modeled less precisely. It's not ideal, but let's return to using `Identified` for now.

Examples/CaseStudies/SwiftUICaseStudies/04-Navigation-Lists-NavigateAndLoad.swift
```diff
@@ -73,25 +73,23 @@ struct NavigateAndLoadListView: View {
...
+        AboutView(readMe: readMe)
+      }
+      ForEach(store.rows) { row in
+        NavigationLink(
+          "Load optional counter that starts from \(row.count)",
+          tag: row.id,
+          selection: .init(
+            get: { store.selectedRowID },
+            set: { id in store.send(.setNavigation(selection: id)) }
+          )
```
stephencelis, line 87: Can this use the new syntax?
```suggestion
          selection: $store.selectedRowID.sending(\.setNavigation)
```

stephencelis (commented):
Thanks for the PR! This case study has definitely fallen behind the other examples and should probably be overhauled from scratch, but I think it'd be good to get these fixes merged.

The one change I think we should revert, though, is the change of `Identified` domain. Can you address that and the other comment below? Then we'll be happy to merge!

Examples/CaseStudies/SwiftUICaseStudies/04-Navigation-Lists-NavigateAndLoad.swift
MaraMincho: Thanks for the code review. As you suggested, I proceeded with the refactor using `Identified`. However, I encountered an issue. When trying to apply the comment(use `sending` method) to the `selection` in `NavigationLink`, refactoring with `Identified` resulted in an inability to use the `sending` method.

``` swift
selection: $store.selection?.id.sending(\.setNavigation) // Compile Error
```

As a workaround, I had to implement the selection using a `get` and `set` approach.
If you have any suggestions for a better solution, I'd appreciate your feedback. Thank you

MaraMincho: Additionally, the original code `state.selection?.value = Counter.State(count: state.rows[id: id]?.count ?? 0)` was causing an Overlapping accesses error, so I assigned the value to a separate variable to resolve this.

```swift
case .setNavigationSelectionDelayCompleted:
    guard let id = state.selection?.id else { return .none }
    // Overlapping accesses to 'state', but modification requires exclusive access; consider copying to a local variable
    state.selection?.value = Counter.State(count: state.rows[id: id]?.count ?? 0)
    return .none
```

stephencelis (approved):
At the end of the day I think this is an improvement to case study that we're not super happy with at the moment, so I think we're good to merge this as a first step to improve things. Thanks!

We'll maybe rethink this case study in the future.

PR #3329, opened 2024-08-30
Address @Shared sendability.

mbrandonw:


Package.swift
```diff
@@ -87,7 +87,7 @@ let package = Package(
 )
 
 #if compiler(>=6)
-  for target in package.targets where target.type != .system {
+  for target in package.targets where target.type != .system && target.type != .test {
```
mbrandonw, line 90: I added this temporarily so that we don't get flooded with warnings in test targets, but maybe we should keep this around until we've addressed all/most warnings in TCA proper, and then we can later concentrate on tests?

Sources/ComposableArchitecture/SharedState/PersistenceKey.swift
```diff
@@ -7,9 +7,9 @@
 ///
 /// See the article <doc:SharingState> for more information, in particular the
 /// <doc:SharingState#Custom-persistence> section.
-public protocol PersistenceReaderKey<Value> {
+public protocol PersistenceReaderKey<Value>: Sendable {
```
mbrandonw, line 10: Persistence keys need to be sendable because they are captured in some `LockIsolated`'s, which currently require `Sendable` values and `@Sendable` transaction closures. If we weaken that, then perhaps this can be weakened too?

Sources/ComposableArchitecture/SharedState/PersistenceKey.swift
```diff
@@ -7,9 +7,9 @@
...
   /// A type that can be loaded or subscribed to in an external system.
-  associatedtype Value
+  associatedtype Value: Sendable
```
mbrandonw, line 12: `Value` is captured in a variety of sendable contexts, such as in the subscription endpoint for persistence strategies. So I think this does really need to be full blown `Sendable`.

Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
```diff
@@ -154,7 +154,7 @@ extension PersistenceReaderKey {
 /// A type defining a user defaults persistence strategy.
 ///
 /// See ``PersistenceReaderKey/appStorage(_:)-4l5b`` to create values of this type.
-public struct AppStorageKey<Value> {
+public struct AppStorageKey<Value: Sendable>: Sendable {
   private let lookup: any Lookup<Value>
   private let key: String
   private let store: UserDefaults
```
mbrandonw, line 160: Technically `UserDefaults` is not `Sendable` even though the docs specifically say "User defaults is thread-safe". So, it should be ok to make this `UncheckedSendable`, but it was strangely causing some test failures. So I need to look at this again.

Sources/ComposableArchitecture/SharedState/References/ValueReference.swift
```diff
@@ -328,15 +328,15 @@ extension SharedReader {
...
+final class ValueReference<Value, Persistence: PersistenceReaderKey<Value>>: Reference,
+  @unchecked Sendable
 {
   private let lock = NSRecursiveLock()
   private let persistenceKey: Persistence?
   #if canImport(Combine)
     private let subject: CurrentValueRelay<Value>
   #endif
-  private var subscription: Shared<Value>.Subscription?
+  private var subscription: Shared<Value>.Subscription!
```
mbrandonw, line 339: we'd prefer for this to be a `let` so that we don't have to do `@unchecked Sendable` above, but we need lazy initialization. And to make that clear I went ahead and made it an implicitly unwrapped optional.

Sources/ComposableArchitecture/SharedState/Shared.swift
```diff
@@ -12,7 +12,7 @@ import IssueReporting
 /// wrapper.
 @dynamicMemberLookup
 @propertyWrapper
-public struct Shared<Value> {
+public struct Shared<Value: Sendable> {
   private let reference: any Reference
   private let keyPath: AnyKeyPath
```
mbrandonw, line 17: Technically this should now be:

```swift
public struct Shared<Value: Sendable> {
```

…and we should drop the conditional conformance below. But there are some weird interactions with `_AnyKeyPath`. Let's chat about this today.

Sources/ComposableArchitecture/SharedState/Shared.swift
```diff
@@ -408,11 +411,10 @@ extension Shared: MutableCollection
 where Value: MutableCollection & RandomAccessCollection, Value.Index: Hashable {
   public subscript(position: Value.Index) -> Shared<Value.Element> {
     get {
-      assertionFailure("Conformance of 'Shared<Value>' to 'MutableCollection' is unavailable.")
-      return self[position, default: DefaultSubscript(self.wrappedValue[position])]
+      fatalError("Conformance of 'Shared<Value>' to 'MutableCollection' is unavailable.")
     }
     set {
-      self._wrappedValue[position] = newValue.wrappedValue
+      fatalError("Conformance of 'Shared<Value>' to 'MutableCollection' is unavailable.")
```
mbrandonw, line 417: This subscript is unavailable so it should be ok to `fatalError`. If we don't fatal error, then we get sendability warnings that can't be fixed because you can't conditionally conform to a protocol with a constraint against a marker protocol, like `Sendable`.

Sources/ComposableArchitecture/Internal/DefaultSubscript.swift
```diff
@@ -1,7 +1,7 @@
-final class DefaultSubscript<Value>: Hashable {
-  var value: Value
+final class DefaultSubscript<Value: Sendable>: Hashable, Sendable {
```
stephencelis, line 1: We know we want to remove these helpers in the future, but since they're use by `store.scope` let's split off a non-sendable version for that.

stephencelis (approved):

PR #3333, opened 2024-08-30
Require main actor isolation in store collection

stephencelis:
We can eliminate sendable warnings and improve the error handling via `MainActor.preconditionIsolated`.


Sources/ComposableArchitecture/Observation/IdentifiedArray+Observation.swift
```diff
@@ -90,6 +90,7 @@
     private let store: Store<IdentifiedArray<ID, State>, IdentifiedAction<ID, Action>>
     private let data: IdentifiedArray<ID, State>
 
+    @MainActor
```
stephencelis, line 93:
```suggestion
    #if swift(<5.10)
      @MainActor(unsafe)
    #else
      @preconcurrency@MainActor
    #endif
```

mbrandonw (approved):

PR #3345, opened 2024-09-04
Fix document in WhatIsNavigation.md

stealmh:
While working through the documentation, I noticed that the items in the array were missing commas, so I fixed that.


mbrandonw (approved):

mbrandonw:
Thanks @stealmh!

PR #3346, opened 2024-09-04
Added failing test to document behavior with observation/identified array

mbrandonw:
Added failing test to document behavior with observation and identified array.


Tests/ComposableArchitectureTests/ObservableTests.swift
```diff
@@ -616,6 +640,7 @@ private struct ChildState: Equatable, Identifiable {
     self
   }
 }
+import ComposableArchitectureMacros
```
stephencelis, line 643:
```suggestion
```

PR #3347, opened 2024-09-04
Warn if bindable store binding action isn't processed

stephencelis:
Looks like the warnings we emit when we detect `BindingReducer` is missing are only applied to view stores, and were not ported over to the newer observable store bindings.

This PR fixes that, though the main caveat is the messages can't seem to point to any good context. These bindings are derived from dynamic member lookup, which can't include source context like file/line.

### DIFF
--- a/Sources/ComposableArchitecture/Observation/Binding+Observation.swift
+++ b/Sources/ComposableArchitecture/Observation/Binding+Observation.swift
@@ -78,12 +78,71 @@
     }
   }
 
+  private final class BindableActionDebugger<Action>: Sendable {
+    let value: any Sendable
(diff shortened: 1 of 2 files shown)


Sources/ComposableArchitecture/Observation/Binding+Observation.swift
```diff
@@ -78,12 +78,81 @@
     }
   }
 
+  private final class BindableActionDebugger<Action>: Sendable {
```
stephencelis, line 81: The work required to reuse the existing debugger seemed to not be worth it, especially if we're removing those APIs in 2.0.

mbrandonw (approved):

PR #3349, opened 2024-09-05
Fix typos

stealmh:
While checking the comments, we noticed that the backtick was missing, so we fixed it. 
Thank you for your time and attention in reviewing this PR.


mbrandonw (approved):
Hi @stealmh, good catches! Thanks!

Issue #3311, opened 2024-08-26
Crash when using @Shared appStorage during certain conditions


zachwaugh:
### Description

I'm seeing a crash with a similar backtrace to https://github.com/pointfreeco/swift-composable-architecture/issues/3247, but appears not to be caused by a thread issue. 

I'm using a @Shared appStorage key roughly like:

```swift
@Shared(.appStorage("key")) var isEnabled: Bool = true
```

When launching the app it crashes. I'm still working on a simplified repro but having a hard time isolating the exact code causing the problem, but thought I'd post the issue in case anyone else has an idea.

<img width="1397" alt="Screenshot 2024-08-26 at 12 24 36 PM" src="https://github.com/user-attachments/assets/5ad06440-b94e-44ce-8f39-7d9160e1269d">

If you look at the backtrace, it appears an `NSAttributedString` is performing layout (which isn't in my code) which triggers an `NSUserDefault` update which in turn causes the `@Shared` storage to update immediately during the same render cycle?The only relevant logs related to the crash are:

```
precondition failure: setting value during update: 219864
```

<img width="660" alt="Screenshot 2024-08-26 at 12 25 55 PM" src="https://github.com/user-attachments/assets/af38c318-15a0-44a7-bc8a-6e85ac6d4f56">

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### The Composable Architecture version information

1.13.1

### Destination operating system

iOS 17.5

### Xcode version information

Xcode 15.4

mbrandonw:
Hi @zachwaugh, thanks for the report, but unfortunately without a repro it's hard to diagnose. The precondition failure message ("precondition failure: setting value during update") seems to clearly describe the problem: in the act of executing a view's body a value is being mutated. If you do this with `@State`, SwiftUI will even produce a purple runtime warning letting you know this isn't a good thing to do.

[...]

PR #3356, opened 2024-09-05
Fix potential deadlock in Shared

### LINKED ISSUE #3355: Deadlock on main thread when loading dependencies and @Shared values
### Description

I noticed my app getting into deadlock on the main thread when using `@Shared`. I think this is probably the same underlying problem and root cause as my other issue (https://github.com/pointfreeco/swift-composable-architecture/issues/3311), but this one I managed to make a repro for.

Screenshot 2024-09-05 at 12 57 51 PM
image

### Checklist

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

Here's a repro similar to my app that I was able to finally reduce into something simpler, but still approximating what happens in my app. There's a lot of things at play and changing some subtle things will cause it not to deadlock, but I was able to reliably deadlock every time with this. Just note if you change something and it works, you *may* need to delete the app and do a fresh install to trigger it again. I didn't have time at the moment to dig further, but I suspect at least some part of it is due to `UserDefaults` initializing the first time?

Run the app in a simulator, and you'll see `Database init() start` log and nothing else and the app will be locked. Hit the pause execution  button and you can see where its locked. I think as @mbrandonw suggested in the other issue that changing that `mainActorNow` to `DispatchQueue.main.async` would fix it and I'll try that later when I have some more time.

```swift
import ComposableArchitecture
import SwiftUI

@main
struct TCATestApp: App {
    var body: some Scene {
        WindowGroup {
            DeadlockView(store: Store(initialState: Deadlock.State(), reducer: {
                Deadlock()
            }))
        }
    }
}

struct DeadlockView: View {
    @Bindable var store: StoreOf<Deadlock>
    @Environment(\.scenePhase) private var scenePhase

    var body: some View {
        NavigationSplitView {
            List(selection: $store.selection) {
                NavigationLink(value: "child") {
                    Text("Go To Child")
                }
            }
            .navigationTitle("Home")
            .listStyle(.sidebar)
            .task {
                await store.send(.task).finish()
            }
        } detail: {
            NavigationStack {
                switch store.destination {
                case .child:
                    if let store = store.scope(state: \.destination?.child, action: \.destination.child) {
                        ChildView(store: store)
                    }
                default:
                    EmptyView()
                }
            }
        }
    }
}


@Reducer
struct Deadlock {
    @Reducer(state: .equatable)
    enum Destination {
        case child(Child)
    }

    @ObservableState
    struct State: Equatable {
        var selection: String? = "child"
        @Presents var destination: Destination.State? = .child(Child.State())
        @Shared(.appStorage("test")) var test: Bool = false
    }

    enum Action: BindableAction {
        case task
    }

    @Dependency(\.database) var database

        BindingReducer()
        Reduce { state, action in
            switch action {
            case .task:
                return .merge(
                    .run { _ in
                        print("Querying database…")
                        await database.query()
                        print("… query complete")
                    }.cancellable(id: CancelId.one, cancelInFlight: true)
                )

            case .destination, .binding:
                return .none
            }
        }
    }

    private enum CancelId {
        case one
    }
}

@Reducer
struct Child {
    @ObservableState
    struct State: Equatable {
        @Shared(.appStorage("count")) var count: Int = 0
    }

    enum Action {
        case incrementButtonTapped
    }

        Reduce { state, action in
            switch action {
            case .incrementButtonTapped:
                state.count += 1
                return .none
            }
        }
    }
}

struct ChildView: View {
    let store: StoreOf<Child>

    var body: some View {
        VStack {
            Text(store.count.formatted())
                .navigationTitle("Child")
            Button("Increment") {
                store.send(.incrementButtonTapped)
            }
            .buttonStyle(.borderedProminent)
        }
    }
}


struct Database: DependencyKey {
    init() {
        print("Database init() start")
        UserDefaults.standard.set("foo-bar", forKey: "test-key")
        // Will never be called, will deadlock on the UserDefaults notification
        print("Database init() end")
    }

    func query() async {}

    static var liveValue: Database = Database()
}

extension DependencyValues {
    var database: Database {
        get { self[Database.self] }
        set { self[Database.self] = newValue }
    }
}
```

1.14.0

iOS 17.5

### Xcode version information

Xcode 15.4

_No response_

mbrandonw:
Fixes #3355.

### DIFF
--- a/Examples/SyncUps/SyncUps/App.swift
+++ b/Examples/SyncUps/SyncUps/App.swift
@@ -2,149 +2,27 @@ import ComposableArchitecture
 import SwiftUI
 
 @main
-struct TCATestApp: App {
+struct SyncUpsApp: App {
+  // NB: This is static to avoid interference with Xcode previews, which create this entry
+  //     point each time they are run.
+  @MainActor
+  static let store = Store(initialState: AppFeature.State()) {
+    AppFeature()
+      ._printChanges()
+  } withDependencies: {
+    if ProcessInfo.processInfo.environment["UITesting"] == "true" {
+      $0.defaultFileStorage = .inMemory
+    }
+  }
+
   var body: some Scene {
     WindowGroup {
-      DeadlockView(store: Store(initialState: Deadlock.State(), reducer: {
-        Deadlock()
-      }))
-    }
-  }
-}
-
-struct DeadlockView: View {
-  @Bindable var store: StoreOf<Deadlock>
-  @Environment(\.scenePhase) private var scenePhase
-
-  var body: some View {
-    NavigationSplitView {
-      List(selection: $store.selection) {
-        NavigationLink(value: "child") {
-          Text("Go To Child")
-        }
-      }
-      .navigationTitle("Home")
-      .listStyle(.sidebar)
-      .task {
-        await store.send(.task).finish()
-      }
-    } detail: {
-      NavigationStack {
-        switch store.destination {
-        case .child:
-          if let store = store.scope(state: \.destination?.child, action: \.destination.child) {
-            ChildView(store: store)
-          }
-        default:
-          EmptyView()
-        }
+      if _XCTIsTesting {
+        // NB: Don't run application in tests to avoid interference between the app and the test.
+        EmptyView()
+      } else {
+        AppView(store: Self.store)
       }
     }
-  }
-}
-
-
-@Reducer
-struct Deadlock {
-  @Reducer(state: .equatable)
-  enum Destination {
-    case child(Child)
-  }
-
-  @ObservableState
-  struct State: Equatable {
-    var selection: String? = "child"
-    @Presents var destination: Destination.State? = .child(Child.State())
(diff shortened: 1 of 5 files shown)


zachwaugh:
Thanks @mbrandonw! Good news is I confirmed it fixes #3355 for me. Unfortunately it doesn't fix #3311. In that case, it's already on the main thread, so it skips the `DispatchQueue.main.async` branch already:

Screenshot 2024-09-05 at 7 47 31 PM

I'll spend some more time trying to make a repro for that one now that I have the other working, hopefully can modify it a bit to trigger that.

stephencelis (approved):

mbrandonw:
> I'll spend some more time trying to make a repro for that one now that I have the other working, hopefully can modify it a bit to trigger that.

Hi @zachwaugh, thanks for confirming #3355, and good to know about #3311. That is a bummer. We could of course go full `main.async`, but it is nice to maintain some synchronicity in the subscription where possible. If you come up with a repo that would be very helpful.

PR #3364, opened 2024-09-07
Fix build issue related to isolated(any)

maximkrouk:
Enable experimental feature `IsolatedAny`

In some environments (at least Xcode 16 beta (16A5171c)) the lack of this flag caused compilation error 😔

`Attribute requires '-enable-experimental-feature IsolatedAny'`

Screenshot 2024-09-07 at 11 54 27 PM

Screenshot 2024-09-08 at 12 14 21 AM

> Issue was present even in standalone swift-tca package build, I didn't check it as a dependency yet, but now local build is successful c:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -91,6 +91,7 @@ let package = Package(
     target.swiftSettings = target.swiftSettings ?? []
     target.swiftSettings?.append(contentsOf: [
       .enableExperimentalFeature("StrictConcurrency"),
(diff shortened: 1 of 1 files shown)


maximkrouk:
Here is a conflict, but instead of fixing it I updated my Xcode and it seems like it solved the problem, I'm closing the PR, at least if someone traps into the same thing they'll find this PR c:

PR #3372, opened 2024-09-09
Add teststore receive action case for non equatable values

Alex293:
Hi,
I often have a couple properties in enum cases and find it a bit painful to have to either wrap those un a equatable struct or use the closure version. Would you be interested by adding an overload for non equatable version which take an equals closure ? like 
```swift
await store.receive(\.delegate.success, ("Hello!", 1), ==)
```
for a action defined like `case success<Result<(String, Int)>, Error>` ?

I just wrote the PR code in a minutes before leaving so it might not work. If you are interested, I know you are chasing down the number of overloads, I'll properly finish the PR.

### DIFF
--- a/Sources/ComposableArchitecture/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestStore.swift
@@ -1476,18 +1476,43 @@ extension TestStore where State: Equatable {
     line: UInt = #line,
     column: UInt = #column
   ) {
+    self._receive(
+      actionCase,
+      value,
+      equals: { $0 == $1 },
+      assert: updateStateToExpectedResult,
+      fileID: fileID,
+      filePath: filePath,
+      line: line,
+      column: column
+    )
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hi @Alex293, thanks for taking the time to explore this!

The main problem with things like this is that the added overloads can severely hurt compile times and error diagnostics. And the latter concern especially goes for this situation, where `receive` now has two trailing closures.

We have run into this problem many, _many_ times, and at this point we really try our hardest to avoid overloading like this when there are alternatives available (such as putting the tuple into a `struct`).

stephencelis:
@Alex293 Gonna close this for now because of the overload issue. I think we could revisit this in 2.0 when the existing overloads are removed, but even then because it's an additional trailing closure to disambiguate, we're not sure it's worth the potential compiler pain 😞

PR #3382, opened 2024-09-13
Soft-deprecate `Store.ifLet`

stephencelis:
We can prefer `observe` and `if let store.scope` now.


mbrandonw (approved):
Should we get a migration guide in for this too?

stephencelis:
@mbrandonw Done!

PR #3406, opened 2024-09-24
Add Swift Macro Compatibility Check GitHub Action

Matejkob:
This PR adds a **Swift Macro Compatibility Check** GitHub Action to ensure that macro changes remain compatible across different versions of `swift-syntax`. The action automatically verifies compatibility with major `swift-syntax` versions (`509.0.0`, `510.0.0`, `600.0.0`), running builds and tests. It's also posible to turn on checking agains all versions of `swift-syntax`.

#### Key Benefits:
- **Automated Compatibility Checks**: No need to manually test changes across different `swift-syntax` versions—this is handled automatically in CI.
- **Prevents Breakages**: Helps avoid accidentally committing changes that break older versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,3 +104,15 @@ jobs:
         run: make SCHEME="Todos" test-example
       - name: VoiceMemos
         run: make SCHEME="VoiceMemos" test-example
+        
+  check-macro-compatibility:
+    name: Check Macro Compatibility
+    runs-on: macos-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Run Swift Macro Compatibility Check
+        uses: Matejkob/swift-macro-compatibility-check@v1.0.0
+        with:
+          run-tests: false
+          major-versions-only: false


Matejkob:
Wrapping up: checking against all versions of `swift-syntax`:

* 509.0.0
* 509.0.1
* 509.0.2
* 509.1.0
* 509.1.1
* 510.0.0
* 510.0.1
* 510.0.2
* 510.0.3
* 600.0.0

for this repository, it takes approximately **8 minutes**,

but checking against only major versions:

* 509.0.0
* 510.0.0
* 600.0.0

takes **3 minutes**. From my experience, running only against major versions is sufficient since swift-syntax does not introduce significant changes across major versions. @stephencelis @mbrandonw, please let me know which version you prefer.

stephencelis:
@Matejkob Might be nice to do major versions only. Should the major versions take the latest up to the next major? _E.g._:

- 509.1.1
- 510.0.3
- 600.0.0

Or do you think compatibility with the earlier version in a major release is better?

Matejkob:
> @Matejkob Might be nice to do major versions only. Should the major versions take the latest up to the next major? _E.g._:
> 
> * 509.1.1
> * 510.0.3
> * 600.0.0
> 
> Or do you think compatibility with the earlier version in a major release is better?

At the time I considered this, I concluded that the "major version" should correspond to the version of `swift-syntax` where a new [version marker module](https://github.com/swiftlang/swift-syntax/tree/main/Sources/VersionMarkerModules) was introduced. I believe this approach makes the most sense. For instance, if at some point `swift-syntax` requires a version marker (empty product) for a non-major version, like `6.0.2`, then it can also be added to the list of "major versions" (it should not be called major version anymore)

Matejkob:
I'm marking it as a draft since I want to make one more change before we land it. But before that let's end discusing the above topic.

stephencelis:
I think it makes sense, just curious about the thought process :smile: Let us know when it's ready and we'll merge!

Matejkob:
@stephencelis it's ready; I've updated the script to use `v1` version

stephencelis (approved):
Thanks!

Issue #3417, opened 2024-09-30
Abnormal behavior for timer in TCA Tutorials


minjunkim-dev:
### Description

I am learning TCA through tutorials.

I found an abnormal behavior for timer in Section 3 of https://pointfreeco.github.io/swift-composable-architecture/main/tutorials/composablearchitecture/01-04-composingfeatures.

After activating each timer in two tabs (tab1, tab2), if you cancel the timer in one tab, the timers in both tabs are canceled.

The same id, "CancelID.timer" (enum case), is registered in the .cancellable(id:) method, so even if only one is canceled, both seem to be cancelled.

I think this behavior is different from what the tutorial intended.

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Expected behavior

The timer on each tab operates “independently.”

### Actual behavior

If you cancel the timer on one tab, the timer on the other tab will also be canceled.

### Reproducing project

Please refer to Section 3(Deriving child stores) of https://pointfreeco.github.io/swift-composable-architecture/main/tutorials/composablearchitecture/01-04-composingfeatures.

### The Composable Architecture version information

1.15.0

### Destination operating system

iOS 18

### Xcode version information

Xcode 16.0

### Swift Compiler version information

```shell
swift-driver version: 1.115 Apple Swift version 6.0 (swiftlang-6.0.0.9.10 clang-1600.0.26.2)
Target: arm64-apple-macosx15.0
```

stephencelis:
Closing per #3418, but we appreciate you taking the time to report and PR!

PR #3418, opened 2024-09-30
Fix issue #3417: Add unique ID for independent timer operation

minjunkim-dev:
Please refer to https://github.com/pointfreeco/swift-composable-architecture/issues/3417.


stephencelis:
@minjunkim-dev Sorry for sitting on this. We discussed this awhile back but never returned to it. Your solution is 100% correct for the library today, where shared effect cancellation identifiers can trample over one another, but we also don't think that the tutorial is the right place to get bogged down by this.

The good news is this behavior is fixed in the upcoming 2.0 of the library with a new effect cancellation system. We hope you'll give it a spin!

PR #3424, opened 2024-10-02
Update benchmarks target

stephencelis:
This PR replaces the outdated suite with a new benchmark using the library adopted by many Apple packages.

A GitHub action will be added in a follow-up PR.


Benchmarks/Benchmarks/swift-composable-architecture-benchmark/ Benchmarks.swift
Matejkob, line 1: Is the space in front of file name ` Benchmarks.swift` intentional?

PR #3429, opened 2024-10-07
Shared: Introduce Persistence Key Lifecycle

stephencelis:
Currently, persistence keys are "globals" that live as long as the application, which is fine when there are a small number of them, but there are two shortcomings that could be improved:

1. Because they can have active subscriptions running in the background, this is work that is performed regardless of if any `Shared` value is currently in existence.
2. Applications with dynamic persistence keys could create an unbounded number of these values in memory, and subscriptions attached to them, over time, which is more memory and work that should be avoided.

Draft PR for now since the implementation is a proof of concept that could probably be cleaned up.


stephencelis:

PR #3437, opened 2024-10-09
Subscribe to `@Shared` strategies with async hop

### LINKED ISSUE #3311: Crash when using @Shared appStorage during certain conditions

stephencelis:
I think we're finding that synchronous subscriptions can be problematic depending on the strategy, for example as we're seeing with user defaults notifications being fired by attributed string's internals.

Fixes #3311.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/DispatchQueue.swift
+++ b/Sources/ComposableArchitecture/Internal/DispatchQueue.swift
@@ -14,20 +14,6 @@ func mainActorNow<R: Sendable>(execute block: @MainActor @Sendable () -> R) -> R
   }
 }
 
-func mainActorAsync(execute block: @escaping @MainActor @Sendable () -> Void) {
-  if isTesting {
-    MainActor._assumeIsolated {
-      block()
-    }
-  } else {
-    DispatchQueue.main.async {
-      MainActor._assumeIsolated {
-        block()
-      }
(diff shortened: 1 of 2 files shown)


stephencelis:
Closing in favor of #3487.

PR #3447, opened 2024-10-16
Fix race conditions in CurrentValueRelay

kabiroberai:
As it's currently written, `CurrentValueRelay` seems to have several race conditions. These are infrequent enough to make them rather elusive in typical usage, but we believe that these are the root cause behind 0.X% of Ramp iOS sessions crashing.

This PR includes a test case that fails on `main`, as well as a fix that solves it, but is slow. It's probably worth you folks writing a fix yourself (without needing the lock to be recursive) to resolve the issue, but I hope that the test case helps. FWIW you should also be able to trip up TSAN when running the failing test case pre-fix; it catches a number of the ways in which the race can be triggered.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/CurrentValueRelay.swift
+++ b/Sources/ComposableArchitecture/Internal/CurrentValueRelay.swift
@@ -55,7 +55,7 @@ extension CurrentValueRelay {
     private var demand = Subscribers.Demand.none
 
     private var _downstream: (any Subscriber<Output, Never>)?
-    var downstream: (any Subscriber<Output, Never>)? {
+    private var downstream: (any Subscriber<Output, Never>)? {
       var downstream: (any Subscriber<Output, Never>)?
       self.lock.sync { downstream = _downstream }
       return downstream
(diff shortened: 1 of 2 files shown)


iampatbrown:
Hey @kabiroberai :) awesome work! Thanks for fixing this. 
Your changes look good and worked well on my end. I think they can be merged as is.
One question related to locking access to the upstream's current value, do you have an example where this is required or possibly a failing test? I'm not saying it's not needed, I'm just trying to better understand for my own curiosity 😅 
Also, is the `_downstream` property mainly so we can lock access to it? It could be worth locking above the guard statements to avoid locking/unlocking twice.

kabiroberai:
appreciate the vote of confidence @iampatbrown!

re locking `upstream.currentValue`, the reason behind this is that, as far as I can tell, it's otherwise possible for a `Subscription.request(_:)` call to read the value in parallel with a call to `CurrentValueRelay.send` which can be writing to it at the same time. This breaks the invariant that there can only be one writer xor multiple readers of shared mutable state.

As for the `_downstream` property, indeed the indirection is mainly a convenience. Moving the `lock()` invocations up is a good call, I'll give that a shot.

stephencelis (approved):
Thanks!

PR #3448, opened 2024-10-16
Shared: Simplify internal reference protocol

stephencelis:
Tackling some small `Shared` refactorings across small PRs. This one simplifies an internal protocol and fixes some potential over-observation.


Tests/ComposableArchitectureTests/SharedTests.swift
```diff
@@ -461,7 +461,7 @@ final class SharedTests: XCTestCase {
     _count = Shared(0)
     let countDidChange = self.expectation(description: "countDidChange")
     withPerceptionTracking {
-      _ = $count
+      _ = $count.wrappedValue
```
stephencelis, line 464: There's no need to observe a `Shared` when its underlying value is never accessed.

stephencelis:

PR #3449, opened 2024-10-16
Give `SharedReader` same `Sendable` treatment as `Shared`

stephencelis:
Some changes made to `Shared` weren't brought over to `SharedReader`.

Would be nice to figure out how to better unify these types to avoid duplication and fragmentation.


stephencelis:

PR #3452, opened 2024-10-18
Call same `receive` method as receiver on wrapped `Publisher`

jhollida24:
Our company is seeing this line cited in a stack overflow. This is ~ a guess, but it would make sense that if the contained `Publisher` is the one being subscribed to that we should pass that along using the same API we are called with.

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Publisher.swift
+++ b/Sources/ComposableArchitecture/Effects/Publisher.swift
@@ -21,7 +21,7 @@ public struct _EffectPublisher<Action>: Publisher {
   }
 
   public func receive(subscriber: some Combine.Subscriber<Action, Failure>) {
-    publisher.subscribe(subscriber)
+    publisher.receive(subscriber: subscriber)
   }
 
   private var publisher: AnyPublisher<Action, Failure> {


jhollida24:
For context, here's the stack overflow: 
image

We have not been able to repro it internally.

Sources/ComposableArchitecture/Effects/Publisher.swift
```diff
@@ -21,7 +21,7 @@ public struct _EffectPublisher<Action>: Publisher {
   }
 
   public func receive(subscriber: some Combine.Subscriber<Action, Failure>) {
-    publisher.subscribe(subscriber)
+    publisher.receive(subscriber: subscriber)
   }
```
pyrtsa, line 25: I don't see why this change would fix the stack overflow, do you?

The docs for [`Publisher.subscribe(_:)`](https://developer.apple.com/documentation/combine/publisher/subscribe(_:)-4u8kn#discussion) read to me like the old implementation was right.

stephencelis:
@jhollida24 According to Apple:

> Always call [subscribe] instead of [receive(subscriber:)](https://developer.apple.com/documentation/combine/publisher/receive(subscriber:)). Adopters of [Publisher](https://developer.apple.com/documentation/combine/publisher) must implement [receive(subscriber:)](https://developer.apple.com/documentation/combine/publisher/receive(subscriber:)). The implementation of [subscribe(_:)](https://developer.apple.com/documentation/combine/publisher/subscribe(_:)-4u8kn) provided by [Publisher](https://developer.apple.com/documentation/combine/publisher) calls through to [receive(subscriber:)](https://developer.apple.com/documentation/combine/publisher/receive(subscriber:)).

—https://developer.apple.com/documentation/combine/publisher/subscribe(_:)-4u8kn

So I don't think that's the problem here. Combine stacks just happen to be deep and so you're more likely to overflow the stack when working with combine publishers and large types on the stack.

TCA uses a lot of value types that can get heavily nested (in state, in actions, the reducers themselves), and so sometimes you may blow the stack just because too much memory is allocated there. The typical way to work around this is to use copy-on-write boxes around parts of your app's state/reducer, or use indirect enums for your actions so that they live on the heap.

TCA's presentation tools try to bake this into the system for you to make stack overflow less likely, but even if you _are_ using `@Presents` and `PresentationAction` you can still end up with value types that are too large and need to be boxed.

jhollida24:
This is helpful, thank you.

My intuition for wrapping something like this says that it likely breaks properties that strict recursion would provide. But this sounds like a more plausible explanation to me.

Do you have any advice on how I might go about profiling the size of stack frames? My understanding is that something like the Allocations tool is for studying what has been allocated to the heap.

stephencelis:
@jhollida24 I don't have any specifics other than printing out the size of the state/action/reducer types in those frames. @tgrapperon shared some tools a long time ago here: https://github.com/pointfreeco/swift-composable-architecture/discussions/752#discussioncomment-2977319

And he has this branch you could explore:

https://github.com/pointfreeco/swift-composable-architecture/compare/main...tgrapperon:swift-composable-architecture:runtime-stack-checks

Given that this PR probably does not address the given issue I'm going to close it, but feel free to start a new discussion to explore diagnosing the issue!

jhollida24:
> @jhollida24 I don't have any specifics other than printing out the size of the state/action/reducer types in those frames. @tgrapperon shared some tools a long time ago here: [#752 (comment)](https://github.com/pointfreeco/swift-composable-architecture/discussions/752#discussioncomment-2977319)
> 
> And he has this branch you could explore:
> 
> [main...tgrapperon:swift-composable-architecture:runtime-stack-checks](https://github.com/pointfreeco/swift-composable-architecture/compare/main...tgrapperon:swift-composable-architecture:runtime-stack-checks)
> 
> Given that this PR probably does not address the given issue I'm going to close it, but feel free to start a new discussion to explore diagnosing the issue!

This is a great lead for us to follow. Thank you so much for your help!

PR #3457, opened 2024-10-21
Update Russian translation in README.md

SubvertDev:
It looks like the Russian translation hasn't been updated since version 0.51, so I’ve bumped it to the latest 1.15. I’ll try to keep it up to date if it gets merged. Thanks!


stephencelis (approved):
Thanks!

PR #3470, opened 2024-10-29
AppStorage: add support for Date values

rudedogdhc:
Since SwiftUI's `AppStorage` now supports `Date` values, it makes sense to add this to TCA as well.


stephencelis (approved):
Thanks!

stephencelis (approved):

PR #3475, opened 2024-11-01
Fix race in DemandBuffer

kabiroberai:
After landing #3447, we found another race condition — this time in `DemandBuffer`, which is marked `@unchecked Sendable` (and should indeed be thread safe) but has no synchronization around `DemandBuffer.buffer` (as well as other slightly less critical properties) in some spots.

A similar fix was actually applied to the original `CombineExt.DemandBuffer` impl back in 2021, but wasn't ported over to TCA. See https://github.com/CombineCommunity/CombineExt/pull/90.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/Create.swift
+++ b/Sources/ComposableArchitecture/Internal/Create.swift
@@ -21,27 +21,23 @@
 // THE SOFTWARE.
 
 @preconcurrency import Combine
-import Darwin
+import Foundation
 
 final class DemandBuffer<S: Subscriber>: @unchecked Sendable {
   private var buffer = [S.Input]()
   private let subscriber: S
   private var completion: Subscribers.Completion<S.Failure>?
   private var demandState = Demand()
-  private let lock: os_unfair_lock_t
+  private let lock = NSRecursiveLock()
 
   init(subscriber: S) {
     self.subscriber = subscriber
-    self.lock = os_unfair_lock_t.allocate(capacity: 1)
-    self.lock.initialize(to: os_unfair_lock())
-  }
-
-  deinit {
-    self.lock.deinitialize(count: 1)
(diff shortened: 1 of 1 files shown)


Sources/ComposableArchitecture/Internal/Create.swift
```diff
@@ -21,27 +21,23 @@
...
-import Darwin
+import Foundation
 
 final class DemandBuffer<S: Subscriber>: @unchecked Sendable {
   private var buffer = [S.Input]()
   private let subscriber: S
   private var completion: Subscribers.Completion<S.Failure>?
   private var demandState = Demand()
-  private let lock: os_unfair_lock_t
+  private let lock = NSRecursiveLock()
```
kabiroberai, line 31: Once again, I wanted to err on the side of caution and changed this to a recursive lock to avoid deadlocks — but it might be worth reasoning about the code more thoroughly and determining if it's possible to change this back. The main concern is that there are calls to `Subscriber.receive` while the lock is held, and afaik these can invoke arbitrary code that might call back into the `DemandBuffer` and deadlock us.

jpsim (approved):
Yeah this does seem to be correct to me. Nice find, Kabir!

freak4pc:
Indeed seems like an important fix to add. Great job @kabiroberai on catching these.

kabiroberai:
CI failure seems unrelated to the PR 🤔

```bash
> sudo xcode-select -s /Applications/Xcode_16.0.app
xcode-select: error: invalid developer directory '/Applications/Xcode_16.0.app'
```

jshier:
GitHub may have changed their images, Xcode 16.0 is now named Xcode_16.app, with a symlink to Xcode_16.0.app, which I guess doesn't work for setting the directory.

PR #3476, opened 2024-11-03
Avoid extra lock when getting `subscriptions` in `CurrentValueRelay.send`

iampatbrown:
Minor improvement I forgot to mention while looking at #3447


kabiroberai (approved):
fwiw this looks good to me!

PR #3479, opened 2024-11-04
Fix race condition in FileStorageKey

mbrandonw:
This fixes a race condition between mutating `@Shared(.fileStorage)` directly and the file system notifying it of changes on disk.


Tests/ComposableArchitectureTests/FileStorageTests.swift
```diff
@@ -251,7 +251,7 @@ final class FileStorageTests: XCTestCase {
 
       try fileStorage.save(Data(), .fileURL)
       scheduler.run()
-      expectNoDifference(users, [])
+      expectNoDifference(users, [.blob])
```
mbrandonw, line 254: Changes made to the file on disk will be ignored if there is an enqueued state change waiting to be saved.

stephencelis (approved):

Issue #3462, opened 2024-10-22
`Shared` subscription issue


ddanilyuk:
### Description

Found some weird behavior with AppStorageKey. The shared `publisher` is triggered after updating some unrelated UserDefaults values.
Here is the XCTest, which is failing but should not.
``` swift
func testSharedSubscription() throws {
    @Dependency(\.defaultAppStorage) var defaults
    var cancellable: (any Cancellable)?
    @Shared(.appStorage("testInt")) var testInt: Int = 0
    var isUpdated = false
    cancellable = $testInt.publisher.sink { _ in
        isUpdated = true
    }
    defaults.set(42, forKey: "randomDefaultsKey")
    XCTAssert(!isUpdated)
    _ = cancellable
}
```

### Checklist

- [X] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Reproducing project

https://github.com/ddanilyuk/SharedSubscriptionIssue/blob/main/TestSharedSubscriptionTests/TestSharedSubscriptionTests.swift

### The Composable Architecture version information

1.15.2

### Xcode version information

16.0

stephencelis:
@ddanilyuk This is the expected behavior as of today, and you'll see similar behavior in SwiftUI's `@AppStorage`. In order to subscribe to updates we use Notification Center because KVO does not work with user defaults keys that contain periods (_e.g._ `"my.key"`), and Notification Center does not allow for granular subscription per key. As such we just always trigger an update when user defaults update, just like `@AppStorage` in SwiftUI.

We are open to improvements here, but we don't consider it a bug, so I'm going to convert this to a discussion.

PR #3487, opened 2024-11-11
Prefer KVO for app storage observation

### LINKED ISSUE #3311: Crash when using @Shared appStorage during certain conditions

stephencelis:
Observing the firehose notification from Notification Center is not only less performant, but it can cause crashes in SwiftUI if a notification is posted during a view body update, which can occur in certain situations.

Instead, we will prefer KVO unless we detect an invalid key, such as one that contains a `.` or is prefixed with an `@`.

Fixes #3311.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@ PLATFORM_WATCHOS = watchOS Simulator,id=$(call udid_for,watchOS,Watch)
 PLATFORM = IOS
 DESTINATION = platform="$(PLATFORM_$(PLATFORM))"
 
+PLATFORM_ID = $(shell echo "$(DESTINATION)" | sed -E "s/.+,id=(.+)/\1/")
+
 SCHEME = ComposableArchitecture
 
 WORKSPACE = ComposableArchitecture.xcworkspace
@@ -36,11 +38,17 @@ endif
 
 TEST_RUNNER_CI = $(CI)
 
-xcodebuild:
+warm-simulator:
+	@test "$(PLATFORM_ID)" != "" \
+		&& xcrun simctl boot $(PLATFORM_ID) \
+		&& open -a Simulator --args -CurrentDeviceUDID $(PLATFORM_ID) \
+		|| exit 0
+
+xcodebuild: warm-simulator
 	$(XCODEBUILD)
 
 # Workaround for debugging Swift Testing tests: https://github.com/cpisciotta/xcbeautify/issues/313
-xcodebuild-raw:
+xcodebuild-raw: warm-simulator
 	$(XCODEBUILD_COMMAND)
 
 build-for-library-evolution:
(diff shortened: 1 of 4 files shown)


Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKey.swift
```diff
@@ -290,42 +290,77 @@ extension AppStorageKey: PersistenceKey {
...
+      store.wrappedValue.addObserver(observer, forKeyPath: key, options: .new, context: nil)
+      removeObserver = { store.wrappedValue.removeObserver(observer, forKeyPath: key) }
     }
     let willEnterForeground: (any NSObjectProtocol)?
     if let willEnterForegroundNotificationName {
       willEnterForeground = NotificationCenter.default.addObserver(
         forName: willEnterForegroundNotificationName,
         object: nil,
-        queue: nil
+        queue: .main
```
stephencelis, line 334: I _believe_ this forces a thread hop, which means we should avoid the crash even on invalid KVO keys, but we could also do an explicit dispatch if we're not confident here.

Sources/ComposableArchitecture/SharedState/PersistenceKey/AppStorageKeyPathKey.swift
```diff
@@ -60,22 +60,6 @@ extension AppStorageKeyPathKey: PersistenceKey, Hashable {
       observer.invalidate()
     }
   }
-
-  private class Observer: NSObject {
```
stephencelis, line 64: This observer wasn't used anymore.

Makefile
```diff
@@ -36,11 +38,17 @@ endif
 
 TEST_RUNNER_CI = $(CI)
 
-xcodebuild:
+warm-simulator:
+	@test "$(PLATFORM_ID)" != "" \
+		&& xcrun simctl boot $(PLATFORM_ID) \
+		&& open -a Simulator --args -CurrentDeviceUDID $(PLATFORM_ID) \
+		|| exit 0
```
stephencelis, line 45: Trying to see if we can reduce CI flakiness by warming the simulators.

denysdanyliukboosters:
@stephencelis Looks like this also fixes [this issue](https://github.com/pointfreeco/swift-composable-architecture/issues/3462).

ddanilyuk:
@stephencelis Looks like this also fixes [this issue](https://github.com/pointfreeco/swift-composable-architecture/issues/3462).

PR #3493, opened 2024-11-14
Add `@ViewBuilder` to `NavigationStack` root view

joshrl:
Adding view builder annotation to `NavigiationStack. init`. This matches the functionality of the built in initializer's root view. See #3483 

Note that I'm not sure if there is a way to add a test for this exactly, but what I did to confirm was to modify `NavigationDemoView` in SwiftUICaseStudies locally (not checked in) like so: 

```
  var body: some View {
    NavigationStack(path: $store.scope(state: \.path, action: \.path)) {
      Text("Fred")
      Form {
        Section { Text(template: readMe) }
      ...
``` 
Screenshot 2024-11-13 at 9 24 19 PM


stephencelis (approved):
Thanks!

PR #3509, opened 2024-11-25
Fix obsolete detail on StackElementID documentation

pyrtsa:
Looks like `StackElementID` [was once backed by a UUID](https://github.com/pointfreeco/swift-composable-architecture/commit/341b09bb4f5eb18a841722599c31ad6f127b3207#diff-41bfac62990a3b8f3c78d97ce4eac2d1a9f94a4a3dab24d797edb0c407d3d021R452), but later on in #1945 the implementation was simplified down to just a wrapped `Int`. This change fixes the confusing documentation accordingly.


mbrandonw (approved):
Good catch @pyrtsa!

PR #3514, opened 2024-11-29
Fix DemandBuffer Data Races

jpsim:
We're still seeing crashes due to what look like data races in `DemandBuffer`. Here, I'm making speculative fixes based on what appears to have helped in https://github.com/pointfreeco/swift-composable-architecture/pull/3447:

In `buffer(value:)`:

* Check if the demand is unlimited and, if so, unlock before calling `subscriber.receive(value)`.
* If the demand is not unlimited, append the value to the buffer and call `flush()` after unlocking.

In `flush(adding:)`:

* Process values in a loop, acquiring and releasing the lock around shared state access.
* Unlock before calling `subscriber.receive(value)` to avoid holding the lock during subscriber calls.
* After sending a value, lock again to update `demandState.requested` with any additional demand returned by the subscriber.
* Ensure that `subscriber.receive(completion:)` is called outside the lock.

Also add a new `DemandBufferTests` test file. I added some tests that I was hoping would fail without the above changes, but they all pass both before and after, even with ThreadSanitizer enabled.

### DIFF
--- a/Sources/ComposableArchitecture/Internal/Create.swift
+++ b/Sources/ComposableArchitecture/Internal/Create.swift
@@ -36,28 +36,25 @@ final class DemandBuffer<S: Subscriber>: @unchecked Sendable {
 
   func buffer(value: S.Input) -> Subscribers.Demand {
     lock.lock()
-    defer { lock.unlock() }
-
     precondition(
-      self.completion == nil, "How could a completed publisher sent values?! Beats me 🤷‍♂️")
+      self.completion == nil, "How could a completed publisher send values?! Beats me 🤷‍♂️")
 
-    switch demandState.requested {
-    case .unlimited:
+    if demandState.requested == .unlimited {
+      lock.unlock()
       return subscriber.receive(value)
-    default:
+    } else {
       buffer.append(value)
+      lock.unlock()
       return flush()
     }
   }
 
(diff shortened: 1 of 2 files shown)


stephencelis:
@jpsim `Create` is more or less vendored from https://github.com/CombineCommunity/CombineExt/blob/main/Sources/Operators/Create.swift

Maybe there are some fixes upstream that we could bring in? And/or maybe you should share these fixes back with CombineExt?

/cc @freak4pc

mbrandonw:
Hey @jpsim, we are definitely interested in fixing any race conditions in the `DemandBuffer`. Have you uncovered anything new since opening this? It does seem that at least one of the tests fails even with this change.

jpsim:
> we are definitely interested in fixing any race conditions in the `DemandBuffer`. Have you uncovered anything new since opening this?

Yeah, just talking to Kabir about this more internally, and even though we see DemandBuffer in the backtrace of the crashes we're trying to fix, we're not convinced that the issue is there or elsewhere.

> It does seem that at least one of the tests fails even with this change.

I agree, I'm not confident in this change. I'll just close this one.

mbrandonw:
Ok thanks for the update, and feel free to hit us up on Slack if you want to discuss further!

PR #3521, opened 2024-12-04
Record state access when constructing _StoreCollection.

mbrandonw:
This fixes an issue brought up in https://github.com/pointfreeco/swift-composable-architecture/discussions/3518. When constructing a `_StoreCollection`, like for a `ForEach`, we should record a state access with the registrar. Currently we use `withState`, which silently accesses state, and so one will not get perception warnings if they forget to use `WithPerceptionTracking`.

### DIFF
--- a/Sources/ComposableArchitecture/Observation/IdentifiedArray+Observation.swift
+++ b/Sources/ComposableArchitecture/Observation/IdentifiedArray+Observation.swift
@@ -100,6 +100,7 @@ public struct _StoreCollection<ID: Hashable & Sendable, State, Action>: RandomAc
   #endif
   fileprivate init(_ store: Store<IdentifiedArray<ID, State>, IdentifiedAction<ID, Action>>) {
     self.store = store
+    store._$observationRegistrar.access(store, keyPath: \.currentState)
     self.data = store.withState { $0 }
   }
 


Sources/ComposableArchitecture/Observation/IdentifiedArray+Observation.swift
```diff
@@ -100,6 +100,7 @@ public struct _StoreCollection<ID: Hashable & Sendable, State, Action>: RandomAc
   #endif
   fileprivate init(_ store: Store<IdentifiedArray<ID, State>, IdentifiedAction<ID, Action>>) {
     self.store = store
+    store._$observationRegistrar.access(store, keyPath: \.currentState)
     self.data = store.withState { $0 }
```
mbrandonw, line 104: This definitely fixes the problem, but I am also wondering if we should be dropping the `_withoutPerceptionChecking` from `withState` as an alternative fix. We introduced `withState` back in the view store days to give people easy access to state when observation wasn't needed. But now with implicit observation everywhere, should we be tracking those accesses with the registrar?

stephencelis: The only reason not to do this that I can think of is that it's a big behavior change for folks not using `@ObservableState`, and they could get really inefficient SwiftUI observation wherever they decide to use `withState`. I think 2.0 will give us a nice opportunity here to opt into `@ObservableState` by default and avoid these questions entirely.

stephencelis (approved):

PR #3523, opened 2024-12-07
Shared ObservableState

todd-weidler:


stephencelis:
@todd-weidler Thanks for sharing a PR. We haven't had a ton of time to think about this yet, but definitely continue discussion over on the Sharing repo if you have any more thoughts or discoveries! Going to close this for now, though.

PR #3540, opened 2024-12-22
Fix NavigationStackController docs

takehilo:


mbrandonw (approved):
Good find, thanks @takehilo!

PR #3548, opened 2025-01-09
Update format action to use Xcode 16

bugix:
This should update the format.yml workflow to use the now bundled swift-format.


stephencelis (approved):
Thanks!

PR #3585, opened 2025-02-05
SyncUps speech client preview

fonkadelic:
I've noticed that there is some unnecessary state in the preview implementation of the `SpeechClient` which is used in the _SyncUps_ example.

However in the _SpeechRecognition_ example, which also contains a `SpeechClient`, the `isRecording` state is used. But i'm wondering why there is an explicit `finishTask` endpoint instead of making use of cancellation.

### DIFF
--- a/Examples/SyncUps/SyncUps/Dependencies/SpeechRecognizer.swift
+++ b/Examples/SyncUps/SyncUps/Dependencies/SpeechRecognizer.swift
@@ -48,7 +48,7 @@ extension SpeechClient: DependencyKey {
               officia deserunt mollit anim id est laborum.
               """
             var text = ""
-            while true {
+            while !finalText.isEmpty {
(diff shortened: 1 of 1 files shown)


Examples/SyncUps/SyncUps/Dependencies/SpeechRecognizer.swift
```diff
@@ -50,7 +48,7 @@ extension SpeechClient: DependencyKey {
               officia deserunt mollit anim id est laborum.
               """
             var text = ""
-            while isRecording.value {
+            while true {
```
stephencelis, line 51:
```suggestion
            while !finalText.isEmpty {
```

stephencelis (approved):
Good catch, thanks! I think we used the mutability of this at one point but can't find it.

PR #3602, opened 2025-02-19
Add assertWithYield helper to TestStore for reliable async testing

OguzYuuksel:
This PR introduces a new asynchronous overload on TestStore named  `assertWithYield(_:)`. The new helper awaits a configurable number of yields (using  `Task.megaYield`) before performing the usual state assertion. This helps alleviate race conditions in tests where asynchronous effects (e.g. from async streams) might not have had enough time to update the state before the assertion is made.

#### Changes:

-   **New API:**  
    Added  `assertWithYield(_:)`  to TestStore. Its purpose is to allow the concurrency runtime extra time to process in-flight asynchronous work before asserting on the store’s state.
-   **Documentation:**  
    Updated the doc comment on the new helper to describe its use cases, parameters, and behavior.
-   **Tests:**  
    Added a new test case (`testStreamDependencyOnReducerWithSharedState`) demonstrating the race condition issue and showing that  `assertWithYield(_:)`  reliably asserts the expected state.

**Rationale:**  
In certain scenarios (such as when testing effects that subscribe to async streams) the state update might not occur before a test assertion is run, leading to flaky tests. This helper gives the runtime a chance to process pending effects (without forcing users to sprinkle manual  `Task.yield()`  calls) thereby making tests more robust.

Please let me know if any additional changes or refinements are needed also feel free to improve documentation and method name, I couldn't overload assert method due to the issue mentioned in https://forums.swift.org/t/async-feedback-overloads-that-differ-only-in-async/49573/15.

### DIFF
--- a/Sources/ComposableArchitecture/TestStore.swift
+++ b/Sources/ComposableArchitecture/TestStore.swift
@@ -1100,6 +1100,63 @@ extension TestStore {
     }
   }
 
+  /// Asserts against the current state of the store after allowing pending asynchronous
+  /// effects additional time to yield.
+  ///
+  /// Use this helper when you experience race conditions in your async tests—when the
+  /// asynchronous effects in your feature haven't had enough time to update the store's
+  /// state before you assert against it. This overload awaits several yields (by default, 20)
+  /// to give the concurrency runtime an opportunity to process in-flight effects.
+  ///
+  /// For example, if your reducer subscribes to an async stream that updates shared state,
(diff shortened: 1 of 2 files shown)


mbrandonw:
Hi @OguzYuuksel, we rather not add more mega yields, and feel that if yields are necessary to make async code testable in Swift, it is best for that to be upfront and obvious in the test code. At the end of the day this is a deficiency in Swift that hopefully gets better over time.

PR #3603, opened 2025-02-19
Add `traitCollection.push(state:)`

stephencelis:
The UIKit equivalent of `NavigationLink(state:)`.


Sources/ComposableArchitecture/UIKit/NavigationStackControllerUIKit.swift
```diff
@@ -69,4 +68,19 @@
...
+  extension UIPushAction: Sendable {
+    /// Pushes an element of ``StackState`` onto the current navigation stack.
+    ///
+    /// This is the UIKit equivalent of
+    /// ``SwiftUI/NavigationLink/init(state:label:fileID:filePath:line:column:)``.
+    ///
+    /// - Parameter state: An element of stack state.
+    public func callAsFunction<Element: Hashable>(state: Element) {
+      @Dependency(\.stackElementID) var stackElementID
+      self(value: StackState.Component(id: stackElementID(), element: state))
```
stephencelis, line 83: This reports an issue under the hood, so we should probably thread through the source context, but we need to make changes to Swift Navigation and cut a release to do that.

mbrandonw (approved):

mbrandonw (approved):

PR #3613, opened 2025-02-27
Cancel effects when root store deallocates

mbrandonw:
We have wanted to do this for a long time, but our last attempt (#2643) needed to be reverted (#2648) because it meant that uncached stores (i.e. stores created using deprecated closure-based `scope`s) would accidentally cancel their effects prematurely since those kinds of stores are created, destroyed and recreated often.

However, in #2664 we separated the concept of the "root store" that powers a composition of features from a "child store" that can be derived and handed off to child features. The root store is what runs effects and holds onto cancellables, and now it is completely legitimate to tear down effects when the root store is deallocated.

### DIFF
--- a/ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -15,8 +15,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/pointfreeco/swift-case-paths",
       "state" : {
-        "revision" : "bc92c4b27f9a84bfb498cdbfdf35d5a357e9161f",
-        "version" : "1.5.6"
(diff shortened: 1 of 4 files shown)


mbrandonw:
This PR may fix #2508

stephencelis (approved):

dk-jlew:
Fantastic, thank you!

dk-jlew:
Can you tell me if this also fixes the situation where multiple stores of the same type, operating in disconnected contexts, will interfere with each others' effects?  For instance, I have:
`StoreA` (root) -> `StoreB` (child)
...and separately (not connected by a common root) the same thing again:
`StoreA` (root) -> `StoreB` (child)

Formerly, if one instance StoreB was to be destroyed, it would potentially kill the effects of the unrelated `StoreB` instance, due to shared storage and identical KeyPaths.

PR #3618, opened 2025-03-06
[Examples] declare store as static to avoid interference with Xcode previews

meltsplit:
Hi!
I am studying the example apps by referring to the tutorial.
I standardized the App.swift file.

**Changes made:**
- Refactored: Declared the store as static to prevent interference with Xcode previews.
- Added: Added an if isTesting condition to prevent interference between the app and the test.

I hope this change will help developers understand the testing and preview mechanisms better.

### DIFF
--- a/Examples/Search/Search/SearchApp.swift
+++ b/Examples/Search/Search/SearchApp.swift
@@ -3,14 +3,20 @@ import SwiftUI
 
 @main
 struct SearchApp: App {
+  
+  @MainActor
+  static let store = Store(initialState: Search.State()) {
+    Search()
+      ._printChanges()
+  }
+  
   var body: some Scene {
     WindowGroup {
-      SearchView(
-        store: Store(initialState: Search.State()) {
-          Search()
-            ._printChanges()
-        }
-      )
+      if isTesting {
+        EmptyView()
+      } else {
+        SearchView(store: Self.store)
+      }
     }
   }
 }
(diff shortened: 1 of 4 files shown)


Examples/Search/Search/SearchApp.swift
```diff
@@ -3,14 +3,20 @@ import SwiftUI
 
 @main
 struct SearchApp: App {
+  
+  @MainActor
```
stephencelis, line 7: I don't believe this is needed since the `App` protocol is `@MainActor`.

```suggestion
```

Examples/Search/Search/SearchApp.swift
```diff
@@ -3,14 +3,20 @@ import SwiftUI
...
-        store: Store(initialState: Search.State()) {
-          Search()
-            ._printChanges()
-        }
-      )
+      if isTesting {
+        EmptyView()
+      } else {
+        SearchView(store: Self.store)
+      }
```
stephencelis, line 19: Since we're not doing UI tests I think we can simplify:

```suggestion
      SearchView(store: Self.store)
```

stephencelis (commented):
@meltsplit Thanks for the PR! I think we'd definitely take the `static` change but I've made a few suggestions in one of the examples. Want to apply them to the rest? Then we'll be happy to merge your contributions :smile:

stephencelis:
Thanks!

meltsplit:
@stephencelis 
I fixed the method name that caused CI failure.

PR #3630, opened 2025-03-20
Added support of ViewAction for preprocessor macros

jasin755:
# Fix: Ensure `@ViewAction(for:)` Macro Compatibility with Preprocessor Directives  

## Summary  
This PR updates the `@ViewAction(for:)` macro to be compatible with preprocessor directives, such as `#if canImport(AppKit)`. Previously, the macro did not handle conditional compilation correctly, which could lead to issues when using platform-specific attributes.  

## Changes  
- Improved handling of `#if canImport(AppKit)` to support platform-dependent attributes.  
- Ensured that attributes like `@Perception.Bindable` and `@Bindable` can be conditionally applied based on platform availability.  
- Refactored macro expansion logic to correctly resolve conditional code paths.  

## Example Usage  
```swift  
#if canImport(AppKit)  
@Perception.Bindable  
var store: StoreOf<MyReducer>  
#else  
@Bindable  
var store: StoreOf<MyReducer>  
#endif  
```

## Impact  
This change improves compatibility when using `@ViewAction(for:)` in cross-platform projects, ensuring that platform-specific attributes are correctly recognized and applied.  

## Testing  
- Verified compilation on different platforms (iOS/macOS).  
- Ensured that the correct attribute is applied depending on platform availability.

### DIFF
--- a/Sources/ComposableArchitectureMacros/ViewActionMacro.swift
+++ b/Sources/ComposableArchitectureMacros/ViewActionMacro.swift
@@ -121,10 +121,24 @@ extension SyntaxProtocol {
   }
 }
 
-extension DeclGroupSyntax {
-  fileprivate var hasStoreVariable: Bool {
-    self.memberBlock.members.contains(where: { member in
-      if let variableDecl = member.decl.as(VariableDeclSyntax.self),
+private extension DeclGroupSyntax {
+  var hasStoreVariable: Bool {
+    let members = memberBlock.members
+    return members.contains(where: { member in
+
+      let variableDecl: VariableDeclSyntax? =
+        if
+          let ifConfigDecl = member.decl.as(IfConfigDeclSyntax.self),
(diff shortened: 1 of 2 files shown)


stephencelis:
@jasin755 Thanks for taking the time to look into this and implement! I think one thing that still needs to be accounted for is nested `#if`s, so some kind of recursion is required.

There is a `SwiftIfConfig` module in SwiftSyntax that provides helpers for this so that you don't need to write manual recursive code: https://github.com/swiftlang/swift-syntax/blob/main/Sources/SwiftIfConfig/SwiftIfConfig.docc/SwiftIfConfig.md

We haven't explored it yet, but do you think you can take a crack at support as well as another test with more complex nested `#if` statements?

mbrandonw:
Hi @jasin755, do you plan on still working on this? If not, what do you think about closing it out until you are ready?

PR #3638, opened 2025-03-27
RootCore crash

arnauddorgans:
Fix RootCore crash when released with effects outside of the MainActor

Because RootCore is a MainActor, every closure is expected to be called on MainActor, so when RootCore deinits outside of the MainActor the AnyCancellable closure is called and makes the app crash

Screenshot 2025-03-27 at 16 27 30

Note: The crash also occurs on the latest release in RootStore

### DIFF
--- a/Sources/ComposableArchitecture/Core.swift
+++ b/Sources/ComposableArchitecture/Core.swift
@@ -128,9 +128,7 @@ final class RootCore<Root: Reducer>: Core {
           }
           boxedTask.wrappedValue = task
           tasks.withValue { $0.append(task) }
-          self.effectCancellables[uuid] = AnyCancellable {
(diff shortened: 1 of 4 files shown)


mbrandonw:
Hi @arnauddorgans, great catch!

There is an alternative way to do this too. The problem stems from `AnyCancellable` being pre-concurrency and having an escaping closure. Swift will infer `@MainActor` for that escaping closure even though it is not enforced by the compiler. By marking that closure as `@Sendable` we can make it so that Swift will check its isolation correctly.

I have made that change in the two `AnyCancellable {…}`s in Core.swift, as well as a few other spots. Will merge soon!

stephencelis (approved):

PR #3640, opened 2025-04-01
Enable binding workaround in release builds

lukeredpath:
The store will suppress warnings for invalid actions emitted by bindings to workaround a SwiftUI issue of bindings firing on dismissal when focussed, but only in debug builds.

We capture issues in release builds in Sentry so this can still be noisy and there doesn't seem to be any need to restrict this to just debug builds.


stephencelis (approved):
Thanks!

PR #3661, opened 2025-04-18
Adjust documentation for binding action

johankool:
Afaict the $ sign shouldn't used to match on a binding as was done in this documentation snippet.


mbrandonw (approved):
Hi @johankool, yeah the `$` is not needed when using `@ObservableState`, but is needed when using view stores. But view stores are deprecated and so it is best to show off the modern APIs. Thanks!

PR #3681, opened 2025-05-10
feat: add onChange overload that ignores old and new values

MojtabaHs:
In my experience, most of the time, the old and new values in `onChange` aren’t needed.
Following [SwiftUI’s design for onChange](https://developer.apple.com/documentation/swiftui/view/onchange(of:initial:_:)-8wgw9), I’ve added a simplified overload that ignores both values.

Internally, this new overload calls [the original function](https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/reduce/onchange(of:_:)) to maintain a single source of truth.

### DIFF
--- a/Sources/ComposableArchitecture/Reducer/Reducers/OnChange.swift
+++ b/Sources/ComposableArchitecture/Reducer/Reducers/OnChange.swift
@@ -116,6 +116,60 @@ extension Reducer {
   ) -> _OnChangeReducer<Self, V, R> {
     _OnChangeReducer(base: self, toValue: toValue, isDuplicate: ==, reducer: reducer)
   }
+
+  /// Adds a reducer to run when this reducer changes the given value in state.
+  ///
+  /// > Important: The `onChange` operator is only capable of detecting changes made by the reducer
+  /// > it is directly attached to. It does not observe changes that are made from other actions,
+  /// > such as parent actions.
+  ///
+  /// Use this operator to trigger additional logic when a value changes, like when a
+  /// ``BindingReducer`` makes a deeper change to a struct held in ``BindingState``.
+  ///
+  /// ```swift
+  /// @Reducer
+  /// struct Settings {
+  ///   struct State {
+  ///     @BindingState var userSettings: UserSettings
+  ///     // ...
+  ///   }
+  ///
+  ///   enum Action: BindableAction {
+  ///     case binding(BindingAction<State>)
+  ///     // ...
+  ///   }
+  ///
+  ///   var body: some Reducer<State, Action> {
+  ///     BindingReducer()
+  ///       .onChange(of: \.userSettings.isHapticFeedbackEnabled) {
+  ///         Reduce { state, action in
+  ///           .run { send in
+  ///             // Persist new value...
+  ///           }
+  ///         }
+  ///       }
+  ///   }
+  /// }
+  /// ```
+  ///
+  /// When the value changes, the new version of the closure will be called, so any captured values
+  /// will have their values from the time that the observed value has its new value.
+  ///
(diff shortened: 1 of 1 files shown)


stephencelis:
Hi @MojtabaHs, thanks for taking the time to contribute! I think we're down to accept this additional API, but could you first provide a motivating example or two where neither the old nor new value are needed?

MojtabaHs:
Hi @stephencelis, this usually needed when the reducer logic is independent of the changing value or when the action is self-contained. Here are a couple of examples to show what I mean:

- Case 1: Triggering Side Effects, Not State Changes
```
.onChange(of: \.level) { _, _ in
    Reduce { state, action in
        .run { send in
            try await leaderboardClient.triggerUpdate()
        }
    }
}
```
In this case, gameCenter (or leaderboardClient) already knows the current score, so we’re just using this as a trigger. We don’t need the new level value itself.

---

- Case 2: Derived Logic from Multiple Properties
```
BindingReducer()
    .onChange(of: \.hours) { _, _ in normalizeTheDuration() }
    .onChange(of: \.minutes) { _, _ in normalizeTheDuration() }
```
Here, `normalizeTheDuration()` uses several properties together. We only care that something changed, not what it changed to.

---

Sure, there are other ways to do this (like a computed property or combining the triggers), but I think this API fits naturally with SwiftUI’s style, and clean.

It’s a helpful way to handle “run logic on change” scenarios without overcomplicating the reducer. Definitely feels like something that belongs in the core.

stephencelis:
And `normalizeTheDuration()` returns a reducer? Can you expand that part of the code so we can better evaluate?

I do think this API is mostly harmless to add, but one concern is that this overload may affect reducer builders using the existing `onChange` operator. Overloads can strain the compiler, especially in result builder contexts, and so merging this does have the potential to break existing code by creating an expression that takes too long to type-check. One option is to merge, release, and revert if there's a regression. It'd be unfortunate to have to do a breaking change via the revert, though, so we're going to think on things just a little longer.

MojtabaHs:
The extracted logic would be something like:
```
func normalizeTheDuration() -> Reduce<State, Action> {
    Reduce { state, action in
        let needsNormalization = state.hours * 60 + state.minutes < minimumMinutes
        if needsNormalization {
            state.minutes = minimumMinutes % 60
            state.hours = Int(Double(minimumMinutes) / 60)
        }
        return .none
    }
}
```
or:
```
.onChange(of: \.hours) { _, _ in  Reduce { state, action in normalizeTheDuration() } }
.onChange(of: \.minutes) { _, _ in  Reduce { state, action in normalizeTheDuration() } }
```

---
We can implement this PR with a simple extension since it literally extends the original behavior:
```
extension Reducer {
    @inlinable
    public func onChange<V: Equatable, R: Reducer>(
        of toValue: @escaping (State) -> V,
        @ReducerBuilder<State, Action> _ reducer: @escaping () -> R
    ) -> _OnChangeReducer<Self, V, R> {
        onChange(of: toValue) { _, _ in reducer() }
    }
}
``` 
I didn't see any issues for auto complete or a warning on `-warn-long-expression-type-checking=100`. Xcode happily shows the auto complete suggestion as:
Screenshot 2025-05-14 at 1 43 00 AM

The environment is a project with multi-module (about 10 SPM targets) architecture and about 30 TCA features.
Please let me know if there is anything I can do to have a reliable benchmark or something.

stephencelis:
@MojtabaHs Sorry for sitting on this so long. We've been hard at work on the next major version of TCA, and `onChange` as defined today is actually not quite right. We just opened #3875, which introduces a streamlined operator that will be supported in 2.0, while we cannot support the current `-> Reducer` shape there. Because of this known migration, I think we can close out this PR, but I hope you'll take 2.0 for a spin when it's available and provide feedback!

PR #3686, opened 2025-05-15
Extend store scoping for custom navigation solutions

sdkdimon:
Extend navigation scoping to support custom navigation solutions i.e. [TCACoordinators.](https://github.com/johnpatrickmorgan/TCACoordinators.git) Discussion: #3685

### DIFF
--- a/Sources/ComposableArchitecture/Observation/NavigationStack+Observation.swift
+++ b/Sources/ComposableArchitecture/Observation/NavigationStack+Observation.swift
@@ -232,25 +232,39 @@ extension Store {
         ],
       action: \.[id: component.id]
     )
+    return scope(
+      cachedState: component.element,
+      scopeId: id,
+      stateKeyPath:
+        \.[
+          id: component.id,
+          fileID: _HashableStaticString(rawValue: fileID),
+          filePath: _HashableStaticString(rawValue: filePath),
+          line: line,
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hi @sdkdimon, these operations were never meant to be public, and it was always dangerous to use `@_spi` in 3rd party libraries. I think you would need to provide a much better use case for why we should expose this in `@_spi`. 

Looking through the code for TCA coordinators I see a lot of index-based subscripting happening in the internals, which means that features built with those tools are susceptible to lots of problems (effect actions being delivered to the wrong feature or actions not being silently dropped). These problems are exactly why `IdentifiedArray` and `StackState` were created.

Is there a reason the library cannot use the safer `IdentifiedArray` and `StackState` tools, and then the public APIs would work for you just fine.

sdkdimon:
@mbrandonw i think that the reason why TCACoordinators is using indexed-based subscripting is that it based on another one navigation library FlowStacks that brings coordinator navigation pattern to the SwiftUI apps. I agree that it is not safe to use index-based subscripting. The concept of TCACoordinators is to hold whole navigation stack (push, sheet, cover) in the some kind of collection. I'll try to setup a bare minimal playground project to see if it possible to use StackState for this purpose as it designed right for navigation.

PR #3688, opened 2025-05-17
Fix for "Your first feature" tutorial Store type mismatch in #Preview

jrhey:
The tutorial uses `Store(...)` to initialize the `CounterFeature` store, which leads to a type mismatch error in the latest versions of TCA.

Screenshot 2025-05-17 at 19 15 02

This commit updates the tutorial code to use the explicit generic form `StoreOf<CounterFeature>(...)`


mbrandonw:
Hi @jrhey, are you sure you don't have another type in your project named `Store`? Or perhaps something else is off in your project, because the stated way in the tutorial is the way to initialize a store. You can see we have hundreds of examples of this throughout the repo that compile just fine:

https://github.com/search?q=repo%3Apointfreeco%2Fswift-composable-architecture+%22Store%28initialState%22&type=code

If you can share a repro of this problem we will take a closer look.

jrhey:
@mbrandonw ah my mistake! You're right, I did have a `StoreKit` Store in this project already. I'll close this PR 👍

PR #3694, opened 2025-05-27
Make our use of SwiftSyntax APIs more resilient to precompiled SwiftSyntax

mbrandonw:


Sources/ComposableArchitectureMacros/Availability.swift
```diff
@@ -86,6 +86,8 @@ extension AttributeListSyntax.Element {
       if let availability = ifConfig.availability {
         return .ifConfigDecl(availability)
       }
+    @unknown default:
+      return nil
```
mbrandonw, line 90: We could also `fatalError` here?

PR #3701, opened 2025-06-05
Remove outdated thread-safety docs

stephencelis:
The Store is now `@MainActor`, so I think we can remove these warnings.


mbrandonw (approved):

stephencelis:
@Matejkob Any idea why the macro job would start to fail but everything else is passing?

Matejkob:
@stephencelis @mbrandonw Everything should be good and passing now; sorry for the inconvenience.

stephencelis:
@Matejkob Worked great, thanks for looking into this!

PR #3707, opened 2025-06-13
Identifiable

RomanPodymov:
Hello.
Thank you for swift-composable-architecture.
For some Swift 5 projects `State` or `Action` should be `Identifiable` (please check `LogInScreen` [here](https://github.com/johnpatrickmorgan/TCACoordinators/blob/0.12.1/TCACoordinatorsExample/TCACoordinatorsExample/Game/LogInCoordinator.swift)), that's why I added it.

### DIFF
--- a/Sources/ComposableArchitecture/Macros.swift
+++ b/Sources/ComposableArchitecture/Macros.swift
@@ -83,6 +83,9 @@ extension _SynthesizedConformance {
   /// Extends the `State` or `Action` types that ``Reducer()`` creates with the `Sendable`
   /// protocol.
   public static let sendable = Self()
+  /// Extends the `State` or `Action` types that ``Reducer()`` creates with the `Identifiable`
+  /// protocol.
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hi @RomanPodymov, it is not appropriate to add `Identifiable` to this list because `Identifiable` cannot be automatically synthesized like the other protocols (`Equatable`, `Hashable`, `Codable`, etc.).

Further, these options were only provided to work around Swift bugs that prevented one from extending macro generated types. That has long been fixed in Swift, and in fact these tools have been [soft-deprecated](https://github.com/pointfreeco/swift-composable-architecture/blob/8bf72ef61f37ad6e7ffa0a80beb4e55e391cd572/Sources/ComposableArchitecture/Macros.swift#L41-L48) in our library. 

I am going to close this PR for now, but feel free to start a [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions) if you have questions.

PR #3710, opened 2025-06-15
Updating swift-navigation to 2.3.1

ignotusverum:
## Context

Updating `swift-navigation` to [2.3.1](https://github.com/pointfreeco/swift-navigation/releases/tag/2.3.1) to avoid error on Xcode 26.


stephencelis:
@ignotusverum Thanks for taking the time to investigate and open a PR!

While it would be nice to avoid this error for our users, in general we try to avoid forcing upgrades on folks unless it's critical to do so, and instead we prefer users have the flexibility to apply bumps locally as they see fit (_e.g._ right-clicking the dependency in Xcode and updating it that way). Given that I think we probably do not want to take this patch at this time, but we're down to reconsider in the future.

PR #3752, opened 2025-08-15
Updates the docs for Explicit shared state.

connor-ricks:
The docs don't currently showcase how to assign the default value for the parent.

### DIFF
--- a/Sources/ComposableArchitecture/Documentation.docc/Articles/SharingState.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Articles/SharingState.md
@@ -87,7 +87,7 @@ the feature's state:
 struct ParentFeature {
   @ObservableState
   struct State {
-    @Shared var count: Int
+    @Shared(value: 0) var count: Int
     // Other properties
   }
   // ...


stephencelis:
@connor-ricks Thanks for taking the time. Our docs here are actually what we want to show, though. Typically (but not always) we find that it should be the parent's responsibility to initialize this state.

We'd be down to _add_ docs to explain both ways, though. It's similar to `@State` in SwiftUI, where when a default is omitted the parent must pass it in, but a feature can also initialize and manage its own local state.

Are you interested in making those changes?

connor-ricks:
> @connor-ricks Thanks for taking the time. Our docs here are actually what we want to show, though. Typically (but not always) we find that it should be the parent's responsibility to initialize this state.
> 
> We'd be down to _add_ docs to explain both ways, though. It's similar to `@State` in SwiftUI, where when a default is omitted the parent must pass it in, but a feature can also initialize and manage its own local state.
> 
> Are you interested in making those changes?

Thanks @stephencelis. I appreciate that context! I think I agree that the docs as stands are probably okay and will help to avoid confusion to others. I'd be concerned that the additional docs my add swirl to first time users of Shared.

PR #3759, opened 2025-08-25
Make `_PrintChangesReducer` conditionally Sendable if `Base` is Sendable

jpsim:


jpsim:
I don't quite understand why but ~as of https://github.com/pointfreeco/swift-composable-architecture/pull/3744~ we need this to satisfy the Swift compiler in Swift 6 mode 🤔

stephencelis:
@jpsim Thanks! Do you happen to have an example of the warning/error you saw? Just want to better understand the issue.

jpsim:
Sure, here's an example of the warning:

```
MyView.swift:266:16: error: type '_PrintChangesReducer<MyReducer>' does not conform to the 'Sendable' protocol
264 | #Preview {
265 |     MyView(
266 |         store: Store(
    |                `- error: type '_PrintChangesReducer<MyReducer>' does not conform to the 'Sendable' protocol
267 |             initialState: .init(

Sources/ComposableArchitecture/Reducer/Reducers/DebugReducer.swift:71:15: note: generic struct '_PrintChangesReducer' does not conform to the 'Sendable' protocol
 69 | }
 70 |
 71 | public struct _PrintChangesReducer<Base: Reducer>: Reducer {
    |               `- note: generic struct '_PrintChangesReducer' does not conform to the 'Sendable' protocol
 72 |   @usableFromInline
 73 |   let base: Base
```

I don't have a full repro yet, but also this change is low risk anyway so I figured it might not be necessary to provide one.

jpsim:
A minimal repro case:

```swift
import ComposableArchitecture

@Reducer
struct TestReducer {
    struct State: Equatable {
        var count = 0
    }

    enum Action {
        case increment
    }

        Reduce { state, action in
            switch action {
            case .increment:
                state.count += 1
                return .none
            }
        }
    }
}

struct SendableWrapper: Sendable {
    static func requireSendable<T: Sendable>(_ value: T) -> T { value }

    static func testSendable() {
        let reducer = TestReducer()._printChanges()
        _ = requireSendable(reducer)
    }
}
```

Before this PR, this fails to build in Swift 6 language mode with the following error:

```
28 |     static func testSendable() {
29 |         let reducer = TestReducer()._printChanges()
30 |         _ = requireSendable(reducer)  // This should trigger the Sendable error
   |             `- error: type '_PrintChangesReducer<TestReducer>' does not conform to the 'Sendable' protocol
31 |     }
32 | }
```

And it builds cleanly after the changes in this PR. It's a contrived example, but demonstrates the issue.

PR #3764, opened 2025-08-29
Skip perception checks in store publishers

stephencelis:


stephencelis:
Closing in favor of including the fix in #3763.

Issue #3771, opened 2025-09-09
Skipped observability notifications since introduction of `shouldNotifyObservers`


jpsim:
### Description

I don't yet have a repro case, but for our iOS app, we've noticed that as of https://github.com/pointfreeco/swift-composable-architecture/pull/3751 we have parts of our app that aren't notified when state changes. Reverting af60ed2d5157a512165613b82b08690ff5235a2a fixes the issue for us.

I will try to get more information to make this bug report more actionable, but in the meantime I wanted to at least flag this regression and give an opportunity for other users of TCA to weigh in if they've seen the same thing in their projects.

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [x] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### The Composable Architecture version information

1.22.2

### Destination operating system

iOS 18.5

### Xcode version information

Version 16.4 (16F6)

### Swift Compiler version information

```shell
swift-driver version: 1.120.5 Apple Swift version 6.1.2 (swiftlang-6.1.2.1.2 clang-1700.0.13.5)
Target: arm64-apple-macosx15.0
```

stephencelis:
@jpsim The `shouldNotifyObservers` affects `Equatable` types and classes. Do you think you might have a custom equatable conformance interfering with observation?

mbrandonw:
Even though #3774 was closed, I am curious to hear from @jpsim when you have a moment. Is there a chance the problem you are seeing is related to reference types with custom `Equatable` conformances?

jpsim:
> Do you think you might have a custom equatable conformance interfering with observation?
> Is there a chance the problem you are seeing is related to reference types with custom `Equatable` conformances?

It's quite possible! The property with skipped observation updates was an `IdentifiedArray` from the [swift-identified-collections](https://github.com/pointfreeco/swift-identified-collections) project, which does indeed have a custom `Equatable` conformance: https://github.com/pointfreeco/swift-identified-collections/blob/main/Sources/IdentifiedCollections/IdentifiedArray/IdentifiedArray%2BEquatable.swift

jpsim:
Here's a complete repro case:

```swift
import ComposableArchitecture
import SwiftUI

[...]

PR #3774, opened 2025-09-14
Fix skipped observations for Equatable class and Optional properties

jeonghi:
### 🐛 Bug
ObservableState macro observation issues
- Fixes skipped observations for Equatable class properties and optional references 
- Updates shouldNotifyObservers to consider identity changes (===) so views update when references change even if == remains true

This ensures that SwiftUI views properly update when class instances are replaced with new instances that have the same value but different identity, which is crucial for proper observation behavior in TCA.

### 🔧 Changes
ObservableState macro updates
Equatable & AnyObject: 
- notify when identity or equality changes (lhs !== rhs || lhs != rhs)

Add optional overloads of shouldNotifyObservers:
- `<T>(_ lhs: T?, _ rhs: T?) -> Bool`
- `<T: Equatable>(_ lhs: T?, _ rhs: T?) -> Bool`
- `<T: AnyObject>(_ lhs: T?, _ rhs: T?) -> Bool`
- `<T: Equatable & AnyObject>(_ lhs: T?, _ rhs: T?) -> Bool`

### 📝 Related
#3771

### DIFF
--- a/Sources/ComposableArchitectureMacros/ObservableStateMacro.swift
+++ b/Sources/ComposableArchitectureMacros/ObservableStateMacro.swift
@@ -113,7 +113,66 @@ public struct ObservableStateMacro {
     let memberGeneric = context.makeUniqueName("Member")
     return
       """
-      private nonisolated func shouldNotifyObservers<\(memberGeneric): Equatable & AnyObject>(_ lhs: \(memberGeneric), _ rhs: \(memberGeneric)) -> Bool { lhs != rhs }
+      private nonisolated func shouldNotifyObservers<\(memberGeneric): Equatable & AnyObject>(_ lhs: \(memberGeneric), _ rhs: \(memberGeneric)) -> Bool { lhs !== rhs || lhs != rhs }
+      """
+  }
+
+  // Optional variants
+  static func shouldNotifyObserversOptionalNonEquatableFunction(
+    _ perceptibleType: TokenSyntax, context: some MacroExpansionContext
+  ) -> DeclSyntax {
+    let memberGeneric = context.makeUniqueName("Member")
+    return
+      """
+       private nonisolated func shouldNotifyObservers<\(memberGeneric)>(_ lhs: \(memberGeneric)?, _ rhs: \(memberGeneric)?) -> Bool {
+         switch (lhs, rhs) {
+         case (.none, .none): return false
+         case (.some, .some): return true
+         default: return true
+         }
+       }
+      """
+  }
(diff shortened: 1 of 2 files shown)


mbrandonw:
Hi @jeonghi, thanks for taking the time to look into this and open a PR.

However, I'm a little confused how this PR can be said to fix #3771 when a reproduction has not yet been given for that issue. Also, some of the `shouldNotifyObservers` are questionable in my opinion:

>   * `<T: Equatable>(_ lhs: T?, _ rhs: T?) -> Bool`

This shouldn't be needed since optionals are already equatable when the underlying wrapped value is equatable.

>   * `<T>(_ lhs: T?, _ rhs: T?) -> Bool`

Why is this needed? This seems to only change the default behavior by special casing `state.value = nil` when `state.value` is already `nil` to not notify observers, but is that actually important in any meaningful way?

> * `Equatable & AnyObject`: notify when identity or equality changes (**lhs !== rhs** || lhs != rhs).
>   
>   * `<T: AnyObject>(_ lhs: T?, _ rhs: T?) -> Bool`
>   * `<T: Equatable & AnyObject>(_ lhs: T?, _ rhs: T?) -> Bool `

These additional overloads could be considered, but honestly it's only covering up for a much bigger problem: reference types should not be given `Equatable` conformances that use their underlying data for `==`. To me it is completely reasonable that this line:

```swift
state.box = Box(payload: state.box.payload)
```

…does not trigger a view update because the data did not change. And in the case of `Box`, which does not have any behavior in it, no bug is introduced by doing this. But if your reference has behavior, then you should only be using the object's identity for `==` and not its data.

---

@jpsim I'm curious if these changes would fix the issue you reported, and if so then does that mean you have classes that have behavior and for which you have defined `==` using the data in the class. If so, I believe the better fix would be to change that to use object identity instead.

jeonghi:
Hi @mbrandonw,
Thank you so much for reviewing this so quickly 🙏.
I realize now that I was too quick to propose an alternative for the issue I mentioned earlier.

As you pointed out:
- The <T: Equatable>? overload is redundant since Optional already provides that conformance.
- For something like Box(payload: state.box.payload), I see that skipping an update is actually reasonable when the data has not changed and the class has no behavior.
- In my case, I ran into problems specifically with classes that define == by value, but where identity changes are still significant. That’s what led me to try to cover it in the macro.

I agree it makes more sense to reconsider how Equatable is implemented on reference types, rather than adding extra overloads here.

Issue #3777, opened 2025-09-18
Broken link in README: Example section returns 404


seosieve:
### Description

While reading the Example/README.md, I noticed that the Tic-Tac-Toe link returns a 404.
After investigating, I found that the iOS part of the repository was split off into RIBs-iOS, which caused the path to change.

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-composable-architecture/issues) or [discussion](https://github.com/pointfreeco/swift-composable-architecture/discussions).

### Expected behavior

Navigate to the tutorial4-completed folder in RIBs-iOS.

### Actual behavior

The link returns a 404 error.

### Reproducing project

1. Navigate to https://github.com/pointfreeco/swift-composable-architecture/tree/main/Examples
2. Open the README.md file
3. In the Tic-Tac-Toe section, click the RIBs (see here) link

### The Composable Architecture version information

1.22.3

### Destination operating system

none

### Xcode version information

none

### Swift Compiler version information

```shell
none
```

PR #3778, opened 2025-09-18
chore: changed libs repository link fix

seosieve:
Resolves <https://github.com/pointfreeco/swift-composable-architecture/issues/3777>

> RIBs link update in Example README

- The original link was created before RIBs was split into RIBs-iOS, and now returns a 404.
- Update the path to point to the Tutorial4-completed folder in RIBs-iOS.


stephencelis (approved):
Thanks!

PR #3780, opened 2025-09-19
Bump swift-navigation for Xcode 26

Betmoong:
Xcode 26 has now been officially released.
To make sure the project compiles correctly with Xcode 26, I updated the swift-navigation dependency to version [2.3.2](https://github.com/pointfreeco/swift-navigation/releases/tag/2.3.2), which includes the necessary fixes.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -26,7 +26,7 @@ let package = Package(
     .package(url: "https://github.com/pointfreeco/swift-dependencies", from: "1.4.0"),
     .package(url: "https://github.com/pointfreeco/swift-identified-collections", from: "1.1.0"),
     .package(url: "https://github.com/pointfreeco/swift-macro-testing", from: "0.2.0"),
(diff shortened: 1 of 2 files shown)


mustafataibah (approved):

mustafataibah:
Might as well bump it to 2.4.2 no?

Betmoong:
@mustafataibah This PR was just to get it compiling on Xcode 26, so I went with 2.3.2, which is the minimum required version.

mbrandonw:
Hi @Betmoong, thanks for taking the time to look into this, but we prefer to keep our dependencies as flexible as possible, and only update them when it's necessary to fix a critical error. That way if there's some reason that people must pin to SwiftNavigation 2.3.1 or earlier they have that choice.

In order to force your own project to SwiftNavigation 2.3.2+ you can add an explicit dependency on swift-navigation (even though TCA already specifies it) and then pin it to the version you need.

PR #3784, opened 2025-09-26
Add support for naming tasks in .run effects

aregler:
Hi 👋🏻

In Swift 6.2 the naming of `Task`s was introduced in [SE-0469](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0469-task-names.md).
This proposal allows developers to assign human-readable names to tasks, which show up in debugging and profiling tools.

I think this also could be useful when working with the `.run` effect in TCA.

### Usage
```swift
case .tapped:
   .run(name: "🔄 Button tapped response-task ") { send in
      await send(.response)
}
```

### Example:
image

Thank you for all your great work,

Arsatius

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -408,6 +408,7 @@ extension Effect {
   // NB: Backwards-compatible shims.
   extension Task {
     @discardableResult
+    @usableFromInline
     init(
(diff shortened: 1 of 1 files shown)


stephencelis (approved):
@aregler Thanks! Pushed a little bit of cleanup, but this looks good to us!

PR #3804, opened 2025-10-24
ifLet CaseReducer

arnauddorgans:


mbrandonw:
Hi @arnauddorgans, can you please provide some context why you are proposing this change?

I general I think we are moving away from overloads like this in the future, and will instead just have people specify the body of `ifLet` by accessing the static` body` property.

arnauddorgans:
@mbrandonw If we want to move away from overloads, maybe we should add something in the documentation about CaseReducer with the static var body property, it is quite unique and the only thing I found in the doc mentioning the @CaseReducer is here https://pointfreeco.github.io/swift-composable-architecture/1.9.0/documentation/composablearchitecture/migratingto1.8#Destination-and-path-reducers with 
```swift
```
but nothing mentioning how to use an enum reducer for a classic usage as an optional/non-optional child feature

mbrandonw:
Ok, we will look into updating the docs, but for now the recommended way is to spell this out explicitly in your feature code:

```swift
.ifLet(\.child, action: \.child) {
  ChildFeature.body
}
```

PR #3805, opened 2025-10-25
WIP Android support

Joannis:
Hey all! I hope you're open to an effort like this. I would love to see TCA support Android, so we can share a huge chunk of our codebase between the two major platforms. This is still WIP, for some reason the Swift compiler doesn't respect `#if os(..)` that well.

Main changes:
- Use OpenCombine on Linux/Android
- Disable SwiftUI specific code using `#if os(macOS) || os(iOS) ...`
- I tried doing `#if canImport(SwiftUI)` but that was also not respected, hence the os path

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "1313e97b3134138f8ca61f73141726a7e426c0097a38fadeba8e33be243b0d2e",
+  "originHash" : "372ba44815bad1dc45c295eaa0194eb2f24433da53e2c0e97e52f6f95f0fe64c",
   "pins" : [
     {
       "identity" : "combine-schedulers",
@@ -8,6 +8,15 @@
       "state" : {
         "revision" : "5928286acce13def418ec36d05a001a9641086f2",
         "version" : "1.0.3"
+      }
+    },
+    {
+      "identity" : "opencombine",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/OpenCombine/OpenCombine.git",
+      "state" : {
+        "revision" : "8576f0d579b27020beccbccc3ea6844f3ddfc2c2",
+        "version" : "0.14.0"
       }
     },
     {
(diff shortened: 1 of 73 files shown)


stephencelis:
@Joannis Exciting! Thanks for starting this and other discussions! We are enthusiastic about this initiative. I think the PR as it stands is doing a lot, though, so it's a bit harder for us to review. Ideally I think it could be split into smaller parts that are each easier to review in isolation.

1. There's a lot of formatting churn, I think due to a newer version of `swift-format`. Are you down to first PR a reformat (and bump of the `format.yml` GitHub action) separately so this PR can be more focused?
2. The `#if canImport(SwiftUI)` branching also feels like it could be pulled out to a separate PR.
3. I asked in the related `combine-schedulers` discussion, but can all of the `#if`-`import OpenCombine`s be eliminated with a module alias to `Combine` on non-Apple platforms, or is there a problem with that?

Once all the above is done/considered I think it'll be easier to review the OpenCombine diff and start reviewing things :smile:

stephencelis:
(As a side note, we are currently in the process of designing 2.0 of the Composable Architecture, which is being approached from first principles with modern Swift and platform agnostics, _i.e._ no dependence on Combine or other Apple-only frameworks. It may be a few months before any of it sees daylight, though, so we're happy to provide earlier support!)

Joannis:
I'm really happy to hear your plans for TCA 2.0, I think that's much more interesting to me by the sounds of it.

mehmetbaykar:
I’m super excited for TCA 2.0 after seeing https://www.pointfree.co/blog#the-composable-architecture-2-0. Out of curiosity, when do you think we’ll be able to see the first beta version of TCA 2.0? Is there an estimated timeline?

Joannis:
I'll close this for now. 2.0 will be much better

stephencelis:
@mehmetbaykar No precise plans at the moment! We're actively working on it and will hopefully have some more to show early next year.

PR #3806, opened 2025-10-27
Fix ShareChangeTracker(reportUnassertedChanges:) build error.

house0000:
# Bug
build error:

スクリーンショット 2025-10-27 13 37 57

# Fix
`reportUnassertedChanges` exists from swift-sharing v1.0.4. So fix minimum version of it in Package.swift


Package@swift-6.0.swift
```diff
@@ -26,9 +26,9 @@ let package = Package(
     .package(url: "https://github.com/pointfreeco/swift-dependencies", from: "1.4.0"),
     .package(url: "https://github.com/pointfreeco/swift-identified-collections", from: "1.1.0"),
     .package(url: "https://github.com/pointfreeco/swift-macro-testing", from: "0.2.0"),
-    .package(url: "https://github.com/pointfreeco/swift-navigation", from: "2.3.0"),
+    .package(url: "https://github.com/pointfreeco/swift-navigation", from: "2.3.2"),
```
house0000: It's trash and I cannot remove it. sorry.(this fix already applied to main, why appear..)

PR #3816, opened 2025-11-08
Allow an optional child scope

larryatkin:
Implement scope, where the closure to access the child state may return nil when the child state changes.  See discussion in [Q&A](https://github.com/pointfreeco/swift-composable-architecture/discussions/3817)


stephencelis:
@larryatkin Replied to your discussion. I think we can close this for now, but let me know if you have any questions!

PR #3834, opened 2026-01-08
Lifecycle action

arnauddorgans:
Inspired by https://github.com/pointfreeco/swift-composable-architecture/pull/3624

> It is common to send an action to a store when it is first created to kick off some long-living effects. While many utilize the onAppear view modifier for this work, we can simplify things by making the store's initializer a little more powerful.

### DIFF
--- a/Sources/ComposableArchitecture/Store.swift
+++ b/Sources/ComposableArchitecture/Store.swift
@@ -353,6 +353,10 @@ public final class Store<State, Action>: _Store {
       }
       self.parentCancellable = subscribeToDidSet(stateType)
     }
+    
+    if let lifecycleAction = (Action.self as? (any LifecycleAction.Type))?.lifecycle as? Action {
+      send(lifecycleAction)
+    }
   }
 
   convenience init<R: Reducer<State, Action>>(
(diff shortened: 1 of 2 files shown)


stephencelis:
@arnauddorgans Thanks for taking the time to share and PR :smile:

It's definitely an interesting idea that aligns with `BindableAction`, but we actually have big changes in the works for 2.0 that negate the need for these extra conformances entirely, and lifecycle actions are built into the system:

https://www.pointfree.co/blog/posts/194-cyber-week-save-30-on-point-free#feature-lifecycle-hooks

To avoid upgrade churn I think we'll be punting on any additional surface area here for 2.0 (I just closed #3624 because of that). Going to close this for now, but please open a discussion if you'd like to talk about these upcoming changes to the next major version of TCA!

arnauddorgans:
@stephencelis can't we imagine adding it under an `@_spi` ? 
it would help me a lot, I have a lot of child features that are not managed by a view, so I have to call an action each time the child feature is not nil etc..

do we have an E.T.A for the 2.0 release?

stephencelis:
@arnauddorgans Does it need anything from `@_spi`? Ideally you would be able to define your own convenience initializer in your project.

> do we have an E.T.A for the 2.0 release?

No ETA, though we hope to have a preview for folks in the next few months.

arnauddorgans:
@stephencelis I would love to use a convenience initializer but it will not work for scoped stores etc.. 
The only workaround I have is to use my own version of the library on my branch (which is okay tho)

stephencelis:
@arnauddorgans Ah yeah I understand after looking at the diff more. I do think the more appropriate option here would be to maintain a fork for now given the roadmap. We don't anticipate a ton of churn on the 1.x mainline right now, and then in 2.0 you could upgrade to the new tools.

PR #3842, opened 2026-02-12
Fix NavigationStackController deprecation warning by migrating to UIBindable

GangWoon:
Since Store is an AnyObject, it’s unavoidable to get the warning when using NavigationStackController.
How about marking the existing UIBinding-based initializer as `@_disfavoredOverload`, and adding a new initializer that takes UIBindable instead?

``` swift
  @available(
    *,
    deprecated,
    message: """
      A '@UIBinding' must be initialized with a value, not an observable reference. Use '@UIBindable', instead.
      """,
    renamed: "UIBindable.init"
  )
  public init(wrappedValue value: Value) where Value: AnyObject {

```

image

### DIFF
--- a/Sources/ComposableArchitecture/UIKit/NavigationStackControllerUIKit.swift
+++ b/Sources/ComposableArchitecture/UIKit/NavigationStackControllerUIKit.swift
@@ -23,6 +23,39 @@
     public convenience init<State, Action>(
       navigationBarClass: AnyClass? = nil,
       toolbarClass: AnyClass? = nil,
+      path: UIBindable<Store<StackState<State>, StackAction<State, Action>>>,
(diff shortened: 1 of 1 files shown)


stephencelis:
@GangWoon I haven't had a chance to spin up a project yet, but I think ideally this would look similar to SwiftUI:

```swift
@UIBindable var store: StoreOf<Feature>

let navigationController = NavigationStackController(
  path: $store.scope(state: \.path, action: \.path),
```

Does this not work as an alternative?

GangWoon:
I hadn’t thought about making the store `@UIBindable`.
I tried it exactly like in your example, and I’m not seeing any warnings now. Thanks! 🙂

PR #3846, opened 2026-02-18
Use DiscardingTaskGroup in Effect.merge(with:)

zekexros:
## Summary

- Replace `withTaskGroup(of: Void.self)` with `withDiscardingTaskGroup` in `Effect.merge(with:)` on iOS 17+ / macOS 14+
- Fall back to `withTaskGroup` on older OS versions via `#available`

## Motivation

`Effect.merge(with:)` uses `withTaskGroup(of: Void.self)` to run two effects concurrently, but never calls `next()` to consume results. In this case, `TaskGroup` retains completed child tasks until the group scope ends.

`DiscardingTaskGroup` eagerly discards and releases child tasks as soon as they complete. When multiple effects are merged via `reduce`, the resulting nested group structure runs all effects concurrently — so effects that finish earlier have their resources reclaimed immediately, rather than waiting for the slowest sibling.

Reference: [DiscardingTaskGroup | Apple Developer Documentation](https://developer.apple.com/documentation/swift/discardingtaskgroup)

## Test plan

- [x] `EffectOperationTests` — 5 tests passed (includes `testMergeDiscardsNones`, `testMergeFuses`)
- [x] `EffectTests` — 9 tests passed (includes `testMerge`)
- [x] Project builds successfully

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -280,12 +280,23 @@ extension Effect {
     ):
       return Self(
         operation: .run { send in
-          await withTaskGroup(of: Void.self) { group in
-            group.addTask(name: lhsName, priority: lhsPriority) {
-              await lhsOperation(send)
+          if #available(iOS 17, macOS 14, tvOS 17, watchOS 10, *) {
+            await withDiscardingTaskGroup { group in
+              group.addTask(name: lhsName, priority: lhsPriority) {
+                await lhsOperation(send)
+              }
+              group.addTask(name: rhsName, priority: rhsPriority) {
+                await rhsOperation(send)
+              }
             }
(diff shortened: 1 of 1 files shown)


stephencelis:
@zekexros Looks like it needs some changes to work.

We'd be happy to merge this for the 1.x branch if you can address, but at the same time we have opened #3879 where the `merge` operation is going fully away, so this patch will mostly apply to apps that don't migrate to 2.0.

Let us know if you still think this is worth shipping! We can close it out if not.

zekexros:
@stephencelis  Thanks for the update\! I checked out #3879 and I'm really excited about the direction. I think this patch doesn't need to be shipped — feel free to close it out.

stephencelis:
@zekexros Thanks again!

PR #3856, opened 2026-02-22
Infrastructure: Shared bindings Xcode 26.4 fix

stephencelis:
Swift 6.3 has fixed some bugs we were accidentally exploiting for deriving shared bindings. This PR has what is a reasonable alternative while we think about the change.

### DIFF
--- a/ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ComposableArchitecture.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,13 +1,13 @@
 {
-  "originHash" : "c55ea51a4817791946e7830a957bd243534ea4fed282d6370350fb9dcb3b15c7",
+  "originHash" : "643912120f2d51dc6f5c8ba4174f7dbaefbbd906d0158db6dc4d4bf6a82c6adb",
   "pins" : [
     {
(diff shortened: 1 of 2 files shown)


dadalar:
Hi @stephencelis. We've hit a similar issue after upgrading to Xcode 26.4 but this suggested change seems to bypass @Bindable and BindableAction altogether. That means we can't observer .binding actions anymore to act on it. An `.onChange` in the Reducer also doesn't work because the changes is not observed by TCA. The only workaround I could find is to create a new Binding which would set the Shared State *AND* send the binding action. It looks like this:
```swift
// Xcode 26.3
Picker("Tab", selection: $store.selectedTab) 

// Xcode 26.4
Picker("Tab", selection: Binding(get: { store.selectedTab }, set: { newValue in store.$selectedTab.withLock { $0 = newValue }; store.send(.binding(.set(\.$selectedTab, store.$selectedTab))) }))
```

Do you think there is a nicer way to bring back to old ergonomics? Thanks already!

stephencelis:
@dadalar You can also use observe `$selectedTab.publisher` in the reducer to detect changes to a shared value.

In 2.0 observation is more built-into the system, so you will be able to do `onChange`:

```swift
.onChange(of: store.selectedTab) { oldTab, state in
  // ...
}
```

PR #3862, opened 2026-02-24
fix: preserve animation context through debounce

### LINKED ISSUE #3586: `debounce` ignores `animation` parameter of `send`
### Description

When adding `debounce` to an effect any actions that are emitted from that effect are not animated even though the `animation` parameter is set when calling `send`.

### Checklist

### Expected behavior

In the sample code, all buttons should animate the square.

### Actual behavior

Only the first and last buttons animate the square.

### Reproducing project

```swift
import ComposableArchitecture
import SwiftUI

@Reducer
struct DebounceAnimationBugFeature {
    @ObservableState
    struct State {
        var toggle = false
    }

    enum Action {
        case toggleWithoutDebounceTapped
        case toggleWithDebounceTapped
        case toggleWithDebounceAndAnimationModifierTapped
        case toggleState
    }

    private enum TaskId {
        case debounce
    }

    @Dependency(\.mainRunLoop) private var mainRunLoop

        Reduce<State, Action> { state, action in
            switch action {
            case .toggleWithoutDebounceTapped:
                return .run { send in
                    await send(.toggleState, animation: .default)
                }
            case .toggleWithDebounceTapped:
                return .run { send in
                    await send(.toggleState, animation: .default)
                }
                .debounce(id: TaskId.debounce, for: 0.1, scheduler: mainRunLoop)
            case .toggleWithDebounceAndAnimationModifierTapped:
                return .run { send in
                    await send(.toggleState)
                }
                .debounce(id: TaskId.debounce, for: 0.1, scheduler: mainRunLoop)
                .animation(.default)
            case .toggleState:
                state.toggle.toggle()
                return .none
            }
        }
    }
}

struct DebounceAnimationBugView: View {
    @Bindable var store: StoreOf<DebounceAnimationBugFeature>

    var body: some View {
        VStack {
            Button("Toggle without debounce") {
                store.send(.toggleWithoutDebounceTapped)
            }

            Button("Toggle with debounce") {
                store.send(.toggleWithDebounceTapped)
            }

            Button("Toggle with debounce and animation modifier") {
                store.send(.toggleWithDebounceAndAnimationModifierTapped)
            }

            Rectangle()
                .fill(Color.blue)
                .frame(width: 50, height: 50)
                .frame(maxWidth: .infinity, alignment: store.toggle ? .leading : .trailing)
        }
    }
}

#Preview {
    DebounceAnimationBugView(store: Store(initialState: .init(), reducer: {
        DebounceAnimationBugFeature()
    }))
}
```

1.17.1

iOS 18.2

### Xcode version information

16.2

```shell

```

WZBbiao:
## What

This PR fixes an issue where the `animation` parameter passed to `send(_:animation:)` inside a `.run` effect is lost when that effect is wrapped with `.debounce`.

## Why

SwiftUI's `withTransaction` stores the transaction (which carries the animation) in a thread-local context. When `Send.callAsFunction(_:animation:)` is invoked inside a `.run` effect, it wraps the `subscriber.send($0)` call with `withTransaction`, making the animation available to the downstream Combine pipeline **synchronously** at the point of delivery.

However, the `.receive(on: scheduler)` call inside `debounce` delivers the action **asynchronously** — queuing it on the scheduler rather than propagating it synchronously. By the time the scheduler delivers the action, the `withTransaction` scope has already exited and the animation context is gone.

Concretely, the old implementation:

```swift
.flatMap { _EffectPublisher(self).receive(on: scheduler) }
```

The `receive(on: scheduler)` introduces an async hop that discards the SwiftUI transaction.

## How

Remove the redundant `.receive(on: scheduler)` from inside the `flatMap`. Thread-safety for publisher effects is already handled by the `receive(on: UIScheduler.shared)` applied in `Core.swift` when the store subscribes to publisher-based effects. `UIScheduler.shared` delivers synchronously when already on the main thread (which is the case for `@MainActor` run effects), preserving the transaction context.

```swift
// Before
.flatMap { _EffectPublisher(self).receive(on: scheduler) }

// After
.flatMap { _EffectPublisher(self) }
```

## How to Test

Use the reproducing project from the issue:

```swift
case .toggleWithDebounceTapped:
    return .run { send in
        await send(.toggleState, animation: .default)
    }
    .debounce(id: TaskId.debounce, for: 0.1, scheduler: mainRunLoop)
```

Before the fix, the square did not animate when using the "Toggle with debounce" button.
After the fix, the square animates correctly.

A regression test `testDebouncePreservesAnimationContext` has been added to `EffectDebounceTests.swift` to verify the animation/transaction context is preserved through debounce.

Fixes #3586

### DIFF
--- a/Sources/ComposableArchitecture/Effects/Debounce.swift
+++ b/Sources/ComposableArchitecture/Effects/Debounce.swift
@@ -42,7 +42,7 @@ extension Effect {
         operation: .publisher(
           Just(())
             .delay(for: dueTime, scheduler: scheduler, options: options)
-            .flatMap { _EffectPublisher(self).receive(on: scheduler) }
+            .flatMap { _EffectPublisher(self) }
             .eraseToAnyPublisher()
         )
       )
--- a/Tests/ComposableArchitectureTests/EffectDebounceTests.swift
+++ b/Tests/ComposableArchitectureTests/EffectDebounceTests.swift
@@ -1,5 +1,6 @@
 import Combine
 @_spi(Internals) import ComposableArchitecture
+import SwiftUI
 import XCTest
 
 final class EffectDebounceTests: BaseTCATestCase {
@@ -93,5 +94,40 @@ final class EffectDebounceTests: BaseTCATestCase {
 
     XCTAssertEqual(values, [1])
     XCTAssertEqual(effectRuns, 1)
+  }
+
+  // NB: Regression test for https://github.com/pointfreeco/swift-composable-architecture/issues/3586
+  @MainActor
+  func testDebouncePreservesAnimationContext() async {
+    let mainQueue = DispatchQueue.test
+    var capturedAnimations: [Animation?] = []
+    var values: [Int] = []
+
+    struct CancelToken: Hashable {}
+
+    let effect = Effect<Int>.send(1, animation: .default)
+      .debounce(id: CancelToken(), for: 1, scheduler: mainQueue)
+
+    let cancellable = _EffectPublisher(effect).sink { value in
+      capturedAnimations.append(Transaction.current.animation)
+      values.append(value)
+    }
+
+    // Nothing emits right away.
+    XCTAssertEqual(values, [])
+    XCTAssertEqual(capturedAnimations.count, 0)
+
+    // Advance past the debounce window.
+    await mainQueue.advance(by: 1)
+
+    // The action should be received and the animation context preserved.
+    XCTAssertEqual(values, [1])
+    XCTAssertEqual(capturedAnimations.count, 1)
+    XCTAssertNotNil(
+      capturedAnimations.first ?? nil,
+      "Animation context should be preserved through debounce"
+    )
+
+    _ = cancellable
   }
 }


stephencelis:
@WZBbiao Thanks for taking the time to look into the problem and suggest a fix. The `debounce(id:)` tool is from pre-concurrency Combine effects, and should probably be deprecated in favor of more modern concurrency tooling. It will definitely be removed in version 2.0 of the library, so I will push a PR soon that prepares folks for just that via deprecation.

Instead, I'd recommend doing something similar to `task(id:)` in SwiftUI:

```swift
// Defined in the `@Reducer struct`:
enum CancelID { case toggleDebounce }
@Dependency(\.continuousClock) var clock

// Used in `Reduce`
return .run { send in
  try await clock.sleep(for: .seconds(0.1)
  await send(.toggleState, animation: .default)
}
.cancellable(id: CancelID.toggleDebounce, cancelInFlight: true)
```

1.0 of the library was developed over a long evolution of Swift and Apple frameworks, which means it has built up a confusing array of APIs for animating and scheduling, and we will hopefully clean it all up in the next major release!

I'm going to close this for now, as I am a little worried about merging in changes that could break scheduling/animation in production applications depending on a particular version of Combine.

PR #3874, opened 2026-03-10
Trait-deprecate `Effect` for `EffectOf`

stephencelis:
The shape of `Effect` is changing in 2.0 to be `Effect<State, Action>`, so to get ahead of the curve we are offering a trait to help folks migrate to `EffectOf<Feature>` (supported today) when explicitly referring to `Effect`.

### DIFF
--- a/Sources/ComposableArchitecture/Effect.swift
+++ b/Sources/ComposableArchitecture/Effect.swift
@@ -15,7 +15,7 @@ public struct _Effect<Action>: Sendable {
     case run(
       name: String? = nil,
       priority: TaskPriority? = nil,
-      operation: @Sendable (_ send: Send<Action>) async -> Void
+      operation: @Sendable (_ send: _Send<Action>) async -> Void
     )
   }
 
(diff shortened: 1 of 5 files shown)


Sources/ComposableArchitecture/Effect.swift
```diff
@@ -2,7 +2,12 @@
 import Foundation
 import SwiftUI
 
-public struct Effect<Action>: Sendable {
+#if ComposableArchitecture2Deprecations
+  @available(*, deprecated, message: "Use 'EffectOf<Feature>' instead")
+#endif
```
stephencelis: It's a pretty invasive change so putting it behind a trait. We could consider soft-deprecating on an `#else` branch...

rcarver:
Just for reference, `EffectOf` can be awkward when defining effect-returning functions on `State`, so I've gotten in the habit of using `Effect<Action>` in general.

```swift
struct State {
  // EffectOf<Self> is won't work
  // EffectOf<ReducerName> feels repetitive
  // EffectOf<State, Action> works
  // EffectOf<Self, Action> works
  func someWork() -> Effect<Action> { } 
}
```

That said, maybe returning effect-returning functions on `State` becomes an anti pattern in 2.0? It's always been a little strange.

stephencelis:
@rcarver Yeah, we really don't think effects should be produced outside of the reducer/feature layer, so this will be a good opportunity to move them over to helpers on your reducer/feature instead.

PR #3879, opened 2026-03-11
Trait-deprecate `Effect.concatenate`, `Effect.map`

stephencelis:
Since Swift concurrency landed, with `async`-`await` and task groups, we have pushed folks towards using these tools in a `.run` effect instead of `.merge` and `.concatenate`. While many still reach for `.merge` and `.concatenate` today, Composable Architecture 2.0 is making some fundamental changes to effects that make them more efficient and powerful but less transformable. To prepare folks, we should deprecate `.concatenate` and `.map` behind our 2.0 trait. As of now we'll retain `.merge`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -23,11 +23,7 @@ let package = Package(
       description: """
         Prepare for the next major release by enabling some of the more viral deprecations.
         """
-<<<<<<< deprecate-merge-concat
-    ),
-=======
     )
->>>>>>> main
   ],
   dependencies: [
     .package(url: "https://github.com/apple/swift-collections", from: "1.1.0"),


jshier:
So what would you suggest for composing reusable `Effect`s that cover unrelated functionality? For instance, we have an analytics `Effect` that can be attached to other effects that fire at the same time. There's no result, it never needs to wait or be waited on, it just needs to fire at the same time. We could force every usage to call the analytics dependency directly, but we have hundreds of calls, so that's not really a good solution.

```swift
return .run { send in 
  await send(.receiveContent(api.fetchContent())
}
.analytics(event: .buttonTap, id: .someButton)
```

Where `analytics` just merges the two `Effect`s. Pulling such a call into the `run` seems rather invasive. I suppose it depends on what functionality `Effect` ultimately gets in v2.

acosmicflamingo:
I am also curious how to handle use-cases where effects are cancellable in-flight. I have many cancellable effects returned from functions that are used in merged effects. Am I going to have to do this now?

```swift
// Old
return .run { _ in
  // Execute some logic
}.cancellable(id: CancelID.foo, inFlight: true)


// New
return .run { _ in
  foo()
}

func foo() async {
  Task.cancel(id: CancelID.foo) {
    // Execute some logic
  }
}
```

stephencelis:
> I suppose it depends on what functionality `Effect` ultimately gets in v2.

@jshier Effects in 2.0 are highly coupled to features and stores now, which we think is a good thing. Effects get tied by the runtime to the action/event that started them, and all that information is surfaced in the debugger (task names, etc.), in test failure messages, and more. And this location-based approach also allows the cancellation model to be much simpler: they are cancellable-in-flight by default, and you can introduce an explicit effect `id` to allow an effect to continue to live even if the same action kicked off something new. As a result it just doesn't make sense for effects to be transformable, since this metadata could be lost.

You could still bring your own static helper:

```swift
return .run(event: .buttonTap, id: .someButton) { store in 
  try await store.send(.receiveContent(api.fetchContent())
}
```

You'd just have to be careful to map things like `fileID`, etc., in the function for it to play nicely with the system.

Alternately, we think it's probably best to do this work inline:

```swift
return .run { store in
  analytics(event: .buttonTap, id: .someButton)
  try await store.send(.receiveContent(api.fetchContent())
}
```

This is the same number of lines that you currently have, but should work the same, and is trivial enough that an agent could probably do the migration for you with no mistakes.

We are definitely open to discussing improvements that we can make to 2.0, and we do sympathize with folks that have a lot of `.merge` and `.concatenate` in their current projects, enough that they're not totally on the chopping block, but there is a lot of uncertainty in how we could support them at the moment:

- `map` is definitely on the chopping block. There's no way to `map` an `Effect<State, Action>`.
- `merge` could potentially be supported with complications: how does it play with debuggability, cancellation, etc.
- `concatenate` could also potentially be supported with those same complications

stephencelis:
@acosmicflamingo Your comment seems to be about effect cancellation, which is unrelated to this PR. While Effects in 2.0 will have a brand new cancellation system, we do plan on having a deprecated port of the 1.0 APIs in the 2.0 package to ease in migration.

acosmicflamingo:
I was trying to write an example away from my keyboard, but this demonstrates what I meant:

 ```swift
case .delegate(.networkStatusDidChange):
    return .merge(
      fetchProductsEffect(productsRef: state.$products),
      validateEntitlementsEffect(),
    )

  private func listenForTransactionsEffect() -> Effect<AppMain.Action> {
    .run { send in
      for await event in storeKit.updates() {
        await send(.storeKit(.updateTransaction(event)))
      }
    }
    .cancellable(id: CancelID.transactionObservation, cancelInFlight: true)
  }
```

I thought it suited this PR because of the use of `Effect.merge`, even though my concern is how cancellation is handled.

stephencelis:
@jshier @acosmicflamingo We've decided to keep `merge` as a utility in 2.0. We still don't think `concatenate` carries its weight, though, and `map` will still be impossible to define.

jshier:
Sounds good, I'll reevaluate our tools once we get a better idea of Effect in TCA 2.

On a related note, with the deprecation / removal of `map` and the `.send` `Effect`, what will be the suggested way to perform child actions from a parent?

acosmicflamingo:
I'm all for removal of .merge, .concatenate, etc. I was just curious what the logic would look like in 2.0. Hopefully my post wasn't pressuring you to keep logic there that you saw was unnecessary.

stephencelis:
> On a related note, with the deprecation / removal of map and the .send Effect, what will be the suggested way to perform child actions from a parent?

We think `.send(.child(.action))` is totally fine. Our guidance away from this was because we saw it being overused in the community.

> I'm all for removal of .merge, .concatenate, etc. I was just curious what the logic would look like in 2.0. Hopefully my post wasn't pressuring you to keep logic there that you saw was unnecessary.

It'll look quite similar with `merge` retained. We've already brought it back in. Everything is still in flight so maybe we'll get rid of it again, but for now we should consider it a holdover :smile:

PR #3889, opened 2026-03-19
Bump swift-navigation to 2.7.0

jagreenwood:
As I was updating to TCA 1.25.2, I ran in to a build error related to enum scoping:

```swift 
.alert($store.scope(state: \.$destination, action: \.destination).alert)
```

> [!CAUTION]
> Cannot convert value of type Binding<AlertState<ScratchReducer.Destination.Alert>?> to expected argument type Binding<AlertState\<Never\>?>

My version of `swift-navigation` was 2.6.0. Updating to 2.7.0 resolved the error. This diff aims to enforce the version requirement for others.

---

Full Snippet for wider context:
```swift
import ComposableArchitecture
import SwiftUI

// MARK: - Reducer

@Reducer
struct ScratchReducer {
    @Reducer
    enum Destination {
        case alert(AlertState<Alert>)

        @CasePathable
        enum Alert {
            case confirm
        }
    }

    @ObservableState
    struct State: Equatable {
        @Presents var destination: Destination.State?
        var items = ["Apple", "Banana", "Cherry"]
    }

    enum Action: Sendable {
        case showConfirmTapped
    }

        Reduce { state, action in
            switch action {
            case .destination(.presented(.alert(.confirm))):
                state.items.removeAll()
                return .none

            case .destination:
                return .none

            case .showConfirmTapped:
                state.destination = .alert(AlertState {
                    TextState("Are you sure?")
                } actions: {
                    ButtonState(action: .confirm) {
                        TextState("Confirm")
                    }
                    ButtonState(role: .cancel) {
                        TextState("Cancel")
                    }
                } message: {
                    TextState("This will remove all items.")
                })
                return .none
            }
        }
    }
}

extension ScratchReducer.Destination.State: Equatable {}

// MARK: - View

struct ScratchView: View {
    @Bindable var store: StoreOf<ScratchReducer>

    var body: some View {
        List {
            Section("Items") {
                ForEach(store.items, id: \.self) { item in
                    Text(item)
                }
            }

            Section("Actions") {
                Button("Show Confirm Alert") {
                    store.send(.showConfirmTapped)
                }
            }
        }
        .alert($store.scope(state: \.$destination, action: \.destination).alert)
    }
}

// MARK: - Preview

#Preview {
    ScratchView(
        store: Store(
            initialState: ScratchReducer.State(),
            reducer: { ScratchReducer() }
        )
    )
}
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -35,7 +35,7 @@ let package = Package(
     .package(url: "https://github.com/pointfreeco/swift-dependencies", from: "1.4.0"),
     .package(url: "https://github.com/pointfreeco/swift-identified-collections", from: "1.1.0"),
     .package(url: "https://github.com/pointfreeco/swift-macro-testing", from: "0.2.0"),
-    .package(url: "https://github.com/pointfreeco/swift-navigation", from: "2.3.2"),
+    .package(url: "https://github.com/pointfreeco/swift-navigation", from: "2.7.0"),
     .package(url: "https://github.com/pointfreeco/swift-perception", "1.3.4"..<"3.0.0"),
     .package(url: "https://github.com/pointfreeco/swift-sharing", "1.0.4"..<"3.0.0"),
     .package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay", from: "1.3.0"),


mbrandonw:
Hey @jagreenwood, thanks for looking into this! I don't have the time to check right this moment, but is 2.7 definitely the lowest version of swift-navigation that fixes this? Just want to try to be as lax as possible with the version to give users more flexibility.

jagreenwood:
> Hey @jagreenwood, thanks for looking into this! I don't have the time to check right this moment, but is 2.7 definitely the lowest version of swift-navigation that fixes this? Just want to try to be as lax as possible with the version to give users more flexibility.

@mbrandonw Totally get it. My version was pinned to 2.6. Seems like the only difference between 2.6 and 2.7 is [.alert(item:)](https://github.com/pointfreeco/swift-navigation/pull/327).

You would absolutely know better than me on the inner workings, but seems possible these are related.

mbrandonw:
Ok cool, if 2.6 has the problem and 2.7 fixes then we can definitely take this. Will take a closer look a little later today. In the meantime you can always add an explicit dependency on swift-navigation and pin it to whichever version you want. Thanks again!

PR #3906, opened 2026-03-29
Fixed: Reading @Shared state from a BindableAction Store raises deprecation warning

seanmrich:
In a `View` context, reading a `@Shared` value from the store's state will raise a deprecation warning when the store's `Action` conforms to `BindableAction`.

```swift
@Reducer
struct Feature {
  @ObservableState
  struct State: Equatable {
    @Shared(.inMemory("value")) var value: Int = 0
  }
  enum Action: BindableAction {
  }
}

struct FeatureView: View {
  let store: StoreOf<Feature>

  var body: some View {
    let _ = store.value
    // 🛑 Setter for 'value' is deprecated: Use '$shared.withLock' to modify a shared value with exclusive access; when constructing a SwiftUI binding, use 'Binding($shared)'
  }
}
```

This is because the dynamicMemberLookup selects an overload that requires a `WritableKeyPath` which is enough to trigger Swift to require a setter for the property which is invalid for a `@Shared` value. Applying `@_disfavoredOverload` will drive the lookup toward an overload requiring a `KeyPath` instead.

### DIFF
--- a/Sources/ComposableArchitecture/Observation/Binding+Observation.swift
+++ b/Sources/ComposableArchitecture/Observation/Binding+Observation.swift
@@ -168,6 +168,7 @@ extension BindableAction where State: ObservableState {
 }
 
 extension Store where State: ObservableState, Action: BindableAction, Action.State == State {
+  @_disfavoredOverload
   public subscript<Value: Equatable & Sendable>(
     dynamicMember keyPath: WritableKeyPath<State, Value>
   ) -> Value {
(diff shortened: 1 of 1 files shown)


pyrtsa:
Same deprecation warning seen in my project too, and just verified this PR fixes it. 👌

samhenrigold:
This has been giving me the runaround this morning, glad someone picked up on this!

PR #3911, opened 2026-04-01
Adds disfavoredOverload annotation to `~=` implementation in Binding.swift

rprince500:
When running versions of TCA >= 1.24.0 on Xcode 26.4, our team's project fails to compile with several instances of the following compiler error: `Ambiguous use of operator '~='`

It looks like the compiler is having trouble disambiguating between this function and the more constrained implementation declared in `Binding+Observation.swift`, where the only difference is the additional `where Root: ObservableState` constraint.

I've confirmed that adding the `@_disfavoredOverload` annotation to the less constrained operator gets us building again.

### DIFF
--- a/Sources/ComposableArchitecture/SwiftUI/Binding.swift
+++ b/Sources/ComposableArchitecture/SwiftUI/Binding.swift
@@ -62,6 +62,7 @@ extension BindingAction {
   /// case .binding(\.enableNotifications):
   ///   // Return an authorization request effect
   /// ```
+  @_disfavoredOverload
   public static func ~= <Value>(
     keyPath: _SendableWritableKeyPath<Root, Value>,
     bindingAction: Self


stephencelis:
@rprince500 Thanks! Can you include a test or code snippet that demonstrates the issue so that we can confirm before merging?

rprince500:
@stephencelis 

CleanShot 2026-03-31 at 22 34 32@2x

Here's a screenshot of one of the errors in our project. When I create a brand new project as an MRE, the compiler doesn't seem to choke up. I guess it's able to parse things just fine up until a certain level of complexity ¯\\\_(ツ)_/¯

For reference, here's the super simple code snippet that utilizes the same functionality but compiles fine in a brand new project.

```
import ComposableArchitecture

@Reducer
struct ExampleReducer {
    @ObservableState
    struct State: Equatable {
        var didReceiveHelloWorld = false
        var searchText = ""
    }

    enum Action: BindableAction, Equatable {
        case helloWorld
    }

        BindingReducer()
        Reduce { state, action in
            switch action {
            case .binding(\.searchText):
                return searchTextUpdated(with: &state)
            case .binding:
                return .none
            case .helloWorld:
                state.didReceiveHelloWorld = true
                return .none
            }
        }
    }

    private func searchTextUpdated(with state: inout State) -> EffectOf<Self> {
        if state.searchText == "Hello World" {
            .send(.helloWorld)
        } else {
            .none
        }
    }
}
```

PR #3919, opened 2026-04-24
Fix Xcode 26 xcodebuild archive failures: add `import Combine` to Logger.swift and `@unknown default` to non-frozen enum switches

### LINKED ISSUE #3918: Error when archiving via xcodebuild archive
### Description

I have a Swift Package module that uses TCA. I want to be able to distribute this as an .xcframework. To achieve this I am using `xcodebuild archive (...)`. When doing so I get an error:

``` 
/SourcePackages/checkouts/swift-composable-architecture/Sources/ComposableArchitecture/Internal/Logger.swift:8:4: 'Published' aliases 'Combine.Published' and cannot be used as property wrapper here because 'Combine' was not imported by this file
  @Published public var logs: [String] = []
   ^
```

I set up a minimal project where the issue is reproducible: https://github.com/joaomvfsantos/TCACompileIssue

The repo contains a README with detailed command instructions.

### Checklist

- [x] If possible, I've reproduced the issue using the `main` branch of this package.

### Expected behavior

I would expect `xcodebuild archive` to finish without errors.

### Actual behavior

The error `'Published' aliases 'Combine.Published' and cannot be used as property wrapper here because 'Combine' was not imported by this file` is thrown

### Reproducing project

https://github.com/joaomvfsantos/TCACompileIssue

1.25.5

iOS 16

### Xcode version information

Xcode 26.4.1

```shell
swift-driver version: 1.148.6 Apple Swift version 6.3.1 (swiftlang-6.3.1.1.2 clang-2100.0.123.102)
Target: arm64-apple-macosx26.0
```

joaomvfsantos:
## Problem

When archiving with `xcodebuild archive` on Xcode 26, the `ComposableArchitecture` target is compiled in **Swift 6 mode** despite the package-level workaround that adds `.swiftLanguageMode(.v5)` to every target via a loop in `Package.swift`.

The root cause is a behaviour difference between xcodebuild archive and the Xcode GUI in Xcode 26: The former reads the package-level `swiftLanguageModes: [.v6]` declaration and **ignores the per-target `.swiftLanguageMode(.v5)` overrides** that are appended dynamically in the `for target in package.targets` loop. The latter does not exhibit this behaviour and archives successfully.

Once `xcodebuild` forces Swift 6 mode on TCA's targets, two classes of code that are valid in Swift 5 become compile errors:

### 1. `Logger.swift` — missing `import Combine`

```
Sources/ComposableArchitecture/Internal/Logger.swift:12:4: error: 'Published' aliases 'Combine.Published' and cannot be used as property wrapper here because 'Combine' was not imported by this file
```

`@Published` is a Combine type. In Swift 5, it could be used without an explicit `import Combine` because Combine was re-exported transitively. In Swift 6, **MemberImportVisibility** is a language requirement: every file must import the module that defines the symbol it uses. `Logger.swift` only imports `OSLog`.

**Fix:** add `import Combine` to `Logger.swift`.

### 2. `Alert+Observation.swift`, `Alert.swift`, `ConfirmationDialog.swift` — non-exhaustive switches on non-frozen enums
```
Sources/ComposableArchitecture/Observation/Alert+Observation.swift:21:13: error: switch covers known cases, but 'ButtonStateAction._ActionType' may have additional unknown values
```

`switch` statements over non-`@frozen` enums without an `@unknown default` case produce a **warning** in Swift 5 but an **error** in Swift 6. Three files contain such switches over `ButtonStateAction._ActionType`.

**Fix:** add `@unknown default: break` to the affected switch statements.

## Repro

1. Xcode 26 (any 26.x build)
2. A project that depends on TCA 1.25.5 as an SPM package inside an Xcode
   workspace, where the native target (`xcodeproj`) uses
   `BUILD_LIBRARY_FOR_DISTRIBUTION=YES`
3. Run:
   ```bash
   xcodebuild archive \
     -workspace MyApp.xcworkspace \
     -scheme MyFramework \
     -destination 'generic/platform=iOS Simulator' \
     SKIP_INSTALL=NO \
     BUILD_LIBRARY_FOR_DISTRIBUTION=YES

Reproducible Repo: https://github.com/joaomvfsantos/TCACompileIssue

Aims to fix #3918

### DIFF
--- a/Sources/ComposableArchitecture/Internal/Deprecations.swift
+++ b/Sources/ComposableArchitecture/Internal/Deprecations.swift
@@ -2079,6 +2079,8 @@ extension View {
                 if let action {
                   store.send(.presented(fromDestinationAction(action)), animation: animation)
                 }
+              @unknown default:
+                break
               }
             } label: {
               Text(button.label)
@@ -2132,6 +2134,8 @@ extension View {
                 if let action {
                   store.send(.presented(fromDestinationAction(action)), animation: animation)
                 }
+              @unknown default:
+                break
               }
             } label: {
               Text(button.label)
--- a/Sources/ComposableArchitecture/Internal/Logger.swift
+++ b/Sources/ComposableArchitecture/Internal/Logger.swift
@@ -1,4 +1,5 @@
 import OSLog
+import Combine
 
 @_spi(Logging)
 @preconcurrency @MainActor
--- a/Sources/ComposableArchitecture/Observation/Alert+Observation.swift
+++ b/Sources/ComposableArchitecture/Observation/Alert+Observation.swift
@@ -22,6 +22,8 @@ extension View {
               if let action {
                 store?.send(action, animation: animation)
               }
+            @unknown default:
+              break
             }
           } label: {
             Text(button.label)
@@ -59,6 +61,8 @@ extension View {
               if let action {
                 store?.send(action, animation: animation)
               }
+            @unknown default:
+              break
             }
           } label: {
             Text(button.label)


joaomvfsantos:
Hello @stephencelis. Any chance to have a look at this PR? If there is something missing from my side, please let me know. Thanks in advance

mbrandonw:
Hi @joaomvfsantos, thanks for this. We don't _technically_ say we support building for library evolution, but for simple fixes like this we are happy to take it. We just can't guarantee that things will always be fixed.

PR #3923, opened 2026-05-07
Add `scope(_:action:)` and `Scope(_:action:)`

stephencelis:
To prepare for changes coming in 2.0, we are introducing new scoping overloads that omit the `state:` parameter label. This label is already omitted for some scoping operations (`ifLet`, `forEach`, etc.) and is just added noise for the dominant part of the operation.

The old API has been trait-deprecated so that folks can migrate at their own pace without too much noise.


mbrandonw (approved):

maximkrouk:
Is there a possibility you could consider providing multiple APIs in 2.0?

I strongly prefer having a labeled arguments (maybe I'm not the only one) for multiline calls like
```swift
SomeView(store.scope(
  state: \.child
  action: \.child
))

Scope(
  state: \.someState
  action: \.someAction,
  child: SomeFeature.init
).ifLet(
  state: \.derived
  action: \.derived
)

// etc.
```

PR #3928, opened 2026-06-02
Stabilize NavigationLink stack element IDs

### LINKED ISSUE #3633: stackElementID breaks List selection
### Description

Normally, pop to a `List` will also deselect the cell along with the pop transition, and this behavior depends on the `value` of `NavigationLink` in SwiftUI.
However, `NavigationLink.init(state:label:fileID:filePath:line:column:)` will increment the `id` each time, making the `value` unique.

https://github.com/pointfreeco/swift-composable-architecture/blob/7c2ed5d62bd91d4ad5b1dcdc5081e05cbcfd2b92/Sources/ComposableArchitecture/Observation/NavigationStack%2BObservation.swift#L249-L250

With this implementation, the cell selection will not be retained during the push and pop, so the cell will be deselected instantly when pop (not during the transition).

### Checklist

- [x] If possible, I've reproduced the issue using the `main` branch of this package.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Reproducing project

_No response_

_No response_

_No response_

### Xcode version information

_No response_

```shell

```

kswift1:
## Summary

- Reset stack element ID generation while building store-powered NavigationStack roots and destinations so NavigationLink(state:) values remain stable across view recomputation.
- Start nested destination links after the current stack component generation to avoid ID collisions.
- Advance the reducer generator when handling push actions with the next expected ID, including live contexts.

Fixes #3633.

## Testing

- swift test
- swift test --filter StackReducerTests
- swift test --filter StackReducerTests/testPush --filter StackReducerTests/testPushIDGreaterThanNextGeneration --filter StackReducerTests/testSendCopiesStackElementIDGenerator

Also attempted the iOS 17 ObservableNavigation integration UI test locally with Xcode 26.5, but that target is currently blocked before execution by unrelated Integration target/toolchain build issues.


stephencelis:
@kswift1 Thanks for taking the time! Unfortunately I don't think it fixes the underlying problem, though. In the NavigationStack case study, I can tap A, then tap A again, and interactive pop will not show the second A highlighted.

We're just not sure SwiftUI's `NavigationLink` is up to the task when it comes to more advanced forms of binding-based navigation, like what we employ in TCA for store scoping. And so far in ComposableArchitecture 2.0 we haven't decided to bring `NavigationLink.init(state:)` along for the ride except as a migratory backfill.

We are open to supporting integration with `NavigationLink.init(value:)` in some capacity in the future, but we think it'll probably look different than the route we took with 1.0.

Given all that, are you still interested in pushing this branch further for 1.x? Or would you rather wait for 2.x and see what can be done there?

kswift1:
@stephencelis Thanks for taking a look!

That makes sense. I was mostly trying to address the original mismatch where `NavigationLink(state:)` produces a new stack element ID across view recomputation, but I understand that the recursive A -> A case still exposes a deeper issue with SwiftUI's selection/highlight behavior.

I'm still interested in giving this one more pass for 1.x, scoped narrowly to the existing `NavigationLink.init(state:)` behavior. I'll try to reproduce the A -> A interactive-pop issue locally and see if there is a focused fix that can cover it without changing the broader navigation model.

If it turns out to require a different design, or if you'd prefer not to invest more in this API for 1.x, I'm also happy to pause this PR and revisit the idea when the 2.x navigation direction is clearer.

kswift1:
Quick update: I took another pass at the A -> A interactive-pop case and was able to reproduce the behavior in the CaseStudies app.

After tracing the ID flow, this branch already appears to give the nested A link the same `StackElementID` as the second path element. So the remaining highlight issue does not seem to be another simple generator-offset bug.

The fixes I can think of from here would require changing what participates in `StackState.Component` identity/equality, requiring user-provided element identity, or moving away from relying on SwiftUI's `NavigationLink(value:)` selection behavior for these store-scoped links. Those all feel broader than a narrow 1.x fix and seem to overlap with the 2.0 navigation work you mentioned.

Given that, I'll close this PR for now rather than keep a partial 1.x fix open. Thanks again for taking a look, and I'd be happy to revisit this when the 2.0 navigation direction is clearer.

PR #3932, opened 2026-06-09
docs: add Persian translation

MojtabaHs:


stephencelis (approved):
Thanks for taking the time! 😃

PR #3935, opened 2026-06-09
Fix syntax errors in bindings documentation examples

indextrown:
## Summary

Fix several syntax issues in the bindings documentation examples:

- Add missing leading dots to enum case patterns.
- Use `isOn:` instead of `text:` for a `Toggle` binding.
- Add a missing `{` after `switch action`.
- Fix a mismatched closing delimiter in a binding test example.

## Testing

- Documentation changes only.


stephencelis (approved):
Thanks!

PR #3936, opened 2026-06-10
Fix SharingState documentation example syntax

opficdev:
## Summary

Fix several syntax issues in the `SharingState.md` documentation examples:

- Add a missing closing delimiter to a `@Shared(.fileStorage(...))` example.
- Normalize a mismatched Swift code fence delimiter.
- Remove an extra closing delimiter from a `URL(/* ... */)` example.
- Update shared state test assertions to use the correct projected value.

## Details

| Change | Reason |
|---|---|
| `@Shared(.fileStorage(URL(/* ... */))) var users` | The example opens `@Shared(`, `.fileStorage(`, and `URL(`, so it needs three closing parentheses before `var users`. |
| Four backticks -> three backticks | The code block opens with three backticks, so closing it with the same delimiter keeps the DocC Markdown example consistent. |
| `static let users = URL(/* ... */)` | The previous example had one extra closing parenthesis, which made the Swift snippet invalid. |
| `$0.$shared` -> `$0.$bool` | The example declares `@Shared(value: false) var bool`, so the projected value is `$bool`. The `_shared` text shown in the failure output is backing storage, not an accessible state property. |

### DIFF
--- a/Sources/ComposableArchitecture/Documentation.docc/Articles/SharingState.md
+++ b/Sources/ComposableArchitecture/Documentation.docc/Articles/SharingState.md
@@ -187,7 +187,7 @@ It works similarly to the in-memory sharing discussed above, but it requires a U
 on disk, as well as a default value that will be used when there is no data in the file system:
 
 ```swift
-@Shared(.fileStorage(URL(/* ... */)) var users: [User] = []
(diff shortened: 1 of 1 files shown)


stephencelis (approved):
Thanks!

PR #3940, opened 2026-06-15
Fix missing await in TestingTCA send examples

opficdev:
## Summary

Fix a few `TestingTCA.md` documentation examples that call `TestStore.send` without `await`:

- Add `await` to examples that explain how `TestStore.state` behaves inside `send` assertions.
- Add `await` to examples that compare exhaustive and non-exhaustive `TestStore` behavior.
- Align the snippets with the rest of the article, where `TestStore.send` is already awaited.

## Details

| Change | Reason |
|---|---|
| `await store.send(.incrementButtonTapped) { ... }` | `TestStore.send` is async and returns `TestStoreTask`, so copied test code needs `await`. |
| `await store.send(.buttonTapped) { ... }` | The surrounding examples use `await` for the same API, and these snippets were the remaining inconsistent calls. |


stephencelis (approved):
Thanks!

PR #3942, opened 2026-06-23
Delete NotificationName.swift

mbrandonw:
This is an old file left over from when the sharing tools were built into TCA directly.


stephencelis (approved):

jshier:
The build failure fixed by this PR still exists for me in Xcode 27b3, so a release would be appreciated. While this seems to be due to Xcode's incorrect decision to build our packages for Intel, it's still a blocker.

gpambrozio:
Still happening on Xcode 27b4 so I need to tag to a hash, not ideal.

mbrandonw:
Sorry, lost track of this. We just released 1.26.1

gpambrozio:
Thank you!

PR #3949, opened 2026-07-13
Remove observable store identity validation for UIBindings

maximkrouk:
Id check in `store.scope` breaks UIKit navigation with `swift-navigation` and `UIBindings`, this PR removes this check specifically for UIBindings. Maybe later I'll manage to create a reproducible repo, but for now just a PR

Some more context: https://pointfreecommunity.slack.com/archives/C04KQQ7NXHV/p1783586402249999


stephencelis:
@maximkrouk Would love to look into this when we have the time but could you attach a small project that demonstrates the problem/fix?

PR #3952, opened 2026-07-16
Fix TestStore initialization in navigation tutorial

opficdev:
## Summary

Fix the `TestStore` initialization in the navigation testing tutorial.

The example previously initialized the test store with an empty state and then placed a second `initialState:` expression inside the reducer closure, which made the tutorial snapshot invalid Swift.

This change:

- Initializes `ContactsFeature.State` with the contact fixtures directly in the existing `initialState` argument.
- Removes the duplicated `initialState:` expression from the reducer closure.
- Keeps the snapshot consistent with the following tutorial step.

## Testing

- `xcrun swiftc -frontend -parse Sources/ComposableArchitecture/Documentation.docc/Tutorials/MeetTheComposableArchitecture/02-Navigation/03-TestingPresentation/02-03-03-code-0002.swift`
- `git diff --check main...HEAD`


mbrandonw (approved):
Good catch @opficdev, thank you!

PR #3954, opened 2026-07-21
Fix reducer in tree-based navigation test examples

indextrown:
## Summary

Fix the reducer used in the tree-based navigation testing examples.

Both examples initialize `TestStore` with `Feature.State`, but incorrectly use `CounterFeature` as the reducer. This updates them to use `Feature` so that the reducer matches the initial state.

## Testing

- Documentation changes only.


mbrandonw (approved):
Good catch @indextrown, thank you!

PR #3955, opened 2026-07-21
Fix errors in stack navigation documentation examples

indextrown:
## Summary

Fix two errors in the stack-based navigation documentation examples.

Changes include:

- Use `state.path.popLast()` instead of `state.popLast()` in the dismissal example.
- Use `Feature` instead of `CounterFeature` as the reducer in testing examples initialized with `Feature.State`.

## Testing

- Documentation changes only.


mbrandonw (approved):
Thanks again!

PR #3956, opened 2026-07-24
Fix missing @ObservableState annotations in bindings examples

indextrown:
## Summary

Fix the ad hoc binding examples by adding the missing `@ObservableState` annotations.

The examples use `@Bindable` and access state through `$store`, which requires
`Settings.State` to conform to `ObservableState`. This makes the examples consistent
with the binding action examples below.

## Testing

- Documentation changes only.


mbrandonw (approved):
Good finds again, thanks @indextrown!

